PR #14, opened 2021-02-24
Feature/condo/sberdoma 15/add basic billing test and access check logic

pahaz:


apps/condo/access.js
```diff
@@ -16,6 +16,30 @@ const permissions = {
...
+    return employeeRole[permission] || false
+}
+
+async function checkBillingIntegrationAccessRight (userId, integrationId) {
+    if (!userId || !integrationId) return false
+    const integration = await getByCondition('BillingIntegrationAccessRight', {
+        integration: { id: integrationId },
+        user: { id: userId },
+    })
+    return Boolean(integration && integration.id)
```
Dimitreee, line 40: What will happen when integration.id will equals `0`?

apps/condo/access.js
```diff
@@ -113,18 +129,101 @@ const rules = {
...
+            //     { integration: { accessRights_some: { user: { id: user.id } } } },
+            // ],
+        }
+    },
+    canManageBillingIntegrationOrganizationContexts: async ({ authentication: { item: user }, originalInput, operation, itemId }) => {
+        if (!user) return false
+        if (user.isAdmin) return true
+        if (operation === 'create') {
+            // NOTE: can create only by the organization integration manager
+            if (!originalInput || !originalInput.organization || !originalInput.organization.connect || !originalInput.organization.connect.id) return false
```
Dimitreee, line 157: better to uze `get` here

Dimitreee:
```
import get from 'lodash/get'
// ...
const organizationId = get(originalInput, ['organization', 'connect', 'id'])
if (!organizationId) return false
// ...
```
 This solution is much readable

apps/condo/access.js
```diff
@@ -113,18 +129,101 @@ const rules = {
...
+            const organizationId = originalInput.organization.connect.id
+            return await checkOrganizationPermission(user.id, organizationId, 'canManageIntegrations')
+        } else if (operation === 'update') {
+            // NOTE: can update by the organization integration manager OR the integration account
+            if (!itemId) return false
+            const context = await getById('BillingIntegrationOrganizationContext', itemId)
+            if (!context) return false
+            const organizationId = context.organization
```
Dimitreee, line 165: Guess its mistake here, u mean `const organizationId = context.organization.connect.id`

apps/condo/access.js
```diff
@@ -113,18 +129,101 @@ const rules = {
...
+            const integrationId = context.integration
```
Dimitreee, line 166: Guess its mistake here, u mean const organizationId = context.integration.connect.id

Otherwise u don't need to define this variables

apps/condo/access.js
```diff
@@ -113,18 +129,101 @@ const rules = {
...
+            // },
+        }
+    },
+    canManageBillingIntegrationLogs: async ({ authentication: { item: user }, originalInput, operation, itemId }) => {
+        if (!user) return false
+        if (user.isAdmin) return true
+        let contextId
+        if (operation === 'create') {
+            // NOTE: can create only by the integration account
+            if (!originalInput || !originalInput.context || !originalInput.context.connect || !originalInput.context.connect.id) return false
```
Dimitreee, line 199: use lodash `get`

apps/condo/access.js
```diff
@@ -113,18 +129,101 @@ const rules = {
...
+            // NOTE: can update only by the integration account
+            if (!itemId) return false
+            const log = await getById('BillingIntegrationLog', itemId)
+            contextId = log.context
+        } else {
+            return false
+        }
+        const context = await getById('BillingIntegrationOrganizationContext', contextId)
+        if (!context) return false
+        const integrationId = context.integration
```
Dimitreee, line 211: Guess u don't need this variable

apps/condo/schema/Billing.js
```diff
@@ -160,6 +204,12 @@ const BillingIntegrationLog = new GQLListSchema('BillingIntegrationLog', {
             schemaDoc: 'Message type. Our clients can use different languages. Sometimes we need to change the text message for the client. The settings for the message texts are in the integration. Ex: WRONG_AUTH_CREDENTIALS',
             type: Text,
             isRequired: true,
+            hooks: {
+                validateInput: ({ resolvedData, fieldPath, addFieldValidationError }) => {
+                    const value = resolvedData[fieldPath]
+                    if (!/^[A-Z_]+$/.test(value)) addFieldValidationError(`${WRONG_TEXT_FORMAT}${fieldPath}] allow only [A-Z_] letters`)
```
Dimitreee, line 210: Can we pick out this regex to named Const. Guess this will improve readability, like `const LATIN_ALPHABET = /^[A-Z_]+$/`

apps/condo/schema/Billing.test.js
```diff
@@ -22,6 +25,62 @@ async function createBillingIntegration (client, extraAttrs = {}) {
     return [obj, attrs]
 }
 
+async function createBillingIntegrationAccessRight (client, user, integration, extraAttrs = {}) {
+    if (!client) throw new Error('no client')
```
Dimitreee, line 29: Later we discussed about this raise exception pattern and we conclude that we will use the Facebook library for this purpose.

apps/condo/access.js
pahaz: It looks like an impossible case. We have database-level constraints

apps/condo/access.js
pahaz: No, it's different. On create we use `context.organization.connect.id`
But on update case, we should get the current db state

apps/condo/access.js
pahaz: no, it a state from the database.

apps/condo/access.js
Dimitreee:
```suggestion
    return get(integration, 'id') !== undefined
```

apps/condo/access.js
Dimitreee:
```suggestion
            const { organization: organizationId } = context
```

apps/condo/access.js
Dimitreee:
```suggestion
        const { integration: integrationId } = context
```

apps/condo/schema/Billing.js
Dimitreee: move to constants

PR #21, opened 2021-03-04
Fix/condo/sberdoma 46/app fixes

Dimitreee:


apps/condo/containers/BBuilder/BBuildingData.ts
```diff
@@ -262,3 +278,44 @@ export function createNewBBuildingSectionData ({ sectionName, maxFloor, minFloor
         }),
     }
 }
+
+type Data = {
+    sections: [{
+        floors: [{
+            units: [{
```
pahaz, line 285: Why you don't want to use already declared Unit interface?

pahaz (approved):

PR #72, opened 2021-04-19
Feature/condo/sberdoma 101/common data visibility fixes

Dimitreee:


apps/condo/domains/user/access/User.js
```diff
@@ -4,22 +4,32 @@
 const access = require('@core/keystone/access')
 
 async function canReadUsers ({ authentication: { item: user } }) {
+
     if (!user || !user.id) return false
-    if (user.isAdmin) return true
-    return true
+
+    if (user.isAdmin) {
+        return {}
```
Dimitreee, line 11: @pahaz should we return `true` for `user.isAdmin` permissions both in `canReadUsers` & `canManageUsers` guards?

pahaz (approved):

apps/condo/domains/common/utils/codegeneration/generate.test.utils.js
```diff
@@ -1,11 +1,12 @@
 const conf = require('@core/config')
 
 const IS_DEBUG = conf.NODE_ENV === 'development' || conf.NODE_ENV === 'test'
+const util = require('util')
```
pahaz, line 4: style (should be upper than consts)

PR #73, opened 2021-04-21
Feature/condo/sberdoma 266/init messages

pahaz:


dkoviazin (approved):
Ok
We need to move phone validation logic somewhere to the common domain (with the clearing of dashes, normalizing 8 to +7 if we have no support for other countries)

Dimitreee (approved):

PR #79, opened 2021-04-23
Feature/condo/sberdoma 151/editing of selected text in address search field

AntonAL:


Dimitreee:
There is no city/region autocomplete suggestions

AntonAL:
A task appeared as too big and has a lot of custom logic.
It's not reasonable to work forward on it now.

PR #89, opened 2021-05-04
Feature/condo/sberdoma 266/init messages

pahaz:


apps/condo/domains/notification/schema/Message.js
```diff
@@ -0,0 +1,108 @@
...
+
+        status: {
+            schemaDoc: 'Message status',
+            type: Select,
+            defaultValue: 'sending',
+            options: 'sending,resending,processing,error,sent,canceled',
+            isRequired: true,
+        },
+
+        processingMeta: {
```
Dimitreee, line 79: What is task processing metadata? What's the reason to add this field in Message model?

apps/condo/domains/notification/schema/Message.js
```diff
@@ -0,0 +1,108 @@
...
+            type: Json,
+            isRequired: false,
+            hooks: {
+                validateInput: (args) => {
+                    if (!hasValidJsonStructure(args, false, 1, {})) return
+                },
+            },
+        },
+
+        sentAt: {
```
Dimitreee, line 90: I'm not sure, but i'll bet that we need handle 'delivered at' time, if it possible.

apps/condo/domains/notification/schema/SendMessageService.js
```diff
@@ -0,0 +1,123 @@
...
+                await checkSendMessageMeta(type, meta)
+
+                const messageAttrs = { dv, sender, status: 'sending', type, meta, lang }
+                if (to.email) messageAttrs.email = to.email
+                if (to.phone) messageAttrs.phone = to.phone
+                if (to.user) messageAttrs.user = { connect: to.user }
+
+                const message = await Message.create(context, messageAttrs)
+
+                await scheduleMessageDelivery(message)
```
Dimitreee, line 89: What about exceptions handle from  `sendMessageByTransport`, should we handle them?

apps/condo/domains/notification/tasks.js
```diff
@@ -0,0 +1,75 @@
...
+    await Message.update(keystone, message.id, {
+        ...baseAttrs,
+        status: 'processing',
+        processingMeta,
+    })
+
+    try {
+        const adapter = TRANSPORTS[transport]
+        const context = await adapter.prepareMessageToSend(message)
+        processingMeta.step = 2
```
Dimitreee, line 41: I'm not sure, but i guess that it is more comfortable and understandable to work with named processing Meta steps, like 'Starting|preparing|delivering' etc.

apps/condo/domains/notification/tasks.js
```diff
@@ -0,0 +1,75 @@
...
+    // update meta
+    await Message.update(keystone, message.id, {
+        ...baseAttrs,
+        status: 'sent',
+        processingMeta,
+    })
+
+    if (processingMeta.error) {
+        throw new Error(processingMeta.error)
+        // TODO(pahaz): need to think about some repeat logic? 2 times?
```
Dimitreee, line 68: I guess task need to be configurable with 'retry times' or stmhng.

apps/condo/domains/notification/transports/email.js
```diff
@@ -0,0 +1,82 @@
...
+
+        if (message.lang === 'en') {
+            return {
+                subject: 'You are invited to join organization as employee',
+                text: `Organization "${organizationName}" invited you as employee.\n` +
+                    `Click to the link to join: ${serverUrl}/auth/invite/${inviteCode}`,
+            }
+        } else if (message.lang === 'ru') {
+            return {
+                subject: 'Вас приглаиси присоединиться к организации в качестве сотрудника',
```
Dimitreee, line 28:
```suggestion
                subject: 'Вас пригласили присоединиться к организации в качестве сотрудника',
```

apps/condo/domains/notification/transports/email.js
```diff
@@ -0,0 +1,82 @@
...
+async function prepareMessageToSend (message) {
+    const email = message.email || (message.user && message.user.email) || null
+    if (!email) throw new Error('on email to send')
+
+    const { subject, text, html } = await renderTemplate(message)
+
+    return { to: email, subject, text, html }
+}
+
+async function send ({ to, cc, bcc, subject, text, html } = {}) {
```
Dimitreee, line 47: As for me, It is not clear enough what these parameters meaning. can we add js doc for `send` function signature?

apps/condo/domains/notification/transports/sms.js
```diff
@@ -0,0 +1,66 @@
...
+    if (!phone) throw new Error('on phone to send')
+
+    const { text } = await renderTemplate(message)
+
+    return { phone, message: text }
+}
+
+async function send ({ phone, message } = {}) {
+    if (!SMS_API_CONFIG) throw new Error('no SMS_API_CONFIG')
+    if (!/^[+]7[0-9]{10}$/g.test(phone)) throw new Error('unsupported phone number')
```
Dimitreee, line 46: move to regexps

apps/condo/domains/notification/utils/clientSchema/Message.ts
```diff
@@ -0,0 +1,66 @@
...
+    // TODO(codegen): write IMessageUIState or extends it from
+}
+
+function convertToUIState (item: Message): IMessageUIState {
+    if (item.dv !== 1) throw new Error('unsupported item.dv')
+    return pick(item, FIELDS) as IMessageUIState
+}
+
+export interface IMessageFormState {
+    id?: undefined
```
Dimitreee, line 27:
```suggestion
    id?: void
```

apps/condo/domains/organization/schema/InviteNewOrganizationEmployeeService.js
```diff
@@ -70,7 +70,7 @@ const InviteNewOrganizationEmployeeService = new GQLCustomSchema('InviteNewOrgan
                 // TODO(pahaz): send email !?!?!
                 console.log('Fake send security email!')
 
-                return await getById('OrganizationEmployee', obj.id)
+                return await getById('OrganizationEmployee', employee.id)
```
Dimitreee, line 73: NICE!!! 😀 😀 😀 😀

packages/@core.keystone/tasks.js
```diff
@@ -121,6 +123,9 @@ function createWorker () {
         console.log('Creating worker process')
     }
 
+    // we needed to prepare keystone to use it inside tasks logic!
+    await prepareKeystoneExpressApp(keystoneModule)
```
Dimitreee, line 127: look's strange, is it correct?

apps/condo/domains/notification/schema/Message.js
pahaz: Yes, I've added two statuses for that.

sent -- everything delivered
error -- something went wrong

apps/condo/domains/notification/schema/Message.js
pahaz: You can see exactly what and where the message was sent. I've updated the text

apps/condo/domains/notification/schema/Message.js
Dimitreee: How should we handle `delivered date` from this statuses?

pahaz: I renamed `sent at` time to `delivered at` time

apps/condo/domains/notification/schema/SendMessageService.js
pahaz: I'm not sure. At the moment it's OK if we throw the error up to GQL.
If some error accrued during the scheduleMessageDelivery call it meant the sendMessage should fail too.

pahaz: And for transports exceptions, we handle all transport exceptions.

apps/condo/domains/notification/tasks.js
pahaz: Yes, I think it's OK to add it in the future. 
Currently, we don't miss any message and if it's required we can reschedule sending easily.

Since multiple resending can burn the service budget, I haven't added it yet. I want to look at the errors in the process.

apps/condo/domains/notification/transports/sms.js
pahaz: It's not a global REGEXP. This is a limitation of the service that is currently used for sending SMS.
This regexp is not reusable.

packages/@core.keystone/tasks.js
pahaz: Yes. The worker is a separate node process and we need to prepare some keystone internal state like database connection, internal apollo gql setup, and so on ...

apps/condo/domains/notification/tasks.js
pahaz: fixed

apps/condo/domains/notification/utils/clientSchema/Message.ts
pahaz: `void` is assignable to null || undefined
And `null` value is not correct in this case

apps/condo/domains/notification/transports/email.js
```diff
@@ -3,13 +3,60 @@ const FormData = require('form-data')
 
 const conf = require('@core/config')
 
+const { INVITE_NEW_EMPLOYEE_MESSAGE_TYPE } = require('../constants')
+
 const EMAIL_API_CONFIG = (conf.EMAIL_API_CONFIG) ? JSON.parse(conf.EMAIL_API_CONFIG) : null
 
-async function prepareMessageToSend (messageId) {
+async function renderTemplate (message) {
+    // TODO(pahaz): we need to decide where to store templates! HArDCODE!
```
dkoviazin, line 11: For emails, it's good to use HTML templates from services 
Mailgun, Mailchimp, Sendgrid - all have an opportunity to store and edit templates on their side. 
In this case, we can pass only template id and variables for replaces 
By doing so, we can also change templates faster

apps/condo/index.js
```diff
@@ -13,15 +13,17 @@ const { EmptyApp } = require('@core/keystone/test.utils')
 const { prepareDefaultKeystoneConfig } = require('@core/keystone/setup.utils')
 const { registerSchemas } = require('@core/keystone/schema')
 
-require('dd-trace').init({
-    logInjection: true,
-})
+if (conf.NODE_ENV === 'production') {
+    require('dd-trace').init({
+        logInjection: true,
```
dkoviazin, line 18: Super

dkoviazin (approved):
Looks good

apps/condo/domains/notification/tasks.js
```diff
@@ -1,19 +1,98 @@
...
+    if (SEND_TO_CONSOLE) {
+        console.info(`MESSAGE by ${transport.toUpperCase()} ADAPTER: ${JSON.stringify(messageContext)}`)
+        return [true, { fakeAdapter: true }]
+    }
+    return await adapter.send(messageContext)
+}
+
+async function deliveryMessage (messageId) {
+    const { keystone } = await getSchemaCtx('Message')
+    const messages = await Message.getAll(keystone, { id: messageId })
```
Dimitreee, line 28:
```suggestion
    const [message] = await Message.getAll(keystone, { id: messageId })
```

apps/condo/domains/notification/tasks.js
```diff
@@ -1,19 +1,98 @@
...
+        processingMeta,
+    })
+
+    try {
+        const adapter = TRANSPORTS[transport]
+        const messageContext = await adapter.prepareMessageToSend(message)
+        processingMeta.step = 'prepared'
+        processingMeta.messageContext = messageContext
+
+        const [isOk, deliveryMetadata] = await _sendMessageByAdapter(transport, adapter, messageContext)
```
Dimitreee, line 63: What's the difference between send/delivery ?

Dimitreee: Seems like it's _deliverMessageByAdapter

Dimitreee: And, why we cannot use adapter as pure dependency of deliveryMessage function?

apps/condo/index.js
Dimitreee: ++

Dimitreee (approved):

apps/condo/domains/notification/transports/email.js
pahaz: Yes, I needed to make a decision with the product. Hardcoded at the moment

apps/condo/domains/notification/tasks.js
pahaz: delivery is an async task. the message delivery process is handled by the worker. I also add a diagram to README.md

PR #126, opened 2021-05-21
Feature/condo/sberdoma 299/add ticket files to sbercloud obs

dkoviazin:
## Preview
1. Add Sbercloud API FileAdapter
2. Add TicketFile schema with migration
3. Add Multiple upload component to use in different forms
4. Add input to upload/delete files on ticket form if ticket already exists


apps/condo/domains/common/utils/sberCloudFileAdapter.js
```diff
@@ -0,0 +1,144 @@
+const ObsClient = require('esdk-obs-nodejs')
+const { createReadStream } = require('fs')
+
+class SberCloudFileAdapter {
+
+    constructor (folder = '') {
+        const sberCloudObsConfig = process.env.SBERCLOUD_OBS_CONFIG ? JSON.parse(process.env.SBERCLOUD_OBS_CONFIG) : {}
```
Dimitreee, line 7: move to constructor args, add private accessor

apps/condo/domains/common/utils/sberCloudFileAdapter.js
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+}
+
+
+
+class SberCloudObs {
+
+    constructor (config) {
+        const sberCloudObsConfig = process.env.SBERCLOUD_OBS_CONFIG ? JSON.parse(process.env.SBERCLOUD_OBS_CONFIG) : {}
```
Dimitreee, line 81: move sberCloudObsConfig to constructor args

apps/condo/domains/common/utils/sberCloudFileAdapter.js
```diff
@@ -0,0 +1,144 @@
...
+
+    async isBucketExists () {
+        const { CommonMsg: { Status } } = await this.obs.headBucket({ Bucket: this.bucket })
+        return Status < 300
+    }
+
+    async createBucket (params = {}) {
+        const { CommonMsg: { Status } } = await this.obs.CreateBucket({
+            Bucket: this.bucket,
+            ACL: this.obs.enums.AclPublicRead,
```
Dimitreee, line 98: What is ACL? add cmment please

apps/condo/domains/ticket/access/TicketFile.js
```diff
@@ -0,0 +1,33 @@
+/**
+ * Generated by `createschema ticket.TicketFile 'file:File;ticket:Relationship:Ticket:PROTECT;'`
+ */
+
+async function canReadTicketFiles ({ authentication: { item: user } }) {
```
Dimitreee, line 5: Why didnt u create FIle model, is this model be differrent from TicketFile?

dkoviazin: I had a discussion with @pahaz  about how it will look like. Here are some points:
1. Access management
If we create only one model named File and have different keys to connect this file to different models and all these fields will contain null except one - every time we will try to check access to this file we need to write lots of "if" conditions
2. S3 files storage
All files from tickets now placed in the "ticket" folder, for users it will be the "user" folder
So, the main part of my task was to create reusable mechanics to work with files in different models and to make a reusable component to easily plug into any form

Dimitreee: You can extend TicketFile model from File model, and write custom points.

How should other Developers works with this model if they will be needed to use Sber bucket to store data there?
@pahaz

dkoviazin: You only need to choose adapter when you create model
```js
const fileAdapter = new SberCloudFileAdapter('ticket')
  fields: {
        dv: DV_FIELD,
        sender: SENDER_FIELD,
        file: {
            schemaDoc: 'File object with meta information and publicUrl',
            type: File,
            adapter: fileAdapter,
            isRequired: true,
        },
```
You can do the same with the user model to store avatar in Sbercloud all you need is to replace 
```

// TODO(pahaz): we should change it to Remote S3 like storage! #scalability
const AVATAR_FILE_ADAPTER = new LocalFileAdapter({
    src: `${conf.MEDIA_ROOT}/avatars`,
    path: `${conf.MEDIA_URL}/avatars`,
})
```
with new one
```
const fileAdapter = new SberCloudFileAdapter('user')
```

dkoviazin: Soon will be:
```js
const AVATAR_FILE_ADAPTER = new SberCloudFileAdapter(process.env.SBERCLOUD_OBS_CONFIG, 'user')
```

PR #141, opened 2021-05-31
SBERDOMA-348 add apollo cache config to useApollo

pahaz:


dkoviazin (approved):
Ok

apps/condo/pages/_app.tsx
```diff
@@ -86,8 +86,10 @@ async function messagesImporter (locale) {
     return { ...locale_data.default }
 }
```
AntonAL, line 88:
```suggestion
/*
    Configuration for `InMemoryCache` of Apollo
    Add fields, related to pagination strategies of Apollo.
    Items of some GraphQL global fields needs to be appended to list,
    when paginated, rather than to be displayed as a slice of data, —
    its like "Infinite scrolling" UI pattern. For example, fetching
    more changes of a ticket on button click.
    For those items, we need to set `concatPagination` strategy.
    https://www.apollographql.com/docs/react/pagination/core-api/
 */
cons
```

AntonAL (approved):
Tested on my example, — it works.
It would be more informative with some comments, I suggested to add.

Dimitreee (approved):

PR #144, opened 2021-06-02
Feature/condo/sberdoma 305/make unit name field not required in ticket

AntonAL:
On create/update ticket form, previously, `unitName` field was required.
Now, when "Address" field is filled in, next field set of a form will get unlocked.


Dimitreee (approved):

dkoviazin (approved):
Ok

PR #150, opened 2021-06-07
SBERDOMA-400 ui colors pallete update

Dimitreee:


apps/condo/domains/common/components/Button.tsx
```diff
@@ -7,57 +7,69 @@ import { colors } from '../constants/style'
 
 const buttonCss = (color) => css`
   background-color: ${color[5]};
-  box-shadow: 0 3px 6px ${color[4]}, 0 6px 16px ${color[3]}, 0 9px 28px ${color[2]};
-  border: 1px solid ${color[5]};
+  border: 2px solid ${color[6]};
```
toplenboren, line 10: What do you think about making color[x] a const? 

So we would do something like: 

`border: 2px solid ${color[secondary]};`

toplenboren (approved):

dkoviazin (approved):
Ок

pahaz (approved):

toplenboren (approved):

PR #158, opened 2021-06-09
Feature/condo/sberdoma 303/phone input enhancment

Dimitreee:


apps/condo/domains/common/components/PhoneLink.tsx
```diff
@@ -8,6 +7,6 @@ interface IPhoneLink {
 
 export const PhoneLink: React.FC<IPhoneLink> = ({ value }) => (
     <a href={`tel:${normalizePhone(value)}`}>
-        {formatPhone(value)}
+        {value}
```
pahaz, line 10: why we don't need to format the phone?

pahaz (approved):

dkoviazin (approved):
Ok

apps/condo/domains/common/components/PhoneLink.tsx
Dimitreee: 30+ countries with unique phone pattern. It's easier to disable phone formatting.

PR #188, opened 2021-06-23
SBERDOMA-86 fix form validation: validate the field by the onСhange event only if an error occurred

leonid-d:


apps/condo/domains/common/components/containers/FormList.tsx
```diff
@@ -410,7 +410,19 @@ const FormWithAction: FunctionComponent<IFormWithAction> = (props) => {
         //}
     }
 
-    function handleChange (changedValues, allValues) {
+    const errors = {}
+
+    const throttledValidateFields = throttle((field) => {
```
pahaz, line 415: why throttle? and no debounce?

pahaz: https://css-tricks.com/debouncing-throttling-explained-examples/

leonid-d: `debounce` will **not** call the validator during the input process, only in the end, but the `throttle` will

PR #201, opened 2021-06-25
SBERDOMA-117 setup keystone sessionStore connection

sitozzz:


apps/condo/package.json
```diff
@@ -48,12 +48,14 @@
     "bull": "^3.22.0",
     "bull-repl": "^0.26.3",
     "classnames": "^2.2.6",
+    "connect-redis": "^6.0.0",
```
dkoviazin, line 51: Is it possible to use 'io-redis' which we already have?

dkoviazin (approved):
Ok

packages/@core.keystone/setup.utils.js
```diff
@@ -2,6 +2,10 @@ const { v5: uuidv5 } = require('uuid')
 
 const { KnexAdapter } = require('@keystonejs/adapter-knex')
 const { MongooseAdapter } = require('@keystonejs/adapter-mongoose')
+const IORedis = require('ioredis')
+const session = require('express-session')
+const { COOKIE_MAX_AGE } = require('@condo/domains/user/constants/common')
```
pahaz, line 7: you don't need to extract a const if you don't know any reusable case for this const

pahaz, line 7: You can't use const form apps in packages!

pahaz (changes requested):

packages/@core.keystone/setup.utils.js
```diff
@@ -53,17 +57,21 @@ function getAdapter (databaseUrl) {
...
+    const redisClient = new IORedis(conf.REDIS_URL)
+    const sessionStore = new RedisStore({ client: redisClient })
+
     return {
         cookieSecret: getCookieSecret(conf.COOKIE_SECRET),
         cookie: {
             sameSite: false,
             secure: false,
-            maxAge: 1000 * 60 * 60 * 24 * 130, // 130 days
+            maxAge: COOKIE_MAX_AGE,
```
pahaz, line 68: you can take it from config or leave the default value

apps/condo/package.json
```diff
@@ -48,12 +48,14 @@
...
     "date-fns": "^2.20.1",
     "dd-trace": "^0.32.0",
     "emotion": "^10.0.27",
     "emotion-server": "^10.0.27",
     "esdk-obs-nodejs": "3.21.3",
     "express": "^4.17.1",
+    "express-session": "^1.17.2",
```
pahaz, line 58: if you put the logic inside @core/keystone you should move the requirenebts inside `packages/@core.keystone/packages.json`

pahaz (approved):

PR #202, opened 2021-06-25
[WIP] SBERDOMA-87 Added notes line-generation for ticket pdf

SavelevMatthew:
amount of lines depends on height of current pdf, so it calculates it and then add lines "while it can"

### DIFF
--- a/apps/condo/domains/common/utils/pdf.ts
+++ b/apps/condo/domains/common/utils/pdf.ts
@@ -41,13 +41,19 @@ export function createPdf (options: ICreatePdfOptions) {
     const settings = format in PDF_FORMAT_SETTINGS ? PDF_FORMAT_SETTINGS[format] : PDF_FORMAT_SETTINGS['a4']
     const { pdfWidth, pdfHeight, elementOffset, firstLineOffset, lineSpace } = settings
 
+    // Calculating image ratio
+    // Width = pdfWidth - leftOffset - rightOffset = pdfWidth - elementOffset * 2
+    // Same for height
     const imageRatio = (pdfHeight - elementOffset * 2) / (pdfWidth - elementOffset * 2)
+    // Now let's define what's max css height with this ratio
     const maxElHeight = (element.clientWidth) * imageRatio
+    // And how much space in css pixels is left for lines
     let freeSpace = maxElHeight - element.clientHeight - firstLineOffset
 
     const linesContainer = element.querySelector('#pdfLineInput')
     let linesCounter = 0
 
+    // Adding lines while we have free space
     while (linesContainer && freeSpace > lineSpace / 2) {
         const marginTop = linesCounter > 0 ? lineSpace : firstLineOffset
         const line = getLine()
--- a/apps/condo/domains/common/utils/pdf.ts
+++ b/apps/condo/domains/common/utils/pdf.ts
@@ -3,12 +3,12 @@ import Jspdf from 'jspdf'
 
 const PDF_FORMAT_SETTINGS = {
     // pdfWidth - width of final image (in mm)
+    // pdfHeight - height of final image (in mm)
     // elementOffset - offset of image (in mm)
     // firstLineOffset - margin right for first line (in css pixels)
     // lineSpace - margin right for rest of lines (in css pixels)
-    // Ratio - how to calculate height (height = pdfWidth * ratio)
-    'a4': { pdfWidth: 190, elementOffset: 10, firstLineOffset: 23, lineSpace: 80, ratio: Math.sqrt(2) },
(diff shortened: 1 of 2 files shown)


apps/condo/domains/common/utils/pdf.ts
```diff
@@ -1,33 +1,71 @@
...
+
 function getPdfHeightFromElement (element: HTMLElement, expectedWidth: number) {
     const { clientWidth, clientHeight } = element
     const originalRatio = clientHeight / clientWidth
     return expectedWidth * originalRatio
 }
 
+function getLine () {
+    const line = document.createElement('div')
+    // line.style.width = 'calc(100% + 48px)'
```
toplenboren, line 22: Lets remove these comments

apps/condo/domains/common/utils/pdf.ts
```diff
@@ -1,33 +1,71 @@
 import html2canvas from 'html2canvas'
 import Jspdf from 'jspdf'
 
+const PDF_FORMAT_SETTINGS = {
+    // pdfWidth - width of final image (in mm)
+    // pdfHeight - height of final image (in mm)
+    // elementOffset - offset of image (in mm)
+    // firstLineOffset - margin right for first line (in css pixels)
+    // lineSpace - margin right for rest of lines (in css pixels)
+    'a4': { pdfWidth: 210, pdfHeight: 297, elementOffset: 10, firstLineOffset: 23, lineSpace: 80, ratio: Math.sqrt(2) },
```
toplenboren, line 10: Maybe we can move `elementOffset`, `firstLineOffset` and `lineSpace` to different constants? They seems to me equal in all cases

toplenboren: `ratio` too)

apps/condo/domains/common/utils/pdf.ts
```diff
@@ -1,33 +1,71 @@
...
+    line.style.width = '100%'
+    // line.style.margin = '0 -24px'
+    line.style.borderTop = '1px solid black'
+    return line
+}
+
 interface ICreatePdfOptions {
     element: HTMLElement
     fileName: string
+    format: string
```
toplenboren, line 32: I think we can make an enumerable or just specify that this interface accepts only `a4` or `a5`
But it's not really necessary. Wyt?

apps/condo/domains/common/utils/pdf.ts
```diff
@@ -1,33 +1,71 @@
...
 }
 
 export function createPdf (options: ICreatePdfOptions) {
```
toplenboren, line 35: I think it would be good to have a great explanation here on what goes where. Code with a lot of math equations might be hard to understand without a docstring

apps/condo/domains/common/utils/pdf.ts
SavelevMatthew: > Maybe we can move `elementOffset`, `firstLineOffset` and `lineSpace` to different constants? They seems to me equal in all cases

`elementOffset` telling us what distance to keep for "safe printing" (some printers cannot print edges correctly, so it may vary....here's just test values of 10 (mm) for both formats, but it can change for others I guess)

`firstLineOffset` and `lineSpace` tell us how far are lines from each other....right now it's proportional (that's why equal, but I'm sure it will change later)

SavelevMatthew: > `ratio` too)

already removed, now using `pdfHeight` instead to calculate the rest

apps/condo/domains/common/utils/pdf.ts
SavelevMatthew: > I think we can make an enumerable or just specify that this interface accepts only `a4` or `a5`
> But it's not really necessary. Wyt?

may be later we'll support other formats as well? like 10x15 of a6....idk

apps/condo/domains/common/utils/pdf.ts
SavelevMatthew: > I think it would be good to have a great explanation here on what goes where. Code with a lot of math equations might be hard to understand without a docstring

added few explanation comments in the most difficult parts

apps/condo/domains/common/utils/pdf.ts
SavelevMatthew: done

apps/condo/domains/common/utils/pdf.ts
toplenboren: But `ratio` is still here :(

SavelevMatthew: oops, removed redundant const

pahaz (approved):

PR #208, opened 2021-06-27
Feature/condo/SBERDOMA-484/loader component

toplenboren:
**New component:**
* Custom loader component based on Ant design `<Spin/>` 
* Component accepts all basic props from Spin as well as:
   * fill - boolean (should component fill its container?)
   * color - string (what color should spinner be?)
* Component features the delay - it will not render until 200 ms are passed to prevent flickering

**New syntax**
`<Loader {...SpinProps} color="red" fill/>`

**Changes around the project**
* Base search form now uses new Loader
* Organization Required and Auth Required uses new Loader
* Update ticket and Update property uses new Loader
* ErrorOrLoading now uses new Loader

**Known issues**
- useObjects hook return `boolean` value, but it also can return [`SpinProps`](https://ant.design/components/spin/#API) as it mentioned [here](https://ant.design/components/table/) and it seems like this is the only way to reliably customize the loader inside Ant. We can easily hack this, but I don't know yet if hacking is good solution

**Yay, screenshots!**
image

### DIFF
--- a/apps/condo/domains/common/components/Loader.tsx
+++ b/apps/condo/domains/common/components/Loader.tsx
@@ -43,5 +43,5 @@ export const Loader: React.FC<ILoaderProps> = (props) => {
 Loader.defaultProps = {
     fill: false,
     delay: DEFAULT_DELAY,
-    color: colors?.sberDefault[5],
+    color: colors.sberDefault[5],
 }
--- a/apps/condo/domains/common/components/Loader.tsx
+++ b/apps/condo/domains/common/components/Loader.tsx
@@ -17,7 +17,6 @@ const FilledLoaderContainer = styled.section`
 
 interface ILoaderProps extends SpinProps {
     fill?: boolean
-    color?: string
 }
 
 export const Loader: React.FC<ILoaderProps> = (props) => {
@@ -25,7 +24,7 @@ export const Loader: React.FC<ILoaderProps> = (props) => {
     // We need this to recolor antd spinner. It's not easily configurable from theme
     const coloredSpinnerStyles = css`
         .ant-spin-dot-item {
-            background-color: ${props.color};
+            background-color: ${colors.sberDefault[5]};
         }
     `
 
@@ -47,5 +46,4 @@ export const Loader: React.FC<ILoaderProps> = (props) => {
 Loader.defaultProps = {
     fill: false,
     delay: DEFAULT_DELAY,
-    color: colors.sberDefault[5],
 }
--- a/apps/condo/domains/common/components/Loader.tsx
+++ b/apps/condo/domains/common/components/Loader.tsx
@@ -3,8 +3,17 @@ import React from 'react'
 import { Spin, SpinProps } from 'antd'
 import { css, jsx } from '@emotion/core'
 import { colors } from '../constants/style'
+import styled from '@emotion/styled'
 
 const DEFAULT_DELAY = 200 // milliseconds
+
+const FilledLoaderContainer = styled.section`
+            display: flex;
+            justify-content: center;
+            align-items: center;
+            height: 100%;
+            width: 100%;
+        `
 
 interface ILoaderProps extends SpinProps {
     fill?: boolean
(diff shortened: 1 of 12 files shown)


apps/condo/domains/common/components/Loader.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <div css={loaderStyles}>
+            <Spin {...props}/>
+        </div>
+    )
+}
+
+Loader.defaultProps = {
+    fill: false,
+    delay: DEFAULT_DELAY,
+    color: colors?.sberDefault[5],
```
dkoviazin, line 46: Why you are not sure that colors will present here? You imported it

apps/condo/domains/property/components/PropertyForm/UpdatePropertyForm.tsx
```diff
@@ -11,6 +11,7 @@ import { FormResetButton } from '@condo/domains/common/components/FormResetButto
 import { runMutation } from '@condo/domains/common/utils/mutations.utils'
 import Modal from 'antd/lib/modal/Modal'
 import ActionBar from '@condo/domains/common/components/ActionBar'
+import { Loader } from '../../../common/components/Loader'
```
dkoviazin, line 14: Better to use paths from @condo/domains

apps/condo/domains/ticket/components/TicketForm/UpdateTicketForm.tsx
```diff
@@ -11,14 +11,14 @@ import { FormResetButton } from '@condo/domains/common/components/FormResetButto
 // @ts-ignore
 import { Ticket, TicketFile } from '@condo/domains/ticket/utils/clientSchema'
 import ActionBar from '@condo/domains/common/components/ActionBar'
+import { Loader } from '../../../common/components/Loader'
```
dkoviazin, line 14: @condo/domains

dkoviazin (commented):
Ok

apps/condo/domains/property/components/PropertyForm/UpdatePropertyForm.tsx
toplenboren: Fixed!

apps/condo/domains/ticket/components/TicketForm/UpdateTicketForm.tsx
toplenboren: Fixed!

apps/condo/domains/common/components/Loader.tsx
toplenboren: For me this was just a safe trigger. If for some reason we do not have colors.sberDefault, then the page will crash with error. Don't really want to kill the page because of the loader

dkoviazin (approved):
👍

AntonAL (approved):

apps/condo/domains/common/components/Loader.tsx
```diff
@@ -0,0 +1,47 @@
...
+const DEFAULT_DELAY = 200 // milliseconds
+
+interface ILoaderProps extends SpinProps {
+    fill?: boolean
+    color?: string
+}
+
+export const Loader: React.FC<ILoaderProps> = (props) => {
+
+    const filledContainerStyles = css`
```
pahaz, line 16: please don't use css call inside components

toplenboren: Moved the filledContainer outside of the component and greatly simplified the code

pahaz (approved):

apps/condo/domains/common/components/Loader.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <section css={coloredSpinnerStyles}>
+            <Spin {...props}/>
+        </section>
+    )
+}
+
+Loader.defaultProps = {
+    fill: false,
+    delay: DEFAULT_DELAY,
+    color: colors.sberDefault[5],
```
pahaz: it's better to remove redundant props. We really don't want to change the Loader color.

nomerdvadcatpyat (approved):

PR #209, opened 2021-06-27
HOTFIX strict version format check

pahaz:


toplenboren (approved):

dkoviazin (approved):

AntonAL (approved):

.github/workflows/deploy.prod.sh
```diff
@@ -42,7 +42,13 @@ if [[ -z "${DOCKER_COMPOSE_MIGRATION_COMMAND}" ]]; then
...
 if [[ -z "${VERSION}" ]]; then
     echo "Use: $0 <VERSION>"
-    echo "Example: $0 v1"
+    echo "Example: $0 v.1.0.0"
+    exit 1
+fi
+
+if [[ ! "${VERSION}" =~ ^v[-][0-9]+[-][0-9]+[-][0-9]+$ ]]; then
+    echo "Use: $0 <VERSION>"
+    echo "Error: wrong version format. Use something like v.1.0.0"
```
AntonAL, line 51:
```suggestion
    echo "Error: wrong version format. Use something like v-1-0-0"
```

.github/workflows/deploy.prod.sh
```diff
@@ -42,7 +42,13 @@ if [[ -z "${DOCKER_COMPOSE_MIGRATION_COMMAND}" ]]; then
 fi
 if [[ -z "${VERSION}" ]]; then
     echo "Use: $0 <VERSION>"
-    echo "Example: $0 v1"
+    echo "Example: $0 v.1.0.0"
```
AntonAL, line 45:
```suggestion
    echo "Example: $0 v-1-0-0"
```

AntonAL (changes requested):

.github/workflows/deploy.prod.sh
pahaz: really it does not matter because we use the normalization function which convert `v.1.0.0` to `v-1-0-0`

PR #214, opened 2021-06-28
SBERDOMA-499 Multiple file choose for upload

dkoviazin:
Multiple files can be chosen to upload

### DIFF
--- a/apps/condo/domains/common/components/MultipleFileUpload.tsx
+++ b/apps/condo/domains/common/components/MultipleFileUpload.tsx
@@ -30,7 +30,7 @@ const reducer = (state, action) => {
     switch (type) {
         case 'delete':
             return {
-                ...state,                    
+                ...state,
                 added: [...state.added].filter(addFile => addFile.id !== file.id),
                 deleted: [...state.deleted, file],
             }
@@ -56,7 +56,7 @@ const convertFilesToUploadFormat = (files: DBFile[]): UploadListFile[] => {
             status: 'done' as UploadFileStatus,
             url: file.publicUrl,
         }
-        return fileInList        
+        return fileInList
     })
     return uploadFiles
 }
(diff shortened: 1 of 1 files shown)


apps/condo/domains/common/components/MultipleFileUpload.tsx
```diff
@@ -164,32 +169,30 @@ const MultipleFileUpload: React.FC<IMultipleFileUploadProps> = (props) => {
...
                 const [uploadFile] = convertFilesToUploadFormat([dbFile])
-                setListFiles([...listFiles.filter(file => file.status !== 'uploading'), uploadFile])
-                onSuccess('Ok', null)
-                onFilesChange({ type: 'add', payload: dbFile })                                
+                onSuccess(uploadFile, null)
+                onFilesChange({ type: 'add', payload: dbFile })
             }).catch(err => {
                 const error = new Error(UploadFailedErrorMessage)
-                setListFiles([...listFiles.filter(file => file.status !== 'uploading'), { name: file.name, uid: file.uid, id: null, status: 'error', error: error }])
+                console.error('Upload failed', err)
```
toplenboren, line 195: I don't know about the policy of console.log messages in our project. So I don't know if this is ok for us to leave console.logs in code

Can you please point me to the source where I can read about it? 🥺

toplenboren (approved):

apps/condo/domains/common/components/MultipleFileUpload.tsx
dkoviazin: Some times we are facing different errors that are not supposed to happen. For example, when we had Firebase we were catching all (4) documented errors. But we cought the  undocumented one. And we can't reproduce it repeating user's actions and we can't ask the user to give us error from console. So, I suppose when we are dealing with third party libraries it's good to show strange errors at least in console

PR #223, opened 2021-06-30
SBERDOMA-486 add employee position column

nomerdvadcatpyat:


AntonAL:
When no position is specified, on page `/user` it states as "Не указано", but in table the value is blank.
It seems, that is should be consistent across all pages.

pahaz (approved):

pahaz (approved):

PR #224, opened 2021-06-30
SBERDOMA-490 Added citizens overview page

SavelevMatthew:
Citizens are now separated with Users (Employees)
Added citizens overview page


apps/condo/domains/contact/hooks/useTableColumns.tsx
```diff
@@ -0,0 +1,120 @@
...
+    clearFilters: () => void
+    showClearButton?: boolean
+}
+
+const FilterContainer: React.FC<IFilterContainerProps> = (props) => {
+    const intl = useIntl()
+    const ResetLabel = intl.formatMessage({ id: 'filters.Reset' })
+
+    return (
+        <div style={{ padding: 16 }}>
```
toplenboren, line 23: I think its better to use emotion's css property when dealing with native browser elements. Inserting styles is more expensive during render time then inserting css class

We can rewrite this code like so:

```
import { css } from 'emotion'

const FilterContainerDiv = css`padding: 16`
```

apps/condo/domains/contact/utils/helpers.ts
```diff
@@ -0,0 +1,148 @@
...
+
+        try {
+            if (CITIZEN_TABLE_COLUMNS.includes(columnKey) && SORT_ORDERS[sortKey]) {
+                return {
+                    columnKey,
+                    order: SORT_ORDERS[sortKey],
+                }
+            }
+        } catch (e) {
+            // TODO(Dimitreee): send error to sentry
```
toplenboren, line 107: 👀

apps/condo/domains/contact/utils/helpers.ts
```diff
@@ -0,0 +1,148 @@
+import { SortOrder } from 'antd/es/table/interface'
```
dkoviazin, line 1: Add tests for utils, like 
```bash
/domains/ticket/utils/helpers.spec.ts
```

pahaz:
We don't have citizen in our architect vocabulary %) It may confuse

pahaz:
It's better to use a `contact` word inside the code.

apps/condo/domains/contact/utils/helpers.ts
SavelevMatthew: in progress :)

SavelevMatthew:
> It's better to use a `contact` word inside the code.

Translation fixed 👍

apps/condo/domains/contact/hooks/useTableColumns.tsx
SavelevMatthew: done 👍

apps/condo/domains/contact/utils/helpers.ts
SavelevMatthew: done

PR #232, opened 2021-07-01
SBERDOMA-562 get deleted objects

toplenboren:
## Remake the soft-delete function:

* give ability to query on deleted objects 
* give ability to restore soft-deleted objects

## To query soft deleted objects:

Explicitly set where=`deltedAt_not = null`

`const count = await Property.count(client, { deletedAt_not: null })`

## To restore soft-deleted objects:

Simply post an update query and set deletedAt to `null`

`updateTestProperty(client, obj.id, { deletedAt: null })`

## Known issues:

The function is dumb :) It simply checks either we have the 'deletedAt' in any depth inside where variable
If the check is successful - remove deletedAt: null rule and return all objects
If the check is not successful - append deletedAt: null rule to return only non-deleted objects

The problem comes when you have complex queries, like:
`where = { role : { deletedAt_not: null } }`

Right now we need to explicitly specify the deletedAt: null property on **every** part of the query like so:
`where = { role : { deletedAt_not: null }, deletedAt: null }`


apps/condo/domains/property/access/Property.js
```diff
@@ -8,6 +8,7 @@ const { checkOrganizationPermission } = require('@condo/domains/organization/uti
 async function canReadProperties ({ authentication: { item: user } }) {
     if (!user) return false
     if (user.isAdmin) return {}
+    const deletedFilters = (user.isAdmin)
```
pahaz, line 11: wtf?

packages/@core.keystone/plugins/softDeleted.js
```diff
@@ -41,9 +42,6 @@ const softDeleted = ({ deletedAtField = 'deletedAt', newIdField = 'newId' } = {}
     fields[newIdField] = { ...newIdOptions }
 
     const newResolveInput = ({ resolvedData, existingItem }) => {
-        if (existingItem && existingItem[deletedAtField]) {
```
pahaz, line 44: you forgot to check double delete action

packages/@core.keystone/plugins/softDeleted.js
```diff
@@ -78,11 +76,26 @@ const softDeleted = ({ deletedAtField = 'deletedAt', newIdField = 'newId' } = {}
...
+    }
+}
+
+function applySoftDeletedFilters (access, deletedAtField, args) {
     const type = getType(access)
     if (type === 'Boolean') {
         return (access) ? { [deletedAtField]: null } : false
     } else if (type === 'Object') {
+        const currentWhereFilters = getWhereVariables(args)
+        if (Object.keys(currentWhereFilters).find((x) => x.startsWith(deletedAtField))) {
```
pahaz, line 95: you need a note for that. it's too tricky

apps/condo/domains/property/access/Property.js
toplenboren: fixed!

packages/@core.keystone/plugins/softDeleted.js
toplenboren: Yep, noted the hell out of it 🔥

packages/@core.keystone/plugins/softDeleted.js
toplenboren: I've created a test case for this lad. Check 

`user: can not soft delete object twice`

apps/condo/domains/property/schema/Property.test.js
```diff
@@ -1,3 +1,5 @@
+import { catchErrorFrom } from '../../common/utils/testSchema'
```
pahaz, line 1: domains!

packages/@core.keystone/plugins/softDeleted.js
```diff
@@ -40,13 +41,13 @@ const softDeleted = ({ deletedAtField = 'deletedAt', newIdField = 'newId' } = {}
     fields[deletedAtField] = { ...datedOptions }
     fields[newIdField] = { ...newIdOptions }
 
-    const newResolveInput = ({ resolvedData, existingItem }) => {
-        if (existingItem && existingItem[deletedAtField]) {
-            throw new Error('Already deleted')
-        }
+    const newResolveInput = ({ existingItem, resolvedData }) => {
         if (existingItem && existingItem[newIdField]) {
```
pahaz, line 45:
```suggestion
        // NOTE: we cant't change and restore an alreadt merged objects!
        if (existingItem && existingItem[newIdField]) {
```

packages/@core.keystone/plugins/softDeleted.js
pahaz: we also have a where arguments like this: `where: { OR: [ {deletedAt_not: null}, {deletedAt: null}] }`
Could you please also add a test for that

apps/condo/domains/property/schema/Property.test.js
toplenboren: Fixed

packages/@core.keystone/plugins/softDeleted.js
toplenboren: This test broke everything :) Thank you

Problem:
We are only able to check the first level of the `where {...}` object:

```
{where: { deletedAt_not: null } } // OK
{where: { property: { deletedAt_not: null } } } // NOT OK
where: { OR: [ {deletedAt_not: null}, {deletedAt: null}] } // NOT OK
```

Chosen solution: 
1. Deep check if the query contains 'deletedAt' in any place

Other possible solutions:
1. Simply check first N levels - No, because it would be super non intuitive

Implementation caveats and issues:
1. It is possible that sending really deep where queries will be long. Do not know if we should fix this on algorithm side.
2. I didn't find anything like deep check in [lodash](https://lodash.com/docs/4.17.15#isPlainObject) which we use. So I implemented the recursive search myself. Tests are here as well, but my fingers are not as trusted source as _ devs :(

dkoviazin (approved):

packages/@core.keystone/plugins/softDeleted.js
```diff
@@ -78,11 +80,98 @@ const softDeleted = ({ deletedAtField = 'deletedAt', newIdField = 'newId' } = {}
...
 }
 
-function applySoftDeletedFilters (access, deletedAtField) {
+/**
+ * Gets "where" variables from http graphQL request
+ * @param {Object} args
+ * @return {Object}
+ */
+function getWhereVariables (args) {
+    const method = get(args, ['context', 'req', 'method'])
```
pahaz, line 89:
```suggestion
    // NOTE: it's a hack! Because we don't have access to GraphQL variables inside Keystone access function!
    const method = get(args, ['context', 'req', 'method'])
```

packages/@core.keystone/plugins/softDeleted.js
```diff
@@ -97,4 +186,5 @@ function applySoftDeletedFilters (access, deletedAtField) {
 
 module.exports = {
     softDeleted,
+    queryHasSoftDeletedFieldDeep,
```
pahaz, line 189: it's better to mark it as `_` private. You expert it only for tests.

pahaz (approved):

packages/@core.keystone/plugins/softDeleted.js
```diff
@@ -78,11 +80,98 @@ const softDeleted = ({ deletedAtField = 'deletedAt', newIdField = 'newId' } = {}
...
+    return Object.keys(whereQuery).find((x) => x.startsWith(deletedAtField))
+}
+
+/**
+ * Checks if query has the soft deleted property on any level of depth
+ * @param {Object} whereQuery
+ * @param {string} deletedAtField
+ * @return {boolean}
+ */
+function queryHasSoftDeletedFieldDeep (whereQuery, deletedAtField) {
```
pahaz, line 120: Example: Filter employee by not deleted roles: `where = { role : { deletedAt: null } }` -- we don't want to remove deletedAt from Employee list

packages/@core.keystone/plugins/softDeleted.spec.js
```diff
@@ -0,0 +1,95 @@
+const softDeleted = require('./softDeleted')
+
+const DELETED_AT_FIELD = 'deletedAt'
+
+test('on empty query => false', () => {
+    expect(softDeleted.queryHasSoftDeletedFieldDeep({}, DELETED_AT_FIELD)).toEqual(false)
+})
+
+const SIMPLE_QUERY = {
```
pahaz, line 9: it's used once. It's better to write it inside the test

packages/@core.keystone/plugins/softDeleted.js
toplenboren: Yep, good case, thank you. Need to rewrite the function

toplenboren: This case is handled by explicitly setting the deletedAt field on all query parts, so I documeted it and created a tech debt task as well as todo

PR #233, opened 2021-07-01
SBERDOMA-568 Fix access rights related to the Organization and Organization Employee

toplenboren:
We had an ability to change the organization even if we were deleted from that organization

Now this is fixed ✨


toplenboren:
PS, code by @pahaz, because of pair programming

AntonAL (approved):

pahaz (approved):

PR #234, opened 2021-07-01
🔥 SBERDOMA-566 removed required validation for `phone` and `name`, don't create contact if these fields are blank

AntonAL:
Video demo:
→ https://disk.yandex.ru/i/FAkpXpLzNEoSUA

Removed required validation for "phone" and "name" fields of a ticket, being created.
When some of the field is blank, `Contact` will not be created, but ticket will.

ASAP Needs reviews and merge.


dkoviazin (approved):
Ok

PR #235, opened 2021-07-01
SBERDOMA-507 fix access to set user field

pahaz:


apps/condo/domains/ticket/access/TicketComment.js
```diff
@@ -46,11 +46,12 @@ async function canManageTicketComments ({ authentication: { item: user }, origin
 }
 
 async function canSetUserField ({ authentication: { item: user }, originalInput, addFieldValidationError }) {
+    if (!user) return false    
     if (user.isAdmin) return true
-    if (get(originalInput, ['user', 'connect', 'id']) !== user.id) {
-        return false
+    if (get(originalInput, ['user', 'connect', 'id']) === user.id) {
```
toplenboren, line 51: I think we can simply 
`return get(originalInput, ['user', 'connect', 'id']) === user.id)`

toplenboren (approved):

apps/condo/domains/ticket/access/TicketComment.js
pahaz: it seems that we have not checked all cases here and we will have to add some code here in the future. And it's easier to leave like so

PR #246, opened 2021-07-05
SBERDOMA-579 Replace side menu shadow with border

MrFoxPro:


dkoviazin (approved):

apps/condo/domains/common/components/containers/BaseLayout/components/styles.ts
```diff
@@ -60,7 +60,9 @@ export const sideMenuDesktopCss = css`
   top: 0;
   left: 0;
   z-index: 10;
-  ${shadow}
+  border-width: 0 1px 0 0;
+  border-color: ${colors.lightGrey[5]};
```
AntonAL, line 64: Color should be `#f0f0f0`, now it has value `#d9d9d9`

MrFoxPro: @AntonAL here it is color from figma
изображение

AntonAL: Hmm, it seems, that color of a border is inconsistent between different pages
[Here](https://www.figma.com/file/P827yyueCLRT6o4gyNtzY6/SberDoma-(ANT)?node-id=3905%3A64386), for example:
image

AntonAL: Waiting for answer from Maks

MrFoxPro: He said to use grey-5 color.

PR #250, opened 2021-07-05
SBERDOMA-545 fix empty units with negative floors

nomerdvadcatpyat:


apps/condo/domains/property/components/panels/Builder/MapConstructor.ts
```diff
@@ -453,10 +453,8 @@ class MapEdit extends MapView {
             const units = []
             for (let unitOnFloor = 0; unitOnFloor < unitsOnFloor; unitOnFloor++) {
                 let label = ' '
-                if (floor > 0) {
-                    label = String(unitNumber)
-                    unitNumber++
-                }
+                label = String(unitNumber)
```
dkoviazin, line 456: Why do you think that flat numbers need to be negative?
When user create property map and add underground floors - he needs to give them names

nomerdvadcatpyat: There are no negative flat numbers here. Due to this change, flats on negative floors will now have a name

dkoviazin: There will be lots of problems with underground units. As they will start to participate in automatic recalculation. They have another naming policy. You need to consult with Anna about  how to name them.

pahaz: It’s better to add prefix for underground floor units. Rather than just increment it

PR #261, opened 2021-07-06
SBERDOMA-598 fix overflowing

MrFoxPro:
изображение

### DIFF
--- a/apps/condo/domains/common/components/containers/BaseLayout/components/styles.ts
+++ b/apps/condo/domains/common/components/containers/BaseLayout/components/styles.ts
@@ -177,6 +177,22 @@ export const StyledMenuItem = styled(Menu.Item)`
     background-color: ${colors.defaultWhite[5]};
     font-weight: 700;
   }
+  > span {
+    width: 100%;
+  }
+  > span > div {
+    width: 100%;
+  }
+  .ant-space-item:nth-child(2) {
+    width: 100%;
+  }
+  .ant-space-item:nth-child(2) {
+    overflow: hidden;
+  }
+  .ant-space-item:nth-child(2) > div {
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
 `
 export const layoutCss = css`
   height: 100%;
--- a/apps/condo/domains/user/components/UserMenu.tsx
+++ b/apps/condo/domains/user/components/UserMenu.tsx
@@ -84,10 +84,10 @@ export const UserMenu = () => {
                 <StyledMenuItem onClick={goToOrganization}>
                     <Space size={16}>
                         <EnvironmentFilled style={menuIconStyles}/>
-                        <>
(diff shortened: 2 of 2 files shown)


apps/condo/domains/user/components/UserMenu.tsx
```diff
@@ -84,10 +84,10 @@ export const UserMenu = () => {
                 <StyledMenuItem onClick={goToOrganization}>
                     <Space size={16}>
                         <EnvironmentFilled style={menuIconStyles}/>
-                        <>
+                        <div title={organization.name}>
```
dkoviazin, line 87: Maybe, you can use Antd tooltips, as we are using them instead of the title attributes

https://ant.design/components/tooltip/

apps/condo/domains/common/components/containers/BaseLayout/components/styles.ts
```diff
@@ -177,6 +177,22 @@ export const StyledMenuItem = styled(Menu.Item)`
     background-color: ${colors.defaultWhite[5]};
     font-weight: 700;
   }
+  > span {
```
dkoviazin, line 180: Please, move this styles to apps/condo/domains/user/components/UserMenu.tsx  
and this is very dangerous to declare such styles without & prefix
are you sure you do not kill styles somewhere else?

AntonAL: Why dangerous? AFAIK, `> span` means `span` as a direct descendant.

apps/condo/domains/user/components/UserMenu.tsx
AntonAL: It seems to me, that tooltips will bring too much of interactivity, that will feel annoying ;)
Asked design team about possibility to make width flexible for menu and for user string in header.

pahaz: As for me it's better to use components instead of div's

pahaz (approved):

apps/condo/domains/common/components/containers/BaseLayout/components/styles.ts
dkoviazin:
```
.ant-space-item:nth-child(2) {
    width: 100%;
  }
.ant-space-item:nth-child(2) {
    overflow: hidden;
}
.ant-space-item:nth-child(2) > div {
    overflow: hidden;
    text-overflow: ellipsis;
}
```

apps/condo/domains/user/components/UserMenu.tsx
AntonAL: So, latest news from design — remove organisation menu item
The menu should have only one item — "logout".

MrFoxPro:
Desing is not actual anymore.

PR #262, opened 2021-07-06
[WIP] SBERDOMA-617 Exporting contacts as Excel sheet

SavelevMatthew:
Now can export contacts as Excel sheet.
Width of columns will be generated automatically


AntonAL:
Checked on my machine, — exported file is blank.
Having one contact displayed in table, I've pressed "Export" button, file has been downloaded and I don't see any contacts.
image

AntonAL:
It will be more convenient, if name of an exported file will contain timestamp and count of exported contacts.
For example:
`57_exported_contacts_202107061857.xlsx`

AntonAL:
Also, exported file contains redundant column in screenshot above.

AntonAL:
I also would name a tab in Excel-file to something meningful, like "Contacts", with translation, of course.

PR #263, opened 2021-07-06
SBERDOMA-620 Removed ability to redirect to contact page

SavelevMatthew:
For now contact profile page is not implemented, but we can redirect there from by clicking on rows from '/contact' page.
We need to disable this until future implementation, so:
1. Table rows were wrapped with tooltip as @dkoviazin asked
2. Disabled ability to be forwarded by clicking on table row


toplenboren (approved):

dkoviazin (approved):
Ok

PR #265, opened 2021-07-07
Feature/condo/sberdoma 415/validate map units

nomerdvadcatpyat:
we need to validate the names of the apartments for uniqueness. did it in this pull request
Снимок экрана 2021-07-07 в 14 18 24

### DIFF
--- a/apps/condo/domains/property/components/BasePropertyForm/index.tsx
+++ b/apps/condo/domains/property/components/BasePropertyForm/index.tsx
@@ -134,15 +134,15 @@ const BasePropertyForm: React.FC<IPropertyFormProps> = (props) => {
                                         rules={[
                                             {
                                                 validator (rule, value) {
-                                                    const unitLabels = get(value, 'sections', [])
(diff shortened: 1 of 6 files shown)


apps/condo/domains/property/components/BasePropertyForm/index.tsx
```diff
@@ -125,15 +128,43 @@ const BasePropertyForm: React.FC<IPropertyFormProps> = (props) => {
...
+                                                },
+                                            },
+                                        ]}
+                                    >
                                         <Input />
                                     </Form.Item>
-                                    <Form.Item shouldUpdate={true}>
+                                    <Form.Item
+                                        shouldUpdate={true}
+                                        onFocus={() => setMapValidationError(null)}
```
toplenboren, line 160: Do we really need to validate the input on focus? I think it would better to validate this on blur. Wyt?

dkoviazin (approved):

pahaz (approved):

PR #267, opened 2021-07-07
SBERDOMA-529 Removed redundant message.success

SavelevMatthew:
We were displaying extra success message, when file "uploaded to process physically", even if data incorrect.
It was unclear for user to see "success" and "error" message at the same time, so
I removed "success on load" message and left only "success on proccessed"


toplenboren (approved):

toplenboren:
Thank you for the adequate description

sitozzz (approved):
ok

PR #275, opened 2021-07-07
[WIP] SBERDOMA-547 Fixed ability to type letter in InputNumber fields

SavelevMatthew:
Before it was possible to press keys and see something like "3asdasd" in InputNumber fields, even if it wasn't possible to select it
Now it's impossible to even type something other than numbers 👍 
Also setting min for "Units on floor" to 1, since it seems like property with less units than 1 is incorrect


dkoviazin (approved):

SavelevMatthew:
Seems like current version works only in Chrome (and maybe mobile browsers?)

sitozzz (approved):

PR #277, opened 2021-07-07
SBERDOMA-478 init throwAuthenticationError helper

pahaz:
For mobile clients, it's important to separate access rights errors from non-authorization errors.
There is a proposal of `throwAuthenticationError` function for access check functions!


dkoviazin (approved):

apps/condo/domains/common/utils/apolloErrorFormatter.js
```diff
@@ -145,8 +147,15 @@ const formatError = error => {
     return safeFormatError(error, IS_HIDE_INTERNALS)
 }
 
+// NEW GraphQL Error standard
+
+function throwAuthenticationError () {
+    throw new AuthenticationError('No or incorrect authentication credentials')
```
toplenboren, line 153: Just a question - how the mobile client is going to handle the errors? Is it OK for us to send them without translation? 

If this is ok and error is handled inside mobile app, then I think we should move these constants to some secure place, like errorConsts.js

toplenboren (approved):

apps/condo/domains/common/utils/apolloErrorFormatter.js
pahaz: Yes, currently we don't have an error translation. At the moment we have an error name.
We will solve it in the future.

apps/condo/domains/common/utils/testSchema/index.js
```diff
@@ -65,6 +65,10 @@ export const expectToThrowAccessDeniedErrorToObj = async (testFunc) => {
 }
 
 /**
+ * @deprecated
+ * THIS FUNCTION IS DEPRECATED. We want to to separate access rights errors from non-authorization errors!
+ * Please use expectToThrowAccessDeniedErrorToObjects!
+ *
```
AntonAL, line 71: Function `expectToThrowAccessDeniedErrorToObjects` should not be deprecated, because it tests use cases, when access rights of **authenticated** user agents are restricted.

PR #278, opened 2021-07-08
SBERDOMA-667 Nulls in ticket table

SavelevMatthew:
Before it was possible to see "null" in ticket overview table in this situations:
1. Unit was not specified
2. When client column was filtered by global search
Now it's fixed 👍


sitozzz (approved):
Seems ok

dkoviazin (approved):

PR #281, opened 2021-07-08
SBERDOMA-532 display max size of uploaded file when catch error

nomerdvadcatpyat:
Снимок экрана 2021-07-08 в 12 58 41


SavelevMatthew (approved):
ok 👍

dkoviazin (approved):

PR #282, opened 2021-07-08
SBERDOMA-685 Added contact email normalization + validation

SavelevMatthew:
Before we just moved email to lower case and this was it.
For now we accept only emptiness or emails that can be normalized.
And we are also changing their domains that can be duplicated (ya.ru -> yandex.ru, hotmail.co.uk -> outlook.com)


dkoviazin (approved):

sitozzz (approved):
Seems ok

PR #302, opened 2021-07-12
SBERDOMA-727 Add "unknown property" label

MrFoxPro:


nomerdvadcatpyat (approved):

AntonAL:
Blank value indicator should be visually distinguishable from real values.
I propose to use just "—" with Tooltip, if we need to annotate with more detail.
Discuss it with Maks, please.
image

apps/condo/domains/ticket/hooks/useTableColumns.tsx
```diff
@@ -196,7 +197,7 @@ export const useTableColumns = (sort: Array<string>, filters: IFilters,
                 render: (record) => {
                     const unitName = get(record, 'unitName')
                     const property = get(record, 'property')
-                    const text = get(property, 'address')
+                    const text = get(property, 'address', NoPropertyMessage)
```
AntonAL, line 200: When address is a blank string `''`, `NoPropertyMessage` is not assigned and the table cell is blank.
image

AntonAL (changes requested):

PR #304, opened 2021-07-12
SBERDOMA-715 add transparent background to header

nomerdvadcatpyat:
Before:
Снимок экрана 2021-07-12 в 16 50 39
After:
Снимок экрана 2021-07-12 в 16 50 26

### DIFF
--- a/apps/condo/domains/user/components/containers/AuthLayout.tsx
+++ b/apps/condo/domains/user/components/containers/AuthLayout.tsx
@@ -36,7 +36,7 @@ export interface AuthPage extends React.FC {
 
 const FOOTER_LINK_STYLE = { color: colors.sberPrimary[7] }
 const AUTH_PAGES_STYLE = { width: '100%', maxWidth: '500px', paddingLeft: '20px', paddingRight: '20px' }
-const HEADER_STYLE = { background: colors.white, padding: '20px', margin: '0px', width: '100%' }
+const HEADER_STYLE = { background: 'transparent', padding: '20px', margin: '0px', width: '100%' }
 const FOOTER_STYLE = { color: colors.lightGrey[7], backgroundColor: colors.white, fontSize: '12px', lineHeight: '20px',  padding: '20px' }
 
 const PageContent: React.FC = ({ children }) => {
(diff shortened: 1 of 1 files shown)


Dimitreee (approved):

AntonAL:
It seems, like solution of the problem is in a wrong way.
The problem is in overlapping of a content part (what we see as clipped h1) with header.
It is a problem of layout, not background.
When I shrink browser window, header goes out of document scroll area.
image

Correct solution is to prevent overlapping between header and content area.

apps/condo/domains/user/components/containers/AuthLayout.tsx
```diff
@@ -36,7 +36,7 @@ export interface AuthPage extends React.FC {
 
 const FOOTER_LINK_STYLE = { color: colors.sberPrimary[7] }
 const AUTH_PAGES_STYLE = { width: '100%', maxWidth: '500px', paddingLeft: '20px', paddingRight: '20px' }
-const HEADER_STYLE = { background: colors.white, padding: '20px', margin: '0px', width: '100%' }
+const HEADER_STYLE = { background: 'transparent', padding: '20px', margin: '0px', width: '100%' }
```
AntonAL, line 39: It seems, like solution of the problem is in a wrong way.
The problem is in overlapping of a content part (what we see as clipped h1) with header.
It is a problem of layout, not background.
When I shrink browser window, header goes out of document scroll area.
image

Correct solution is to prevent overlapping between header and content area.

AntonAL (changes requested):

dkoviazin (approved):

apps/condo/domains/user/components/containers/AuthLayout.tsx
nomerdvadcatpyat: maybe it will be correct in this case to remove fixed position of AuthLayout header?

pahaz (approved):

toplenboren (approved):

AntonAL:
The problem is still not solved.
Look at screenshot
image

Content of `.ant-layout-content` get's centered vertically and in case of low browser window height, it gets clipped because of `align-items: center` property. You should align it to top, when there is not enough height to fit content.

pahaz (approved):

AntonAL (approved):
Now the layout works correctly!
👍

PR #308, opened 2021-07-13
SBERDOMA-730 change placeholder color

nomerdvadcatpyat:
**Before:** 
изображение
**After:**
Снимок экрана 2021-07-13 в 14 59 49

### DIFF
--- a/apps/condo/domains/common/components/containers/GlobalStyle.tsx
+++ b/apps/condo/domains/common/components/containers/GlobalStyle.tsx
@@ -152,6 +152,14 @@ export default function GlobalStyle () {
                   cursor: pointer;
                 }
               }
+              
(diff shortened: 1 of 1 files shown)


apps/condo/domains/common/components/containers/GlobalStyle.tsx
```diff
@@ -152,6 +152,14 @@ export default function GlobalStyle () {
                   cursor: pointer;
                 }
               }
+              
+              .ant-input-affix-wrapper > input.ant-input, .ant-form-item-has-error .ant-input-affix-wrapper > input.ant-input {
```
dkoviazin, line 156: demo.dev.doma.ai and master branch already has a good looking input only  v1 has black text placeholder
Do we need these changes?

PR #317, opened 2021-07-14
SBERDOMA-63 add gql query generation

MrFoxPro:


MrFoxPro:
Is it good gql query?
```
const MANAGE_EXAMPLE_BY_ID_QUERY = gql`
    query getManageExampleById ($data: ManageExampleInput!) {
        obj: ManageExample(where: {id: $id}) ${COMMON_FIELDS}
    }
`
```

dkoviazin (commented):
There will be no common fields there as it's not storing in the database.
It will look like a mutation.
The only reason we are not using mutations everywhere is that mutations mean that data will be somehow modified but queries mean that data will be fetched in some complex way. But in syntaxis, there will be no differences

Dimitreee (approved):

dkoviazin (approved):

pahaz:
We already have getAll util. What is really purpose for this query?

PR #320, opened 2021-07-15
SBERDOMA-766 fixed contact is not created in ticket create page

AntonAL:


dkoviazin (approved):

sitozzz (approved):

apps/condo/domains/contact/components/ContactsEditor/index.tsx
```diff
@@ -159,7 +159,7 @@ export const ContactsEditor: React.FC<IContactEditorProps> = (props) => {
             [fields.phone]: contact.phone,
         })
         setValue(contact)
-        const isNew = !has(contact, 'id')
+        const isNew = !contact.id
```
Dimitreee, line 162: What is this hotfix about?

AntonAL: It's about checking, whether `contact` is saved in database (has `id` value).
This line checks presence of `contact.id` property, but not value.
Therefore, when `contact.id` is `undefined`, `isNew` gets `false` and contact will not be created.

PR #327, opened 2021-07-16
SBERDOMA-776 documented solution of error by running makemigrations

AntonAL:
Anyone, having following error by running `makemigrations`:

> UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 3131: invalid start byte

Do the following steps, described in docs.

### DIFF
--- a/apps/condo/docs/migrations.md
+++ b/apps/condo/docs/migrations.md
@@ -109,7 +109,7 @@ It's better to roll back your migration and recreate it. But sometimes you shoul
 docker-compose run app python apps/condo/.kmigrator/manage.py makemigrations --merge
 ```
 
-If the util ask You to provide a default value, it expects it in Python syntax.
+If the util asks You to provide a default value, it expects it in Python 3 syntax.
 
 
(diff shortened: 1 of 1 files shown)


dkoviazin (approved):

apps/condo/docs/migrations.md
```diff
@@ -109,4 +109,27 @@ It's better to roll back your migration and recreate it. But sometimes you shoul
 docker-compose run app python apps/condo/.kmigrator/manage.py makemigrations --merge
 ```
 
-If the util ask You to provide a default value, it expects it in Python syntax.
\ No newline at end of file
+If the util ask You to provide a default value, it expects it in Python syntax.
```
toplenboren, line 112:
```suggestion
If the util asks to provide a default value, it expects it in Python 3 syntax.
```

toplenboren (approved):
😃

apps/condo/docs/migrations.md
AntonAL: Fixed in `0de86d28`

PR #368, opened 2021-07-27
SBERDOMA-666 Create custom middleware for large JSON bodysizes

toplenboren:
## Problem:
When chessboard is too large - the system fails to process it. Keystone.js does not offer a trivial and reliable way to customize it

## Solution
Create a custom middleware in which we explicitly set maximum body size to 100Mb


dkoviazin (approved):

SavelevMatthew (approved):
LGTM :)

PR #370, opened 2021-07-27
SBERDOMA-874 add .vscode to .gitignore

MrFoxPro:


SavelevMatthew (changes requested):
I think you should also add more lines for VSCode users, based on this (https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) solution
(Idk, why is "!" rules included...)

MrFoxPro:
> I think you should also add more lines for VSCode users, based on this (https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) solution
> (Idk, why is "!" rules included...)

Do we really need launch.json and other settings? It is worth to add it after someone finds solution how to configure vscode properly for debugging
Also `.history` folder is related to vs code extension, not to vscode itself

AntonAL (approved):

SavelevMatthew:
> > I think you should also add more lines for VSCode users, based on this (https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore) solution
> > (Idk, why is "!" rules included...)
> 
> Do we really need launch.json and other settings? It is worth to add it after someone finds solution how to configure vscode properly for debugging
> Also `.history` folder is related to vs code extension, not to vscode itself

I'm not VSCode user, just thought something like "*.code-workspace" and ".history/" will help someone with VSCode extensions, if your changes is enough, than ok 👍

SavelevMatthew (approved):

PR #389, opened 2021-07-30
SBERDOMA-905 Common validations for frontend forms

SavelevMatthew:
Moved common form validations in useValidations hook...
I'll add specific confluence page later 👍

### DIFF
--- a/apps/condo/domains/common/hooks/useValidations.ts
+++ b/apps/condo/domains/common/hooks/useValidations.ts
@@ -10,6 +10,7 @@ type ValidatorHookReturnType = {
     requiredValidator: Rule
     phoneValidator: Rule
     emailValidator: Rule
+    trimValidator: Rule
 }
 
 const combiner = (...rules: CombinerParams) => {
@@ -33,6 +34,7 @@ export const useValidations: () => ValidatorHookReturnType = () => {
 
     const phoneValidator: Rule = {
         validator: (_, value) => {
+            if (!value) return Promise.resolve()
             const v = normalizePhone(value)
             if (!v) return Promise.reject(PhoneIsNotValidMessage)
             return Promise.resolve()
@@ -44,11 +46,19 @@ export const useValidations: () => ValidatorHookReturnType = () => {
         message: EmailErrorMessage,
     }
 
+    const trimValidator: Rule = {
+        validator: (_, value) => {
+            if (!value || value.trim().length === 0) return Promise.reject(ThisFieldIsRequiredMessage)
+            return Promise.resolve()
+        },
+    }
+
     return {
         combiner,
         messageChanger,
         requiredValidator,
         phoneValidator,
         emailValidator,
+        trimValidator,
     }
 }
--- a/apps/condo/domains/common/hooks/useValidations.ts
+++ b/apps/condo/domains/common/hooks/useValidations.ts
@@ -2,30 +2,25 @@ import { useIntl } from '@core/next/intl'
 import { Rule } from 'rc-field-form/lib/interface'
 const { normalizePhone } = require('@condo/domains/common/utils/phone')
 
-type CombinerParams = Array<Rule | Array<Rule>>
-
-type ValidatorHookReturnType = {
-    combiner: (...rules: CombinerParams) => Rule[]
-    messageChanger: (rule: Rule, message: string) => Rule
+type ValidatorTypes = {
+    changeMessage: (rule: Rule, message: string) => Rule
     requiredValidator: Rule
     phoneValidator: Rule
     emailValidator: Rule
     trimValidator: Rule
+    minLengthValidator: (length: number) => Rule
 }
 
-const combiner = (...rules: CombinerParams) => {
-    return rules.map(rulesPart => Array.isArray(rulesPart) ? rulesPart : [rulesPart]).flat()
-}
-
-const messageChanger = (rule: Rule, message: string) => {
+const changeMessage = (rule: Rule, message: string) => {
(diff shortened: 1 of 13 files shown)


apps/condo/domains/common/hooks/useValidations.ts
```diff
@@ -0,0 +1,64 @@
...
+    const emailValidator: Rule = {
+        type: 'email',
+        message: EmailErrorMessage,
+    }
+
+    const trimValidator: Rule = {
+        validator: (_, value) => {
+            if (!value || value.trim().length === 0) return Promise.reject(ThisFieldIsRequiredMessage)
+            return Promise.resolve()
+        },
```
MrFoxPro, line 53: Maybe it is worth to make function `validator` async and return pure value, so you wouldn't wrap all in `Promise.resolve()` and `Promise.reject()`
E.g.:
```ts
    const phoneValidator: Rule = {
        validator: async (_, value) => {
            if (!value) return
            const v = normalizePhone(value)
            if (!v) throw new Error(PhoneIsNotValidMessage)
        },
    }
```

apps/condo/domains/common/hooks/useValidations.ts
```diff
@@ -0,0 +1,64 @@
...
+type ValidatorHookReturnType = {
+    combiner: (...rules: CombinerParams) => Rule[]
+    messageChanger: (rule: Rule, message: string) => Rule
+    requiredValidator: Rule
+    phoneValidator: Rule
+    emailValidator: Rule
+    trimValidator: Rule
+}
+
+const combiner = (...rules: CombinerParams) => {
```
Dimitreee, line 16: combine

apps/condo/domains/common/hooks/useValidations.ts
```diff
@@ -0,0 +1,64 @@
...
+    return rules.map(rulesPart => Array.isArray(rulesPart) ? rulesPart : [rulesPart]).flat()
+}
+
+const messageChanger = (rule: Rule, message: string) => {
```
Dimitreee, line 20: changeMessage

apps/condo/domains/common/hooks/useValidations.ts
Dimitreee: We need to return rejected Promise from validators. We will include promise in any case.

apps/condo/domains/common/hooks/useValidations.ts
```diff
@@ -0,0 +1,64 @@
+import { useIntl } from '@core/next/intl'
+import { Rule } from 'rc-field-form/lib/interface'
+const { normalizePhone } = require('@condo/domains/common/utils/phone')
+
+type CombinerParams = Array<Rule | Array<Rule>>
+
+type ValidatorHookReturnType = {
```
Dimitreee, line 7: ValidationsType

apps/condo/domains/property/components/BasePropertyForm/index.tsx
```diff
@@ -67,6 +66,19 @@ const BasePropertyForm: React.FC<IPropertyFormProps> = (props) => {
         return formData
     }, [initialValues])
 
+    const { combiner, requiredValidator } = useValidations()
+    const addressValidator = {
```
Dimitreee, line 70: Why didnt we add this validator in useValidations?

apps/condo/pages/auth/change-password.tsx
```diff
@@ -133,10 +130,7 @@ const ChangePasswordPage: AuthPage = () => {
                     style={{ marginTop: '40px' }}
                     dependencies={['password']}
                     rules={[
-                        {
-                            required: true,
-                            message: PleaseConfirmYourPasswordMsg,
-                        },
+                        messageChanger(requiredValidator, PleaseConfirmYourPasswordMsg),
```
Dimitreee, line 133: Add declaration to validations as confirmPassword

apps/condo/domains/common/hooks/useValidations.ts
Dimitreee: Why should we introduce this combiner function if as the result we will get an array of validators? As i can imagine we can use validators like: 

`
const { combiner, messageChanger, phoneValidator, emailValidator, requiredValidator } = useValidations()
const validators = {
    phone: [requiredValidator, phoneValidator, alreadyRegisteredPhoneValidator],
}
`

SavelevMatthew: Idea was that common rules can be "Rule" or "Rule[]", so, combine was good idea in my head, it will make you not think about lists deconstruction...so, should we left it?

apps/condo/domains/property/components/BasePropertyForm/index.tsx
SavelevMatthew: since it's page-related and uses states inside

apps/condo/domains/common/hooks/useValidations.ts
SavelevMatthew: resolved 👍

apps/condo/domains/common/hooks/useValidations.ts
SavelevMatthew: resolved 👍

apps/condo/pages/auth/change-password.tsx
SavelevMatthew: resolved 👍

Dimitreee (approved):

sitozzz (approved):

PR #393, opened 2021-08-01
SBERDOMA-915 add /create_release_report command, DRY, KISS

Dimitreee:


apps/notification_bot/utils.ts
```diff
@@ -1,8 +1,95 @@
...
+                        commitsList.push(commit)
+                        isLastCommitFound = true
+                    }
+                }
+            })
+
+            page++
+        }
+    } else {
+        while (!isLastCommitFound) {
```
dkoviazin, line 78: Better to make some limitations to "while" when you are working with third parties libraries

apps/notification_bot/utils.ts
```diff
@@ -1,8 +1,95 @@
...
+        })
+
+    return Array.from(new Set<string>(formattedCommit))
+        .sort((leftCommit, rightCommit) => {
+            const [, leftCommitNumber] = leftCommit.split('-')
+            const [, rightCommitNumber] = rightCommit.split('-')
+
+            return Number(leftCommitNumber) > Number(rightCommitNumber) ? 1 : -1
+        }).map((commit) => {
+            return `https://doma.atlassian.net/browse/${commit}`
```
dkoviazin, line 35: May be move to constant

dkoviazin (approved):

apps/notification_bot/utils.ts
```diff
@@ -1,8 +1,95 @@
...
+        })
+}
+
+const fetchCommitsList = async (page: number, per_page = 100) => {
+    return await octokit.request('GET /repos/{owner}/{repo}/commits', {
+        owner: 'open-condo-software',
+        repo: 'condo',
```
toplenboren, line 42: Why not move to const?

toplenboren (approved):

apps/notification_bot/utils.ts
Dimitreee: i'm not sure we need it here

apps/notification_bot/utils.ts
Dimitreee: i'm not sure we need it here

apps/notification_bot/utils.ts
```diff
@@ -1,8 +1,102 @@
...
+                    isFirstCommitFound = true
+                }
+
+                if (isFirstCommitFound && !isLastCommitFound) {
+                    if (sha !== lastCommitSha) {
+                        commitsList.push(commit)
+                    } else {
+                        commitsList.push(commit)
+                        isLastCommitFound = true
+                    }
```
AntonAL, line 74:
```suggestion
                    if (sha === lastCommitSha) {
                        isLastCommitFound = true
                    }
                    commitsList.push(commit)
```

AntonAL (changes requested):

apps/notification_bot/utils.ts
```diff
@@ -1,8 +1,102 @@
...
+    const formattedCommit = commits
+        .map(({ message }) => {
+            const [task_number] = message.split(' ')
+            return task_number
+        })
+        .filter((task_number) => {
+            const [taskType] = task_number.split('-')
+
+            return AVAILABLE_TASK_TYPES.includes(taskType)
+        })
```
AntonAL, line 25: It seems, that more configurable approach is to use `reduce` instead of `map` with `TASK_NUMBER_REGEXP` to capture numbers instead of using `AVAILABLE_TASK_TYPES`. This way we will have one pass over array instead of two passes (with `map` and `filter`) and will not be tied to `split` and `-` char in algorithm.

apps/notification_bot/utils.ts
AntonAL: I think, better place for configuration is `config` node in `package.json`

apps/notification_bot/utils.ts
Dimitreee: :)

PR #395, opened 2021-08-01
SBERDOMA-889 fix suggest 

leonid-d:
bonus: fix affix layout :)

### DIFF
--- a/apps/condo/domains/common/components/ActionBar.tsx
+++ b/apps/condo/domains/common/components/ActionBar.tsx
@@ -7,7 +7,7 @@ import { css, jsx } from '@emotion/core'
 
 const actionBar = css`
     position: relative;
-    padding: 24px 24px 32px 24px;
+    padding: 24px;
     margin: 0 -24px;
     transition: box-shadow 0.6s ease-out;
 
@@ -25,7 +25,7 @@ const ActionBar: React.FC<IActionBarProps> = ({ children, fullscreen = false }) 
     return (
         <Affix offsetBottom={24} style={{ width: fullscreen ? '100%' : 'unset' }}>
             <div css={actionBar}>
-                <Space wrap={true} size={[40, 24]}>
+                <Space wrap={true} size={[40, 24]} style={{ marginBottom: '0 !important' }}>
                     { children }
                 </Space>
             </div>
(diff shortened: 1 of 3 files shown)


apps/condo/domains/common/components/ActionBar.tsx
```diff
@@ -24,7 +24,7 @@ const ActionBar: React.FC<IActionBarProps> = ({ children, fullscreen = false })
     return (
         <Affix offsetBottom={24} style={{ width: fullscreen ? '100%' : 'unset' }}>
             <div css={actionBar}>
-                <Space wrap={true} size={[40, 24]}>
+                <Space wrap={true} size={[40, 24]} style={{ marginBottom: '0 !important' }}>
```
Dimitreee, line 27: Bad practise. We shouldnt use !important here. Use common ant'd components and props.

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -65,6 +65,11 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
     function handleSelect (value, option) {
         setSelected(option.children)
 
+        if (props.mode === 'multiple') {
```
Dimitreee, line 68: What is this fix about? Please describe your solution

Dimitreee (changes requested):

apps/condo/domains/common/components/GraphQlSearchInput.tsx
leonid-d: In shareTIcketModal, we have a multiselect for user choice, in which we need to clear the search field after selecting the user with whom we want to share the ticket.

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -65,6 +65,11 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
+            setValue('')
+            handleSearch('')
```
AntonAL, line 70: It triggers extra query on select item.
```suggestion
```
Can `autoClearSearchValue` property of `Select` be used in your case?
https://ant.design/components/select/
I've tried locally, it seems to be working.

AntonAL (approved):

Dimitreee (approved):

PR #398, opened 2021-08-02
SBERDOMA-912 - add ability to select multiple property addresses

sitozzz:
example

### DIFF
--- a/apps/condo/domains/common/components/GraphQlSearchInput.tsx
+++ b/apps/condo/domains/common/components/GraphQlSearchInput.tsx
@@ -6,7 +6,7 @@ import { Select, SelectProps } from 'antd'
 import { ApolloClient } from '@apollo/client'
 
 // TODO: add apollo cache shape typings
-interface ISearchInputProps extends SelectProps {
+interface ISearchInputProps extends SelectProps<string> {
     search: (client: ApolloClient<Record<string, unknown>>, queryArguments: string) => Promise<Array<Record<string, unknown>>>
     onSelect?: (...args: Array<unknown>) => void
     onChange?: (...args: Array<unknown>) => void
@@ -20,7 +20,6 @@ interface ISearchInputProps extends SelectProps {
     autoFocus?: boolean
     initialValue?: string
     formatLabel?: (option: { value: string, text: string }) => JSX.Element
-    maxTagCount?: 'responsive' | number
 }
 
 export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
--- a/apps/condo/domains/common/components/GraphQlSearchInput.tsx
+++ b/apps/condo/domains/common/components/GraphQlSearchInput.tsx
@@ -20,6 +20,7 @@ interface ISearchInputProps extends SelectProps {
     autoFocus?: boolean
     initialValue?: string
     formatLabel?: (option: { value: string, text: string }) => JSX.Element
+    maxTagCount?: 'responsive' | number
 }
 
 export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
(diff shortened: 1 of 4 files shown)


dkoviazin (approved):

apps/condo/pages/analytics/index.tsx
```diff
@@ -397,13 +412,14 @@ const TicketAnalyticsPageFilter: React.FC<ITicketAnalyticsPageFilterProps> = ({
...
                     <Form.Item label={AddressTitle} {...FORM_ITEM_STYLE}>
-                        {/* TODO (sitozzz): change to multiple mode */}
-                        <PropertyAddressSearchInput
+                        <GraphQlSearchInput
+                            allowClear
+                            search={searchAddress}
+                            mode={'multiple'}
+                            value={addressList}
+                            onChange={onAddressChange}
+                            maxTagCount={'responsive' as const}
```
Dimitreee, line 421: Why did you use type cast here?

apps/condo/pages/analytics/index.tsx
```diff
@@ -317,9 +318,12 @@ const TicketAnalyticsPageFilter: React.FC<ITicketAnalyticsPageFilterProps> = ({
     const PresetQuarter = intl.formatMessage({ id: 'pages.condo.analytics.TicketAnalyticsPage.Filter.PeriodPreset.Quarter' })
     const PresetYear = intl.formatMessage({ id: 'pages.condo.analytics.TicketAnalyticsPage.Filter.PeriodPreset.Year' })
 
+    const userOrganization = useOrganization()
+    const userOrganizationId = get(userOrganization, ['organization', 'id'])
     const [dateRange, setDateRange] = useState<[Moment, Moment]>([moment().subtract(1, 'week'), moment()])
     const [dateRangePreset, setDateRangePreset] = useState<null | string>(null)
     const [addressList, setAddressList] = useState([])
+    const addressListRef = useRef([])
```
Dimitreee, line 326: Why did you use Ref, not state Here?

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -20,6 +20,7 @@ interface ISearchInputProps extends SelectProps {
     autoFocus?: boolean
     initialValue?: string
     formatLabel?: (option: { value: string, text: string }) => JSX.Element
+    maxTagCount?: 'responsive' | number
```
Dimitreee, line 23: SelectProps interface did not provide maxTagCount typedefs?

apps/condo/pages/analytics/index.tsx
sitozzz: So it is written in the antd documentation https://ant.design/components/select/#components-select-demo-responsive

apps/condo/domains/common/components/GraphQlSearchInput.tsx
sitozzz: Yes, antdesign type `SelectProps` contain that property, but `GraphQlSearchInput` component overrides all default props, so we have error with unknown prop type

apps/condo/pages/analytics/index.tsx
sitozzz: I used ref here because I don't need to subscribe on that variable changes. In that case I need to add additional variable for store full address info. This info will be used if user reopen page with same url

apps/condo/domains/common/components/GraphQlSearchInput.tsx
sitozzz: @Dimitreee ok, so I found solution to solve this. Please review again

Dimitreee (approved):

PR #406, opened 2021-08-03
SBERDOMA-751 Information about lock time remain

dkoviazin:
image

image

### DIFF
--- a/apps/condo/pages/auth/register.tsx
+++ b/apps/condo/pages/auth/register.tsx
@@ -2,7 +2,7 @@ import { useIntl } from '@core/next/intl'
 import { MaskedInput } from 'antd-mask-input'
 import { FormattedMessage } from 'react-intl'
 import Router, { useRouter } from 'next/router'
-import { Form, Input, Typography } from 'antd'
+import { Form, Input, Typography, notification } from 'antd'
 import { Button } from '@condo/domains/common/components/Button'
 import AuthLayout, { AuthLayoutContext, AuthPage } from '@condo/domains/user/components/containers/AuthLayout'
 import React, { createContext, useEffect, useState, useContext, useCallback, Dispatch, SetStateAction } from 'react'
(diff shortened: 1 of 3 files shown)


apps/condo/pages/auth/register.tsx
```diff
@@ -253,10 +247,21 @@ const InputPhoneForm = ({ onFinish }): React.ReactElement<IInputPhoneFormProps>
...
+                const smsTimeoutError = error.graphQLErrors.find((err) => (err.name.indexOf(TOO_MANY_REQUESTS) === 0))
+                if (smsTimeoutError) {
+                    const { data: { timeRemain = 0 } } = smsTimeoutError
+                    notification.error({
+                        message: SMSTooManyRequestsError,
+                        description: intl.formatMessage({ id: 'pages.auth.register.error.smstimeout' }, { timeRemain }),
+                    })
+                }
+            }
             setIsLoading(false)
```
AntonAL, line 265: If another kind of error occurs on server, nothing will be displayed to user, because you have suppressed default errors. Tested on my machine by adding `throw new Error('test')` into resolver of `startConfirmPhoneAction` mutation.

apps/condo/pages/auth/register.tsx
```diff
@@ -385,9 +386,18 @@ const ValidatePhoneForm = ({ onFinish, onReset }): React.ReactElement<IValidateP
...
+            if (has(error, 'graphQLErrors')) {
+                const smsTimeoutError = error.graphQLErrors.find((err) => (err.name.indexOf(TOO_MANY_REQUESTS) === 0))
+                if (smsTimeoutError) {
+                    const { data: { timeRemain = 0 } } = smsTimeoutError
+                    notification.error({
+                        message: SMSTooManyRequestsError,
+                        description: intl.formatMessage({ id: 'pages.auth.register.error.smstimeout' }, { timeRemain }),
+                    })
+                }
+            }
```
AntonAL, line 400: If another kind of error occurs on server, nothing will be displayed to user, because you have suppressed default errors.

apps/condo/pages/auth/register.tsx
Dimitreee: We don't need to react at every I'm unknown server error.

Dimitreee (approved):

pahaz (approved):

apps/condo/pages/auth/register.tsx
pahaz: it's better to give feedback about errors. rather than just hide it

PR #409, opened 2021-08-04
SBERDOMA-942 - PDF generation with pagebreaks

sitozzz:
Add helper method for creating infinite PDF document

image

### DIFF
--- a/apps/condo/domains/common/utils/pdf.ts
+++ b/apps/condo/domains/common/utils/pdf.ts
@@ -1,5 +1,6 @@
 import html2canvas from 'html2canvas'
 import Jspdf from 'jspdf'
+import html2pdf from 'html2pdf.js'
 
 const PDF_FORMAT_SETTINGS = {
     // pdfWidth - width of final image (in mm)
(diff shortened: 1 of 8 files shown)


yarn.lock
```diff
@@ -3643,6 +3643,77 @@
     mkdirp "^1.0.4"
```
Dimitreee, line 3643: Revert this, please

Dimitreee (approved):

apps/condo/domains/common/utils/pdf.ts
```diff
@@ -9,6 +9,7 @@ const PDF_FORMAT_SETTINGS = {
     // lineSpace - margin right for rest of lines (in css pixels)
     'a4': { pdfWidth: 210, pdfHeight: 297, elementOffset: 10, firstLineOffset: 23, lineSpace: 80 },
     'a5': { pdfWidth: 148, pdfHeight: 210, elementOffset: 10, firstLineOffset: 23, lineSpace: 80 },
+    'fullscreen': { pdfWidth: PDF_REPORT_WIDTH, pdfHeight: 297, elementOffset: 15, firstLineOffset: 23, lineSpace: 80 },
```
toplenboren, line 12: I saw that we named this one `fullscreen` but below width is defined as 1920px. Maybe its worth to rename it to smth like `1080p`?

toplenboren (approved):

yarn.lock
sitozzz: Done

apps/condo/domains/common/utils/pdf.ts
sitozzz: Fixed

dkoviazin (approved):

PR #433, opened 2021-08-06
SBERDOMA-942 - add method for creating paged pdf docs 

sitozzz:
- [x] unlock access to pdf generation page
- [x] add helper method for creating paged pdf
image


apps/condo/domains/common/utils/pdf.ts
```diff
@@ -81,3 +83,28 @@ export const createPdf: ICreatePdf = (options) => {
         return doc.save(fileName, { returnPromise: true })
     })
 }
+
+export const createPdfWithPageBreaks: ICreatePdfWithPageBreaks = (options) => {
+    const { element, fileName } = options
+    const elementWidth = element.clientWidth
+    const elementHeight = element.clientHeight
+    const topLeftMargin = 15
```
Dimitreee, line 91: get from options

apps/condo/domains/common/utils/pdf.ts
```diff
@@ -81,3 +83,28 @@ export const createPdf: ICreatePdf = (options) => {
...
+    const pdfWidth = elementWidth + (topLeftMargin * 2)
+    const pdfHeight = (pdfWidth * 1.5) + (topLeftMargin * 2)
+
+    const totalPDFPages = Math.ceil(elementHeight / pdfHeight) - 1
+
+    return html2canvas(element, { allowTaint: true }).then(canvas => {
```
Dimitreee, line 97: i'm not sure allowTaint need here

apps/condo/domains/common/utils/pdf.ts
```diff
@@ -81,3 +83,28 @@ export const createPdf: ICreatePdf = (options) => {
...
+        const imgData = canvas.toDataURL('image/jpeg', 1.0)
+        const pdf = new Jspdf('p', 'pt',  [pdfWidth, pdfHeight])
```
Dimitreee, line 99: get pdfWIdth, pdfHeight from settings

apps/condo/domains/common/utils/pdf.ts
sitozzz: Did you mean to get element properties from options?

apps/condo/domains/common/utils/pdf.ts
sitozzz: I can't get pdfWIdth, pdfHeight params from constant because thats related to problem with chart display. In this case document properties should be calculated from element rendered before pdf save.

dkoviazin (approved):

Dimitreee (approved):

PR #440, opened 2021-08-09
Fix/condo/sberdoma 721/resident api fixes soft delete

AntonAL:
Only soft-delete update-operation is allowed for resident.
Resident can now be updated from current user in session, which is equal to Resident's user.

### DIFF
--- a/apps/condo/domains/resident/access/Resident.js
+++ b/apps/condo/domains/resident/access/Resident.js
@@ -3,6 +3,7 @@
  */
 
 const { RESIDENT } = require('@condo/domains/user/constants/common')
+const { Resident } = require('../utils/serverSchema')
 
 async function canReadResidents ({ authentication: { item: user } }) {
     if (!user) return false
@@ -22,6 +23,16 @@ async function canManageResidents ({ authentication: { item: user }, originalInp
         if (operation === 'create') return true
         // Only soft-delete is allowed for current resident
         if (operation === 'update') {
+            if (!itemId) {
+                return false
+            }
+            const [resident] = await Resident.getAll(context, { id: itemId })
+            if (!resident) {
+                return false
+            }
+            if (user.id !== resident.user.id) {
+                return false
+            }
             if (originalInput.deletedAt !== null) {
                 return true
             }
--- a/apps/condo/domains/resident/access/Resident.js
+++ b/apps/condo/domains/resident/access/Resident.js
@@ -4,7 +4,7 @@
 
 const { RESIDENT } = require('@condo/domains/user/constants/common')
 const { Resident } = require('../utils/serverSchema')
-const { has } = require('lodash')
+const { isSoftDelete } = require('@core/keystone/access')
 
 async function canReadResidents ({ authentication: { item: user } }) {
     if (!user) return false
@@ -33,12 +33,7 @@ async function canManageResidents ({ authentication: { item: user }, originalInp
             if (user.id !== resident.user.id) {
                 return false
             }
-            if (
-                originalInput.deletedAt !== null &&
-                Object.keys(originalInput).length === 3 &&
-                has(originalInput, 'dv') &&
-                has(originalInput, 'sender')
-            ) {
+            if (isSoftDelete(originalInput)) {
                 return true
             }
         }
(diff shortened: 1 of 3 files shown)


apps/condo/domains/resident/access/Resident.js
```diff
@@ -18,8 +19,24 @@ async function canReadResidents ({ authentication: { item: user } }) {
 async function canManageResidents ({ authentication: { item: user }, originalInput, operation, itemId, context }) {
     if (!user) return false
     if (user.isAdmin) return true
-    if (operation === 'create' && user.type === RESIDENT) {
-        return true
+    if (user.type === RESIDENT) {
+        if (operation === 'create') return true
```
pahaz, line 23: We have register resident ? We don’t need to create it

apps/condo/domains/resident/access/Resident.js
```diff
@@ -18,8 +19,24 @@ async function canReadResidents ({ authentication: { item: user } }) {
...
+        // Only soft-delete is allowed for current resident
+        if (operation === 'update') {
```
pahaz, line 25: It’s better to check only deletedAt ticket

toplenboren: You also should check if the deletedAt is the only key in the query

Right now I can update some properties while deleting an object

apps/condo/domains/resident/schema/Resident.test.js
```diff
@@ -412,5 +424,32 @@ describe('Resident', () => {
             expect(objUpdated.updatedAt).toMatch(DATETIME_RE)
             expect(objUpdated.updatedAt).not.toEqual(objUpdated.createdAt)
         })
+
+        it('can be soft-deleted using update operation by current user with type resident', async () => {
```
toplenboren, line 428: Try also to delete user while updating it

apps/condo/domains/resident/access/Resident.js
AntonAL: Hmm, I thought, that access control hooks are being called, when executing custom mutation (with passing not modified context). I see, that they are bypassed.

AntonAL: Removed in `dc1fc488`.

apps/condo/domains/resident/schema/Resident.test.js
AntonAL: Added extra checks in `737cbaa1`

apps/condo/domains/resident/access/Resident.js
AntonAL: Added extra checks in `737cbaa1`

apps/condo/domains/resident/access/Resident.js
AntonAL: Fixed in `058686b1`

apps/condo/domains/resident/access/Resident.js
```diff
@@ -18,8 +20,28 @@ async function canReadResidents ({ authentication: { item: user } }) {
...
+            }
+            const [resident] = await Resident.getAll(context, { id: itemId })
+            if (!resident) {
+                return false
+            }
+            if (user.id !== resident.user.id) {
+                return false
+            }
+            if (
+                originalInput.deletedAt !== null &&
```
dkoviazin, line 37: Maybe there is a more simple way to check - without checking Object.keys(originalInput).length === 3
At least we can define some function to check if update action was in fact softDelete to use this later

dkoviazin (approved):

apps/condo/domains/resident/access/Resident.js
AntonAL: Extracted in `9eb94e62`

packages/@core.keystone/access.js
```diff
@@ -54,6 +55,15 @@ const readOnlyField = {
     update: false,
 }
 
+const isSoftDelete = (originalInput) => {
+    return (
+        originalInput.deletedAt !== null &&
```
pahaz, line 60: newId is also related to softDelete

pahaz (approved):

packages/@core.keystone/access.js
AntonAL: Added in `066ce93b`

PR #443, opened 2021-08-10
SBERDOMA-962 Extends fields validations from cookies.

MrFoxPro:
This fixes validation error in admin UI when trying to change OrganizationEmployee entity because of lack of fields 'dv', 'sender' in request.

### DIFF
--- a/apps/condo/domains/common/utils/makeid.utils.js
+++ b/apps/condo/domains/common/utils/makeid.utils.js
@@ -1,9 +1,10 @@
+const { sample } = require('lodash')
+
 function makeId (length) {
     let result = ''
     const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789'
-    const charactersLength = characters.length
     for (let i = 0; i < length; i++) {
-        result += characters.charAt(Math.floor(Math.random() * charactersLength))
+        result += sample(characters)
     }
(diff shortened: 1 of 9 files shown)


MrFoxPro:
I suppose this should be refactored

apps/condo/domains/common/utils/makeid.utils.js
```diff
@@ -0,0 +1,13 @@
+function makeId (length) {
+    let result = ''
+    const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789'
+    const charactersLength = characters.length
+    for (let i = 0; i < length; i++) {
+        result += characters.charAt(Math.floor(Math.random() * charactersLength))
```
AntonAL, line 6: An util `sample` from Lodash can be used here.

MrFoxPro:
@pahaz There are no ways to throw validation error outside of validateInput hook, so I think it is the only way to do this. Or maybe we can extend GQLCustomSchema and inject in every validateInput hook.

apps/condo/domains/common/utils/makeid.utils.js
MrFoxPro: Done

Dimitreee (approved):

toplenboren (approved):

PR #444, opened 2021-08-10
SBERDOMA-818 Fix Resend message test

dkoviazin:
Some times worker changed the status of the message immediately to "processing" so this test was failing ~ 1/10 times


sitozzz (approved):
One less unstable test case!

Dimitreee (approved):

PR #452, opened 2021-08-12
SBERDOMA-999 - fix analytics page tabs default margin

sitozzz:
- [x] Fix top tabs margin

before:
image
after:
image

### DIFF
--- a/apps/condo/pages/analytics/detail/report_by_tickets/index.tsx
+++ b/apps/condo/pages/analytics/detail/report_by_tickets/index.tsx
@@ -98,6 +98,9 @@ const tabsCss = css`
   & .ant-tabs-tab.ant-tabs-tab-active {
     font-weight: bold;
   }
+  & .ant-tabs-nav {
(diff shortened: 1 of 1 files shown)


MrFoxPro (approved):

apps/condo/pages/analytics/detail/report_by_tickets/index.tsx
```diff
@@ -98,6 +98,9 @@ const tabsCss = css`
   & .ant-tabs-tab.ant-tabs-tab-active {
     font-weight: bold;
   }
+  & .ant-tabs-nav {
+    margin: 0;
```
SavelevMatthew, line 102: please notice, that our design system uses 2 margins for tab now: 40px for pages and 16px for components....I do not recommend to change margins to 0....

Fix margin-top of child container instead

(I'll do our custom tabs in free time, but for now it's like this)

SavelevMatthew (changes requested):

PR #451, opened 2021-08-12
Feature/condo/sberdoma 137/new division domain and schema

AntonAL:
New domain `division` with one first schema `Division`.

CRUD implementation of `Division` respects some of the following rules:
it can be created by employee with `canManageDivisions` ability in organization in question
it can be connected only to responsible which has `canBeAssignedAsResponsible` ability
it cannot be connected to responsible from another organization

PR #463, opened 2021-08-13
Feature/condo/sberdoma 137/new schema division property

AntonAL:
This PR depends on #451

New schema `DivisionProperty` which acts as a join table for many association.

CRUD implementation of `DivisionProperty` respects some of the following rules:
it can be created by employee with `canManageDivisions` ability in organization in question
it cannot be connected to property or division from another organization


AntonAL:
Keystone does not allows to have custom join schemas.
It's strange, because it is a must for a mature ORM…

PR #468, opened 2021-08-16
SBERDOMA-1021 restore soft-deleted resident on subsequent register

AntonAL:
If resident gets soft-deleted and registered on the same address and unitName, `deletedAt` gets cleared.

### DIFF
--- a/apps/condo/domains/resident/schema/RegisterResidentService.js
+++ b/apps/condo/domains/resident/schema/RegisterResidentService.js
@@ -6,6 +6,7 @@ const { getById, GQLCustomSchema } = require('@core/keystone/schema')
 const access = require('../access/RegisterResidentService')
 const { Resident } = require('../utils/serverSchema/index')
 const { Property } = require('@condo/domains/property/utils/serverSchema')
+const { Resident: ResidentAPI } = require('../utils/serverSchema')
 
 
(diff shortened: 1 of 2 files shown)


toplenboren (approved):

apps/condo/domains/resident/schema/RegisterResidentService.js
```diff
@@ -30,14 +31,28 @@ const RegisterResidentService = new GQLCustomSchema('RegisterResidentService', {
...
+                        attrs.property = { connect: { id: property.id } }
+                        attrs.organization = { connect: { id: property.organization.id } }
+                    }
+                    const resident = await Resident.create(context, attrs)
+                    id = resident.id
                 }
-                const resident = await Resident.create(context, attrs)
                 // Hack that helps to resolve all subfields in result of this mutation
-                return await getById('Resident', resident.id)
+                return await getById('Resident', id)
```
dkoviazin, line 55: better not to use return await
[link](http://hassansin.github.io/Why-return-await-Is-a-Bad-Idea)

dkoviazin (approved):

apps/condo/domains/resident/schema/RegisterResidentService.js
AntonAL: Changed at `0a38781`.

PR #478, opened 2021-08-19
Fix/condo/sberdoma 1046/wrong data merging in use create client util

AntonAL:
If on a form we set initial values and initializing create action in a following way:

```
const initialValues = {
// some fields
}

const handleComplete = () => {/* ... */}

const action = SomeSchema.useCreate(initialValues, handleComplete)
```

Then on executing this action, initial values will override actual form values.
Merging should be in opposite direction.


Dimitreee (approved):

nomerdvadcatpyat (approved):

AntonAL:
Ticket create page needs to be refactored.

dkoviazin (approved):

PR #482, opened 2021-08-19
Fix/condo/sberdoma 1057/missing address in duplicate check of resident

AntonAL:
## Added `address` field on duplicates check

Previously, resident was checked on duplicates using a set of `property`, `unitName` and `user` fields.
When we are trying to create a resident on different address, but with same unitName (both without properties), false-positive duplicates check error will be returned.

## Removed `property` field from duplicates check

Logically, no resident of the same user can exist in the same address and unit name (flat).
Therefore, check for the same property (of organization) is not needed.

## Implemented address cleaning up to building on create resident

Input data of Resident `address` is out of control for mobile client, because they can enter `address` by scanning a QR-code of a receipt. Address in receipt is not standardised.
It can be a case, when a flat number will slip into an `address` field of Resident and starting from that, all checks and validations will fail.

So, we need to make sure, that address of Resident is only up to building and does not contains flat number.

PS: refactored a little bit.

### DIFF
--- a/apps/condo/domains/property/schema/CheckPropertyWithAddressExistService.js
+++ b/apps/condo/domains/property/schema/CheckPropertyWithAddressExistService.js
@@ -7,11 +7,11 @@ const access = require('@condo/domains/property/access/CheckPropertyWithAddressE
 const { Property } = require('@condo/domains/property/utils/serverSchema')
 const get = require('lodash/get')
 const { jsonAddressMetaValidator } = require('../utils/validation.utils')
+const { getAddressUpToBuildingFrom, FLAT_WITHOUT_FLAT_TYPE_MESSAGE } = require('../utils/helpers')
 
 const NO_OBJECT_MESSAGE = 'No object specified!'
 const META_INCORRECT_JSON_MESSAGE = 'AddressMeta Json had incorrect format!'
 const DV_VERSION_MISMATCH_MESSAGE = 'Unknown version of addressMeta! Expected "dv" to equal 1'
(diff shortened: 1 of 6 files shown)


apps/condo/domains/property/utils/testSchema/index.js
```diff
@@ -9,10 +9,11 @@ const { CHECK_PROPERTY_WITH_ADDRESS_EXIST_QUERY } = require('../../gql')
 const { throwIfError } = require('@condo/domains/common/utils/codegeneration/generate.test.utils')
 const { buildingMapJson } = require('@condo/domains/property/constants/property')
 const { generateGQLTestUtils } = require('@condo/domains/common/utils/codegeneration/generate.test.utils')
-const { buildFakeAddressMeta } = require('@condo/domains/common/utils/testSchema/factories')
+const { buildFakeAddressAndMeta } = require('@condo/domains/common/utils/testSchema/factories')
 const { Property: PropertyGQL } = require('@condo/domains/property/gql')
 const { makeClientWithRegisteredOrganization } = require('@condo/domains/organization/utils/testSchema/Organization')
 const { EXPORT_PROPERTIES_TO_EXCEL } = require('@condo/domains/property/gql')
+const {  } = require('@condo/domains/common/utils/testSchema/factories')
```
dkoviazin, line 16: looks strange

dkoviazin (approved):

apps/condo/domains/property/utils/testSchema/index.js
AntonAL: Fixed `b6267890`

sitozzz (approved):

PR #487, opened 2021-08-20
Refactor/condo/sberdoma 1036/connect and disconnect of relationships for template of client schema hooks

AntonAL:
**Changes in `createschema` template, that allows to properly work with single and many associations**

To properly connect and disconnect items in Keystone `Relationship` fields, it now uses initial and final form state.
Initial form state is a set of form values, that was passed to form during initialization.
Final form state is a set of form values that we have for form submision.

Based on difference between initial and final form state:
1. It connects and disconnects items in "one-to-many" relationships,
2. It disconnects item in "one-to-zero" and "one-to-one" relationship
3. It does not clutters GraphQL input, if there is no changes between form states

To utilize all of this, we need to provide initial values into `useUpdate` hook, that comes after `convertToUIState`.
Example code for update-form:

```
const { obj } = MySchema.useObject({ where: { id } })
let initialValues = MySchema.convertToUIFormState(obj)
const action = MySchema.useUpdate(initialValues)
```

👉 Working example in #488


packages/@core.codegen/templates/schema00/domains/{{domain}}/utils/clientSchema/{{name}}.ts
```diff
@@ -38,12 +38,38 @@ function convertToUIFormState (state: I{{ name }}UIState): I{{ name }}FormState
     return result as I{{ name }}FormState
 }
 
-function convertToGQLInput (state: I{{ name }}FormState): {{ name }}UpdateInput {
+function convertToGQLInput (state: I{{ name }}FormState, obj: I{{ name }}UIState, initialState: I{{ name }}FormState): {{ name }}UpdateInput {
```
dkoviazin, line 41: You need to test it on tickets first. When you save a ticket you do not always have all information of initial state. 
1. Changing ticket status
2. Editing ticket without touching existed client

Dimitreee:
https://github.com/open-condo-software/condo/blob/master/apps/condo/domains/organization/utils/clientSchema/OrganizationEmployee.ts#L57

apps/condo/domains/organization/utils/clientSchema/OrganizationEmployee.ts
```diff
@@ -54,35 +55,40 @@ function convertGQLItemToFormSelectState (item: OrganizationEmployee): { value:
...
-        if (Object(state).hasOwnProperty(fieldName) && RELATIONS.includes(fieldName)) {
-            const fieldValue = get(state, fieldName)
-            const id = get(fieldValue, 'id')
-
-            if (id) {
-                result[fieldName] = {
-                    connect: {
-                        id,
-                    },
+    for (const attr of Object.keys(state)) {
```
Dimitreee, line 62: What about tickets?

pahaz:
It's not active anymore ...

PR #492, opened 2021-08-23
SBERDOMA-1039 Fix corrupt migration directory

toplenboren:
Fixes corrupt migration directory, that was resulted in `not` rebasing #481 after #490 merge to master

@MrFoxPro ^^

PR #493, opened 2021-08-23
SBERDOMA-1039 Fix corrupt migration directory by regenerating last migration

toplenboren:
This PR solves the same problem as #492 but in different way. Instead of creating merge migration - we regenerate the last one


dkoviazin (approved):

MrFoxPro (approved):
gj

AntonAL (approved):
Changing names of migrations

PR #515, opened 2021-08-30
SBERDOMA-992 Export tickets to excel fixes

dkoviazin:
1. Add an "updatedAt" column to the template
2. Fix spelling in the template column
3. Fix bug when not all ticket comments exports to excel
4. Fix bug when not all status changes exports to excel

The problem was that TicketComments.getAll and TicketChange.getAll have a limit of 100 records - so, when more than 1000 tickets were exported - we didn't have all comments and dates of changing status in export. 
SBERDOMA-992, SBERDOMA-990


apps/condo/domains/common/utils/serverSchema/index.js
```diff
@@ -0,0 +1,25 @@
...
+}) => {
+    let skip = 0
+    let maxiterationsCount = Math.floor(limit / chunkSize)
+    let newchunk = []
+    let all = []
+    do {
+        newchunk = await model.getAll(context, where, { sortBy, first: chunkSize, skip: skip })
+        all = all.concat(newchunk)
+        skip += newchunk.length
+    } while (--maxiterationsCount > 0 && newchunk.length)
```
AntonAL, line 18:
```suggestion
    let maxiterationsCount = Math.floor(limit / chunkSize)
    let newchunk = []
    let all = []
    do {
        newchunk = await model.getAll(context, where, { sortBy, first: chunkSize, skip: all.length })
        all = all.concat(newchunk)
    } while (--maxiterationsCount > 0 && newchunk.length)
```

apps/condo/domains/common/utils/serverSchema/index.js
```diff
@@ -0,0 +1,25 @@
...
+const loadModelsByChunks = async ({
+    context,
+    model,
+    where = {},
+    sortBy = ['createdAt_ASC'],
+    chunkSize = 100,
+    limit = 100000,
+}) => {
+    let skip = 0
+    let maxiterationsCount = Math.floor(limit / chunkSize)
```
AntonAL, line 11: Are you sure about `Math.floor(limit / chunkSize)`?
For example, a query of 199 items, by 100 items in one chunks requires 2 queries.
We have 199/100 = 1.99 and floored — `1`.

Condition in `while` using `--` prefixed decrement, will get false and we will never have second query.

I think, `Math.ceil` should be used in this case.

PR #520, opened 2021-08-31
SBERDOMA-1128 property range limit

sitozzz:
### Problem: 
Now we can't load huge data with existing getAll util
### Solution: 
Use raw db query with knex adapter


dkoviazin (approved):

SavelevMatthew (approved):
LGTM

PR #526, opened 2021-09-01
SBERDOMA-1136 fix organization create modal view logic

Dimitreee:


AntonAL (approved):
Tested on localhost, it works.

PR #534, opened 2021-09-02
Refactor/condo/sberdoma 1122/error with validation of property map

AntonAL:
Last commit [5e983c08](https://github.com/open-condo-software/condo/pull/534/commits/5e983c0899f7a4d0ce3c20623d95ea63347aa050) demonstrates a strange error with validation of `Property.map`
See Property.test.js:L22

Ajv validator correctly validates a sample `map` object.
But when the same `map` object gets validated in `validateInput` hook of `Property`, — we get following validation error:

```json
[
  {
    "instancePath": "/sections/0/floors/0/units/0/name",
    "schemaPath": "#/properties/sections/items/properties/floors/items/properties/units/items/properties/name/type",
    "keyword": "type",
    "params": {
      "type": "string"
    },
    "message": "must be string"
  }
]
```


AntonAL:
Problem is solved with help of @dkoviazin and @pahaz

AntonAL:
Problem is solved with help of @dkoviazin and @pahaz

PR #537, opened 2021-09-03
SBERDOMA-1122 added type `AddressMetaDataMetroField` to `addressMeta.data.metro` field

AntonAL:


apps/condo/domains/property/schema/fields/AddressMetaField.js
```diff
@@ -1,5 +1,11 @@
 const Ajv = require('ajv')
 
+const AddressMetaDataMetroField = {
```
toplenboren, line 3: Are we sure that we need to use the word `Metro` here? Why not `Underground`? It seems to be more international

toplenboren (approved):

apps/condo/domains/property/schema/fields/AddressMetaField.js
AntonAL: It is a mapping of fields from third-party service with no extra additions.

dkoviazin (approved):

PR #538, opened 2021-09-03
SBERDOMA-1037 One date library for all our needs

SavelevMatthew:
**Problem**: currently using > 1 date library, so...bundle size is increasing

**Solution**: move everything to one library

<hr>

### Options
We're using antd and it's support only 3 libraries for date-management.
1. momentjs (by default)
2. date-fns
3. dayjs

<hr>

### Why not moment js?
- Too large bundle size
- Date objects are mutable, which can be dangerous and lead to some bugs

<hr>

#### dayjs:
+ the smallest bundle size, plugins toggle, when needed
+ chaining syntax, just like momentjs (increases code readability)
+ immutable objects
#### date-fns
+ immutable objects
+ a bit faster than dayjs, because manupulates with native date itself (btw it's hard to notice in reallife scale experience)
- functional syntax (decrease code readability)

More complex comparison you can read here: https://dev.to/dmtrkovalenko/you-might-not-need-date-fns-23f7

<hr>

### What're we choosing: dayjs

<hr>

## TODO:
- [x] Replace Luxon usage with dayjs
- [x] Remove Luxon from packages
- [x] Create custom pickers for antd
- [x] Replace current pickers with new ones
- [x] Replace moment usage with dayjs
- [x] Remove moment from packages
- [x] Replace date-fns locales with dayjs locales
- [x] Replace locales usage with new ones
- [x] Replace date-fns usage with dayjs
- [x] Remove date-fns from packages


sitozzz (approved):
Tested it localy. Looks like everething working as before refactor

nomerdvadcatpyat:
on `/ticket/[id]` page, if ticket was created 1 minute ago, then the number `1` is lost
изображение

nomerdvadcatpyat (approved):

SavelevMatthew:
> on `/ticket/[id]` page, if ticket was created 1 minute ago, then the number `1` is lost
> изображение
> 
> Otherwise, everything is ok

It's not a bug right here, it' usual behaviour of dayjs lib, when working with durations

PR #541, opened 2021-09-03
SBERDOMA-1157 Added colors for antd typography.text + alert

SavelevMatthew:
### Problem
Colors from Figma didn't match with antd Colors, that's why we were used to hard-code values / invent a bicycle

### Solution
Fix colors in antd theme / global styles

image

### DIFF
--- a/apps/condo/domains/common/components/containers/GlobalStyle.tsx
+++ b/apps/condo/domains/common/components/containers/GlobalStyle.tsx
@@ -161,6 +161,36 @@ export default function GlobalStyle () {
                 display: none;
               }
               
+              a.ant-typography[disabled], 
+              .ant-typography a[disabled], 
+              a.ant-typography.ant-typography-disabled, 
+              .ant-typography a.ant-typography-disabled {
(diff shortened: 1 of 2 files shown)


toplenboren (approved):

sitozzz (approved):

apps/condo/domains/common/components/containers/GlobalStyle.tsx
```diff
@@ -161,6 +161,36 @@ export default function GlobalStyle () {
...
+              a.ant-typography.ant-typography-disabled, 
+              .ant-typography a.ant-typography-disabled {
+                color: ${colors.green[6]};
+                &:hover {
+                  color: ${colors.green[6]};
+                }
+              }
+              
+              .ant-typography strong {
+                font-weight: 800;
```
Dimitreee, line 175: move fontWeight to const

SavelevMatthew: done 👍

PR #557, opened 2021-09-09
DOMA-1177 Meter model updates

nomerdvadcatpyat:
1. Change MeterReading value types from Integer to Decimal
2. Write tests for it
3. Add `sealingDate` to Meter schema


dkoviazin (approved):

SavelevMatthew (approved):
Make sure to check migration before merging

PR #559, opened 2021-09-09
Feature/condo/doma 1134/ticket assignment logic based on divisions

AntonAL:
Following cases are covered:
1. When property is selected, then from first matched division first executor from this division will be assigned;
2. When property and category classifier are selected, then from first matched division executor with matched specialization will be assigned.


pahaz (approved):

apps/condo/domains/ticket/components/BaseTicketForm/AutoAssignerByDivisions.tsx
```diff
@@ -0,0 +1,117 @@
...
+        message = FoundManyDivisionsMessage
+            .replace('{address}', addressWithoutCity)
+            .replace('{divisions}', divisionNames)
+    }
+
+    if (responsibleUserId) {
+        message += '. ' + AssignedResponsibleMessage
+    }
+
+    message += executorUserId
```
dkoviazin, line 100: Please, do not use multiple ternary operators  - it's hard to read such code

dkoviazin (approved):

apps/condo/domains/ticket/components/BaseTicketForm/AutoAssignerByDivisions.tsx
AntonAL: Why? It is a shortest possible chain of conditions, being read in a sequence…

PR #561, opened 2021-09-10
DOMA-1127 MVP of BillingReceipts JSON

SavelevMatthew:
Added types for BillingReceipt JSON fields

Tested:

- [x] GraphiQL
- [x] Fetching from admin-panel
- [x] Displaying on admin-panel
- [x] Updating from admin-panel
- [x] Receiving data on client
- [x] Receiving data through allResidentBillingReceipts
- [x] allResidentBillingReceipts tests
- [x] generating receipts of all detailing levels via "yarn generate-receipts" 
- [x] BillingReceipts tests


nomerdvadcatpyat (approved):

apps/condo/domains/billing/schema/fields/BillingReceipt/Services.js
```diff
@@ -0,0 +1,96 @@
...
+const validateServices = getValidator(ServicesSchemaValidator)
+const SERVICE_TO_PAY_DETAILS_QUERY_LIST = Object.keys(ServiceToPayDetailsFields).join(' ')
+const SERVICES_QUERY_LIST = `${Object.keys(ServiceFields).join(' ')} { ${SERVICE_TO_PAY_DETAILS_QUERY_LIST} }`
+
+const SERVICES_FIELD = {
+    schemaDoc: 'Services to pay for. Every service has id, name and toPay. Service may or may not have toPay detail. Detail level 3 and 4',
+    type: Json,
+    isRequired: false,
+    extendGraphQLTypes: [SERVICES_GRAPHQL_TYPES],
+    graphQLInputType: `[${BILLING_RECEIPT_SERVICE_INPUT_NAME}!]`,
```
pahaz, line 85: it looks like a substitution of definitions. The name `BillingReceiptServiceFieldInput` assume to be an input of the field. But, the real input type is `[BillingReceiptServiceFieldInput!]`

pahaz: It's confusing me

SavelevMatthew: Why? Field called services, type is service….So services =[service]…seems like intuitive

apps/condo/domains/billing/schema/fields/BillingReceipt/Recipient.js
```diff
@@ -0,0 +1,55 @@
...
+    graphQLAdminFragment: `{ ${RECIPIENT_QUERY_LIST} }`,
+    hooks: {
+        validateInput: validateRecipient,
+    },
+}
+
+module.exports = {
+    RECIPIENT_FIELD,
+    RECIPIENT_QUERY_LIST,
+}
```
toplenboren, line 55: Please add a newline

apps/condo/domains/billing/schema/fields/BillingReceipt/Services.js
toplenboren: The problem here is that variable

`BILLING_RECEIPT_SERVICE_INPUT_NAME` is simply `BillingReceiptServiceFieldInput`

But if real world it should be used like this: `[ + BILLING_RECEIPT_SERVICE_INPUT_NAME + ]` which is super counter-intuitive

Imagine if we need to use this variable in other context or other file, in that case we should somehow know that we need to surround it with `[ ]`

Maybe we need to make `BILLING_RECEIPT_SERVICE_FIELD_NAME` equal to real value of the field `[BillingReceiptServiceFieldInput!]`

apps/condo/domains/billing/schema/fields/BillingReceipt/ToPayDetailsField.js
```diff
@@ -0,0 +1,59 @@
...
+    hooks: {
+        validateInput: validatePaymentDetails,
+    },
+}
+
+module.exports = {
+    TO_PAY_DETAILS_FIELD,
+    TO_PAY_DETAILS_QUERY_LIST,
+    ToPayDetailsFields,
+}
```
toplenboren, line 59: Please add newline

apps/condo/domains/common/utils/omitRecursively.js
```diff
@@ -1,12 +1,12 @@
-import { isArray, isObject, omit, transform } from 'lodash'
+const { isArray, isObject, omit, transform } = require('lodash')
 
 /**
```
toplenboren, line 3: Why do we have `omitRecursively` here and in packages/@core.keystone/fields/Json/utils/cleaner.js ?

packages/@core.keystone/fields/Json/utils/cleaner.js
```diff
@@ -0,0 +1,20 @@
+const { isArray, isObject, omit } = require('lodash')
```
toplenboren, line 1: Its a good practice to write a `.spec` with tests if you write utils

SavelevMatthew: tests added 👍

apps/condo/domains/billing/schema/fields/BillingReceipt/Services.js
SavelevMatthew: Moved to const 👍

apps/condo/domains/common/utils/omitRecursively.js
SavelevMatthew: removed duplicate

PR #565, opened 2021-09-13
DOMA-1206 Fix "/property" layout

SavelevMatthew:
Fix "import + create" buttons alignment
Fix property table columns widths

Before/after

image

image


sitozzz (approved):
Seem's ok) But may you provide some screenshots that could be show what exactly you are fixed

SavelevMatthew:
> Seem's ok) But may you provide some screenshots that could be show what exactly you are fixed

sure....added to description

dkoviazin (approved):

PR #582, opened 2021-09-16
Feature/condo/doma 1227/mobile api enhancements

toplenboren:
## Problem:

Mobile registration flow has screen when it prompts user to enter his account number

After this we call mutation `registerServiceConsumer` which finishes with `OK` only if:
- Organization (`УК`) has billing set up (please also see notes for EPS case)
- Billing has `BillingAccount` with the same `accountNumber` and `unitName`

We do not want to show this screen to users whose `Organization` does not have billing set up

### Notes on EPS (unified payment system) case:

in future`registerServiceConsumer` should create EPS service consumer if billing is not setUp. But in this case we are going to call this mutation in `Оплаты` section of the app, not `Регистрация`. So this query is needed

Check this stuff on **App Client Prod** figma page

## Solution

Create custom query which returns if organization has set up billing integration

### Usage

`checkBillingIntegrationOrganizationContextExist( data: { organizationId: <UUID> }) { isFound }`


apps/condo/domains/billing/access/CheckOrganizationIntegrationContextExistService.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Generated by `createservice billing.CheckOrganizationIntegrationContextExistService --type queries`
+ */
+const { throwAuthenticationError } = require('@condo/domains/common/utils/apolloErrorFormatter')
+async function canCheckOrganizationIntegrationContextExist ({ authentication: { item: user } }) {
```
pahaz, line 5:
```suggestion

async function canCheckOrganizationIntegrationContextExist ({ authentication: { item: user } }) {
```

pahaz (changes requested):
It's a little confusing me. Why do you want some extras for the EPS case?

apps/condo/domains/billing/schema/CheckOrganizationIntegrationContextExistService.js
```diff
@@ -0,0 +1,51 @@
...
+            schema: 'checkOrganizationIntegrationContextExist (data: CheckOrganizationIntegrationContextExistInput!): CheckOrganizationIntegrationContextExistOutput',
+            resolver: async (parent, args, context, info, extra = {}) => {
+                const { data: inputData } = args
+                const { organizationId } = inputData
+
+                if (!organizationId) throw new Error(NO_ORGANIZATION_MESSAGE)
+
+                const count = await BillingIntegrationOrganizationContext.count(context, {
+                    organization: { id: organizationId },
+                    status: 'finished',
```
SavelevMatthew, line 38: use const from billing domain

apps/condo/domains/billing/schema/CheckOrganizationIntegrationContextExistService.js
```diff
@@ -0,0 +1,51 @@
...
+                })
```
SavelevMatthew, line 39: should we also check last report here?

PR #602, opened 2021-09-20
DOMA-1251 amoCRM integration

MrFoxPro:


.env.example
```diff
@@ -24,3 +24,7 @@ DOCKER_COMPOSE_SERVER_URL=https://condo.dev.doma.ai
 
 # feature flags configuration
 FEATURE_FLAGS_CONFIG='["billing"]'
+NGROK_URL=https://5f4a-80-83-252-85.ngrok.io
```
toplenboren, line 27: Ngrok is paid little buddy
Should we force to use him, or maybe we can simply use any software that can tunnel localhost?

MrFoxPro: no, this is just example

PR #603, opened 2021-09-20
DOMA-1224 multiple filters modal

nomerdvadcatpyat:
Description of filters using filterMeta + modal window with filters

[how to use](https://doma.atlassian.net/wiki/spaces/SBER/pages/820117521)

### DIFF
--- a/apps/condo/domains/common/components/TableFilter.tsx
+++ b/apps/condo/domains/common/components/TableFilter.tsx
@@ -5,6 +5,7 @@ import { Button } from './Button'
 import { FilterFilled } from '@ant-design/icons'
 import { colors } from '../constants/style'
 import { FilterDropdownProps } from 'antd/lib/table/interface'
+import styled from '@emotion/styled'
 
 export interface IFilterContainerProps {
     clearFilters: () => void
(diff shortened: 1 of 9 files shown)


apps/condo/domains/common/components/TableFilter.tsx
```diff
@@ -35,6 +36,34 @@ export const FilterContainer: React.FC<IFilterContainerProps> = (props) => {
     )
 }
 
+export const SelectFilterContainer: React.FC<IFilterContainerProps> = (props) => {
+    const intl = useIntl()
+    const ResetLabel = intl.formatMessage({ id: 'filters.Reset' })
+
+    return (
+        <div style={{
```
SavelevMatthew, line 44: Why not styled component?

apps/condo/domains/common/utils/filters.utils.ts
```diff
@@ -0,0 +1,128 @@
+import { QueryMeta } from './tables.utils'
```
SavelevMatthew, line 1: Should it be covered with test?

SavelevMatthew (approved):

apps/condo/domains/common/utils/filters.utils.ts
```diff
@@ -0,0 +1,128 @@
...
+            const search = get(props, 'search')
+            const mode = get(props, 'mode')
+            return getGQLSelectFilterDropdown(search, mode, columnFilterComponentWrapperStyles)
+        }
+
+        case ComponentType.Custom: {
+            return get(component, 'getComponentFilterDropdown')
+        }
+
+        default: return
```
sitozzz, line 126: May be better to throw an error like `Unknown component type`?

dkoviazin (approved):

apps/condo/domains/common/utils/filters.utils.ts
nomerdvadcatpyat: It seems to me it is better to return void, because in this case we simply will not have a component for filtering (because we did not find the right one by type). If you throw out an error, the client part will fall in runtime
изображение

sitozzz (approved):

PR #617, opened 2021-09-23
DOMA-818 Cypres E2E Tests

dkoviazin:
1. Added "data-cy" attributes to key elements in all auth scenarios
2. Used keystone test utils to access data
3. Auth scenarios  Sign in/Forgot Password/Change Password/Register

Test video

https://dashboard.cypress.io/projects/dc36jj/runs/2/test-results/1f3bce67-bab9-4186-81f5-0d7b93e5135d/video?utm_source=Dashboard&utm_medium=Share+URL&utm_campaign=Video

to run it locally (better to build and start first as in dev mode it will be longer)
```bash
yarn run cypress open
```


apps/condo/cypress/objects/Auth.js
```diff
@@ -0,0 +1,195 @@
+/* eslint-disable @typescript-eslint/explicit-module-boundary-types */
+/* eslint-disable no-undef */
```
sitozzz, line 2: We need to find way for writing tests without casting `no-undef` on each test file

sitozzz (approved):

PR #619, opened 2021-09-23
DOMA-1287 add canManageMeterReading role

nomerdvadcatpyat:
Now there is a role "canManageMeters", which applies to both readings and meters. There are cases when an employee can take readings, but cannot create meters. We need to make support for this.


pahaz (approved):

SavelevMatthew (approved):
LGTM

PR #629, opened 2021-09-26
DOMA-1291 add createCronTask and removeCronTask

pahaz:
image

### DIFF
--- a/packages/@core.keystone/tasks.js
+++ b/packages/@core.keystone/tasks.js
@@ -18,6 +18,8 @@ const taskQueue = new Queue('tasks', WORKER_REDIS_URL)
 const taskLogger = pino({ name: 'worker', enabled: falsey(process.env.DISABLE_LOGGING) })
 const globalTaskOptions = {}
 const TASKS = new Map()
+const CRON_TASKS = new Map()
+const REMOVE_CRON_TASKS = []
 let isWorkerCreated = false
 
 function createTask (name, fn, opts = {}) {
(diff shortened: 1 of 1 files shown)


packages/@core.keystone/tasks.js
```diff
@@ -155,11 +173,28 @@ async function createWorker (keystoneModule) {
...
     await taskQueue.isReady()
     taskLogger.info({ message: 'Worker: ready to work!' })
+    const cronTasksNames = [...CRON_TASKS.keys()]
+    if (cronTasksNames.length > 0) {
+        taskLogger.info({ message: 'Worker: add repeatable tasks!', names: cronTasksNames })
+        cronTasksNames.forEach((name) => {
+            const fn = TASKS.get(name)
+            fn.delay()
+        })
+    }
```
AntonAL, line 183: It's unclear, how a task is set as repeatable. There is no mention of this.
Could you please write an instruction

AntonAL (commented):
Out of scope of this PR:
Could you please write instruction of how to use cron or non-cron tasks?
Why `prepareKeystoneExpressApp` is imported from test helpers, but it seems, like it is used in non-test environment in `apps/condo/worker.js`.
How to use `removeCronTask`?

packages/@core.keystone/tasks.js
pahaz: I think we can discuss it on out tech tolk

AntonAL:
Can number of wrapper functions be reduced to make it less complex?
For example, can you merge body of `createTaskWrapper` into `createTask`?

AntonAL (approved):

PR #630, opened 2021-09-27
DOMA-1293 added build step to checklists/pull-request

AntonAL:


nomerdvadcatpyat (approved):

pahaz:
We have this step in our CI/CD. Do you really think that we need to check it by hand?

AntonAL:
> We have this step in our CI/CD. Do you really think that we need to check it by hand?

It is just a recommendation to save time, because all the operations will be completed faster, when run locally.

toplenboren (approved):

PR #680, opened 2021-10-06
DOMA-1371 Billing goes Decimal

toplenboren:
## Problem

- Billing stores toPay as string

## Solution

- Make toPay a Decimal 
- Implement Currency code in billing
- Make internationalization of the text on frontend using Intl

image

### DIFF
--- a/apps/condo/domains/billing/utils/testSchema/index.js
+++ b/apps/condo/domains/billing/utils/testSchema/index.js
@@ -45,12 +45,14 @@ async function createTestBillingIntegration (client, extraAttrs = {}) {
     const sender = { dv: 1, fingerprint: faker.random.alphaNumeric(8) }
     const name = faker.company.companyName().replace(/ /, '-').toUpperCase() + ' TEST INTEGRATION'
     const detailsTitle = faker.animal.dog() + ' TEST INTEGRATION'
+    const currencyCode = 'RUB'
 
     const attrs = {
         dv: 1,
         sender,
         name,
         detailsTitle,
+        currencyCode,
         ...extraAttrs,
     }
     const obj = await BillingIntegration.create(client, attrs)


apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -55,32 +56,22 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
...
             schemaDoc: 'Total sum to pay. Usually counts as the sum of all services. Stored like this: "123.30" Detail level 1.',
-            type: Text,
+            type: Decimal,
             isRequired: true,
+            knexOptions: {
+                scale: 2,
+            },
+        },
+
+        commission: {
```
SavelevMatthew, line 68: why is commission here?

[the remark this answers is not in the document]
pahaz: `to pay sum` = `property management company part` + `doma service part` + `bank service part` ?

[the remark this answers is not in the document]
pahaz: We have mobe than one comission

pahaz: As for me, the commissions are related to payment rather than receipts. It's a good question

apps/condo/domains/billing/schema/BillingReceipt.test.js
```diff
@@ -208,6 +208,23 @@ describe('BillingReceipt', () => {
...
+            const [obj] = await createTestBillingReceipt(managerUserClient, context, property, billingAccount)
+
+            const payload = {
+                toPay: '22.92',
+            }
+
+            const [objUpdated] = await updateTestBillingReceipt(managerUserClient, obj.id, payload)
+
+            expect(obj.id).toEqual(objUpdated.id)
+            expect(objUpdated.toPay).toEqual('22.92000000')
```
SavelevMatthew, line 226: Is that correct way to check decimals? at least it's not obvious for me

apps/condo/domains/common/schema/fields.js
```diff
@@ -122,6 +123,15 @@ const CLIENT_PHONE_LANDLINE_FIELD = {
     },
 }
 
+const MONEY_FIELD = {
+    schemaDoc: 'Money field',
+    type: Decimal,
+    isRequired: false,
+    knexOptions: {
+        scale: 8,
```
SavelevMatthew, line 131: Scale = digits after delimiter?

toplenboren: Yep

apps/condo/domains/billing/schema/BillingReceipt.test.js
toplenboren: Well, it looks odd, yep

The idea here is that you save something like 0.23 (note that it has 2 digits on decimal part) and DB parses it into 0.23000000 (note that it has 8 digits on decimal part as specified)

SavelevMatthew (changes requested):
some important notes:
Use currencyCode instead of name and currency itself, it's more clear and obvious.
Don't store anything else...Currency formatting is too complicated, and all that additional field s will not help you to format it...just use native way (like intl), it can work with all locales and currencies out of the box: https://formatjs.io/docs/react-intl/api/#formatnumber

apps/condo/domains/common/utils/tables.utils.ts
```diff
@@ -150,6 +150,10 @@ export const getFilter: (
     }
 }
 
+export const getDecimalFilter: (dataIndex: DataIndexType) => FilterType = (dataIndex) => {
```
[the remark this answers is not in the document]
toplenboren: No, it shouldn't be upscaled, since Decimal serach does it under the hood 

From https://v5.keystonejs.com/keystonejs/fields/src/types/decimal/
```
Despite being specified as strings, all comparisons are performed on numerical equivalence. In essence, this means leading and trailing zeros are ignored. Eg. 04.53000 === 4.53.
```

apps/condo/domains/common/constants/currencies.js
```diff
@@ -0,0 +1,1077 @@
+const CURRENCIES = {
```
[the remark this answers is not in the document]
toplenboren: We might have a situation when we really need to send some currency-specific information to some client which cannot access this libraries

Example -- some quirky embedded meter reading software?

Please also note that these currencies are located only in backend and does not ship to client

apps/condo/migrations/20211015160205-0067_auto_20211015_1102.js
```diff
@@ -0,0 +1,62 @@
...
+
+exports.up = async (knex) => {
```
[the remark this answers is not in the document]
toplenboren: Good point, gonna regenerate

apps/condo/domains/common/components/Table/Renders.tsx
```diff
@@ -89,79 +90,62 @@ export const renderMeterReading = (values: string[], resourceId: string, measure
...
 }
 
-const fillMoneyWithSpaces = (substring: string, startIndex: number, spaces: Array<number>) => {
-    const chars = []
-    for (let i = 0; i < substring.length; i++) {
-        const absoluteIndex = startIndex + i
-        const char = substring.charAt(i)
-        if (spaces.includes(absoluteIndex)) {
-            chars.push(MONEY_PARTS_SEPARATOR)
+const getCurrencySymbol = (currencyName) => {
```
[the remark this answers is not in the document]
toplenboren: This function gets the currency mark, like $ for dollar. The reason for this function is because native Intl returns string in this format:

`<currencyMark><value>.<decimalPart> // $ 20.00` but we need to have it this way: `20.00$`

While it is obvious that we can simply take the first symbol and put it in the end of the string, it also might not work with specific locales, which does not have a narrow symbol (ETH, BTC) or for negative numbers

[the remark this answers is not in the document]
toplenboren: Well, yes, but for some annoying reason Intl thinks that russian locale should start with currency mark :(

I think amount is by default more important then currency in our service, since you only work with one currency at the time, so let's leave a template like this: `amount + code` 

image

apps/condo/domains/billing/schema/BillingIntegration.js
```diff
@@ -87,12 +89,11 @@ const BillingIntegration = new GQLListSchema('BillingIntegration', {
...
-        currency: {
+        currencyCode: {
+            ...CURRENCY_CODE_FIELD,
             schemaDoc: 'Currency which this billing uses',
-            type: Relationship,
-            ref: 'BillingCurrency',
-            isRequired: false,
-            kmigratorOptions: { null: true, on_delete: 'models.CASCADE' },
+            isRequired: true,
+            defaultValue: DEFAULT_CURRENCY_CODE,
```
SavelevMatthew, line 96: I think we should force integration to specify currencyCode values, because default value can be used to create mistakes. Clarifying it will decrease amount of errors

apps/condo/domains/common/components/Table/Renders.tsx
```diff
@@ -89,79 +90,62 @@ export const renderMeterReading = (values: string[], resourceId: string, measure
...
-            chars.push(MONEY_PARTS_SEPARATOR)
+const getCurrencySymbol = (currencyCode) => {
+    return (0).toLocaleString(
+        undefined,
+        {
+            style: 'currency',
+            currency: currencyCode,
+            minimumFractionDigits: 0,
+            maximumFractionDigits: 0,
+            currencyDisplay: 'narrowSymbol',
```
SavelevMatthew, line 101: What's this? sounds like custom logic...

apps/condo/domains/common/components/Table/Renders.tsx
```diff
@@ -89,79 +90,62 @@ export const renderMeterReading = (values: string[], resourceId: string, measure
...
-}
+const dimText = (text: string, extraStyles: React.CSSProperties = {}) => (
+    <Typography.Text type={'secondary'} style={extraStyles}>
+        {text}
+    </Typography.Text>
+)
 
-export const getMoneyRender = (search: string, currencyMark = '₽', partSeparator = DEFAULT_CURRENCY_SEPARATOR) => {
+const renderMoney = (currencyValuePart: string, currencyDecimalPart: string, currencySymbol: string) => (
+    <>{currencyValuePart}{dimText(',' + currencyDecimalPart)}{NBSP}{currencySymbol}</>
```
SavelevMatthew, line 113: "," ?

SavelevMatthew: You should better use `Intl.formatNumber`, then find index of `['.', ',']`, then find digits after....and use calculated index to recolour string

SavelevMatthew: Or may be I'm a bit confused about this logic....Can you provide more detailed explanation what's going on?

apps/condo/migrations/20211020003436-0069_auto_20211019_1934.js
```diff
@@ -0,0 +1,93 @@
...
...bicsCiAgICAgICAgICAgIG5hbWU9J2N1cnJlbmN5Q29kZScsCiAgICAgICAgICAgIGZpZWxkPW1vZGVscy5UZXh0RmllbGQoZGVmYXVsdD0nUlVCJyksCiAgICAgICAgICAgIHByZXNlcnZlX2RlZmF1bHQ9RmFsc2UsCiAgICAgICAgKSwKICAgICAgICBtaWdyYXRpb25zLkFkZEZpZWxkKAogICAgICAgICAgICBtb2RlbF9uYW1lPSdiaWxsaW5naW50ZWdyYXRpb25oaXN0b3J5cmVjb3JkJywKICAgICAgICAgICAgbmFtZT0nY3VycmVuY3lDb2RlJywKICAgICAgICAgICAgZmllbGQ9bW9kZWxzLlRleHRGaWVsZChibGFuaz1UcnVlLCBudWxsPVRydWUpLAogICAgICAgICksCiAgICAgICAgbWlncmF0aW9ucy5BbHRlckZpZWxkKAogICAgICAgICAgICBtb2RlbF9uYW1lPSdiaWxsaW5ncmVjZWlwdCcsCiAgICAgICAgICAgIG5hbWU9J3RvUGF5JywKICAgICAgICAgICAgZmllbGQ9bW9kZWxzLkRlY2ltYWxGaWVsZChkZWNpbWFsX3BsYWNlcz04LCBtYXhfZGlnaXRzPTE4KSwKICAgICAgICApLAogICAgICAgIG1pZ3JhdGlvbnMuQWx0ZXJGaWVsZCgKICAgICAgICAgICAgbW9kZWxfbmFtZT0nYmlsbGluZ3JlY2VpcHRoaXN0b3J5cmVjb3JkJywKICAgICAgICAgICAgbmFtZT0ndG9QYXknLAogICAgICAgICAgICBmaWVsZD1tb2RlbHMuRGVjaW1hbEZpZWxkKGJsYW5rPVRydWUsIGRlY2ltYWxfcGxhY2VzPTQsIG1heF9kaWdpdHM9MTgsIG51bGw9VHJ1ZSksCiAgICAgICAgKSwKICAgICAgICBtaWdyYXRpb25zLkRlbGV0ZU1vZGVsKAogICAgICAgICAgICBuYW1lPSdiaWxsaW5nY3VycmVuY3knLAogICAgICAgICksCiAgICAgICAgbWlncmF0aW9ucy5EZWxldGVNb2RlbCgKICAgICAgICAgICAgbmFtZT0nYmlsbGluZ2N1cnJlbmN5aGlzdG9yeXJlY29yZCcsCiAgICAgICAgKSwKICAgIF0K
```
SavelevMatthew, line 2: Don't forget to check migrations before merging

apps/condo/domains/common/components/Table/Renders.tsx
toplenboren: https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormatNo, this is Intl format logic

apps/condo/domains/common/components/Table/Renders.tsx
toplenboren: input: `2400.21000000`

1) Use formatNumber: `// P2,400.21`
2) Replace P with '' `// 240.21`
3) Replace "," with " " `// 2 400.21`
4) Replace "." with "," `// 2 400,21`
5) Find decimal part and real part of the number
6) Recolor the decimal part

apps/condo/domains/common/components/Table/Renders.tsx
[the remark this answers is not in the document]
SavelevMatthew: > <currencyMark><value>.<decimalPart> // $ 20.00 but we need to have it this way: 20.00$`
Are we? I guess we should not fix position of currency. Instead we can create locale-depended resolver, which generates value (and use Intl as fallback?)....But not hard-coded

apps/condo/domains/billing/schema/BillingIntegration.js
SavelevMatthew: > We already force integration to specify (isRequired: true)

Are we? I thought it's working just like default value in DB, so it will be passed, if not specified....It means if I forgot to specify it somewhere, defaultValue will be passed....But no error will be thrown, it can lead to mistakes later

toplenboren: Oh, let me double check it

apps/condo/domains/common/components/Table/Renders.tsx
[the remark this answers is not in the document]
toplenboren: Please see the latest commits, the i18n is completly rewritten

apps/condo/domains/common/components/Table/Renders.tsx
```diff
@@ -89,79 +89,64 @@ export const renderMeterReading = (values: string[], resourceId: string, measure
...
-        const separatorIndex = text.indexOf(partSeparator)
-        const lastSymbol = text.startsWith('-') ? 2 : 1
-        const spaces = []
-        for (let i = separatorIndex - 3; i >= lastSymbol; i -= 3) {
-            spaces.push(i)
-        }
-        const markStyles = { backgroundColor: colors.markColor }
+
+        const formattedCurrency = intl.formatNumberToParts(
+            parseFloat(text),
```
SavelevMatthew, line 137: I guess it whole also work with string...doesn't it?

apps/condo/domains/common/components/Table/Renders.tsx
```diff
@@ -89,79 +89,64 @@ export const renderMeterReading = (values: string[], resourceId: string, measure
...
+            return renderMoney(formattedCurrency)
         }
         return (
             <>
                 <TextHighlighter
                     text={text}
                     search={search}
-                    renderPart={(part, startIndex, marked) =>
-                        recolorMoney(part, startIndex, separatorIndex, spaces, marked ? markStyles : {})}
+                    renderPart={() => renderMoney(formattedCurrency)}
```
SavelevMatthew, line 148: Disable global search? Then may be remove highlighter?

SavelevMatthew (approved):

PR #700, opened 2021-10-11
DOMA 1392 upgrade billing tables

toplenboren:
## Problem

We need to display `unitName` for our user

## Solution

- Add an ability to display `unitName` in table 
- Add an ability to display `unitName` in receipts modal

image

image


dkoviazin (approved):

SavelevMatthew (changes requested):
Table view uses space to separate address from unitName, and in modal there's ", ". Please ask Max to determine correct way of join

SavelevMatthew (approved):
LGTM

PR #726, opened 2021-10-18
DOMA-831 Deny redis connection at build time

MrFoxPro:
Please suggest some better names for env variable at build time.


packages/@core.keystone/setup.utils.js
```diff
@@ -56,9 +56,14 @@ function getAdapter (databaseUrl) {
 }
 
 function prepareDefaultKeystoneConfig (conf) {
```
pahaz, line 58:
```
if (conf.PHASE === 'build') {
    console.warn('prepareDefaultKeystoneConfig() return nothing at build time')
    return {} // really we don't need any keystone settings at build time
}
```

MrFoxPro: No, thats not working, keystone requires session store to be configured event at build time

MrFoxPro:
> 
> 
> Probably, it's better to just don't register any schema in index.js `if (conf.PHASE === 'build')` It's easy to add some checks inside `index.js`.

That's not working, keystone requires schema even at build time

pahaz (approved):

PR #731, opened 2021-10-20
DOMA-1467 Global permissions access plugin

0x241F31:
This API changes allow to manage permission for users. Main goal of this feature is to create service users with limited access to lists and mutations.
Feature logic:
If we have defined field 'permissions' on request user object, we are checking this field and deciding if request is to List/Service legitimate.
If field 'permission' is undefined, we are performing default access checking.

### DIFF
--- a/apps/condo/domains/common/utils/testSchema/index.js
+++ b/apps/condo/domains/common/utils/testSchema/index.js
@@ -132,11 +132,12 @@ export const expectToThrowAccessDeniedErrorToResult = async (testFunc ) => {
  * @param {TestFunc} testFunc - Function, expected to throw an error
  * @return {Promise<void>}
  */
-export const expectToThrowForbiddenErrorToResult = async (testFunc) => {
+export const expectToThrowForbiddenError = async (testFunc, path) => {
     await catchErrorFrom(testFunc, ({ errors, data }) => {
         expect(errors[0]).toMatchObject({
             'message': 'You are not allowed to perform this operation',
             'name': 'ForbiddenError',
+            path,
         })
     })
 }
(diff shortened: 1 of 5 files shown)


apps/condo/domains/common/utils/accessControl.js
```diff
@@ -0,0 +1,54 @@
...
+function accessControl (schemaType, schemaName, schema){
+    if (schemaType === 'GQLListSchema' && isObject(schema.access)) {
+        Object.getOwnPropertyNames(schema.access).forEach(crudOperation => {
+            const originalAccess = schema.access[crudOperation]
+            if (isBoolean(originalAccess)){
+                return
+            }
+            schema.access[crudOperation] = function ({ authentication: { item: user } } ) {
+                if (user && user.permissions) {
+                    const thisPermissionName = `can${capitalize(crudOperation)}${schemaName}s`
```
pahaz, line 15: ${schemaName}s Is not a common case. The keystone have the right model name for such cases

apps/condo/domains/common/utils/accessControl.js
```diff
@@ -0,0 +1,54 @@
...
+                    if (user.permissions[thisPermissionName] !== true) {
```
pahaz, line 16: Why not return just false?

MrFoxPro: @pahaz conceptual solution

apps/condo/domains/common/utils/accessControl.js
```diff
@@ -0,0 +1,54 @@
...
+                }
+                return originalAccess(...arguments)
+            }
+        })
+    }
+    else if (schemaType === 'GQLCustomSchema' && schema.mutations){
+        schema.mutations.forEach(mutation => {
+            const gqlMutationSchema = gql('type A { ' + mutation.schema + ' }')
+            const mutationName = gqlMutationSchema.definitions[0].fields[0].name.value
+            const permissionName = `can${capitalize(mutationName)}`
```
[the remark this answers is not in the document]
MrFoxPro: fixed by upperFirst

PR #774, opened 2021-11-02
DOMA-1526 fetch contacts only for unit

nomerdvadcatpyat:
When we have >1000 contacts in the property, an error occurs about exceeding the server limit (1000 objects)

### DIFF
--- a/apps/condo/domains/contact/components/ContactsEditor/index.tsx
+++ b/apps/condo/domains/contact/components/ContactsEditor/index.tsx
@@ -71,7 +71,7 @@ export const ContactsEditor: React.FC<IContactEditorProps> = (props) => {
     searchContacts(client, {
         organizationId: organization,
         propertyId: property ? property : undefined,
-        unitName: unitName ? unitName : undefined,
+        unitName: unitName ? unitName : null,
     })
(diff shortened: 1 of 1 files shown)


Dimitreee (approved):

sitozzz (approved):

apps/condo/domains/contact/components/ContactsEditor/index.tsx
```diff
@@ -71,7 +71,7 @@ export const ContactsEditor: React.FC<IContactEditorProps> = (props) => {
     searchContacts(client, {
         organizationId: organization,
         propertyId: property ? property : undefined,
-        unitName: unitName ? unitName : undefined,
+        unitName: unitName ? unitName : null,
```
AntonAL, line 74: I think, a case, when no `unitName` is provided, should be handled differently in some explicit way.
Currently, a `Contact` schema does't requires `unitName` to be set.
So, we potentially can have more than 1000 contacts without `unitName` and the same error will arise.
Maybe to not fetch contacts at all?
For example, when `unitName` is not set, call `setContacts([])`.

pahaz (approved):

PR #776, opened 2021-11-03
HOTFIX add robots.txt

pahaz:


apps/condo/public/robots.txt
```diff
@@ -0,0 +1,5 @@
+User-agent: *
+Allow: /auth/
+
+User-agent: *
```
AntonAL, line 4: It seems, like that second `User-agent` declaration is redundant.
```suggestion
```

pahaz: I'm not sure. Lots of examples contain robots like this. Even if it's a redundant block, I don't think it will create any problems.

dkoviazin (approved):

sitozzz (approved):

PR #779, opened 2021-11-08
DOMA-1462 Bug when properties do not load in edit form

dkoviazin:
Bug description
1. GET_ALL_PROPERTIES_BY_VALUE_QUERY - returns only 10 records
2. When GraphQlSearchInput initialized no requests were made to load chosen properties

Before fixing:
УК Изумруд

After fixing:
image

### DIFF
--- a/apps/condo/domains/common/components/GraphQlSearchInput.tsx
+++ b/apps/condo/domains/common/components/GraphQlSearchInput.tsx
@@ -1,7 +1,7 @@
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-nocheck
 import { useApolloClient } from '@core/next/apollo'
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useState, useCallback } from 'react'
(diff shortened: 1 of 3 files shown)


sitozzz (approved):

Dimitreee (approved):

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -64,6 +64,18 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
         ? renderOptions(data, renderOption)
         : data.map(renderOption)
 
+    const initialValueLoad =  useCallback(async () => {
```
Dimitreee, line 67: loadInitialOptions

pahaz (approved):

apps/condo/domains/common/components/GraphQlSearchInput.tsx
dkoviazin: renamed

PR #781, opened 2021-11-08
DOMA-1544 Service Consumer Upgrades 

toplenboren:
## Problem

Our current service consumer model can only be used for billing and acquiring domains, as it requires and explicit link to `billingAccount`

## Solution

Rethink and refactor service consumer model and mutations to allow:

#### Being able to use service consumer for payments

For this to work we need serviceConsumer to have:

- `accountNumber` 
- `billingAccount` link
- `acquiringIntegration` link

#### Being able to use serviceConsumer for meter readings

For this to work we need serviceConsumer to have:

- `accountNumber`
- `organiaztion` equals resident.organization

### Model changes

- `billingAccount` field is not required no more
- new field: `billingIntegrationOrgqnizationContext` - for cases when `billingAccount` is deleted and we need to guess the billingContext
- new field: `acquiringIntegrationContext`
- new required field: `organization` 

### RegisterServiceConsumer changes

The only way to create a serviceConsumer without hacks is `registerServiceConsumer` mutation. Signature (arguments + return value) stays the same, while business logic changes:

#### Before: 

- BEGIN
- Try to find `billingAccount` for this `accountNumber`, `unitName` and `resident`
- If not found: error
- If found: create (or restore) `serviceConsumer` and return it
- END

#### After:

- BEGIN
- Anyways create or restore ServiceConsumer
- END

#### Case 1: Mobile user tries to pay, then tries to send meter readings

- Register user
- Register resident (RegisterResidentMutation)
- Wait until Resident.Organization has billing data
- Register ServiceConsumer (RegisterServiceConsumerMutation) 
- ConnectServiceConsumerToBillingAccount
- If service consumer has `billingAccount` and `acquiringIntegrationContext`, then you can see receipts and pay!
- Created service consumer should have `accountNumber` as it is required field, so you can send meter readings!

#### Case 2: Mobile user tries to send meter readings

- Register user
- Register resident (RegisterResidentMutation)
- Register ServiceConsumer (RegisterServiceConsumerMutation) 
- You can send meter readings
- Wait until Resident.Organization has billing data
- ConnectServiceConsumerToBillingAccount (RegisterServiceConsumerMutation)
- If everything is ok - you can see receipts and pay!

---

> Please be advised that this PR is partial by design :) we need to unblock @nomerdvadcatpyat 

Todo in further tasks

- Add `connectServiceConsumerToBillingAccount` mutation
- Add `magic` of automatic billing changing


apps/condo/domains/resident/schema/RegisterServiceConsumerService.js
```diff
@@ -2,28 +2,20 @@
  * Generated by `createservice resident.RegisterServiceConsumerService --type mutations`
  */
 
+const { AcquiringIntegrationContext } = require('../../acquiring/utils/serverSchema')
```
SavelevMatthew, line 5: Use domain import

nomerdvadcatpyat (approved):

apps/condo/domains/resident/schema/ServiceConsumer.js
```diff
@@ -56,6 +56,15 @@ const ServiceConsumer = new GQLListSchema('ServiceConsumer', {
             type: Text,
             isRequired: true,
         },
+
+        organization: {
+            schemaDoc: 'Organization to which this service consumer pays to object',
```
SavelevMatthew, line 61: `models.CASCADE`? Hmmm 🤔 Sounds like it's ORGANIZATION_OWNED_FIELD

SavelevMatthew (approved):
Don't forget to check migration actuality before merging :)

apps/condo/domains/resident/schema/ServiceConsumer.js
toplenboren: I think we should be able to update the organization, but not quite sure. If so - this is really ORGANIZATION OWNED FIELD

PR #797, opened 2021-11-11
DOMA-1512 Support can create subscriptions

dkoviazin:
For an old organization, without trial subscriptions, we want to set it.

now, we add access for supports to create subscriptions in admin UI
Later we will make a script in /bin to set all subscriptions for old organizations


SavelevMatthew (approved):
Seems ok if it's approved by our Products

AntonAL (approved):

PR #802, opened 2021-11-11
DOMA-1593 function `buildBicryptId` with tests

AntonAL:
Builds bicryptId string in format, required by SBBOL API
TODO: use `Ajv` for arguments validation.


AntonAL:
Will open new pull-request with all stages of getting Certificate Signing Request (CSR)

PR #804, opened 2021-11-11
DOMA-1611 normalize filter select options

nomerdvadcatpyat:
The antd select displays the object's `label` field by default. Brought all the select options that are used in filters to one type.

also update columns width in `meter` page


apps/condo/domains/common/utils/filters.utils.ts
```diff
@@ -174,6 +174,10 @@ export function getFilterDropdownByKey <T> (filterMetas: Array<FiltersMeta<T>>,
     }
 }
 
+export function convertToOptions <T> (objects: T[], labelField: string, valueField: string): OptionType[] {
+    return objects.map(object => ({ label: object[labelField], value: object[valueField] }))
```
akarjakin, line 178: if `object` is  not an `Object` really, it will throw here. so please use `get(object, labelField)` instead

akarjakin (approved):

apps/condo/domains/common/utils/filters.utils.ts
nomerdvadcatpyat: fixed in 8b91092

dkoviazin (approved):

PR #806, opened 2021-11-12
DOMA-1613 adde apartment number to contacts table in callcenter, refactoring

akarjakin:


nomerdvadcatpyat (approved):

apps/condo/domains/common/components/TableFilter.tsx
```diff
@@ -79,14 +90,41 @@ export const getTextFilterDropdown = (columnName: string, setFiltersApplied: Rea
                 // @ts-ignore
                 value={selectedKeys}
                 onChange={(e) => {
-                    // @ts-ignore
-                    setSelectedKeys(e.target.value)
-                    setFiltersApplied(true)
-                    confirm({ closeDropdown: false })
+                    if (isFunction(setSelectedKeys)) setSelectedKeys(e.target.value)
```
MrFoxPro, line 93: Is it ok to shorten this to `setSelectedKeys && setSelectedKeys(e.target.value)`?

yarn.lock
```diff
@@ -565,11 +565,6 @@
   dependencies:
```
MrFoxPro, line 565: Not sure we need lock file in this pr

.gitignore
```diff
@@ -16,7 +16,6 @@ __media
 _example.js
 codegen.yaml
 schema.graphql
-
```
MrFoxPro, line 19: ?

apps/condo/domains/common/components/TableFilter.tsx
akarjakin: actually we want to check if setSelectedKeys not only exists, bat also is a function, that's why I always recommend using such checks in places, where code can throw exception.

.gitignore
akarjakin: it's after yarn install, anyway someone will merge this sooner or later :)

yarn.lock
akarjakin: it's after yarn install, anyway someone will merge this sooner or later :)

.gitignore
akarjakin: I added submodules here, then removed (not needed actually), and looks like I just removed meaningless empty row too.

yarn.lock
akarjakin: looks like it's better to commit yarn.lock https://stackoverflow.com/questions/39990017/should-i-commit-the-yarn-lock-file-and-what-is-it-for

PR #832, opened 2021-11-19
DOMA-1356 mobile menu redesign

MrFoxPro:
image

PR #846, opened 2021-11-22
DOMA-1634 add click trigger to dropdowns

nomerdvadcatpyat:
By default, dropdown is triggered only on `hover`, on mobile devices it did not always work, added `click` trigger

### DIFF
--- a/apps/condo/domains/common/components/ResidentActions/ResidentActions.tsx
+++ b/apps/condo/domains/common/components/ResidentActions/ResidentActions.tsx
@@ -44,6 +44,8 @@ interface IResidentActionsProps {
     minified: boolean
 }
 
+const RESIDENT_ACTIONS_OPEN_DROPDOWN_TRIGGERS: ('hover' | 'click' | 'contextMenu')[] = ['hover', 'click']
+
 export const ResidentActions: React.FC<IResidentActionsProps> = (props) => {
     const intl = useIntl()
     const { minified } = props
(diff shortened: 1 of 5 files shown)


apps/condo/domains/common/components/ResidentActions/ResidentActions.tsx
```diff
@@ -44,6 +44,8 @@ interface IResidentActionsProps {
     minified: boolean
 }
 
+const RESIDENT_ACTIONS_OPEN_DROPDOWN_TRIGGERS: ('hover' | 'click' | 'contextMenu')[] = ['hover', 'click']
```
sitozzz, line 47: isn't this type defined in the library and we need to describe it manually?

sitozzz (approved):

apps/condo/domains/common/components/ResidentActions/ResidentActions.tsx
sitozzz: May be better to import `DropDownProps` from `antd` and use `DropDownProps['trigger']` as type for your constant?

nomerdvadcatpyat: > isn't this type defined in the library and we need to describe it manually?

This type is not defined in the library:
изображение



> May be better to import `DropDownProps` from `antd` and use `DropDownProps['trigger']` as type for your constant?

Yes, it's a good idea, I'll do it now.

MrFoxPro:
This will conflict with https://github.com/open-condo-software/condo/pull/832

dkoviazin (approved):

PR #847, opened 2021-11-22
DOMA-1615 mobile api upgrades

toplenboren:
## Problem:

IOS and Android clients need a way to explicitly tell whether result of `registerServiceConsumer` mutation (created serviceConsumer) has `billingAccount` field

Right now `ServiceConsumer.billingAccount` is an optional field, thus when we fail to get it we do not throw an Error, but instead create ServiceConsumer without `billingAccount`

> Creating separate mutation, like `connectServiceConsumer` is possilbe, but opens up a new `possible fail point` - new http request

## Solution

Add `validations:[String]` argument to the `registerServiceConsumer` mutation. If you fill it with fields then fields that are specified are going to be checked for nil. If check fails -> mutation would throw an error

```(js)
> await registerServiceConsumer(... validations:['billingAccount'])
> Error: billingAccount not found

> await registerServiceConsumer(...) // no validations are specified
> pass // serviceConsumer is created
```

## PS

I fixed the error when we did not use the `tin` argument, because we did not know the way to filter an `Organizaiton` by its `tin`. It seems like this is critical now :(


apps/condo/domains/resident/schema/RegisterServiceConsumerService.js
```diff
@@ -56,35 +60,33 @@ const RegisterServiceConsumerService = new GQLCustomSchema('RegisterServiceConsu
...
                 if (!tin) { throw new Error(`${REQUIRED_NO_VALUE_ERROR}tin] tin null or empty: ${tin}`) }
 
-                // Todo (DOMA-1604) add filtering by tin
-                // Right now we don't actaully use tin for getting an organization, but get organization from resident. Reasons:
-                // 1. We can't optimally filter on tin: (DOMA-1604)
-                // 2. Right now it's okay to get organiation from resident. We would need to get organization by tin when we add new Acquiring
-                // const [ organization ] = await Organization.getAll(context, { meta_in: { tin: tin } })
-                const [ organization ] = await Organization.getAll(context, { id: get(resident, ['organization', 'id']) })
+                // GraphQL from Keystone does not supports querying of database fields of type JSON.
+                const { keystone } = await getSchemaCtx('Organization')
```
dkoviazin, line 64: Better to move this to some function, as it will be refactored after TIN will have its own field in the Organization model

dkoviazin (approved):

apps/condo/domains/resident/schema/RegisterServiceConsumerService.js
```diff
@@ -96,9 +98,22 @@ const RegisterServiceConsumerService = new GQLCustomSchema('RegisterServiceConsu
                     organization: { connect: { id: organization.id } },
                 }
 
+                const optionalAttrs = {
+                    ...paymentFeatureAttrs,
+                }
+
+                // User can request explicit validation of some fields. This is required for iOS and Android mobile clients
+                if (Array.isArray(validations)) {
```
pahaz, line 106: Why it’s optional? Can we just use it by default?

toplenboren: I don't want to supply an empty array as a validations argument, if I don't want any oprtional validations to be run, thus why it is not required

nomerdvadcatpyat (approved):

PR #852, opened 2021-11-23
Doma-1653 | Connect registers

SavelevMatthew:
**Problem**: some of billing integrations can have multiple variants (like format of registers), but flows (at least for condo) are basically the same. Needed to move all repeated common integrations into single one.

**Solution**: add options to integration, to make user choose it after reading integration description...like this
<details>
image
</details>
and if options inside integration is specified => billing context should have it too...
<br><br>

**Possible usage**: (choose data source for automatic integration, like "get data from GIS" and "get data from BRIZ"

**IMPORTANT NOTE**: All of fields have Text format....changing it to LocalizedText will be done in scope of all billing in another task

TODO:

- [x] Add options to Integration
- [x] Add option to context
- [x] Add validations to context options
- [x] Change `yarn generate-billing`
- [x] Change front-end behaviour for option-containing integrations

 ### IMPORTANT NOTE 2
Current layout has some new styles in it...and seem like it will be global, that's why I not implemented it yet before request from design arrived

image


bin/generate-billings.js
```diff
@@ -56,30 +56,56 @@ const InProgressBilling = {
...
+    detailsText: 'Выберите формат, в котором хотите загружать ваши реестры в Домá.\n' +
+        'Мы запомним ваш выбор и в следующий раз не будем спрашивать про формат файлов.',
     detailsConfirmButtonText: 'Подключить',
-    detailsInstructionButtonText: 'Подробнее про шаблон',
-    detailsInstructionButtonLink: 'https://www.sberbank.ru/ru/s_m_business/new_sbbol',
     contextDefaultStatus: BILLING_INTEGRATION_ORGANIZATION_CONTEXT_FINISHED_STATUS,
     billingPageTitle: 'Биллинг, реестровый обмен',
-    dataFormat: Lvl1DataFormat,
+    dataFormat: Lvl3DataFormat,
+    availableOptions: {
```
toplenboren, line 68: What if we have a billing integration which would require more in-depth setup? Like an input field with special API key?

apps/condo/domains/billing/schema/BillingIntegration.test.js
```diff
@@ -48,6 +49,33 @@ describe('BillingIntegration', () => {
                 expect(err).toBeDefined()
             })
         })
+
+        test('Can be created with options', async () => {
+            const support = await makeClientWithSupportUser()
+            const firstOption = { name: '1C', billingPageTitle: 'Биллиг "Реестры 1C"', descriptionDetails: { detailsText: 'о формате', detailsLink: faker.internet.url() } }
+            const secondOption = { name: 'Сббол 9_2', billingPageTitle: 'Биллиг "Реестрыыыыыы"' }
+            const thirdOption = { name: 'Сббол 8_1', descriptionDetails: { detailsText: 'о формате', detailsLink: faker.internet.url() } }
```
toplenboren, line 57: I need some documentation on how to create billing with options

bin/generate-billings.js
toplenboren: If we use a reusable solution I'd saw it like this

A form that is encoded into JSON, which is rendered on stage of creating a BillingContext and supplied to BillingContext.meta

Pros:
- It would cover potentially all cases
- It would be reusable right from the start (_in Registry Exchange billing form input_)

Cons:
- It would require more time (Code + Docs)
- Support or non-tech teammates wont be able to create billing context manually without extensive documentation

> Maybe use this https://nikolamargit.dev/dynamic-form-using-ant-design/

> I did not do any research but have a hypothesis that it's easy to find a library that does this form rendering

bin/generate-billings.js
```diff
@@ -56,30 +56,56 @@ const InProgressBilling = {
...
+        title: 'Формат ваших реестров',
+        options: [
+            {
+                name: '1C',
+                displayName: '1C',
+                billingPageTitle: 'Интеграция через загрузку реестров 1С',
+                descriptionDetails: {
+                    detailsText: 'о формате',
+                    detailsLink: 'https://help.doma.ai',
+                },
```
toplenboren, line 78: Why not use some common JSON format for this? Like ANT?

bin/generate-billings.js
```diff
@@ -56,30 +56,56 @@ const InProgressBilling = {
...
+    dataFormat: Lvl3DataFormat,
+    availableOptions: {
+        title: 'Формат ваших реестров',
+        options: [
+            {
+                name: '1C',
+                displayName: '1C',
+                billingPageTitle: 'Интеграция через загрузку реестров 1С',
+                descriptionDetails: {
+                    detailsText: 'о формате',
```
toplenboren, line 76: detailsText -> name (since its under *descriptionDetails*) and you use name instead of text on root

bin/generate-billings.js
```diff
@@ -56,30 +56,56 @@ const InProgressBilling = {
...
+                    detailsLink: 'https://help.doma.ai',
```
toplenboren, line 77: detailsLink -> url (since its under *descriptionDetails*)

apps/condo/domains/billing/schema/BillingIntegrationOrganizationContext.js
```diff
@@ -74,6 +80,38 @@ const BillingIntegrationOrganizationContext = new GQLListSchema('BillingIntegrat
             },
         },
 
+        integrationOption: {
```
toplenboren, line 83: Why not place this data to settings ?

toplenboren: BillingIntegrationOrganiationContext.settings

We use this field specifically for this and it is not changed

apps/condo/domains/billing/schema/BillingIntegrationOrganizationContext.js
```diff
@@ -74,6 +80,38 @@ const BillingIntegrationOrganizationContext = new GQLListSchema('BillingIntegrat
...
+            schemaDoc: 'Name of billing integration option, if it has more than 1 variants. Example: registry format. Using "name" as unique identifier inside single billing',
+            isRequired: false,
+            type: Text,
+
```
toplenboren, line 87:
```suggestion
```

toplenboren (changes requested):

apps/condo/pages/settings/integration/[id]/index.tsx
```diff
@@ -140,7 +162,7 @@ const BillingIntegrationDetailsPage = () => {
                                     />
                                     <PageContent>
                                         <Col span={20}>
-                                            <Row gutter={[0, 50]}>
+                                            <Row gutter={[0, 40]}>
```
Comment (akarjakin) [inline apps/condo/pages/settings/integration/[id]/index.tsx:165]: it's better to move such unchanging values (literal array/object declarations) to constants, to persist variable link between renders in order not to cause unnecessary re-renders

apps/condo/pages/settings/integration/[id]/index.tsx
```diff
@@ -162,6 +184,47 @@ const BillingIntegrationDetailsPage = () => {
...
                                                 }
+                                                {
+                                                    options && optionsTitle && (
+                                                        <>
+                                                            <Col span={24}>
+                                                                <Typography.Title level={4}>
+                                                                    {optionsTitle}
+                                                                </Typography.Title>
+                                                            </Col>
+                                                            <Col span={24} style={{ paddingBottom: 20 }}>
```
Comment (akarjakin) [inline apps/condo/pages/settings/integration/[id]/index.tsx:195]: constant styles would be better

apps/condo/pages/settings/integration/[id]/index.tsx
```diff
@@ -162,6 +184,47 @@ const BillingIntegrationDetailsPage = () => {
...
+                                                                            })}
+                                                                        </Space>
+                                                                    </Radio.Group>
+                                                                </Tooltip>
+                                                            </Col>
+                                                        </>
+                                                    )
+                                                }
                                                 <Col span={24}>
                                                     <Space size={20} style={{ width: '100%', flexWrap: 'wrap' }}>
```
Comment (akarjakin) [inline apps/condo/pages/settings/integration/[id]/index.tsx:229]: constant styles would be better

bin/generate-billings.js
SavelevMatthew: What do you mean by that?

apps/condo/domains/billing/schema/BillingIntegration.test.js
SavelevMatthew: I'll write confluence page if needed, but for now example can be found in `bin/generate-billings`

bin/generate-billings.js
SavelevMatthew: renamed to urlText

bin/generate-billings.js
SavelevMatthew: renamed to `url`

apps/condo/domains/billing/schema/BillingIntegrationOrganizationContext.js
SavelevMatthew: settings is more about keys, links.....generally speaking, it's using like .env

But I think it's not good idea to store billing format together :(

bin/generate-billings.js
SavelevMatthew: I think support should be able to get / change it without lots of problems...Since there'll be some cases when clients will want to change billing formats

SavelevMatthew: Think like dynamic forms here are a bit yagni

SavelevMatthew: ## Result after conversation:
We don't know for now, if we will even need something like this later.....
and we are not even sure that the current interface and flow will remain so

We decided to left this variant for MVP. And we'll change it later

bin/generate-billings.js
SavelevMatthew: We'll describe format in confluence page

apps/condo/domains/billing/schema/BillingIntegrationOrganizationContext.js
SavelevMatthew: fixed

apps/condo/pages/settings/integration/[id]/index.tsx
Comment (SavelevMatthew) [reply apps/condo/pages/settings/integration/[id]/index.tsx]: fixed

apps/condo/pages/settings/integration/[id]/index.tsx
Comment (SavelevMatthew) [reply apps/condo/pages/settings/integration/[id]/index.tsx]: fixed

apps/condo/pages/settings/integration/[id]/index.tsx
Comment (SavelevMatthew) [reply apps/condo/pages/settings/integration/[id]/index.tsx]: fixed

toplenboren (approved):

dkoviazin (approved):

PR #856, opened 2021-11-24
DOMA-1517 infinity address select scrolling

nomerdvadcatpyat:
**Added infinity scrolling to property Select on:**
- Division form
- Ticket filters modal
- Meter filters modal
- Reports page

---
**Why is the scrolling logic in `GraphQLSearchInput` and `BaseSearchInput` repeated?**

Currently, `PropertyAddressSearchInput` and, in some places, `GraphqlSearchInput` are used to find `properties` in the organization. `PropertyAddressSearchInput` is based on `BaseSearchInput` and `GraphQLSearchInput` is based on antd `Select` component. Converting `GraphQLSearchInput` to `BaseSearchInput` is not a quick task and affects many of the components that use `GraphQLSearchInput`.

---
**Why is there no scroll on the `property/create` or `property/update` pages?**
Because the Dadata service can return a maximum 20 addresses per search and does not support pagination.

изображение

### DIFF
--- a/apps/condo/domains/common/components/BaseSearchInput/index.tsx
+++ b/apps/condo/domains/common/components/BaseSearchInput/index.tsx
@@ -169,12 +169,17 @@ export const BaseSearchInput = <S extends string>(props: ISearchInput<S>) => {
     const options = useMemo(
         () => {
             const dataOptions = data.map((option, index) => renderOption(option, value, index))
+            
+            if (!fetching) return dataOptions
 
-            return fetching ? [...dataOptions, (
-                <Option key={'loader'} value={null} disabled>
-                    <Loader style={SELECT_LOADER_STYLE}/>
-                </Option>
-            )] : dataOptions
+            return [
+                ...dataOptions,
+                (
+                    <Option key={'loader'} value={null} disabled>
+                        <Loader style={SELECT_LOADER_STYLE}/>
+                    </Option>
+                ),
+            ]
         },
         [data, fetching, value],
     )
--- a/apps/condo/domains/common/components/BaseSearchInput/index.tsx
+++ b/apps/condo/domains/common/components/BaseSearchInput/index.tsx
@@ -50,7 +50,7 @@ export const BaseSearchInput = <S extends string>(props: ISearchInput<S>) => {
     const [selected, setSelected] = useState('')
     const [fetching, setFetching] = useState(false)
     const [data, setData] = useState([])
-    const [isAllData, setIsAllData] = useState(false)
+    const [isAllDataLoaded, setIsAllDataLoaded] = useState(false)
     const [searchValue, setSearchValue] = useState('')
     const [initialOptionsLoaded, setInitialOptionsLoaded] = useState(false)
     const [initialValue, isInitialValueFetching] = useInitialValueGetter(value, initialValueGetter)
@@ -58,7 +58,7 @@ export const BaseSearchInput = <S extends string>(props: ISearchInput<S>) => {
 
     const searchSuggestions = useCallback(
         async (value) => {
-            setIsAllData(false)
+            setIsAllDataLoaded(false)
             setFetching(true)
             const data = await search((selected) ? selected + ' ' + value : value)
             setFetching(false)
@@ -76,7 +76,7 @@ export const BaseSearchInput = <S extends string>(props: ISearchInput<S>) => {
 
     const searchMoreSuggestions = useCallback(
         async (value, skip) => {
-            if (isAllData) return
+            if (isAllDataLoaded) return
 
             setFetching(true)
             const data = await search(selected ? undefined : value, skip)
@@ -85,10 +85,10 @@ export const BaseSearchInput = <S extends string>(props: ISearchInput<S>) => {
             if (data.length > 0) {
                 setData(prevData => [...prevData, ...data])
             } else {
-                setIsAllData(true)
+                setIsAllDataLoaded(true)
             }
         },
-        [isAllData, search, selected],
+        [isAllDataLoaded, search, selected],
     )
 
     const throttledSearchMore = useMemo(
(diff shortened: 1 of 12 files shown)


apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -53,15 +58,17 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
         initialValue,
         getInitialValueQuery,
+        infinityScroll,
         ...restProps
     } = props
     const client = useApolloClient()
     const [selected, setSelected] = useState('')
     const [isLoading, setLoading] = useState(false)
     const [data, setData] = useState([])
+    const [isAllData, setIsAllData] = useState(false)
```
sitozzz, line 68: If this flag have meaning like "when all data has been loaded to list" may be better to name it like `isAllDataLoaded`?

sitozzz (approved):

apps/condo/domains/common/components/BaseSearchInput/index.tsx
```diff
@@ -98,15 +133,23 @@ export const BaseSearchInput = <S extends string>(props: ISearchInput<S>) => {
...
 
     const handleChange = (value, options) => {
-        setSearchValue(value) 
+        setSearchValue(value)
         onChange(value, options)
     }
-    
+
+    const handleScroll = useCallback(async (scrollEvent) => {
+        const lastElementId = String((data || []).length - 1)
```
akarjakin, line 141: `lastElementIdx` array index is not an id actually :)

apps/condo/domains/common/components/BaseSearchInput/index.tsx
```diff
@@ -124,8 +167,16 @@ export const BaseSearchInput = <S extends string>(props: ISearchInput<S>) => {
     )
 
     const options = useMemo(
-        () => data.map((option) => renderOption(option, restSelectProps.value)),
-        [data, fetching, restSelectProps.value],
+        () => {
+            const dataOptions = data.map((option, index) => renderOption(option, value, index))
+
+            return fetching ? [...dataOptions, (
```
akarjakin, line 173: This way it would be a bit more readable
```
            if (!fetching) return dataOptions

            return [
                ...dataOptions, 
                (
                    <Option key={'loader'} value={null} disabled>
                        <Loader style={SELECT_LOADER_STYLE}/>
                    </Option>
                )
            ]
```

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -130,6 +139,46 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
+            if (data.length > 0) {
+                setData(prevData => [...prevData, ...data])
+            } else {
+                setIsAllData(true)
+            }
+        },
+        [],
+    )
+
+    const throttledSearchMore = useMemo(
```
akarjakin, line 167: looks like some of the functions here could be made reusable by moving to useInfiniteScroll hook

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -130,6 +139,46 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
+        () => {
+            return throttle(searchMoreSuggestions, DEBOUNCE_TIMEOUT)
+        },
+        [searchMoreSuggestions],
+    )
+
+    const handleScroll = async (scrollEvent) => {
```
akarjakin, line 174: useCallback

apps/condo/domains/common/components/BaseSearchInput/index.tsx
```diff
@@ -98,15 +133,23 @@ export const BaseSearchInput = <S extends string>(props: ISearchInput<S>) => {
     )
 
     const handleChange = (value, options) => {
```
akarjakin, line 135: useCallback

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -130,6 +139,46 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
+        const lastElementId = String((data || []).length - 1)
```
akarjakin, line 175: `lastElementIdx`

apps/condo/domains/common/utils/select.utils.ts
```diff
@@ -0,0 +1,10 @@
+
+export function isNeedToLoadNewElements (scrollEvent, lastElementId: string, isLoading: boolean) {
```
akarjakin, line 2: what about `checkIfScrollChanged`?

apps/condo/domains/property/components/PropertyAddressSearchInput.tsx
```diff
@@ -53,13 +53,14 @@ export const PropertyAddressSearchInput: React.FC<IAddressSearchInput> = (props)
      * TODO(DOMA-1513) replace HighLighter with apps/condo/domains/common/components/TextHighlighter.tsx and renderHighlightedPart
      */
     const renderOption = useCallback(
-        (dataItem, searchValue) => {
+        (dataItem, searchValue, index) => {
             return (
                 <Select.Option
                     style={{ direction: 'rtl', textAlign: 'left', color: grey[6] }}
```
akarjakin, line 59: constant please

apps/condo/domains/common/utils/select.utils.ts
nomerdvadcatpyat: no, it's not just check if scroll changed. This function returns true if we scrolled to the last element

apps/condo/domains/common/components/GraphQlSearchInput.tsx
SavelevMatthew: Seems like yes, but I guess it should be in another task, if current is having increased priority, isn't it?

apps/condo/domains/common/utils/select.utils.ts
```diff
@@ -0,0 +1,10 @@
+
+export function isNeedToLoadNewElements (scrollEvent, lastElementIdx: string, isLoading: boolean) {
```
SavelevMatthew, line 2: Hoe about adding some tests for utils as we usually do?

SavelevMatthew (approved):
LGTM, but I think utils should be covered with tests

PR #861, opened 2021-11-25
DOMA-1690 add 'active' flag to the payment category

toplenboren:
We currently support only Квартплата -- so let's return only Квартплата

image


nomerdvadcatpyat (approved):

SavelevMatthew (approved):
LGTM, but make tests great again (I also recommend to avoid notation like `[0]`, since test can be broken, if order in consts changes )

PR #878, opened 2021-11-26
DOMA-1661 Logging integration actions for create receipt

dkoviazin:
Add select to choose resident 
Fix signin page (smsCode needs to be Integer)
Fix saveEpsReceipt (missing await)
Add Logs at last steps = on service consumer create and receipt create


AntonAL (approved):
You can use Pino JSON-logger. It's easier to track, because for each record a timestamp is saved and any metadata can be attached.
Example of initialization can be found in `apps/condo/domains/organization/integrations/sbbol/common.js` and usage in `apps/condo/domains/organization/integrations/sbbol/sync/syncPaymentRequestsForSubscriptions.js`.

PR #884, opened 2021-11-29
DOMA-1521 property map copy section

sitozzz:
- [x] Copy existing section from creation form
- [x] Some menu items will be disabled based on the current state of PropertyMap


apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -279,6 +279,10 @@ export const BuildingPanelEdit: React.FC<IBuildingPanelEditProps> = (props) => {
     }, [id, push])
 
     const menuClick = useCallback((event) => {
+        if (INSTANT_ACTIONS.includes(event.key)) {
+            Map[event.key]()
```
akarjakin, line 283: Naming convention that is used for `Map` is better to use for classes. Class instance is an object, so `mapEdit` would be better here.

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -379,6 +383,8 @@ export const BuildingPanelEdit: React.FC<IBuildingPanelEditProps> = (props) => {
                             addUnit: <UnitForm Builder={Map} refresh={refresh}/>,
                             editSection: <EditSectionForm Builder={Map} refresh={refresh}/>,
                             editUnit: <UnitForm Builder={Map} refresh={refresh}/>,
+                            removeBasement: <RemoveBasementForm Builder={Map} refresh={refresh} />,
```
akarjakin, line 386: all of props are elements of an object actually, so `Builder` should be `builder` actually

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -657,44 +707,69 @@ const AddSectionForm: React.FC<IAddSectionFormProps> = ({ Builder, refresh }) =>
...
+
+    useEffect(() => {
+        if (copyId !== null) {
+            const sectionToCopy = Builder.map.sections.find((section) => section.id === copyId)
+            const floorIndexes = sectionToCopy.floors.map((floor) => floor.index)
+            setMinFloor(Math.min(...floorIndexes))
+            setMaxFloor(Math.max(...floorIndexes))
+            setUnitsOnFloor(get(sectionToCopy, 'floors.0.units.length', 0))
+        }
+    }, [copyId])
```
akarjakin, line 734: please make sure all function member variables and props are included in deps

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -863,53 +938,86 @@ const MODAL_FORM_BUTTON_GUTTER: RowProps['gutter'] = [0, 16]
...
         </Row>
     )
 }
+
+interface IRemoveBasementForm {
+    Builder: MapEdit
+    refresh(): void
+}
+
+const RemoveBasementForm: React.FC<IRemoveBasementForm> = ({ Builder, refresh }) => {
```
akarjakin, line 972: `Builder` -> `builder `

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -863,53 +938,86 @@ const MODAL_FORM_BUTTON_GUTTER: RowProps['gutter'] = [0, 16]
...
+        </Row>
+    )
+}
+
+interface IRemoveAtticForm {
+    Builder: MapEdit
+    refresh(): void
+}
+
+const RemoveAtticForm: React.FC<IRemoveAtticForm> = ({ Builder, refresh }) => {
```
akarjakin, line 1001: `Builder` -> `builder `

apps/condo/domains/property/components/panels/Builder/BuildingPanelView.tsx
```diff
@@ -55,8 +58,14 @@ const CHESS_SCROLL_CONTAINER_STYLE: React.CSSProperties = {
     width: '100%',
     overflowY: 'hidden',
 }
+const FULL_SIZE_UNIT_STYLE: React.CSSProperties = { width: '100%', marginTop: '8px', display: 'block' }
 
 export const PropertyMapView: React.FC<IPropertyMapViewProps> = ({ Builder, refresh }) => {
```
akarjakin, line 63: `Builder` -> `builder `

apps/condo/domains/property/components/panels/Builder/BuildingPanelView.tsx
```diff
@@ -92,50 +101,74 @@ export const PropertyMapView: React.FC<IPropertyMapViewProps> = ({ Builder, refr
...
-                                                >{section.name}</UnitButton>
-                                            </div>
-                                        )
-                                    })
+                                                    style={FULL_SIZE_UNIT_STYLE}
+                                                >{BasementTitlePrefix} {basement.label}</UnitButton>
+                                            ))}
+                                            <UnitButton
+                                                secondary
+                                                style={{ width: '100%', marginTop: '8px' }}
```
akarjakin, line 167: constant please

apps/condo/lang/en.json
```diff
@@ -325,8 +328,10 @@
   "pages.condo.property.select.option.ceil": "Ceil",
   "pages.condo.property.modal.title.addSection": "Add section",
   "pages.condo.property.modal.title.addUnit": "Add unit",
-  "pages.condo.property.modal.title.editSection": "Edit section",
+  "pages.condo.property.modal.title.editSection": "Remove section",
```
akarjakin, line 331: maybe it'd be better to rename `editSection` to `removeSection` too ?

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -265,20 +265,24 @@ export const BuildingPanelEdit: React.FC<IBuildingPanelEditProps> = (props) => {
...
+    }, [mapEdit, refresh])
 
 
     const onCancel = useCallback(() => {
         push(`/property/${id}`)
     }, [id, push])
 
     const menuClick = useCallback((event) => {
+        if (INSTANT_ACTIONS.includes(event.key)) {
+            mapEdit[event.key]()
```
akarjakin, line 283: `if (isFunction(mapEdit[event.key])) mapEdit[event.key]()`

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -453,27 +457,27 @@ const SCROLL_CONTAINER_EDIT_PADDING = '330px'
 const MENU_COVER_MAP_WIDTH = 800
 
 const ChessBoard: React.FC<IChessBoardProps> = (props) => {
```
akarjakin, line 459: can we move components to separate files, so that 1 component = 1 file ?

apps/condo/domains/property/components/panels/Builder/BuildingPanelView.tsx
```diff
@@ -92,55 +95,53 @@ export const PropertyMapView: React.FC<IPropertyMapViewProps> = ({ Builder, refr
...
+                                        <MapSectionContainer
+                                            key={section.id}
+                                            visible={builder.isSectionVisible(section.id)}
+                                        >
+                                            {
+                                                builder.possibleChosenFloors.map(floorIndex => {
+                                                    const floorInfo = section.floors.find(floor => floor.index === floorIndex)
+                                                    if (floorInfo && floorInfo.units.length) {
+                                                        return (
+                                                            <div key={floorInfo.id} style={{ display: 'block' }}>
```
akarjakin, line 113: constant styles please

apps/condo/domains/property/components/panels/Builder/BuildingPanelView.tsx
```diff
@@ -92,55 +95,53 @@ export const PropertyMapView: React.FC<IPropertyMapViewProps> = ({ Builder, refr
...
-                                        )
-                                    })
+                                    builder.sections.map(section => (
+                                        <MapSectionContainer
+                                            key={section.id}
+                                            visible={builder.isSectionVisible(section.id)}
+                                        >
+                                            {
+                                                builder.possibleChosenFloors.map(floorIndex => {
+                                                    const floorInfo = section.floors.find(floor => floor.index === floorIndex)
```
akarjakin, line 110: can we sort `section.floors` once and get data just using index like `section.floors[floorIndex]` ?
searching array this way is quite expensive

other possible solution is to convert `section.floors` to object with keys = floorIndexes

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
sitozzz: @akarjakin IMO it might make sense if it were reusable components. So there just local components, that could be used once

apps/condo/domains/property/components/panels/Builder/BuildingPanelView.tsx
sitozzz: @akarjakin  we can't use index here because of `floor.index` is a label, that might be negative

sitozzz: @akarjakin Solution with change `sectiion.floors` data type to object is good if not for the need to explicitly define the type of each component. Main problem is that we cannot explicitly describe the data type, because it would be `Record<string, floor>` with an infinite number of keys

pahaz (approved):

akarjakin (approved):

PR #886, opened 2021-11-29
Doma-1709 Fix broken tickets

SavelevMatthew:
**Problem**: 9 of our tickets in prod environment have client linked, but clientName + clientPhone was not setup for them (they were created before fix), so.....
**Solution**: this script will fix an issues, similar to fix property maps


pahaz (commented):
You need to add it to condo/package.json

akarjakin (approved):

PR #887, opened 2021-11-29
Feature/condo/doma 1289/prolong subscriptions till 20211201 for some organizations

AntonAL:
A following script has been implemented, which sets "finishAt" to specified value for ServiceSubscription records,
which belonging to organizations with ids from specified file,
and which satisfies filtering conditions
Usage:
 yarn workspace @app/condo subscription:prolong <type> <isTrial> <date> <finishAt> <filePathWithOrganizationIds>
 <type>
     Value for `ServiceSubscription.type` for filtering condition of records, that will be affected with this script
 <isTrial>
     Value for `ServiceSubscription.isTrial` for filtering condition of records, that will be affected with this scrip
 <date>
     New value of `ServiceSubscription.finishAt` for records, that will be affected with this script
 <filePathWithOrganizationIds>
     Path to file, that contains a list of organization ids, whose subscriptions should be altered. Each id should be 
Example:
     yarn workspace @app/condo subscription:prolong default true 2021-12-01T12:00:00.000Z ./ids.txt

### DIFF
--- a/apps/condo/package.json
+++ b/apps/condo/package.json
@@ -25,7 +25,8 @@
     "refresh-sbbol-tokens": "node ./../../bin/sbbol.js",
     "create-integration": "node ./../../bin/create-integration.js",
     "sbbol:crypto:get": "node ./../../bin/sbbol/crypto.js",
-    "fix:ticket:contacts": "node ./../../bin/fix-tickets-contacts.js"
+    "fix:ticket:contacts": "node ./../../bin/fix-tickets-contacts.js",
+    "subscription:prolong": "node ../../bin/prolong-subscription.js"
   },
   "dependencies": {
     "@2gis/mapgl": "^1.12.0",
--- a/bin/prolong-subscription.js
+++ b/bin/prolong-subscription.js
@@ -1,0 +1,131 @@
+/**
+ * Sets "finishAt" to specified value for ServiceSubscription records,
+ * which belonging to organizations with ids from specified file,
+ * and which satisfies filtering conditions
+ *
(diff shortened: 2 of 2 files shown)


apps/condo/package.json
```diff
@@ -24,7 +24,8 @@
     "lint-i18n-translations": "node ./../../bin/lint-i18n-translations.js",
     "refresh-sbbol-tokens": "node ./../../bin/sbbol.js",
     "create-integration": "node ./../../bin/create-integration.js",
-    "sbbol:crypto:get": "node ./../../bin/sbbol/crypto.js"
+    "sbbol:crypto:get": "node ./../../bin/sbbol/crypto.js",
+    "subscription:prolong": "node ../../bin/prolong-subscription.js"
```
pahaz, line 28: what do you think about domain folders?

AntonAL: Initially, I have done it for sbbol, but due to small amount of scripts, I've reverted that approach to avoid having most folders with one file.

AntonAL: Btw, scripts can be moved from `bin` to `domains/…/scripts`!
It seems like a good idea.

AntonAL: But, it depends… For example SBBOL features covering many domains, — authentication belongs to `user` domain, organization syncing belongs to `organisation` domain, subscriptions etc.

AntonAL: I think, to avoid complexity, — to leave things as is.
When we will have tens scripts, — some sort of organising will be actual.

PR #896, opened 2021-11-30
DOMA-376 Removed error throwing on cookie resetting

SavelevMatthew:
**Problem**: 
Screenshot 2021-11-30 at 17 41 31
**Solution**: do not throw this error and let system do the rest :)))))


toplenboren (approved):

dkoviazin (approved):

AntonAL:
Are you sure, that this is a right place to fix this issue?
It seems, that cookies should be cleaned in `onCompleted` handler of `SIGNOUT_MUTATION`.

akarjakin (approved):

pahaz (approved):

PR #907, opened 2021-12-02
DOMA-1735 fix wrong prop passed to dom warning

MrFoxPro:


AntonAL (commented):
In a task it is a link to "Viewport meta tags should not be used in _document.js's <Head>" documentation page, which says:
> Adding `<meta name="viewport" ...>` in `pages/_document.js` will lead to unexpected results since it cannot be deduped. The viewport tag should be handled by `next/head` in `pages/_app.js`.

How does this changes relates to it?

MrFoxPro:
Wrong task number.

PR #921, opened 2021-12-03
HOTFIX Setting all existing dataFormats to null

SavelevMatthew:


toplenboren (approved):

apps/condo/migrations/20211204030000-87_manual_set_dataFormat_null.js
```diff
@@ -0,0 +1,11 @@
+exports.up = async (knex) => {
+    await knex.raw(`
+        BEGIN;
+        UPDATE "BillingIntegration" SET "dataFormat" = null WHERE true;
```
pahaz, line 4: Do you want to set it for all `BillingIntegration`?

SavelevMatthew: yep

PR #927, opened 2021-12-07
DOMA-1743 Added importId fields to MultiPayment and Payment

SavelevMatthew:


dkoviazin (approved):

apps/condo/domains/acquiring/schema/fields/common.js
```diff
@@ -0,0 +1,11 @@
...
+
+const IMPORT_ID_FIELD = {
+    schemaDoc: 'Id of `acquiring object` which represents current item in base of external service. Used for internal needs of `acquiring integration service`',
+    type: Text,
+    isRequired: false,
+}
+
+module.exports = {
+    IMPORT_ID_FIELD,
+}
```
toplenboren, line 11: We might want to move this field to common, since it is used in billing and acquiring model

toplenboren (approved):
Ok, but please look at the comments

apps/condo/domains/acquiring/schema/fields/common.js
SavelevMatthew: done 👍

PR #928, opened 2021-12-07
Fix/condo/doma 1767/sticky floor axis

sitozzz:
#### Problem: at the moment user cannot see what floor number actually corresponds to the visible section units
image

#### Solution: make floor axis sticky to the left side of parent container
image

### DIFF
--- a/apps/condo/pages/property/[id]/index.tsx
+++ b/apps/condo/pages/property/[id]/index.tsx
@@ -207,7 +207,11 @@ const PropertyIdPage: IPropertyIdPage = () => {
     </>
 }
 
-PropertyIdPage.headerAction = <ReturnBackHeaderAction descriptor={{ id: 'menu.AllProperties' }} path={'/property/'}/>
+PropertyIdPage.headerAction = <ReturnBackHeaderAction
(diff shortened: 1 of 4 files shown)


dkoviazin (approved):

pahaz (approved):

apps/condo/pages/property/[id]/index.tsx
```diff
@@ -207,7 +207,11 @@ const PropertyIdPage: IPropertyIdPage = () => {
     </>
 }
 
-PropertyIdPage.headerAction = <ReturnBackHeaderAction descriptor={{ id: 'menu.AllProperties' }} path={'/property/'}/>
+PropertyIdPage.headerAction = <ReturnBackHeaderAction
+    descriptor={{ id: 'menu.AllProperties' }}
```
Comment (akarjakin) [inline apps/condo/pages/property/[id]/index.tsx:211]: descriptor here is static, please make it constant.

akarjakin (approved):

apps/condo/pages/property/[id]/index.tsx
Comment (sitozzz) [reply apps/condo/pages/property/[id]/index.tsx]: Done

PR #943, opened 2021-12-09
DOMA-1783 redirect from error page if cookie onErrorPage is true

nomerdvadcatpyat:
Problem: if error throws in ssr mode, then app redirects to /500 page and when user reload this page, it still stays on the /500 page

### DIFF
--- a/apps/condo/pages/500.tsx
+++ b/apps/condo/pages/500.tsx
@@ -1,12 +1,14 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import { Col, Row, Typography } from 'antd'
 import styled from '@emotion/styled'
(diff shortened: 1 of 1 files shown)


sitozzz (approved):

apps/condo/pages/500.tsx
```diff
@@ -20,6 +22,16 @@ export default function Custom500 () {
     const DescriptionMessage = intl.formatMessage({ id: 'pages.condo.error.Description' })
 
     const { isSmall } = useLayoutContext()
+    const router = useRouter()
+
+    useEffect(() => {
+        if (cookie.get('onErrorPage')) {
```
MrFoxPro, line 28: Can we rely on browser history instead of saving into cookies?

PR #947, opened 2021-12-10
DOMA-1605 filter tooltip

nomerdvadcatpyat:
If some of the applied filters are not in the filters in the table, but are in the modal (in the control room, for example, the `author`, `source`, `division` ...), hen when you hover over the table row, the tooltip should be displayed

### DIFF
--- a/apps/condo/domains/ticket/hooks/useFiltersTooltipData.ts
+++ b/apps/condo/domains/ticket/hooks/useFiltersTooltipData.ts
@@ -1,12 +1,13 @@
 import { useOrganization } from '@core/next/organization'
-import { Division } from '@condo/domains/division/utils/clientSchema'
 import isEqual from 'lodash/isEqual'
-import { parseQuery } from '@condo/domains/common/utils/tables.utils'
 import { useRouter } from 'next/router'
 import { useCallback, useMemo } from 'react'
 import { useIntl } from '@core/next/intl'
 import get from 'lodash/get'
-import { Ticket } from '../../../schema'
+import { Ticket } from '@app/condo/schema'
+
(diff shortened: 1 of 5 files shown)


AntonAL (commented):
A tooltip, appearing for each hovered row looks strange. We have the same information, popped out over and over, when we navigate through table.
Is it discussed with design team?
Why we need it, when a "Filters" button already displays a number of applied filters?

nomerdvadcatpyat:
> A tooltip, appearing for each hovered row looks strange. We have the same information, popped out over and over, when we navigate through table. Is it discussed with design team? Why we need it, when a "Filters" button already displays a number of applied filters?

The tooltip displays fields that are not in the table columns, but that are present in the applied filters.
This is done so that the operator can understand why this ticket was displayed without going to the page of this ticket.
Yes, this is agreed with the design team.
изображение

apps/condo/domains/ticket/hooks/useFiltersTooltipData.ts
```diff
@@ -0,0 +1,65 @@
+import { useOrganization } from '@core/next/organization'
+import { Division } from '@condo/domains/division/utils/clientSchema'
+import isEqual from 'lodash/isEqual'
+import { parseQuery } from '@condo/domains/common/utils/tables.utils'
+import { useRouter } from 'next/router'
+import { useCallback, useMemo } from 'react'
+import { useIntl } from '@core/next/intl'
+import get from 'lodash/get'
+import { Ticket } from '../../../schema'
```
sitozzz, line 9: `@app/condo/schema`

sitozzz (approved):

PR #949, opened 2021-12-10
DOMA-1732 save table filters

nomerdvadcatpyat:
previously filters were saved in localStorage only if you click the confirm button in the modal window


sitozzz (approved):

MrFoxPro:
изображение
This happens when I loading tickets page. This is because of
```js
 useEffect(() => {
        FiltersStorage.saveFilters(organization.id, FILTER_TABLE_KEYS.TICKET, filters)
    }, [filters])
``` 
this effect is being triggered every time **link** to `filters`  object changes. But we need deep value compare. Also it could be hard to track `filters` object change, mainly because it is not reactive value, I'm not sure it would track correctly

MrFoxPro (approved):

PR #968, opened 2021-12-14
Feature/condo/doma 1627/meters import

AntonAL:
1. Open `/meter` page
2. Push "import" button (from right of "add" button)
3. Select a file `apps/condo/public/meter-import-example.xlsx`

Meter readings should be imported and properly connected to current organization.

There is one 🐞 in import algorithm under investigation, — only one meter will be created and reused for each row, regardless of unique combination of search conditions. It looks strange…

Added new permanent `MeterReadingSource` record with type `import_condo`.

Separate key set is used for import feature exclusively to keep a set of column names in import requirements safe from changes in other places.
Import feature will be used by many clients, so, a change in set of columns will cause problems.


pahaz (approved):

akarjakin (approved):

apps/condo/domains/meter/hooks/useImporterFunction.ts
```diff
@@ -0,0 +1,221 @@
...
+
+        addons.propertyId = propertyId
+
+        const searchMeterWhereConditions = {
+            organization: userOrganizationId,
+            property: propertyId,
+            unitName,
+            accountNumber,
+            number: meterNumber,
+        }
```
nomerdvadcatpyat, line 126:
```suggestion
        const searchMeterWhereConditions = {
            organization: { id: userOrganizationId },
            property: { id: propertyId },
            unitName,
            accountNumber,
            number: meterNumber,
        }
```

apps/condo/domains/meter/utils/clientSchema/search.ts
```diff
@@ -0,0 +1,72 @@
...
+        $number: String
+    ) {
+        objs: allMeters(
+            where: {
+                organization: { id: $organization },
+                property: { id: $property },
+                unitName: $unitName,
+                accountNumber: $accountNumber,
+                number: $number
+            }
```
nomerdvadcatpyat, line 18:
```suggestion
    query getMeter (
        $where: MeterWhereInput,
        $orderBy: String
    ) {
        objs: allMeters(
            where: $where,
            orderBy: $orderBy
        ) {
```

nomerdvadcatpyat (commented):
Found what the error was: in query `getMeter`, the arguments were `$organization: ID`, `$property: ID`, `$unitName: String`, `$accountNumber: String`, `$number: String`, and passed `{ where, orderBy, first, skip }`

apps/condo/domains/meter/hooks/useImporterFunction.ts
```diff
@@ -0,0 +1,221 @@
...
+    const client = useApolloClient()
+    const { addressApi } = useAddressApi()
+
+    const userOrganizationId = get(userOrganization, ['organization', 'id'])
+
+    const meterCreateAction = Meter.useCreate({},
+        () => Promise.resolve())
+
+    const meterReadingCreateAction = MeterReading.useCreate({},
+        () => Promise.resolve())
```
nomerdvadcatpyat, line 67: can we refetch `meter readings` after all `readings` are imporded?

PR #977, opened 2021-12-17
HOTFIX make meters import date not required

MrFoxPro:


apps/condo/domains/meter/hooks/useImporterFunction.ts
```diff
@@ -154,7 +154,11 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
         }
         return Promise.resolve(true)
     }
-
+    const validateDate = (str) =>  {
+        if (!str) return null
+        const djs = dayjs(String(str))
+        return djs.isValid() && djs.toISOString() || null
```
toplenboren, line 160: I'd rather do djs.isValid() ?? djs.toISOString() : null, seems to be more obvious

toplenboren (approved):

apps/condo/domains/meter/hooks/useImporterFunction.ts
```diff
@@ -188,12 +192,12 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
...
-                installationDate: dayjs(installationDate).toISOString(),
-                commissioningDate: dayjs(commissioningDate).toISOString(),
-                sealingDate: dayjs(sealingDate).toISOString(),
-                controlReadingsDate: dayjs(controlReadingsDate).toISOString(),
+                verificationDate: validateDate(verificationDate),
+                nextVerificationDate: validateDate(nextVerificationDate),
+                installationDate: validateDate(installationDate),
+                commissioningDate: validateDate(commissioningDate),
+                sealingDate: validateDate(sealingDate),
+                controlReadingsDate: validateDate(controlReadingsDate) || dayjs().toISOString(),
```
AntonAL, line 200: With this solution in case of invalid dates, a user will not see errors and will think, that everything is ok, — but dates in created `Meter` records will be `null`. To inform user about errors, you can check dates in `meterReadingValidator` and push error. These errors will be rendered to a new column in exported spreadsheet with not imported rows.

apps/condo/domains/meter/hooks/useImporterFunction.ts
```diff
@@ -154,7 +154,11 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
         }
         return Promise.resolve(true)
     }
-
+    const validateDate = (str) =>  {
```
AntonAL, line 157: Function name does't corresponds to what it actually does.
Use it for validation only.
In place, where dates are actually passed to GraphQL input variables, you can format it.

apps/condo/domains/meter/hooks/useImporterFunction.ts
MrFoxPro: Done

apps/condo/domains/meter/hooks/useImporterFunction.ts
MrFoxPro: Renamed

AntonAL (approved):

toplenboren:
Rebasing and merging into master, since tests fails are random.

apps/condo/domains/meter/hooks/useImporterFunction.ts
```diff
@@ -213,7 +229,7 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
             // @ts-ignore
             value4,
             // @ts-ignore
-            date: controlReadingsDate,
+            date: controlReadingsDate ? toISO(controlReadingsDate) : dayjs().toISOString(),
```
pahaz, line 232: Date.now() !

PR #989, opened 2021-12-21
Feature/condo/odma 1872/clean comments and remove multitariff values from non electricity meters in meter import example excel file

AntonAL:
Changes in import example file:
1. Removed multi-tariff values from non-electricity meterr
2. Removed cell comments


AntonAL:
Spell in branch name.

PR #993, opened 2021-12-21
Feature/condo/doma 1858/new meter reading date column for meter import

AntonAL:
New column "Reading submission date"
Parsing of "Reading submission date" column value either in "YYYY-MM-DD" or "YYYY-MM-DD" format


apps/condo/domains/meter/hooks/useImporterFunction.ts
```diff
@@ -137,6 +156,14 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
         }
         addons.meterResourceId = METER_RESOURCE_ABBREVIATION_TO_ID[String(meterResourceTypeAbbr)]
 
+        let parsedReadingSubmissionDate = dayjs(readingSubmissionDate, DATE_PARSING_FORMAT)
+        if (!parsedReadingSubmissionDate.isValid()) {
+            parsedReadingSubmissionDate = dayjs(readingSubmissionDate, MONTH_PARSING_FORMAT)
+        }
+        if (parsedReadingSubmissionDate.isValid()) {
```
pahaz, line 163: It looks like a runtime error, but I’m not sure

pahaz (approved):

apps/condo/domains/meter/hooks/useImporterFunction.ts
AntonAL: No, it does not produces a runtime error.
In case of incorrect value, it will be handled and pushed to errors array of current row in `meterReadingValidator`.

PR #1025, opened 2021-12-24
DOMA-1891: disable build time redundant actions

pahaz:


apps/condo/index.js
```diff
@@ -62,7 +56,13 @@ const keystone = new Keystone({
     },
 })
 
-if (!IS_BUILD_PHASE) {
+if (IS_BUILD_PHASE) {
+    // NOTE: we need to register User list for `createAuthStrategy`
+    //  and we need the AuthStrategy because it's required for /admin/ static files build
+    registerSchemas(keystone, [
+        require('@core/keystone/schemas/User'),
+    ])
```
AntonAL, line 64: Why during build phase other schemas are omitted?
Why only `User` schema is registered?
Why it is registered from core?

AntonAL: Could you write extended documentation about that?

AntonAL (changes requested):

apps/condo/index.js
pahaz: Look at `NOTE` comment on line 60

apps/condo/index.js
```diff
@@ -23,18 +22,13 @@ const FileAdapter = require('@condo/domains/common/utils/fileAdapter')
 const IS_ENABLE_DD_TRACE = conf.NODE_ENV === 'production'
 const IS_ENABLE_APOLLO_DEBUG = conf.NODE_ENV === 'development' || conf.NODE_ENV === 'test'
 
-let IS_BUILD_PHASE = conf.PHASE === 'build'
-if (IS_BUILD_PHASE) {
-    process.env.FILE_FIELD_ADAPTER = 'local' // Test
-}
-// TODO(DOMA-1614): if set to true adminUi will failed to load after build + start is launched
```
pahaz, line 30: @dkoviazin look at this ^^

apps/condo/index.js
```diff
@@ -62,7 +56,13 @@ const keystone = new Keystone({
     },
 })
 
-if (!IS_BUILD_PHASE) {
+if (IS_BUILD_PHASE) {
```
AntonAL, line 59:
```suggestion
// Because Babel is used only for frontend to transpile and optimise code,
// backend files will bring unnecessary workload to building stage
if (IS_BUILD_PHASE) {
```

apps/condo/index.js
```diff
@@ -62,7 +56,13 @@ const keystone = new Keystone({
     },
 })
```
AntonAL, line 58:
```suggestion

// Because Babel is used only for frontend to transpile and optimise code,
// backend files will bring unnecessary workload to building stage
```

apps/condo/index.js
```diff
@@ -62,7 +56,13 @@ const keystone = new Keystone({
...
 
-if (!IS_BUILD_PHASE) {
+if (IS_BUILD_PHASE) {
+    // NOTE: we need to register User list for `createAuthStrategy`
+    //  and we need the AuthStrategy because it's required for /admin/ static files build
```
AntonAL, line 61:
```suggestion
    // `User` list is used by `createAuthStrategy`, that creates `AuthStrategy`, required to build /admin/ static files
```

apps/condo/index.js
```diff
@@ -79,35 +81,35 @@ if (!IS_BUILD_PHASE) {
         require('@condo/domains/acquiring/schema'),
     ])
 
+    const { registerTasks } = require('@core/keystone/tasks')
+    // NOTE(pahaz): we need to put it here because it init the readis connection and we don't want it at build time
```
akarjakin, line 85: we put it here because it inits the redis connection and we don't want it at build time

pahaz:
```suggestion
    // NOTE(pahaz): we put it here because it inits the redis connection and we don't want it at build time
```

AntonAL (approved):

PR #1031, opened 2021-12-29
Bug/condo/DOMA-1824/search by any field displaying

AleX83Xpert:
# Where
The `Tickets` page, text box `Search by any field...`
image

# Before
1. Write something in the text box
2. Go to any other page
3. back to the `Tickets` page

Result: the text box is empty (see ticket in Jira)

# After
Repeat same steps
Result: text box is filled by previous value

### DIFF
--- a/apps/condo/domains/common/components/Table/Index.tsx
+++ b/apps/condo/domains/common/components/Table/Index.tsx
@@ -2,7 +2,7 @@ import React from 'react'
 import { ColumnsType } from 'antd/es/table/interface'
 import { Table as DefaultTable, TableProps } from 'antd'
 import get from 'lodash/get'
-import { debounce } from 'lodash'
+import debounce from 'lodash/debounce'
 import { useRouter } from 'next/router'
 import {
     getPageIndexFromOffset,
@@ -10,7 +10,7 @@ import {
     FULL_TO_SHORT_ORDERS_MAP,
     FiltersFromQueryType,
 } from '@condo/domains/common/utils/tables.utils'
-import { isEqual } from 'lodash'
+import isEqual from 'lodash/isEqual'
 import { updateQuery } from '@condo/domains/common/utils/filters.utils'
 
 export type TableRecord = any
--- a/apps/condo/domains/common/components/Table/Index.tsx
+++ b/apps/condo/domains/common/components/Table/Index.tsx
@@ -10,8 +10,8 @@ import {
     FULL_TO_SHORT_ORDERS_MAP,
     FiltersFromQueryType,
 } from '@condo/domains/common/utils/tables.utils'
-import qs from 'qs'
 import { isEqual } from 'lodash'
+import { updateQuery } from '@condo/domains/common/utils/filters.utils'
 
 export type TableRecord = any
 
@@ -100,8 +100,9 @@ export const Table: React.FC<ITableProps> = ({
             return applyQuery(queryParams)
         }
         else {
-            const newQuery = qs.stringify({ ...queryParams }, { arrayFormat: 'comma', skipNulls: true, addQueryPrefix: true })
-            return router.push(router.route + newQuery)
+            // The `queryParams` contains only filters, sort and offset, so we can use `updateQuery`.
+            // In case of additional parameters, we have to use custom code or modify `updateQuery` signature.
+            return updateQuery(router, newFilters, newSorters, newOffset)
         }
     }, 400)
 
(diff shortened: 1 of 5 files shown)


toplenboren (approved):
lgtm

apps/condo/domains/common/hooks/useSearch.ts
```diff
@@ -21,5 +21,5 @@ export const useSearch = <F> (loading: boolean): [string, (search: string) => vo
         searchChange(value)
     }
 
-    return [search, handleSearchChange]
+    return [search ?? searchValueFromQuery, handleSearchChange]
```
dkoviazin, line 24: we had some problems with compatibility for nullish coalescing operator. We need some tests and an agreement for using it. Better, to use 
```js
[search || searchValueFromQuery, handleSearchChange] 
```
for now

akarjakin: It's better not to use `??` here, because we manage to keep similar code style both on back and front, and back doesn't have TS yet, so `??` can cause problems in some cases.

apps/condo/domains/common/components/Table/Index.tsx
```diff
@@ -10,8 +10,8 @@ import {
     FULL_TO_SHORT_ORDERS_MAP,
     FiltersFromQueryType,
 } from '@condo/domains/common/utils/tables.utils'
-import qs from 'qs'
 import { isEqual } from 'lodash'
```
akarjakin, line 13: import isEqual from 'lodash/isEqual'

apps/condo/domains/common/utils/FiltersStorage.ts
```diff
@@ -25,7 +25,17 @@ export class FiltersStorage {
     public static saveFilters (organizationId: string, tableKey: FILTER_TABLE_KEYS, filters: FiltersFromQueryType | string): void {
         const oldFilters = this.getFilters(tableKey)
 
-        localStorage.setItem(tableKey, JSON.stringify(pickBy({ ...oldFilters, [organizationId]: filters })))
+        const data2Save = pickBy({ ...oldFilters, [organizationId]: filters })
```
akarjakin, line 28: `data` would be enough actually, and we don't use shortcuts like `2` instead of `To` :)

apps/condo/domains/common/utils/filters.utils.ts
```diff
@@ -184,11 +185,20 @@ export function convertToOptions <T> (objects: T[], labelField: string, valueFie
 }
 
 export async function updateQuery (router: NextRouter, newFilters?: FiltersFromQueryType, sort?: string[], offset?: number): Promise<boolean> {
-    if (!offset && 'offset' in router.query)
+    if (!offset && 'offset' in router.query) {
         router.query['offset'] = '0'
+    }
+
+    const possibleFilters = pickBy(newFilters, (value, key) => !isEmpty(value))
```
akarjakin, line 192: key is not used here and can be removed.

apps/condo/domains/common/utils/filters.utils.ts
```diff
@@ -184,11 +185,20 @@ export function convertToOptions <T> (objects: T[], labelField: string, valueFie
...
+    const possibleQueryData = { ...router.query, sort, offset }
+    if (isEmpty(possibleFilters)) {
+        delete possibleQueryData['filters']
```
akarjakin, line 195: maybe `possibleQueryData.filters` ?

apps/condo/domains/common/hooks/useSearch.ts
AleX83Xpert: @dkoviazin, fixed

akarjakin (approved):

dkoviazin (approved):

apps/condo/domains/common/hooks/useSearch.ts
AleX83Xpert: @akarjakin, `??` replaced with `||`

apps/condo/domains/common/components/Table/Index.tsx
AleX83Xpert: Fixed

apps/condo/domains/common/utils/FiltersStorage.ts
AleX83Xpert: `data2Save` replaced with `dataToSave`
In case of short `data` there is not understandable purpose of the variable

apps/condo/domains/common/utils/filters.utils.ts
AleX83Xpert: The `possibleQueryData` variable has no type, so I decided to use `['filters']` to not make the linter angry to me :)

apps/condo/domains/common/utils/filters.utils.ts
AleX83Xpert: Fixed

PR #1033, opened 2021-12-29
Fix/condo/doma 1907/missing classifier in print version of ticket page

AntonAL:


sitozzz (approved):

apps/condo/pages/ticket/[id]/pdf.tsx
```diff
@@ -158,7 +159,21 @@ const PdfView = () => {
                     <Col span={6}>
                         <TicketDescriptionField
                             title={ClassifierMessage}
-                            value={get(ticket, ['classifier', 'name'])}
+                            value={
+                                <Breadcrumb separator={<>≫<br/></>}>
+                                    {
+                                        compact([
```
Comment (akarjakin) [inline apps/condo/pages/ticket/[id]/pdf.tsx:165]: useMemo

akarjakin (approved):

nomerdvadcatpyat (approved):

apps/condo/pages/ticket/[id]/pdf.tsx
Comment (AntonAL) [reply apps/condo/pages/ticket/[id]/pdf.tsx:165]: It's irrelevant here, because a pdf version is not going to be changed to be rerendered

PR #1039, opened 2022-01-11
DOMA-1750 [Backend] Temporary auth flow remake 

MrFoxPro:
TODO
- [x] Write test

### DIFF
--- a/apps/condo/domains/user/schema/ForgotPasswordService.js
+++ b/apps/condo/domains/user/schema/ForgotPasswordService.js
@@ -171,7 +171,7 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
                     })
                 }
                 else if (action.user) {
-                    const userId = action.user.id
+                    userId = action.user.id
                     phone = await getById('User', userId).then(p => p.phone)
                     const tokenId = action.id
 
--- a/apps/condo/domains/user/schema/ForgotPasswordService.js
+++ b/apps/condo/domains/user/schema/ForgotPasswordService.js
@@ -162,10 +162,11 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
                         throw new Error(`${RESET_TOKEN_NOT_FOUND}] Unable to find valid token`)
                     }
                     phone = action.phone
-                    userId = await User.getAll(context, {
+                    const [user] = await User.getAll(context, {
                         type: STAFF,
                         phone,
-                    }).then(([u]) => u && u.id)
+                    })
+                    userId = user && user.id
 
                     if (!userId) {
                         throw new Error(`${RESET_TOKEN_NOT_FOUND}] Unable to find valid token`)
--- a/apps/condo/domains/user/schema/ForgotPasswordService.js
+++ b/apps/condo/domains/user/schema/ForgotPasswordService.js
@@ -146,6 +146,7 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
                         token, 
                         expiresAt_gte: now,
                         completedAt: null,
+                        isPhoneVerified: false,
                     })
                 }
                 if (!action) {
@@ -155,11 +156,9 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
                 let phone, userId
 
                 if (typeof action.smsCode === 'number') {
-                    if (action.isPhoneVerified === false) {
-                        throw new Error(`${RESET_TOKEN_NOT_FOUND}] Unable to find valid token`)
-                    }
                     phone = action.phone
                     userId = await User.getAll(context, {
+                        type: STAFF,
                         phone,
                     }).then(([u]) => u && u.id)
 
@@ -169,8 +168,7 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
                     await ConfirmPhoneActionUtil.update(context, action.id, {
                         completedAt: now,
                     })
-                }
-                else if (action.user) {
+                } else if (action.user) {
                     userId = action.user.id
                     phone = await getById('User', userId).then(p => p.phone)
                     const tokenId = action.id
--- a/apps/condo/domains/user/schema/ForgotPasswordService.js
+++ b/apps/condo/domains/user/schema/ForgotPasswordService.js
@@ -140,12 +140,14 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
                     token,
                     expiresAt_gte: now,
                     usedAt: null,
-                }) || await ConfirmPhoneActionUtil.getAll(context, {
-                    token, 
-                    expiresAt_gte: now,
-                    completedAt: null,
                 })
-
+                if (!action) {
+                    [action] = await ConfirmPhoneActionUtil.getAll(context, {
+                        token, 
+                        expiresAt_gte: now,
+                        completedAt: null,
+                    })
+                }
                 if (!action) {
                     throw new Error(`${RESET_TOKEN_NOT_FOUND}] Unable to find valid token`)
                 }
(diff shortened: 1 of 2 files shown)


apps/condo/domains/user/schema/ForgotPasswordService.js
```diff
@@ -133,25 +135,49 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
...
+                    const userId = action.user.id
+                    phone = await getById('User', userId).then(p => p.phone)
+                    const tokenId = action.id
+
+                    // mark token as used
+                    await ForgotPasswordActionUtil.update(context, tokenId, {
+                        usedAt: now,
+                    })
+                }
+
```
MrFoxPro, line 180: I think we need to set `ConfirmPhoneAction.completedAt` here to current date. But this flag is not set anywhere at all :/

apps/condo/domains/user/schema/ForgotPasswordService.js
```diff
@@ -75,7 +75,9 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
                 const users = await User.getAll(context, { phone, type: STAFF })
 
                 if (isEmpty(users)) {
-                    throw new Error(`${WRONG_PHONE_ERROR}] Unable to find user when trying to start password recovery`)
+                    // We don't need to say to client whether we have users with this phone number
+                    // because it is not safety.
+                    return { status: 'ok' }
```
pahaz, line 80: these changes will not help in any way. they only hide the problem

apps/condo/domains/user/schema/ForgotPasswordService.js
```diff
@@ -133,25 +135,52 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
...
                     usedAt: null,
                 })
-
+                if (!action) {
+                    [action] = await ConfirmPhoneActionUtil.getAll(context, {
+                        token, 
+                        expiresAt_gte: now,
+                        completedAt: null,
+                    })
+                }
```
pahaz, line 150:
```suggestion
                }

```

apps/condo/domains/user/schema/ForgotPasswordService.js
```diff
@@ -133,25 +135,52 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
...
+                    }).then(([u]) => u && u.id)
+
+                    if (!userId) {
+                        throw new Error(`${RESET_TOKEN_NOT_FOUND}] Unable to find valid token`)
+                    }
+                    await ConfirmPhoneActionUtil.update(context, action.id, {
+                        completedAt: now,
+                    })
+                }
+                else if (action.user) {
```
pahaz, line 173: Please, don't change the if/else style. Use `} else if (...) {` style (one line). 
```suggestion
                } else if (action.user) {
```

apps/condo/domains/user/schema/ForgotPasswordService.js
```diff
@@ -133,25 +135,52 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
...
-                const userId = action.user.id
-                const { phone } = await getById('User', userId)
-                const tokenId = action.id
+                let phone, userId
 
-                // mark token as used
-                await ForgotPasswordActionUtil.update(context, tokenId, {
-                    usedAt: now,
-                })
+                if (typeof action.smsCode === 'number') {
```
pahaz, line 157: This is an implicit approach. The data model may change. We can add this field to ForgotPasswordAction as well.

apps/condo/domains/user/schema/ForgotPasswordService.js
```diff
@@ -133,25 +135,52 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
...
+                    if (action.isPhoneVerified === false) {
+                        throw new Error(`${RESET_TOKEN_NOT_FOUND}] Unable to find valid token`)
+                    }
+                    phone = action.phone
+                    userId = await User.getAll(context, {
```
pahaz, line 162: It's a bad style to use `async/await` with Promises.

apps/condo/domains/user/schema/ForgotPasswordService.js
MrFoxPro: That's true but we are using javascript and don't have classes metainformation. I can wrap logic inside if condition but it will be complex to read, although it is temporary solution. 
See example in https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates
изображение

pahaz (changes requested):
This PR adds vulnerability!

apps/condo/domains/user/schema/ForgotPasswordService.js
pahaz: This is true. But you can use an explicit approach. You know the model for which you found the token.

pahaz: In this case, you are trying to implicitly select a branch of the solution. It's not clear.

apps/condo/domains/user/schema/ForgotPasswordService.js
```diff
@@ -133,25 +135,52 @@ const ForgotPasswordService = new GQLCustomSchema('ForgotPasswordService', {
...
+                let [action] = await ForgotPasswordActionUtil.getAll(context, {
                     token,
                     expiresAt_gte: now,
                     usedAt: null,
                 })
-
+                if (!action) {
+                    [action] = await ConfirmPhoneActionUtil.getAll(context, {
+                        token, 
+                        expiresAt_gte: now,
```
pahaz, line 147: you can just add `isPhoneVerified` here

apps/condo/domains/user/schema/ForgotPasswordService.js
AntonAL, line 80: Throwing an exception interrupts execution flow and `return` will never be reached.

apps/condo/domains/user/schema/ForgotPasswordService.test.js
```diff
@@ -165,5 +170,27 @@ describe('ForgotPasswordAction Service', () => {
             expect(result.errors).toHaveLength(1)
             expect(result.errors[0].message).toEqual(`${PASSWORD_TOO_SHORT}] Password is too short`)
         })
+        
+        it('changing password with ConfirmPhoneAction', async () => {
```
AntonAL, line 174: What about testing a case, when an action is not found by specified phone?

apps/condo/domains/user/schema/ForgotPasswordService.js
MrFoxPro: Remade

apps/condo/domains/user/schema/ForgotPasswordService.js
pahaz: It's a new style for this project. And I don't understand the profit of this.

apps/condo/domains/user/schema/ForgotPasswordService.test.js
```diff
@@ -165,5 +170,92 @@ describe('ForgotPasswordAction Service', () => {
...
+        
+        it('can change password with ConfirmPhoneAction', async () => {
+            const admin = await makeLoggedInAdminClient()
+            const [user, userAttrs] = await createTestUser(admin)
+            const client = await makeClient()
+
+            const [{ token }] = await createTestConfirmPhoneAction(admin, {
+                phone: userAttrs.phone,
+            })
+            const [actionBefore] = await ConfirmPhoneAction.getAll(admin, { token })
```
pahaz, line 182: it's unstable test!

pahaz: You already have this object and you don't need to use getAll. Use object from 179 line

apps/condo/domains/user/schema/ForgotPasswordService.test.js
```diff
@@ -165,5 +170,92 @@ describe('ForgotPasswordAction Service', () => {
...
+            expect(status).toBe('ok')
+            
+            const password = `new_${userAttrs.password}`
+            const { data } = await client.mutate(CHANGE_PASSWORD_WITH_TOKEN_MUTATION, { data: { token, password } })
+            expect(data.result).toEqual({ status: 'ok',  phone:  userAttrs.phone })
+
+            const { errors } = await client.mutate(CHANGE_PASSWORD_WITH_TOKEN_MUTATION, { data: { token, password } })
+            expect(errors[0].message).toEqual(`${RESET_TOKEN_NOT_FOUND}] Unable to find valid token`)
+        })
+        it('cannot change when action token expired', async () => {
```
pahaz, line 232:
```suggestion

        it('cannot change when action token expired', async () => {
```

pahaz (approved):

nomerdvadcatpyat (approved):

PR #1056, opened 2022-01-17
fix/condo/DOMA-1752/no new ticket at page of contact

AleX83Xpert:
Steps to reproduce bug and to see it solved:
- Open contact page
- Create new ticket from this contact
- Go to contact's page again
- Observe tickets list
  - **Before** bug was solved: no ticket in the list
  - **After** bug solved: ticket presented in the list
---
[RUN CONDO TESTS](https://github.com/open-condo-software/condo/actions/workflows/nodejs.condo.test.yml)

### DIFF
--- a/apps/condo/domains/common/components/TicketCard/TicketCardList.tsx
+++ b/apps/condo/domains/common/components/TicketCard/TicketCardList.tsx
@@ -11,11 +11,9 @@ import groupBy from 'lodash/groupBy'
 import pickBy from 'lodash/pickBy'
 import Link from 'next/link'
 import qs, { IStringifyOptions } from 'qs'
-import React, { useEffect, useMemo } from 'react'
-import { Ticket as TicketGQL } from '../../../ticket/gql'
+import React, { useMemo } from 'react'
 import { useLayoutContext } from '../LayoutContext'
 import { TicketOverview } from './TicketOverview'
-import { useApolloClient } from '@core/next/apollo'
 
 interface IContainerProps {
     isSmall: boolean
--- a/apps/condo/domains/common/components/TicketCard/TicketCardList.tsx
+++ b/apps/condo/domains/common/components/TicketCard/TicketCardList.tsx
@@ -1,19 +1,21 @@
-import React, { useMemo } from 'react'
-import { useIntl } from '@core/next/intl'
-import Link from 'next/link'
-import qs, { IStringifyOptions } from 'qs'
-import { Row, Col, Typography, RowProps, Space } from 'antd'
-import get from 'lodash/get'
-import pickBy from 'lodash/pickBy'
-import groupBy from 'lodash/groupBy'
-import styled from '@emotion/styled'
-import { colors } from '@condo/domains/common/constants/style'
-import { Ticket } from '@condo/domains/ticket/utils/clientSchema'
+import { green } from '@ant-design/colors'
 import { SortTicketsBy, Ticket as TicketSchema } from '@app/condo/schema'
 import { Loader } from '@condo/domains/common/components/Loader'
-import { TicketOverview } from './TicketOverview'
-import { green } from '@ant-design/colors'
+import { colors } from '@condo/domains/common/constants/style'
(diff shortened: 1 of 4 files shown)


apps/condo/domains/common/components/TicketCard/TicketCardList.tsx
```diff
@@ -1,19 +1,21 @@
...
-import { Ticket } from '@condo/domains/ticket/utils/clientSchema'
-import { SortTicketsBy, Ticket as TicketSchema } from '@app/condo/schema'
-import { Loader } from '@condo/domains/common/components/Loader'
-import { TicketOverview } from './TicketOverview'
-import { green } from '@ant-design/colors'
+import pickBy from 'lodash/pickBy'
+import Link from 'next/link'
+import qs, { IStringifyOptions } from 'qs'
+import React, { useEffect, useMemo } from 'react'
+import { Ticket as TicketGQL } from '../../../ticket/gql'
```
SavelevMatthew, line 15: `@condo/domains/ticket`
```suggestion
import { Ticket as TicketGQL } from '@condo/domains/ticket/gql'
```

SavelevMatthew: Why do you even need this import?)

apps/condo/domains/ticket/components/TicketForm/CreateTicketForm.tsx
```diff
@@ -57,13 +59,16 @@ export const CreateTicketForm: React.FC = () => {
...
+    const { addTicketToQueryCacheForTicketCardList } = useCacheUtils(client.cache)
 
     const action = Ticket.useCreate(
         {
             status: OPEN_STATUS,
             source: DEFAULT_TICKET_SOURCE_CALL_ID,
         },
-        () => {
+        (ticket) => {
+            addTicketToQueryCacheForTicketCardList(ticket)
```
SavelevMatthew, line 71: As I see, the solution does not solve problem, if ticket was created by user A, while we're on user B account from same organization, am I right?

May be then disabling cache in `/contact/[id]/index` ticket query was "goto solution"?

SavelevMatthew (approved):

apps/condo/domains/common/components/TicketCard/TicketCardList.tsx
AleX83Xpert: Fixed. Unused imports removed.

apps/condo/domains/ticket/components/TicketForm/CreateTicketForm.tsx
AleX83Xpert: > As I see, the solution does not solve problem, if ticket was created by user A, while we're on user B account from same organization, am I right?
The solution modifies query cache used at particular contact page.
image
> May be then disabling cache in /contact/[id]/index ticket query was "goto solution"?
If you mean this cache setting
image
I thought about this, but I try to use cache in right way to prevent server overloading. Let's say, disabling of cache is the fall back decision.

apps/condo/domains/ticket/hooks/useCacheUtils.tsx
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        const cachedData: ICachedData = cache.readQuery(queryData)
+
+        if (cachedData) {
+            cache.writeQuery({
+                ...queryData,
+                data: {
+                    ...cachedData,
+                    objs: [...cachedData.objs, ticket],
```
AntonAL, line 28: It seems, like you should insert new record at the beginning of array to respect a sort order `TICKET_SORT_BY = ['createdAt_DESC']` used in `generateQueryVariables`. Check this, please.

AntonAL:
I've checked locally and did't see any difference, — selection of another executor, having another already selected, does't work.

apps/condo/domains/ticket/hooks/useCacheUtils.tsx
AleX83Xpert: It's not necessary, because because data sorted by apollo. I've tested, all is ok.

PR #1063, opened 2022-01-19
DOMA-2009 add signinas button to user item view in admin ui

dkoviazin:
AdminUi allows to override actions for list or single item. So we can extend actions for user and add SignInAs mutation call
I placed hooks to the root of condo app as different domains can extend admin UI
image

### DIFF
--- a/apps/condo/admin-ui/index.js
+++ b/apps/condo/admin-ui/index.js
@@ -19,7 +19,7 @@ const ICON_STYLE = {
 function SignInAsUser () {
     const location = useLocation()
     const [signinAs] = useMutation(SIGNIN_AS_USER_MUTATION, {
-        onCompleted: () => window.location = '/',
+        onCompleted: () => window.location.href = '/',
     })
     const onClick = useCallback(() => {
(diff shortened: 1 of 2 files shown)


sitozzz (approved):

apps/condo/admin-ui/index.js
```diff
@@ -0,0 +1,50 @@
...
+    })
+    const onClick = useCallback(() => {
+        const sender = getClientSideSenderInfo()
+        const path = location.pathname.split('/').splice(2, 2)
+        const userId = (path[0] === 'users' && path[1]) ? path[1] : null
+        const data = { dv: 1, sender, id: userId }
+        signinAs({ variables: { data } }).catch(error => {
+            console.log('Failed to signin', error)
+        })
+    }, [location])
```
akarjakin, line 32: }, [location, signinAs])

akarjakin (approved):

apps/condo/admin-ui/index.js
dkoviazin: Fixed

PR #1095, opened 2022-01-31
DOMA-1564 Ticket deadline

nomerdvadcatpyat:
- Added a `deadline` field to the `Ticket` schema. By default, the deadline occurs 2 days after the `ticket` is created.
- Changed the layout in `BaseTicketForm` in accordance with layouts
- Made `tickets` sorting based on `deadline`. Now the `tickets` with the status "Open" are pinned on top. Next are expired `tickets`. Next come `tickets` that have less than 24 hours left to complete. All other `tickets` follow.
- Fixed `DatePicker` localization by adding `AntdDayjsWebpackPlugin`

---
- [x] make support for `date` in `TicketChange`
- [x] Update layout in `CreateTicketForm` in `callcenter`


apps/condo/next.config.js
```diff
@@ -44,4 +45,11 @@ module.exports = withTM(withLess(withCSS({
             },
         ]
     },
+    webpack: (config) => {
+        const plugins = config.plugins
+
+        config.plugins = [ ...plugins, new AntdDayjsWebpackPlugin() ]
```
pahaz, line 51: Please add `NOTE` what it is and why we need it?

pahaz, line 51:
```suggestion
        // NOTE: Replace Moment.js with Day.js in antd project
        config.plugins = [ ...plugins, new AntdDayjsWebpackPlugin() ]
```

apps/condo/domains/ticket/constants/statusTransitions.js
```diff
@@ -28,6 +28,10 @@ const COUNTRY_RELATED_STATUS_TRANSITIONS = {
 
 const TICKET_ORDER_BY_STATUS = {
     [STATUS_IDS.OPEN]: 100,
+    [STATUS_IDS.IN_PROGRESS]: 200,
```
pahaz, line 31: we can create our own status ids for organizations. You need to pay attantion

pahaz (approved):

apps/condo/domains/ticket/constants/statusTransitions.js
nomerdvadcatpyat: In this case, I think I will make `order = 100` only for the "Open" status, for all others set 200 (not only for default ones)

sitozzz (changes requested):
Need to recreate migration file, because at origin we already have `0094`

sitozzz (approved):

PR #1115, opened 2022-02-06
DOMA-2135 Add reset user feature

toplenboren:
# Problem

1. QA buys new SIM-cards for testing whole system

# Solution

A `ResetUser` mutation, which takes userId and renders `user` unusable.

From: `ResetUserService.js` ...

```
/**
 * Reset user mutation removes phone, email and password from user, making it impossible to use this user.
 *
 * Mutation also sets name to "Deleted User" and do not set deletedAt.
 * Reason for this is the fact that if we soft-delete user, then we are about to face a lot of "user is null" kind of errors in unpredictable places.
 * Name is changed so we can distinguish reset users.
 *
 * Telegram uses the same approach: if user is deleted, then all his messages and other content are available, but his name is now "Deleted user"
 *
 * Primary use cases:
 * 1. QA Wants to test SMS services, but dont want to get multiple phone numbers
 * 2. User wants to remove all his data from our system.
 */
```


apps/condo/domains/user/schema/ResetUserService.js
```diff
@@ -0,0 +1,76 @@
...
+                }
+
+                const user = await getById('User', id)
+                if (!user) {
+                    throw new Error(`${NOT_FOUND_ERROR}user] No user found for this id`)
+                }
+
+                const adminContext = await context.createContext({ skipAccessControl: true })
+                await User.update(adminContext, id, {
+                    phone: null,
```
dkoviazin, line 59: We have constraints for unique combination userType+phone and userType+email = 2 null phone users will fail to set

SavelevMatthew, line 59: Setting all user identical to null = no user in db...
Since if we clean UserHistory we cannot restore it...
So now it's equal to deletedAt performance...

May be @pahaz can resolve our conversation?

toplenboren: No, it's not identical, since deletedAt means that we will delete user from db at some point. In this case we do not delete user from db

This solution was initially proposed by @pahaz, as it is also used in telegram

toplenboren: They won't, combined constants do not work, if either userType or phone is set to NULL

toplenboren: Gonna add new test for that

apps/condo/domains/common/utils/testSchema/index.js
```diff
@@ -235,3 +235,7 @@ export const expectToThrowAuthenticationErrorToObj = async (testFunc) => {
 export const expectToThrowAuthenticationErrorToObjects = async (testFunc) => {
     return await expectToThrowAuthenticationError(testFunc, 'objs')
 }
+
+export const expectToThrowAuthenticationErrorToResult = async (testFunc) => {
```
pahaz, line 239: Omg :))

apps/condo/domains/user/schema/ResetUserService.js
```diff
@@ -0,0 +1,76 @@
...
+const { User } = require('@condo/domains/user/utils/serverSchema')
+
+const { DV_UNKNOWN_VERSION_ERROR, NOT_FOUND_ERROR } = require('@condo/domains/common/constants/errors')
+
+
+const ResetUserService = new GQLCustomSchema('ResetUserService', {
+    types: [
+        {
+            access: true,
+            type: 'input ResetUserInput { dv: Int!, sender: JSON! id: String!}',
```
pahaz, line 32:
```suggestion
            type: 'input ResetUserInput { dv: Int!, sender: SenderInput!, id: String!}',
```

pahaz: It’s better to Use user link type instead of just Id

apps/condo/domains/user/schema/ResetUserService.js
```diff
@@ -0,0 +1,76 @@
...
+                    throw new Error(`${DV_UNKNOWN_VERSION_ERROR}dv] Unknown \`dv\``)
+                }
+
+                const user = await getById('User', id)
+                if (!user) {
+                    throw new Error(`${NOT_FOUND_ERROR}user] No user found for this id`)
+                }
+
+                const adminContext = await context.createContext({ skipAccessControl: true })
+                await User.update(adminContext, id, {
```
pahaz, line 58: You need to use serverUtils for that

toplenboren: razobralis

apps/condo/domains/user/schema/ResetUserService.js
```diff
@@ -0,0 +1,78 @@
...
+const { User } = require('@condo/domains/user/utils/serverSchema')
+
+const { DV_UNKNOWN_VERSION_ERROR, NOT_FOUND_ERROR } = require('@condo/domains/common/constants/errors')
+
+
+const ResetUserService = new GQLCustomSchema('ResetUserService', {
+    types: [
+        {
+            access: true,
+            type: 'input ResetUserInput { dv: Int!, sender: JSON! id: String!}',
```
pahaz, line 32:
```suggestion
            type: 'input ResetUserInput { dv: Int!, sender: JSON! id: String! }',
```

pahaz, line 32:
```suggestion
            type: 'input ResetUserInput { dv: Int!, sender: JSON!, user: UserWhereUniqueInput!}',
```

apps/condo/domains/user/schema/ResetUserService.js
```diff
@@ -0,0 +1,78 @@
...
+                }
+
+                const adminContext = await context.createContext({ skipAccessControl: true })
+                await User.update(adminContext, id, {
+                    phone: null,
+                    email: null,
+                    password: null,
+                    name: DELETED_USER_NAME,
+                    isPhoneVerified: false,
+                    isEmailVerified: false,
```
pahaz, line 64: you also need to remove the `importId` and import remove system

pahaz (approved):

apps/condo/domains/user/schema/ResetUserService.js
```diff
@@ -0,0 +1,78 @@
...
+                if (dv !== 1) {
+                    throw new Error(`${DV_UNKNOWN_VERSION_ERROR}dv] Unknown \`dv\``)
+                }
+
+                const user = await getById('User', id)
+                if (!user) {
+                    throw new Error(`${NOT_FOUND_ERROR}user] No user found for this id`)
+                }
+
+                const adminContext = await context.createContext({ skipAccessControl: true })
```
pahaz, line 57: remove context, you don't need it

apps/condo/domains/user/schema/ResetUserService.js
pahaz: Really we have two cases:
 1) we need to release the existing user phones for test purposes
 2) we need to arise user data by user request

It's hard to really remove use because of lots of FKs. For that reason, we use things like that. Erase the user data and not change the business logic

apps/condo/domains/user/schema/ResetUserService.js
toplenboren: we discussed it in calll

apps/condo/domains/user/utils/serverSchema/index.js
```diff
@@ -49,6 +50,20 @@ async function registerNewServiceUser (context, data) {
     })
 }
 
+async function resetUser (context, data) {
+    if (!context) throw new Error('no context')
+    if (!data) throw new Error('no data')
+    if (!data.sender) throw new Error('no data.sender')
+    // TODO(codegen): write resetUser serverSchema guards
```
akarjakin, line 57: is this comment still needed?

apps/condo/domains/user/schema/ResetUserService.js
```diff
@@ -0,0 +1,78 @@
...
+            access: access.canResetUser,
+            schema: 'resetUser(data: ResetUserInput!): ResetUserOutput!',
+            resolver: async (parent, args, context, info, extra = {}) => {
+                const { data } = args
+                const { dv, id } = data
+
+                if (dv !== 1) {
+                    throw new Error(`${DV_UNKNOWN_VERSION_ERROR}dv] Unknown \`dv\``)
+                }
+
```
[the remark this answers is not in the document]
dkoviazin: then it's better to set restrictions for support users to reset admins

SavelevMatthew (approved):

SavelevMatthew:
Please fix "support can delete admin" issue

toplenboren:
> Please fix "support can delete admin" issue

Removed an ability to reset admin users

PR #1120, opened 2022-02-07
Refactor/condo/doma 1976/standardize error reporting

AntonAL:
Abstract basic error kinds are declared along with a custom `GraphQLError` implementation.
A preprocessor `schemaDoc` takes each custom schema and embeds documentation, that will be displayed in GraphiQL.
Used concerns about errors and documentation for a custom mutation `registerNewUserService`.


apps/condo/domains/user/schema/RegisterNewUserService.test.js
```diff
@@ -45,9 +45,14 @@ describe('RegisterNewUserService', () => {
 
         expect(data).toEqual({ 'user': null })
         expect(errors).toMatchObject([{
-            'message': '[format:phone] invalid format',
-            'name': 'GraphQLError',
-            'path': ['user'],
+            message: 'Wrong format of provided phone number',
+            name: 'GraphQLError',
+            path: ['user'],
```
pahaz, line 50: should be `['user','phone']`

AntonAL: No, it's an Apollo's part, where it points a node of GraphQL request, where error was occurred. We have `['user']` here, because we have declared our mutation as `user: registerNewUser`.

apps/condo/domains/user/schema/RegisterNewUserService.test.js
```diff
@@ -34,7 +33,17 @@ describe('RegisterNewUserService', () => {
                 email,
             },
         })
-        expect(JSON.stringify(errors)).toMatch(PHONE_ALREADY_REGISTERED_ERROR)
+        expect(errors).toMatchObject([{
+            message: 'User with specified phone already exists',
+            name: 'GraphQLError',
```
pahaz, line 38: We don't really deen the name. It's not a part of GQL standart

apps/condo/domains/user/schema/RegisterNewUserService.test.js
```diff
@@ -71,6 +86,16 @@ describe('RegisterNewUserService', () => {
...
         })
-        expect(JSON.stringify(errors)).toMatch(EMAIL_ALREADY_REGISTERED_ERROR)
+        expect(errors).toMatchObject([{
+            message: 'User with specified email already exists',
+            name: 'GraphQLError',
+            path: ['user'],
+            extensions: {
+                mutation: 'registerNewUser',
+                variable: ['data', 'email'],
+                code: 'CONFLICT',
```
pahaz, line 96: Should be `BAD_USER_INPUT`

packages/@core.keystone/errors.js
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * First level of error classification, used in custom GraphQL queries or mutations
+ * Only generic error kinds are listed
+ * Conceptually, it conforms to HTTP standard for error codes
+ * @readonly
+ * @enum {String}
+ */
+const GQLErrorCode = {
+    NOT_FOUND,
+    CONFLICT,
```
pahaz, line 86: https://www.apollographql.com/docs/apollo-server/data/errors/#error-codes 
There is no conflict here. It's better to use simplified APOLLO codes. It's really not the same as HTTP codes but similar to them

pahaz: I think the CONFLICT is the same as BAD_USER_INPUT. I don't know any useful case for the code. Do you know any?

packages/@core.keystone/errors.js
```diff
@@ -0,0 +1,128 @@
...
+    constructor (fields) {
+        super(fields.message, fields.code, fields)
+        Object.defineProperty(this, 'name', { value: 'GraphQLError' })
+    }
+}
+
+module.exports = {
+    GQLError,
+    GQLErrorCode,
+    GQLErrorType,
```
pahaz, line 127: You don't need to put the `GQLErrorType` here. It's a part of app domain logic. Not a package stuff

packages/@core.keystone/errors.js
AntonAL: No, they are different.

`BAD_USER_INPUT` is about input data, that does not matches required format.

`CONFLICT` is about state of a storage, that violates existing constraints
Take a look at a [W3C Standard](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html)
> The request could not be completed due to a conflict with the current state of the resource

AntonAL: > https://www.apollographql.com/docs/apollo-server/data/errors/#error-codes
> There is no CONFLICT here.
Apollo errors are about handling some abstract GraphQL, so, we see a limited set of errors.
I'm developing errors standard, that is more about business logic and Apollo errors don't fit for that.

packages/@core.keystone/errors.js
AntonAL: `NOT_UNIQUE`, `WRONG_FORMAT` are general-purpose codes.
Btw, it seems strange to me to implement all of this in `@core.keystone` package.
Logically, it is an errors module for our GraphQL API, not a keystone.

apps/condo/domains/user/schema/AuthenticateUserWithPhoneAndPasswordService.test.js
```diff
@@ -33,17 +31,33 @@ describe('Auth by phone and password', () => {
...
         const client = await makeClient()
-        const res1 = await client.mutate(SIGNIN_BY_PHONE_AND_PASSWORD_MUTATION, { phone: phone + Math.random(), password })
-        expect(JSON.stringify(res1.errors)).toContain(WRONG_EMAIL_ERROR)
+        const { errors } = await client.mutate(SIGNIN_BY_PHONE_AND_PASSWORD_MUTATION, { phone: phone + Math.random(), password })
+        expect(errors).toMatchObject([{
+            message: 'Wrong format of provided phone number',
+            name: 'GraphQLError',
+            path: ['obj'],
+            extensions: {
+                mutation: 'authenticateUserWithPhoneAndPassword',
```
pahaz, line 58: you need to add `code` and `type` here

apps/condo/domains/user/schema/AuthenticateUserWithPhoneAndPasswordService.test.js
```diff
@@ -33,17 +31,33 @@ describe('Auth by phone and password', () => {
...
         const client = await makeClient()
-        const res1 = await client.mutate(SIGNIN_BY_PHONE_AND_PASSWORD_MUTATION, { phone, password: password + Math.random() })
-        expect(JSON.stringify(res1.errors)).toContain(WRONG_PASSWORD_ERROR)
+        const { errors } = await client.mutate(SIGNIN_BY_PHONE_AND_PASSWORD_MUTATION, { phone, password: password + Math.random() })
+        expect(errors).toMatchObject([{
+            message: 'Wrong password',
+            name: 'GraphQLError',
+            path: ['obj'],
+            extensions: {
+                mutation: 'authenticateUserWithPhoneAndPassword',
```
pahaz, line 40: you need to add `code` and `type`

pahaz: and probably translations ?

apps/condo/domains/user/schema/ChangePhoneNumberResidentUserService.test.js
```diff
@@ -55,8 +52,17 @@ describe('ChangePhoneNumberResidentUserService', () => {
...
                 }
-                const { errors: [error] } = await client.mutate(CHANGE_PHONE_NUMBER_RESIDENT_USER_MUTATION, { data })
-                expect(error.message).toContain(CONFIRM_PHONE_ACTION_EXPIRED)
+                const { errors } = await client.mutate(CHANGE_PHONE_NUMBER_RESIDENT_USER_MUTATION, { data })
+                expect(errors).toMatchObject([{
+                    message: 'Unable to find a non-expired confirm phone action, that corresponds to provided token',
+                    name: 'GraphQLError',
+                    path: ['result'],
+                    extensions: {
+                        mutation: 'changePhoneNumberResidentUser',
```
pahaz, line 61: `code` and `type`

packages/@core.keystone/errors.js
```diff
@@ -0,0 +1,136 @@
...
+/**
+ * First level of error classification, used in custom GraphQL queries or mutations
+ * Second level of classification will be specific to domain in question
+ * Only generic error kinds are listed
+ * Conceptually, it conforms to HTTP standard for error codes
+ * @readonly
+ * @enum {String}
+ */
+const GQLErrorCode = {
+    NOT_FOUND,
```
pahaz, line 93: I don't understand how to unambiguously match the error to the NOT_FOUND code. For example, a user passes a link to an object to the API, but he does not have access to it. For example, this is the creation of a ticket, and he passes the ID of the organization to which he does not have access. 

I think it's a BAD_USER_INPUT. Because it's a hacker who tries to check the API. 

To be honest, I do not know why we might need this code. And I would introduce it if we find a case when we use it.

pahaz: And the default Apollo codes: https://www.apollographql.com/docs/apollo-server/data/errors/#error-codes doesn't contain NOT_FOUND code

pahaz (approved):

apps/condo/domains/user/schema/AuthenticateUserWithPhoneAndPasswordService.test.js
AntonAL: Please, take a look at the `changePasswordWithToken` mutation in `ForgotPasswordActionService`.
Yet, this is only one mutation, that utilises all implemented features.
Next step will utilize the features in all another custom mutations, including this one.

apps/condo/domains/user/schema/AuthenticateUserWithPhoneAndPasswordService.test.js
AntonAL: Yes, later.

apps/condo/domains/user/schema/ChangePhoneNumberResidentUserService.test.js
AntonAL: Yes, later.

packages/@core.keystone/errors.js
AntonAL: Yes, they don't have it, because GraphQL returns either empty arrays or null.
I will get the `NOT_FOUND` out of set.

AntonAL: But, in some mutations we returning errors about something is not found. Will see, whether we need `NOT_FOUND` error code during transition of all custom mutations and queries from all domains.

AntonAL: I think, it will be a type `NOT_FOUND` (2-nd level classification), not the code.

packages/@core.codegen/templates/service00/domains/{{domain}}/schema/{{name}}.js
```diff
@@ -4,7 +4,27 @@
...
+ * They will be rendered in documentation section in GraphiQL for this custom schema
+ */
+const errors = {
+    NAME_OF_ERROR_FOR_USAGE_INSIDE_THIS_MODULE_ONLY: {
+        {%- if type == "mutations" -%}
+        mutation: '{{ convertFirstLetterToLower(name.replace("Service", "")) }}',
+        {%- else %}
+        query: '{{ convertFirstLetterToLower(name.replace("Service", "")) }}',
+        {%- endif %}
+        variable: ['data', 'someVar'], // TODO: Provide path to a query/mutation variable, whose value caused this error. Remove this property, if variables are not relevant to this error
```
pahaz, line 21: ‘TODO(codegen)’

packages/@core.codegen/templates/service00/domains/{{domain}}/schema/{{name}}.js
```diff
@@ -4,7 +4,27 @@
 
 const { GQLCustomSchema } = require('@core/keystone/schema')
 const access = require('@{{app}}/domains/{{ domain }}/access/{{name}}')
+const { GQLError, GQLErrorCode: { BAD_USER_INPUT, INTERNAL_ERROR } } = require('@core/keystone/errors')
+const { NOT_FOUND } = require('@condo/domains/common/constants/errors')
 
+/**
```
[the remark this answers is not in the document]
AntonAL: We need, because display of error in GraphiQL documentation is not obvious

PR #1122, opened 2022-02-08
[BUG 🐞] DOMA-2196 Removed link to contact on ticket page, if there's no contact or it's deleted

SavelevMatthew:


apps/condo/domains/ticket/gql.js
```diff
@@ -11,7 +11,7 @@ const { generateGqlQueries } = require('@condo/domains/common/utils/codegenerati
...
+const TICKET_FIELDS = `{ deadline organization { id name } property { id name address deletedAt addressMeta { ${ADDRESS_META_SUBFIELDS_QUERY_LIST} } } unitName sectionName floorName status { id name type organization { id } colors { primary secondary additional } } statusReopenedCounter statusUpdatedAt statusReason number client { id name } clientName clientEmail clientPhone contact { id name deletedAt } operator { id name } assignee { id name } executor { id name } watchers { id name } classifier { id name } ${THREE_LVL_CLASSIFIER_FIELDS} details related { id details } isEmergency isPaid isWarranty meta source { id name type } sourceMeta ${COMMON_FIELDS} }`
```
pahaz, line 14: 👍

pahaz: Usually, you should not get contact if it's deleted. You need to remove any `deletedAt` fields from the frontend query.

SavelevMatthew:
> Usually, you should not receive contact if it's deleted.
You need to remove any deletedAt fields from the frontend query.

I'm receiving deleted objects, since softDelete does not check nested objs

toplenboren:
> > Usually, you should not receive contact if it's deleted.
> > You need to remove any deletedAt fields from the frontend query.
> 
> I'm receiving deleted objects, since softDelete does not check nested objs

We are going to delete deleted contacts for real each year, so if we think that we can get softDeleted objects, we are going to get a lot of errors in the future

SavelevMatthew:
PR is closed...need to redo deleted property task and then combine contact on top

PR #1134, opened 2022-02-09
[REFACTOR 👨‍💻] DOMA-2196 Remove deletedAt logic from client

SavelevMatthew:
**Initial problem**: Contact link on ticket/id page was active if contact was deleted
**Problem source**: query using `OR: [{ deletedAt: null }, {deletedAt_not: null}]` for obtaining tickets (to write address, if it was deleted)
**Solution**: Move address info close to ticket and sync with Property


apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -304,6 +334,23 @@ const Ticket = new GQLListSchema('Ticket', {
                         }
                     }
                 }
+                const propertyChanged = resolvedData['property'] && resolvedData['property'] !== get(existingItem, 'property')
+                const propertyAddressChanged = resolvedData['propertyAddress'] && resolvedData['propertyAddress'] !== get(existingItem, 'propertyAddress')
```
pahaz, line 338: I don't think that we need to give anyone access to change the `propertyAddress` directly. 
It's auto-generated field.

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -304,6 +334,23 @@ const Ticket = new GQLListSchema('Ticket', {
...
+                const propertyAddressMetaChanged = resolvedData['propertyAddressMeta'] && resolvedData['propertyAddressMeta'] !== get(existingItem, 'propertyAddressMeta')
```
pahaz, line 339: the same for meta. we should update these fields automatically.

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -304,6 +334,23 @@ const Ticket = new GQLListSchema('Ticket', {
...
+                const newItem = { ...existingItem, ...resolvedData }
+                if (propertyChanged || propertyAddressChanged || propertyAddressMetaChanged) {
+                    const propertyId = get(newItem, 'property', null)
+                    const property = await getById('Property', propertyId)
```
pahaz, line 343: you also need to check access to that property. Or check the property.organisation.

apps/condo/domains/ticket/schema/Ticket.test.js
```diff
@@ -5,550 +5,656 @@ const { createTestContact } = require('@condo/domains/contact/utils/testSchema')
...
+            expect(obj.isPaid).toEqual(false)
+            expect(obj.isEmergency).toEqual(false)
+            expect(obj.isWarranty).toEqual(false)
+            expect(obj.meta).toEqual(null)
+            expect(obj.client).toEqual(null)
+            expect(obj.contact).toEqual(expect.objectContaining({ id: attrs.contact.connect.id }))
+            expect(obj.operator).toEqual(null)
+            expect(obj.assignee).toEqual(null)
+            expect(obj.executor).toEqual(null)
+            expect(obj.watchers).toEqual([])
```
pahaz, line 60: you need to add new fields here

apps/condo/domains/ticket/schema/Ticket.test.js
```diff
@@ -5,550 +5,656 @@ const { createTestContact } = require('@condo/domains/contact/utils/testSchema')
...
-            role: { connect: { id: role.id } },
-        })
-        await createTestTicket(clientFrom, organizationTo, propertyTo)
-        await updateTestOrganizationEmployee(admin, employeeFrom.id, {
-            isBlocked: true,
+            await expectToThrowAccessDeniedErrorToObj(async () => {
+                await createTestTicket(clientFrom, organizationTo, propertyTo)
+            })
         })
+    })
```
pahaz, line 606: Do you have a test for "property soft delete case"?

apps/condo/domains/ticket/schema/Ticket.js
SavelevMatthew: Right now you cannot change it it manually to anything except to current ticket property.
Reason, why I left this editable: 
Somehow in db will be ticket with property with `address` `A`, but `propertyAddress` is `B`.
Then it can be fixed manually by setting `propertyAddress` to `A` (and nothing else is not allowed and covered by tests)

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -236,6 +238,34 @@ const Ticket = new GQLListSchema('Ticket', {
             knexOptions: { isNotNullable: true }, // Relationship only!
             kmigratorOptions: { null: false, on_delete: 'models.PROTECT' },
         },
+        propertyAddress: {
+            schemaDoc: 'Address of property, which synced with property and displayed, if property is deleted',
+            type: Text,
+            isRequired: true,
+            hooks: {
+                resolveInput: async ({ resolvedData, existingItem, fieldPath }) => {
```
pahaz, line 246: usually, you don't need to set these fields by hand. And resolve should not be called

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -236,6 +238,34 @@ const Ticket = new GQLListSchema('Ticket', {
             knexOptions: { isNotNullable: true }, // Relationship only!
             kmigratorOptions: { null: false, on_delete: 'models.PROTECT' },
         },
+        propertyAddress: {
+            schemaDoc: 'Address of property, which synced with property and displayed, if property is deleted',
+            type: Text,
+            isRequired: true,
```
pahaz, line 244: It's should be a read-only field. You should not have access to change this by API.

apps/condo/migrations/20220209123400_0099_manual_set_ticket_address_info.js
```diff
@@ -0,0 +1,16 @@
+exports.up = async (knex) => {
+    await knex.raw(`
+        BEGIN;
+            UPDATE "Ticket" AS t
+            SET "propertyAddress" = p.address,
+                "propertyAddressMeta" = p."addressMeta"
+            FROM "Property" AS p
```
pahaz, line 7: I'm expecting something like `join` here. Is this query really work well? Did you test it?

apps/condo/pages/ticket/[id]/index.tsx
```diff
@@ -383,7 +397,7 @@ export const TicketPageContent = ({ organization, employee, TicketContent }) =>
     const { query: { id } } = router as { query: { [key: string]: string } }
 
     const { refetch: refetchTicket, loading, obj: ticket, error } = Ticket.useObject({
-        where: { id: id, property: { OR: [{ deletedAt: null }, { deletedAt_not: null }] }, deletedAt: null },
```
Comment (pahaz) [inline apps/condo/pages/ticket/[id]/index.tsx:386]: ❤️❤️❤️

apps/condo/pages/ticket/[id]/pdf.tsx
```diff
@@ -101,7 +101,7 @@ const PdfView = () => {
     // NOTE: cast `string | string[]` to `string`
     const { query: { id } } = router as { query: { [key: string]: string } }
 
-    const { loading, obj: ticket, error } = Ticket.useObject({ where: { id:id, property: { OR: [{ deletedAt_not: null }, { deletedAt: null }] } } })
```
Comment (pahaz) [inline apps/condo/pages/ticket/[id]/pdf.tsx:104]: ❤️❤️❤️

apps/condo/pages/ticket/[id]/pdf.tsx
```diff
@@ -124,7 +124,7 @@ const PdfView = () => {
     }
 
     const TicketCreationDate = getTicketCreateMessage(intl, ticket)
-    const ticketAddress = get(ticket, ['property', 'address'])
+    const ticketAddress = get(ticket, ['propertyAddress'])
```
Comment (pahaz) [inline apps/condo/pages/ticket/[id]/pdf.tsx:127]: it's better to use `property.address` here and user `propertyAddress` as fallback in case of `property === null`. It means the property is DELETED. And we ned to show it to user

apps/condo/migrations/20220209123400_0099_manual_set_ticket_address_info.js
SavelevMatthew: Tested it locally in datagrip

SavelevMatthew: https://stackoverflow.com/questions/7869592/how-to-do-an-update-join-in-postgresql

apps/condo/domains/ticket/schema/Ticket.js
pahaz: But you can use `afterChange` hook like in Property model. Or you can use `resolveInput` hook like in createdAt, updatedAt, createdBy, updatedBy, v and so on fields work.

apps/condo/domains/ticket/schema/Ticket.test.js
SavelevMatthew: Now I do 😃

apps/condo/pages/ticket/[id]/pdf.tsx
Comment (SavelevMatthew) [reply apps/condo/pages/ticket/[id]/pdf.tsx]: changed

apps/condo/domains/ticket/schema/Ticket.test.js
SavelevMatthew: done 👍

apps/condo/domains/ticket/schema/Ticket.js
pahaz: Probably, it's better to use `resolveInput` because it allows you to avoid the redundant updates.

SavelevMatthew: didn't understand what do you want me to change?

Current logic of validation is: if at least 1 of 3 fields was changed (property link, address or meta), make sure it matches....otherwise throw error

`afterChange` is not an option, since we cannot prevent update at this stage, and `resolveInput` is not an option, since we cannot find property by other 2 fields

SavelevMatthew: And logic of resolvers are:
If detected changes in property-link, set meta and address from new property

If address of current property is changed, make sure to update fields in linked tickets

apps/condo/domains/ticket/schema/Ticket.js
SavelevMatthew: Case when meta/address changes without property is when we change address of property itself...It cannot be detected here...That's why `afterChange` hook in property is used and that's why we're accepting 'controllable manual change'

apps/condo/domains/ticket/schema/Ticket.js
SavelevMatthew: see comments above

apps/condo/domains/ticket/schema/Ticket.js
SavelevMatthew: see comments above

apps/condo/domains/ticket/schema/Ticket.js
SavelevMatthew: Moved logic to model resolveInput....set up fields accesses to remove this fields from inputs

dkoviazin (approved):

pahaz (approved):
LGTM!

PR #1135, opened 2022-02-09
Keystone v5/v6 plugins and schema compatibility layer

pahaz:


apps/condo/domains/notification/schema/Message.js
```diff
@@ -20,47 +18,47 @@ const Message = new GQLListSchema('Message', {
 
         organization: {
             schemaDoc: 'This message is related to some organization. Organization can manage their messages',
-            type: Relationship,
+            type: 'Relationship',
```
pahaz, line 21: We changed the keystone v5 types to strings! And use this string as our own DSL.
For keystone v5 we convert these strings to `@keystonejs/fields`.
For keystone v6 we need to convert these strings to `@keystone-6/core` fields

sitozzz (approved):

dkoviazin (approved):

PR #1138, opened 2022-02-10
DOMA-2136 Acquiring init 

toplenboren:


dkoviazin (approved):

apps/condo/domains/billing/schema/BillingIntegration.js
```diff
@@ -28,23 +30,11 @@ const BillingIntegration = new GQLListSchema('BillingIntegration', {
             isRequired: true,
         },
 
-        shortDescription: {
```
pahaz, line 31: Why do you want to move these fields? Do you want to reuse it?

pahaz (approved):

apps/condo/domains/billing/schema/BillingIntegration.js
toplenboren: Yes, we might reuse it after some time. Cant tell for sure untill I see makety

PR #1147, opened 2022-02-11
DOMA-2235 added meter account number client validation

nomerdvadcatpyat:
added meter account number client validation on create and update meter forms


pahaz (approved):

pahaz:
It's also better to catch the error from back-end side and show it on the front-end

PR #1157, opened 2022-02-15
DOMA-2261 Add amountWithImplicitFee to Payment

dkoviazin:
1. MultiPayment has "amountWithExplicitFee" field - it's the same as "amount of money - payer needs to pay for bill"
2. Payment has  "amountWithImplicitFee" field - it's the same as "recipient will get amount of money from bill"

We add this field because we do not want to calculate this field every time we need this information


apps/condo/domains/acquiring/schema/Payment.js
```diff
@@ -51,7 +51,13 @@ const Payment = new GQLListSchema('Payment', {
...
-            schemaDoc: 'Amount of money from MultiPayment.amountWithOutExplicitFee to pay for billing receipt',
+            schemaDoc: 'Amount of money to pay for billing receipt',
+            isRequired: true,
+        },
+
+        amountWithImplicitFee: {
+            ...POSITIVE_MONEY_AMOUNT_FIELD,
+            schemaDoc: 'Amount of money recipient will get from payment. amountWithImplicitFee = amount - implicitFee',
             isRequired: true,
         },
```
toplenboren, line 62: Why not calculate the value of this field on save?

pahaz: Why it's not virtual field?

PR #1174, opened 2022-02-18
DOMA-2291 contact is created for new clientPhone within ticket create/update, or existing connected

akarjakin:
Problem: organizations loose contacts (contact is not being created/connected if ticked is created/updated by employee)
Solution: create new contact if no existing contact present with same phone within property/organization, or find existing one and connect to ticket


akarjakin:
TODO: port these changes to cc

nomerdvadcatpyat (approved):

PR #1177, opened 2022-02-18
Add dvAndSender plugin! Apply to Organisation

pahaz:


toplenboren (approved):

AntonAL:
Can we utilize this plugin to all schemas, that are using this logic?

PR #1190, opened 2022-02-24
DOMA-2243 add argument `organizationId` to `bin/generate-tickets.js`

AntonAL:
We need to generate tickets for specified organization to test performance of `exportTicketsToExcel` custom query

### DIFF
--- a/apps/condo/bin/generate-tickets.js
+++ b/apps/condo/bin/generate-tickets.js
@@ -1,8 +1,8 @@
-const { 
-    Ticket, 
-    TicketStatus, 
+const {
+    Ticket,
+    TicketStatus,
     TicketClassifierRule,
-    TicketComment, 
+    TicketComment,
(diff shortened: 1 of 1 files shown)


apps/condo/bin/generate-tickets.js
```diff
@@ -171,7 +177,12 @@ class TicketGenerator {
 }
 
 const createTickets = async () => {
-    const TicketManager = new TicketGenerator({ ticketsByDay: { min: 20, max: 50 } })
+    const [organizationId] = process.argv.slice(2)
```
AleX83Xpert, line 180: I think it's better to pass `organizationId` to `createTicket`

AntonAL: I've followed already existing approach to use class and assignments to `this`.
I would implement everything as functions, because here there is no use cases for classes as instances of things, with some life cycle.

AleX83Xpert (approved):

PR #1211, opened 2022-03-09
DOMA-2244 cypress base user flow

sitozzz:
This PR adds two base user scenarios:

**Auth scenario**: 
1. Login as an existing user
2. Register a new user
3. Reset password

**Ticket scenario**:
1. Create ticket
2. Filtering with global and local controls
3. Update ticket fields


dkoviazin (approved):

toplenboren (approved):

.env.example
```diff
@@ -6,6 +6,7 @@ COOKIE_SECRET=random
 SERVER_URL=http://localhost:3000
 DEFAULT_LOCALE=ru
 FILE_FIELD_ADAPTER=local
+GOOGLE_RECAPTCHA_CONFIG='{"SITE_KEY":"6LcPRvQaAAAAAJRyxsFIB4rP5VH036pFOkNH8lgh", "SERVER_KEY":"6LcPRvQaAAAAADn_h1440Es7fXIGD0E4lpXR_FyF"}'
```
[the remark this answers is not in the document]
dkoviazin: Yes. Without reCaptcha we can not run e2e tests in browser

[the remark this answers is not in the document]
sitozzz: > do we need to commit this?

Google recaptcha used at user restore and register scenario. So this key was configured to work with `localhost`.
Another way is save this value to github secret storage and then pass it to the workflow script.
But after discussing we decided to save this config as example string at environment file

apps/condo/bin/generate-initial-data.js
```diff
@@ -0,0 +1,30 @@
+const path = require('path')
+const { GraphQLApp } = require('@keystonejs/app-graphql')
+
+class InitialDataGenerator {
+    async connect () {
+        const resolved = path.resolve('./index.js')
+        const { distDir, keystone, apps } = require(resolved)
```
pahaz, line 7: Could you please use prepareKeystoneApp from test utils

pahaz: https://github.com/open-condo-software/condo/blob/master/apps/condo/bin/test-sbbol-oauth-by-hands.js#L22 Example

apps/condo/bin/generate-initial-data.js
```diff
@@ -0,0 +1,30 @@
...
+        const graphqlIndex = apps.findIndex(app => app instanceof GraphQLApp)
+        await keystone.prepare({ apps: [apps[graphqlIndex]], distDir, dev: true })
+        await keystone.connect()
+    }
+}
+
+if (process.env.NODE_ENV === 'production') {
```
pahaz, line 14: We use the production mod inside the CI

apps/condo/cypress.json
```diff
@@ -0,0 +1,4 @@
+{
```
pahaz, line 1: Is it possible to move it inside cypress folder

apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
```diff
@@ -333,6 +333,7 @@ const ResetFiltersModalButton: React.FC<ResetFiltersModalButtonProps> = ({
             type={'text'}
             onClick={handleReset}
             size={size}
+            data-cy={'filters-button-reset'}
```
pahaz, line 336: It’s better to use something like ‘<domain>__<component name>—<element>’ format for such cases

apps/condo/bin/generate-initial-data.js
sitozzz: Done

apps/condo/cypress.json
sitozzz: Move to cypress related dir

apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
sitozzz: Refactored this

PR #1216, opened 2022-03-10
DOMA-2392 Add fields to BillingRecipient

dkoviazin:
1. Add bankName territoryCode (ОКТМО) offsettingAccount to BillingRecipient
2. Add receiver to frozenReceipt
3. Add deletedAt check on BillingRecipient creation


apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -129,7 +129,9 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
                 iec: get(recipient, 'iec'),
                 bic: get(recipient, 'bic'),
                 bankAccount: get(recipient, 'bankAccount'),
+                deletedAt: null, // TODO(zuch): DOMA-2395 Move deletedAt filter to getOne
```
SavelevMatthew, line 132: It should be passed automatically I guess, you should add deletedAt filter only to native keystone request…to something like getByCondition…and how about getByCondition here?

SavelevMatthew (commented):
Please add new fields to freezeBillingReceipt.spec.js

apps/condo/domains/billing/schema/BillingReceipt.js
dkoviazin: getOne uses getAll and getAll uses execGqlWithoutAccess
I added todo to move this filter to getOne function

apps/condo/domains/billing/schema/BillingRecipient.js
```diff
@@ -45,6 +45,24 @@ const BillingRecipient = new GQLListSchema('BillingRecipient', {
...
         },
 
+        bankName: {
+            schemaDoc: 'Bank name',
+            type: Text,
+            isRequired: false,
+        },
+
+        offsettingAccount: {
+            schemaDoc: 'Bank account',
```
AntonAL, line 55: Could you please write a description of what does it mean?
I've looked up a term "offsetting account banking" and found following "A savings or transaction account, that's typically linked to an eligible home loan".
Is it it?

AntonAL (approved):

apps/condo/domains/billing/schema/BillingRecipient.js
dkoviazin: Will add all translates to one file (iec, tin, etc ...) DOMA-2401

SavelevMatthew (approved):

PR #1252, opened 2022-03-17
Doma-1823 search tickets by date

dkoviazin:
1. added search tickets by date
2. fixed search by number that breaks graphql request on floating point numbers

image

### DIFF
--- a/apps/condo/domains/common/utils/tables.utils.ts
+++ b/apps/condo/domains/common/utils/tables.utils.ts
@@ -122,7 +122,7 @@ export const getFilter: (
         }
         switch (argData) {
             case 'number':
-                args = search.filter(Number).map(Number)
+                args = search.filter(value => Number(value) && String(value).indexOf('.') === -1).map(Number)
                 break
             case 'dateTime':
                 args = search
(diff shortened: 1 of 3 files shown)


apps/condo/domains/common/utils/tables.utils.ts
```diff
@@ -211,10 +214,39 @@ export const getDayLteFilter: (dataIndex: DataIndexType) => FilterType = (dataIn
     }
 }
 
+const datesFromTextQuery = (text) => {
+    const possibleDates = text.match(/[0-9.]+/g) || []
```
sitozzz, line 218: May this regexp be stored and reused at some common cases?

sitozzz (approved):

apps/condo/domains/common/utils/tables.utils.ts
dkoviazin: This is too common regex but we will need to move it to constants - when we will face another ticket's dates display. Maybe based on user locale as different countries have different formats for dates

pahaz (approved):

PR #1255, opened 2022-03-18
Make sbbol subscription free of charge

AntonAL:
Simplified subscription process.
Instead of creating trial SBBOL subscription, posting payment request, periodically checking its status, creating a non-trial SBBOL subscription, — now we just create a non-trial SBBOL subscription for 1 year.

Following modules are to "deprecated" folder:
- `syncPaymentRequestsForSubscription`
- `syncSbbolSubscriptionPaymentRequestState`

I would completely wipe it out, but it can be used later and to be explicit, I've kept it.
Another way is to mark the commit with a tag, but it will be quickly forgotten.

Following tasks are deleted:
- `syncSbbolPaymentRequestsForSubscription`
- `syncSbbolSubscriptionPaymentRequestState`


apps/condo/domains/organization/tasks/syncSbbolSubscriptionPaymentRequestsState.js
```diff
@@ -1,15 +0,0 @@
-const dayjs = require('dayjs')
-const { createCronTask } = require('@core/keystone/tasks')
-const { syncSbbolSubscriptionPaymentRequestsState } = require('@condo/domains/organization/integrations/sbbol/sync/syncSbbolSubscriptionPaymentRequestsState')
-const { getSchemaCtx } = require('@core/keystone/schema')
-
-/**
- * Syncs new and cancelled subscriptions
- */
-const task = createCronTask('syncSbbolSubscriptionPaymentRequestsState', '10 9 * * *', async () => {
```
pahaz, line 9: It's not so simple to just delete job! You already may have it inside a queue. It's better to just change it to `return true`. And remove it after some time.

PR #1266, opened 2022-03-23
DOMA-1964 fixing a bug where stubs were not displayed with empty data

DanilovMaxim13:


apps/condo/domains/division/components/DivisionsTable.tsx
```diff
@@ -67,75 +72,83 @@ export default function DivisionTable (props: BuildingTableProps) {
     }
 
     const [search, handleSearchChange] = useSearch<IFilters>(loading)
+    const isNoDivisionsData = !divisions.length && isEmpty(filters) && !loading
```
AntonAL, line 75: When wi will have empty divisions list for some filter query, a stub will not be displayed, according to this logic.
Are you sure that this is correct? Ask design team.

dkoviazin (approved):

PR #1272, opened 2022-03-24
Fix/condo/DOMA-2376/fix_layout

gegius:
Due to the fact that it was decided to raise the footer closer to the form, at a large screen resolution, the footer remained on the left.

image
image

I didn't follow this, I fixed it.

image


toplenboren (approved):

apps/condo/domains/user/components/auth/RegisterForm.tsx
```diff
@@ -43,6 +43,9 @@ export const RegisterForm: React.FC<IRegisterFormProps> = ({ onFinish }) => {
     const EmailIsAlreadyRegisteredMsg = intl.formatMessage({ id: 'pages.auth.EmailIsAlreadyRegistered' })
     const ConfirmActionExpiredError = intl.formatMessage({ id: 'pages.auth.register.ConfirmActionExpiredError' })
 
+    const PASSWORD_MSG_LABEL = <label style={{ whiteSpace:'break-spaces' }}>{PasswordMsg}</label>
```
pahaz, line 46: Why it’s looks like const? Do we really use this style ?

gegius: If u're talking about making it a constant, this is for readability by the advice of Pavel.
If u're talking about creating a label, this is because the styles of the label embedded in the form cannot be changed

PR #1286, opened 2022-03-28
DOMA-2552 increase records limit for contacts import

nomerdvadcatpyat:
Added feature-flag `bigger_limit_for_import`, used it in contacts and meter readings import
This flag increases maximum number of records from Excel-file from 500 to 10000.


sitozzz (approved):

apps/condo/pages/contact/index.tsx
```diff
@@ -126,6 +128,10 @@ export const ContactsPageContent = ({
                                                                         accessCheck={canManageContacts}
                                                                         onFinish={refetch}
                                                                         columns={columns}
+                                                                        maxTableLength={hasFeature('bigger_limit_for_contacts_import') ?
```
SavelevMatthew, line 131: How about common feature-flag for meters and contacts? Seems like it be more comfortable....and people who load meters, load contacts as well

nomerdvadcatpyat: added common feature flag, updated .helm

SavelevMatthew (commented):
.helm should point to it's master....I can't see you commit there

SavelevMatthew (approved):

PR #1297, opened 2022-03-30
fix/condo/doma-2487/payments registry default filter

AleX83Xpert:
The solved problem: data for payments registry did loading twice (before default filter applied and after that)

### DIFF
--- a/apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
+++ b/apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
@@ -34,6 +34,9 @@ const ROW_GUTTER: [Gutter, Gutter] = [0, 40]
 const TAP_BAR_ROW_GUTTER: [Gutter, Gutter] = [0, 20]
 const DATE_PICKER_COL_LAYOUT = { span: 11, offset: 1 }
 
+/**
+ * Next two variables need for keeping data about default filters during component lifetime
+ */
 let isDefaultFilterApplied = false
 let shouldApplyDefaultFilter = true
 
@@ -106,6 +109,9 @@ const PaymentsTable: React.FC<IPaymentsTableProps> = ({ billingContext, contexts
     const [search, handleSearchChange] = useSearch<IFilters>(loading)
     const [dateRange, setDateRange] = useDateRangeSearch('advancedAt', loading)
 
+    /**
+     * We need to check if default filters should be applied only at first render
+     */
     useEffect(() => {
         if (!hasFilters && shouldApplyDefaultFilter && !isDefaultFilterApplied) {
             isDefaultFilterApplied = true
--- a/apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
+++ b/apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
@@ -21,7 +21,7 @@ import { useOrganization } from '@core/next/organization'
 import { Col, Input, Row } from 'antd'
 import { Gutter } from 'antd/lib/grid/row'
 import dayjs, { Dayjs } from 'dayjs'
-import { get } from 'lodash'
+import { get, isEmpty } from 'lodash'
 import { useRouter } from 'next/router'
 import React, { useEffect } from 'react'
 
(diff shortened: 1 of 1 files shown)


apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
```diff
@@ -28,10 +28,14 @@ import React, { useEffect } from 'react'
 const SORTABLE_PROPERTIES = ['advancedAt', 'amount']
 const PAYMENTS_DEFAULT_SORT_BY = ['advancedAt_DESC']
 const DEFAULT_DATE_RANGE: [Dayjs, Dayjs] = [dayjs().subtract(1, 'week'), dayjs()]
+const DEFAULT_DATE_RANGE_STR: [string, string] = [String(DEFAULT_DATE_RANGE[0]), String(DEFAULT_DATE_RANGE[1])]
 
 const ROW_GUTTER: [Gutter, Gutter] = [0, 40]
 const TAP_BAR_ROW_GUTTER: [Gutter, Gutter] = [0, 20]
 
+let isDefaultFilterApplied = false
```
akarjakin, line 36: IS_DEFAULT_FILTER_APPLIED
SHOULD_APPLY_DEFAULT_FILTER

constants shouldn't look like variables for readability sake. :)

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
```diff
@@ -47,6 +51,16 @@ const PaymentsTable: React.FC<IPaymentsTableProps> = ({ billingContext, contexts
     const userOrganization = useOrganization()
 
     const { filters, sorters, offset } = parseQuery(router.query)
+    const hasFilters = Object.keys(filters).length > 0
```
akarjakin, line 54: !isEmpty(filters)

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
```diff
@@ -92,8 +106,11 @@ const PaymentsTable: React.FC<IPaymentsTableProps> = ({ billingContext, contexts
     const [dateRange, setDateRange] = useDateRangeSearch('advancedAt', loading)
 
     useEffect(() => {
-        if (!dateRange && appliedFiltersCount < 1 && !search) {
+        if (!hasFilters && shouldApplyDefaultFilter && !isDefaultFilterApplied) {
+            isDefaultFilterApplied = true
```
akarjakin, line 110: this would affect all instances of this interface, are you sure you want such behaviour ?

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
```diff
@@ -92,8 +106,11 @@ const PaymentsTable: React.FC<IPaymentsTableProps> = ({ billingContext, contexts
...
             setDateRange(DEFAULT_DATE_RANGE)
+        } else {
+            shouldApplyDefaultFilter = false
```
akarjakin, line 113: the same.

I think such flags should be individual for every instance

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
AleX83Xpert: The same behavior underhood. But fixed.

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
AleX83Xpert: These variables are really variables.

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
AleX83Xpert: This hook works only once (`deps = []`) within the table component. It's like old-style `componentDidMount` hook.

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
AleX83Xpert: No, this is mandatory behavior. Here we need to keep value between table rerenders to prevent setting default filters after all filters will be cleaned.

pahaz (approved):

PR #1352, opened 2022-04-06
DOMA-2645 Add Recipient to AcquiringIntegrationContext

dkoviazin:
This recipient will be used if billing uses transit bank account (from another integration)
+ Added createTestRecipient to utils


apps/condo/domains/acquiring/schema/AcquiringIntegrationContext.js
```diff
@@ -55,6 +57,31 @@ const AcquiringIntegrationContext = new GQLListSchema('AcquiringIntegrationConte
...
         },
 
+        reason: {
+            schemaDoc: 'Contract number and date. Basis for invoicing',
+            type: Text,
+            isRequired: false,
+        },
+
+        email: {
+            schemaDoc: 'Email. To send acquiring payments report',
```
AntonAL, line 67:
```suggestion
            schemaDoc: 'Acquiring payments report will be sent to this email',
```

AntonAL (approved):

apps/condo/domains/acquiring/schema/AcquiringIntegrationContext.js
```diff
@@ -55,6 +57,31 @@ const AcquiringIntegrationContext = new GQLListSchema('AcquiringIntegrationConte
...
+                    return normalizeEmail(resolvedData['email']) || resolvedData['email']
+                },
+            },
+        },
+
+        recipient: {
+            ...RECIPIENT_FIELD,
+            schemaDoc: 'Requisites from contract. Are used for invoicing in case of transit accounts or missing requisites',
+            isRequired: false,
+        },
```
toplenboren, line 83: Usually in case of transit bank you have multiple recipients, how are we going to handle this case? 

Also, just a take to think on: maybe this data should lie inside `rb` service? As some kind of mapper, which sees recipient, parse that this is transit bank recipient and then maps it to some other place?

dkoviazin: Changed documentation for the field. Recipient for context will be used in case of dealing with receipts connected to not confirmed recipients. Transit account is only a special case.

toplenboren (approved):

apps/condo/schema.ts
```diff
@@ -448,6 +448,12 @@ export type AcquiringIntegrationContext = {
   settings?: Maybe<Scalars['JSON']>;
```
SavelevMatthew, line 448: Rebase schema)

apps/condo/migrations/20220406105217-0117_auto_20220406_0552.js
```diff
@@ -0,0 +1,66 @@
+// auto generated by kmigrator
```
SavelevMatthew, line 1: don't forget to rebase (services are coming)

SavelevMatthew (approved):

PR #1359, opened 2022-04-06
DOMA-2684 fixed access to update ticket 

nomerdvadcatpyat:
we need to give the resident access to update `reviewValue` and `reviewComment` for tickets that are created in crm and visible to the resident in the mobile app

### DIFF
--- a/apps/condo/domains/ticket/access/Ticket.js
+++ b/apps/condo/domains/ticket/access/Ticket.js
@@ -94,6 +94,7 @@ async function canManageTickets ({ authentication: { item: user }, operation, it
     if (!user) return throwAuthenticationError()
     if (user.deletedAt) return false
     if (user.isAdmin) return true
+    let ticket
 
     if (user.type === RESIDENT) {
         let unitName, propertyId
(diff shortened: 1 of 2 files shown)


apps/condo/domains/ticket/access/Ticket.js
```diff
@@ -123,7 +121,39 @@ async function canManageTickets ({ authentication: { item: user }, operation, it
             deletedAt: null,
         })
 
-        return residents.length > 0
+        if (residents.length === 0) {
+            return false
+        }
+
+        if (operation === 'create') {
```
pahaz, line 128: Looks like resident can create any ticket to any organization

apps/condo/domains/ticket/access/Ticket.js
```diff
@@ -123,7 +121,39 @@ async function canManageTickets ({ authentication: { item: user }, operation, it
...
+
+            if (ticket.createdBy === user.id) {
+                inaccessibleUpdatedFields = omit(originalInput, ['dv', 'sender', 'details', 'reviewValue', 'reviewComment'])
+            } else {
+                const ticketContactId = get(ticket, 'contact')
+                if (!ticket.canReadByResident) return false
+                if (!ticketContactId) return false
+
+                const ticketContact = await getById('Contact', ticketContactId)
+                if (user.phone !== ticketContact.phone) {
```
pahaz, line 141: Tricky access :) do we really need it?)

apps/condo/domains/ticket/access/Ticket.js
nomerdvadcatpyat: No, on line 126, if there are no residents by property, false is returned.
Added test for this case

apps/condo/domains/ticket/access/Ticket.js
nomerdvadcatpyat: We need to give residents access to update `reviewComment` and `reviewValue` in order to review such tickets.

PR #1360, opened 2022-04-06
DOMA-2685 added sms text normalization and test, fixed some notification templates

akarjakin:


SavelevMatthew (approved):

apps/condo/domains/notification/templates.js
```diff
@@ -97,9 +99,13 @@ function translationStringKeyForPushTitle (messageType) {
     return `notification.messages.${messageType}.${PUSH_TRANSPORT}.title`
 }
 
+function normalizeSMSText (text) {
+    return unescape(text).replace(SMS_FORBIDDEN_SYMBOLS_REGEXP, '*')
```
AleX83Xpert, line 103: I think it's not good to replace symbols to star. It's better to write tests that control these symbols. This approach gives the ability developer to choose any other symbols.

PR #1366, opened 2022-04-07
DOMA-2686 added forced convertation of data fields to string for push notifications

akarjakin:


apps/condo/domains/notification/adapters/firebaseAdapter.js
```diff
@@ -118,6 +119,32 @@ class FirebaseAdapter {
...
+    static prepareData (data = {}) {
+        const result = {}
+        const invalidFields = []
+
+        for (const key in data) {
+            if (data.hasOwnProperty(key)) {
+                result[key] = data[key]
+
+                if (!isString(data[key])) {
+                    result[key] = data[key].toString()
```
AleX83Xpert, line 135: What if `data[key]` is `Number`?
image
```suggestion
                    result[key] = String(data[key])
```

apps/condo/domains/notification/adapters/firebaseAdapter.js
```diff
@@ -118,6 +119,32 @@ class FirebaseAdapter {
...
+                    invalidFields.push([key, data[key]])
```
AleX83Xpert, line 136: Why not Object? Object is more readable after stringifying than an array of arrays.
```suggestion
                    invalidFields[key] = data[key]
```

akarjakin: invalidFields is an Array

AleX83Xpert: I see, but why not Object instead?

AleX83Xpert (approved):

apps/condo/domains/notification/adapters/firebaseAdapter.js
akarjakin: actually no profit of having object here, also calculating object emptiness is more expensive then array emptiness :)

PR #1377, opened 2022-04-11
DOMA-1648 remove back button and title header from all pages

DanilovMaxim13:
### **Back button**

before

image

after
image

### **Title header**
before
image

after
image


AleX83Xpert (changes requested):
You have removed `ReturnBackHeaderAction` and `TitleHeaderAction`. But these components still used in submodules: `callcenter` and `eps`
image
image
These modules won't work.

AleX83Xpert (approved):
Please do not forget about submodules :)

nomerdvadcatpyat (approved):

PR #1391, opened 2022-04-12
DOMA-2722 fix size footer considering margin

gegius:
Happy pullrequestday

### DIFF
--- a/apps/condo/domains/user/components/containers/styles.tsx
+++ b/apps/condo/domains/user/components/containers/styles.tsx
@@ -17,7 +17,7 @@ export const Footer = styled.div<IFooterProps>`
             position: fixed;
           `
 }
-  width: 100%;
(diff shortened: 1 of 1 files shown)


apps/condo/domains/user/components/containers/styles.tsx
```diff
@@ -17,7 +17,7 @@ export const Footer = styled.div<IFooterProps>`
             position: fixed;
           `
 }
-  width: 100%;
+  width: 95%;
```
pahaz, line 20: :👍

pahaz (approved):

apps/condo/domains/user/components/containers/styles.tsx
pahaz: it looks little strange

PR #1403, opened 2022-04-13
HOTFIX Added support access for creating billing / acquiring context

SavelevMatthew:


Kanol (approved):

apps/condo/domains/acquiring/schema/AcquiringIntegrationContext.js
```diff
@@ -33,6 +33,11 @@ const AcquiringIntegrationContext = new GQLListSchema('AcquiringIntegrationConte
             isRequired: true,
             knexOptions: { isNotNullable: true }, // Required relationship only!
             kmigratorOptions: { null: false, on_delete: 'models.PROTECT' },
+            access: {
+                create: true,
+                read: true,
+                update: false,
+            },
```
toplenboren, line 40: Maybe we can use `ORGANIZATION_OWNED_FIELD` as it is in `BillingContext` ?

toplenboren (approved):

apps/condo/domains/acquiring/schema/AcquiringIntegrationContext.js
SavelevMatthew: Not sure about `model.Cascade` there, since it's not safe enough (at least for now)

PR #1407, opened 2022-04-14
DOMA-2430 Add link to responsible employee

Kanol:
Problem: In division card we have links to executors, but doesn't have link to responsible employee
Solution: Add link

### DIFF
--- a/apps/condo/pages/division/[id]/index.tsx
+++ b/apps/condo/pages/division/[id]/index.tsx
@@ -65,6 +65,8 @@ export const DivisionPageContent = ({ division, loading, columns, role }: Divisi
         ...executor,
     }))
 
+    const responsible = get(division, 'responsible')
+
     const dataSource = useMemo(() => {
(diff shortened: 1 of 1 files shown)


apps/condo/pages/division/[id]/index.tsx
```diff
@@ -76,7 +78,11 @@ export const DivisionPageContent = ({ division, loading, columns, role }: Divisi
             </Typography.Title>
             <Row gutter={[0, 20]}>
                 <PageFieldRow labelSpan={5} title={ResponsibleLabelMessage} highlight>
-                    {get(division, ['responsible', 'name'])}
+                    <Link href={`/employee/${responsible.id}`}>
```
Comment (dkoviazin) [inline apps/condo/pages/division/[id]/index.tsx:81]: responsible can be null

Comment (Kanol) [reply apps/condo/pages/division/[id]/index.tsx:81]: Responsible is required field
image

dkoviazin (approved):

PR #1414, opened 2022-04-14
DOMA-2598 connect yandex metrica

DanilovMaxim13:


.env.example
```diff
@@ -32,3 +32,6 @@ FIREBASE_CONFIG_JSON='{"type": "service_account", "project_id": "<PROJECT_ID>",
 
 # real Firebase push token for testing real push notification delivery to device, would be used only if provided
 # FIREBASE_PUSH_TOKEN_TEST=flt0weSOS8eROf6OEZAmHp:APA91bG0I-QBvGjCL5jblzDoOuVV6bZ1x4dyRmyPqr2iZBYyFbDJcqtKw0vvzo4MH-PaLiVQJbLfAGCoryYXykdqCKXdx2QqYAk6vE4pmxrKk6RHe33-mVNHNTB0HxYI9KUsb21CHfPp
+
+#
+YANDEX_METRIKA_ID="88253038"
```
SavelevMatthew, line 37: I guess it should not be here, since it's not mandatory...Instead you should connect metrics only if it's specified in config (optional)

SavelevMatthew (commented):
And where is submodule update for .helm?)

toplenboren (approved):

.env.example
toplenboren: Yes, please delete this

SavelevMatthew (approved):

PR #1417, opened 2022-04-15
DOMA-2704 Add refetching query on finish

Kanol:
Problem: After loading data from excel file the last record doesn't shown during the page manual reload. It happens because of data doesn't refetching after upload (only refetching during upload because of parent rerender)
Solution: Add refetch after upload complete

### DIFF
--- a/apps/condo/domains/common/components/Import/Index.tsx
+++ b/apps/condo/domains/common/components/Import/Index.tsx
@@ -156,6 +156,7 @@ const ImportWrapper: React.FC<IImportWrapperProps> = (props) => {
                 const config = getUploadSuccessModalConfig(ImportTitle, message, ImportOKMessage)
                 activeModal.current = modal.success(config)
             }
+            onFinish()
         },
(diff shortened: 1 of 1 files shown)


gegius (approved):

dkoviazin (approved):

apps/condo/domains/common/components/Import/Index.tsx
```diff
@@ -156,6 +156,7 @@ const ImportWrapper: React.FC<IImportWrapperProps> = (props) => {
                 const config = getUploadSuccessModalConfig(ImportTitle, message, ImportOKMessage)
                 activeModal.current = modal.success(config)
             }
+            onFinish()
```
akarjakin, line 159: naming collision, makes confusion, can you please deconstruct `props.onFinish` as `handleFinish`, for example?

Kanol: done

PR #1435, opened 2022-04-18
DOMA-2726 disappearing of number on register layout

gegius:
https://doma.atlassian.net/browse/DOMA-2726

### DIFF
--- a/apps/condo/domains/user/components/auth/ValidatePhoneForm.tsx
+++ b/apps/condo/domains/user/components/auth/ValidatePhoneForm.tsx
@@ -145,11 +145,7 @@ export const ValidatePhoneForm = ({ onFinish, onReset, title }): React.ReactElem
 
     useEffect(() => {
         const formattedPhone = formatPhone(phone)
-        const phoneVisible =
-            isPhoneVisible ?
-                formattedPhone
-                :
-                `${formattedPhone.substring(0, 9)}***${formattedPhone.substring(12)}`
+        const phoneVisible = isPhoneVisible ? formattedPhone : `${formattedPhone.substring(0, 9)}***${formattedPhone.substring(12)}`
         setShowPhone(phoneVisible)
     }, [isPhoneVisible, phone, setShowPhone])
 
--- a/apps/condo/domains/user/components/auth/ValidatePhoneForm.tsx
+++ b/apps/condo/domains/user/components/auth/ValidatePhoneForm.tsx
@@ -1,5 +1,5 @@
 import { Col, Form, Row, Space, Typography, Input, RowProps } from 'antd'
-import React, { useCallback, useContext, useMemo, useState } from 'react'
+import React, { useCallback, useContext, useEffect, useMemo, useState } from 'react'
 import { FormattedMessage } from 'react-intl'
 import { useMutation } from '@core/next/apollo'
 import { useIntl } from '@core/next/intl'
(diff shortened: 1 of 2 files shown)


gegius:
https://doma.atlassian.net/browse/DOMA-2726

https://doma.atlassian.net/browse/DOMA-2724 (?)

The problem is very old. Register as change-password use a validatePhoneForm. Apparently, in the RegisterContextProvider, when the captcha fails, we erase the user's phone number (why?)
We erase it not even at the stage of working with sending a message, but just like that

As we say, the author quit a long time ago of his own free will

apps/condo/domains/user/components/auth/ValidatePhoneForm.tsx
```diff
@@ -143,13 +143,14 @@ export const ValidatePhoneForm = ({ onFinish, onReset, title }): React.ReactElem
...
-    useEffect(() => {
-        if (isPhoneVisible) {
-            setShowPhone(formatPhone(phone))
-        } else {
-            const unHidden = formatPhone(phone)
-            setShowPhone(`${unHidden.substring(0, 9)}***${unHidden.substring(12)}`)
-        }
+    const handleNumberVisible = useCallback(() => {
+        const formattedPhone = formatPhone(phone)
+        isPhoneVisible
```
akarjakin, line 148: please use ternary operator only for providing values, in this case you need to use classical `if`

akarjakin: thou you're calling the same functions, other possibility here is to precalculate value using ternary operator and then call the function once.

akarjakin:
```
const phoneValue = isPhoneVisible ? formattedPhone : `${formattedPhone.substring(0, 9)}***${formattedPhone.substring(12)}`

setShowPhone(phoneValue)
```

apps/condo/domains/user/components/auth/ValidatePhoneForm.tsx
```diff
@@ -177,12 +178,11 @@ export const ValidatePhoneForm = ({ onFinish, onReset, title }): React.ReactElem
...
                                         phone: (
                                             <span style={{ whiteSpace: 'nowrap' }}>
-                                                {`${showPhone} `}
+                                                {`${formatPhone(showPhone)} `}
                                                 <Typography.Link
                                                     underline
                                                     style={{ color: 'black' }}
-                                                    onClick={() =>
-                                                        setIsPhoneVisible(!isPhoneVisible)}
+                                                    onClick={handleNumberVisible}
```
akarjakin, line 185: More straight forward strategy here would be handling only flag value by this callback, and format value according to this flag, using useMemo

gegius:
Reverted useEffect back as we need to call setShowPhone on start

apps/condo/domains/user/components/auth/ValidatePhoneForm.tsx
```diff
@@ -144,12 +144,13 @@ export const ValidatePhoneForm = ({ onFinish, onReset, title }): React.ReactElem
...
-        } else {
-            const unHidden = formatPhone(phone)
-            setShowPhone(`${unHidden.substring(0, 9)}***${unHidden.substring(12)}`)
-        }
+        const formattedPhone = formatPhone(phone)
+        const phoneVisible =
+            isPhoneVisible ?
+                formattedPhone
+                :
+                `${formattedPhone.substring(0, 9)}***${formattedPhone.substring(12)}`
```
pahaz, line 152:
```suggestion
        const phoneVisible = isPhoneVisible ? formattedPhone : `${formattedPhone.substring(0, 9)}***${formattedPhone.substring(12)}`
```

pahaz (approved):

sitozzz (approved):

PR #1454, opened 2022-04-20
DOMA-2772 toggle visibility of `<EmptyListView/>` when using with `<ImportWrapper />`

sitozzz:


pahaz:
What's going on here?)

sitozzz:
> What's going on here?)

This PR fixes the problem when data is imported into an empty table. In the previous implementation, this case was working and the import component was removed from the DOM

pahaz (approved):

dkoviazin (approved):

PR #1470, opened 2022-04-21
DOMA-2503 Fix tables scrolling in low descktop resolution

VKislov:
Screenshot from 2022-04-21 17-49-17
Screenshot from 2022-04-21 17-50-49
Screenshot from 2022-04-21 17-53-37

### DIFF
--- a/apps/condo/domains/common/components/LayoutContext.tsx
+++ b/apps/condo/domains/common/components/LayoutContext.tsx
@@ -9,6 +9,7 @@ const { useBreakpoint } = Grid
 interface ILayoutContext {
     isMobile?: boolean
     isSmall?: boolean
+    tableScrollInLowResolution?: boolean
     breakpoints?: ScreenMap
     isCollapsed?: boolean
     toggleCollapsed?: () => void
@@ -34,6 +35,7 @@ export const LayoutContextProvider: React.FC = (props) => {
     }
 
     const isSmall = (breakpoints.md || breakpoints.xs || breakpoints.sm) && !breakpoints.lg
+    const tableScrollInLowResolution = (breakpoints.md || breakpoints.xs || breakpoints.sm) && !breakpoints.xl
 
     useEffect(() => {
         const isCollapsed = localStorage.getItem('isCollapsed') === 'true'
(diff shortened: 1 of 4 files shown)


apps/condo/domains/common/components/LayoutContext.tsx
```diff
@@ -34,6 +35,7 @@ export const LayoutContextProvider: React.FC = (props) => {
     }
 
     const isSmall = (breakpoints.md || breakpoints.xs || breakpoints.sm) && !breakpoints.lg
+    const tableScrollInLowResolution = (breakpoints.md || breakpoints.xs || breakpoints.sm) && !breakpoints.xl
```
akarjakin, line 38: maybe `lowResScroll` ? It's anyway used within tables, so we can make it a bit shorter.

akarjakin (approved):

apps/condo/domains/common/components/LayoutContext.tsx
VKislov: Fixed

sitozzz (approved):
LGTM!

apps/condo/domains/common/components/LayoutContext.tsx
```diff
@@ -9,6 +9,7 @@ const { useBreakpoint } = Grid
 interface ILayoutContext {
     isMobile?: boolean
     isSmall?: boolean
+    lowResScroll?: boolean
```
AleX83Xpert, line 12: This variable is defined within `LayoutContext`, so it's good to know without debugging what this variable using for.
Also, this is a boolean variable. A good practice to name such variables starting with `is`, `has`, `should`.
As for this particular one, I suggest `shouldTableScroll`.

apps/condo/domains/common/components/LayoutContext.tsx
AleX83Xpert: I disagree. See my [comment below](#discussion_r855346923).

apps/condo/domains/common/components/LayoutContext.tsx
```diff
@@ -34,6 +35,7 @@ export const LayoutContextProvider: React.FC = (props) => {
     }
 
     const isSmall = (breakpoints.md || breakpoints.xs || breakpoints.sm) && !breakpoints.lg
+    const shouldTableScroll = (breakpoints.md || breakpoints.xs || breakpoints.sm) && !breakpoints.xl
```
pahaz, line 38: why you don't use `isSmall` in this case?

PR #1506, opened 2022-04-29
DOMA-2887 fixed infinity scroll in gql search

nomerdvadcatpyat:


sitozzz (approved):

akarjakin (approved):

dkoviazin (approved):

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -88,8 +88,9 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
             const initialOptions = await search(client, null, initialValueQuery, initialValue.length)
```
AntonAL, line 88: Maybe to move `setLoading(true)` inside condition and keep `setLoading(false)` also inside it to avoid setting and resetting loading state when condition is not met.

AntonAL (approved):

PR #1519, opened 2022-05-04
DOMA-2527 fixed cancle edit action. Add state editCanсeled

VKislov:
There were two options to write a handleChange function and use it instead of handleSave in onChange. Or just add an non-save condition to handleSave.
Initially, it was just changing the mod to 'display' to cancel, but since onChange uses handleSave, any content change is always saved

### DIFF
--- a/apps/condo/domains/common/components/Comments/Comment.tsx
+++ b/apps/condo/domains/common/components/Comments/Comment.tsx
@@ -95,17 +95,21 @@ export const Comment: React.FC<ICommentProps> = ({ comment, updateAction, delete
 
     const [mode, setMode] = useState<CommentMode>('display')
     const [content, setContent] = useState(comment.content)
+    const [editCanceled, setEditCanceled] = useState(false)
 
     const [dateShowMode, setDateShowMode] = useState<'created' | 'updated'>('created')
     const handleSave = (newContent) => {
-        updateAction({ content: newContent }, comment)
-            .then(() => {
-                setMode('display')
-                setContent(newContent)
-            })
+        if (!editCanceled) {
+            updateAction({ content: newContent }, comment)
+                .then(() => {
+                    setMode('display')
+                    setContent(newContent)
+                })
+        }
     }
 
     const handleCancelSave = () => {
+        setEditCanceled(true)
         setMode('display')
     }
 


apps/condo/domains/common/components/Comments/Comment.tsx
```diff
@@ -95,17 +95,22 @@ export const Comment: React.FC<ICommentProps> = ({ comment, updateAction, delete
 
     const [mode, setMode] = useState<CommentMode>('display')
     const [content, setContent] = useState(comment.content)
+    const [editCanceled, setEditCanceled] = useState<boolean>(false)
```
SavelevMatthew, line 98: If you provide `false` as default values, hook will automatically generate boolean type

SavelevMatthew (approved):

apps/condo/domains/common/components/Comments/Comment.tsx
```diff
@@ -95,17 +95,22 @@ export const Comment: React.FC<ICommentProps> = ({ comment, updateAction, delete
...
 
     const [dateShowMode, setDateShowMode] = useState<'created' | 'updated'>('created')
+
     const handleSave = (newContent) => {
-        updateAction({ content: newContent }, comment)
-            .then(() => {
-                setMode('display')
-                setContent(newContent)
-            })
+        if (editCanceled) {
```
dkoviazin, line 103: Maybe, 
```js
if (!editCanceled) { ...
```

sitozzz (approved):

pahaz:
Why you add a new submode here?

pahaz:
Remove merge commit. Use rebase

dkoviazin (approved):

dkoviazin (changes requested):
Condo submodule

dkoviazin (approved):

AntonAL (commented):
You have two buttons with same `key="save"`

apps/condo/domains/common/components/Comments/Comment.tsx
```diff
@@ -95,17 +95,21 @@ export const Comment: React.FC<ICommentProps> = ({ comment, updateAction, delete
...
+            updateAction({ content: newContent }, comment)
+                .then(() => {
+                    setMode('display')
+                    setContent(newContent)
+                })
+        }
     }
 
     const handleCancelSave = () => {
+        setEditCanceled(true)
```
AntonAL, line 112: I've tested it locally, when you click cancel, you cannot save edited text anymore.
It's because `editCanceled` is set permanently here.

AntonAL (changes requested):

VKislov:
Comment component has been changed and now this PR is not relevant. The task was solved with another PR

PR #1525, opened 2022-05-04
DOMA-2269 Remove Consts with deprecated tag

VKislov:


pahaz:
Your sub module should point to master. You need to create pr to each sub module and merge it. Then create the condo pr

PR #1526, opened 2022-05-04
HOTFIX Excluded middlewares on type generation

SavelevMatthew:


akarjakin (approved):

packages/@core.keystone/test.utils.js
```diff
@@ -88,15 +88,16 @@ function setFakeClientMode (path) {
     __isAwaiting = true
 }
 
-const prepareKeystoneExpressApp = async (entryPoint, { excludeApps } = {}) => {
+const prepareKeystoneExpressApp = async (entryPoint, { excludeApps, remainingApps } = {}) => {
```
AntonAL, line 91: Let's shorten it:

```suggestion
const prepareKeystoneExpressApp = async (entryPoint, { excludeApps, apps: remainingApps } = {}) => {
```

bin/create-graphql-schema.js
```diff
@@ -22,7 +22,9 @@ generates:
 
 
 async function getGraphQLSchema (keystoneModule) {
-    const { keystone } = await prepareKeystoneExpressApp(keystoneModule)
+    const { keystone } = await prepareKeystoneExpressApp(keystoneModule, {
+        remainingApps: ['GraphQLApp'],
```
AntonAL, line 26:
```suggestion
        apps: ['GraphQLApp'],
```

pahaz:
It’s better to use excluded apps. You really don’t know about future apps. If some app will change the GQl schema? 

You will hide the error in such cases. I’m add excluded apps especially for that reason rather than remaining apps.

pahaz:
I have use excluded just for performance optimization. Why you need to another one option? It may hide an error

PR #1588, opened 2022-05-12
DOMA-2489 Added a field and a resolver to the schema. Created a migra…

VKislov:
…tion.


apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -243,6 +243,12 @@ const Ticket = new GQLListSchema('Ticket', {
             defaultValue: false,
             isRequired: true,
         },
+        files: {
```
pahaz, line 246: Don't add M2M to Ticket! It's too heavy

VKislov:
The solution is not optimal. The task was frozen

PR #1589, opened 2022-05-12
DOMA-2987 fix admin ui on build phaze

dkoviazin:
Register all schema in build phase (if not - they will not appear in admin ui)

### DIFF
--- a/apps/condo/admin-ui/index.js
+++ b/apps/condo/admin-ui/index.js
@@ -1,3 +1,4 @@
+// @ts-ignore
 import { ItemId, AddNewItem } from '@keystonejs/app-admin-ui/components'
 import React, { useCallback } from 'react'
 import { useMutation, gql } from '@apollo/client'
--- a/apps/condo/index.js
+++ b/apps/condo/index.js
@@ -67,32 +67,29 @@ const keystone = new Keystone({
 // Because Babel is used only for frontend to transpile and optimise code,
 // backend files will bring unnecessary workload to building stage.
 // They can be safely ignored without impact on final executable code
-if (IS_BUILD_PHASE) {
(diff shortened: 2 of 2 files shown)


apps/condo/admin-ui/index.js
```diff
@@ -1,3 +1,4 @@
+// @ts-ignore
```
pahaz, line 1: do we really need it for js files?

apps/condo/index.js
```diff
@@ -67,33 +67,30 @@ const keystone = new Keystone({
...
 // They can be safely ignored without impact on final executable code
-if (IS_BUILD_PHASE) {
-    // `User` list is used by `createAuthStrategy`, that creates `AuthStrategy`, required to build /admin/ static files
-    registerSchemas(keystone, [
-        require('@core/keystone/schemas/User'),
-    ])
-} else {
+
+// We need to register all schemas as they will appear in admin ui
+registerSchemas(keystone, [
```
pahaz, line 72: do we really need all these domains on the build stage?

pahaz (approved):

apps/condo/admin-ui/index.js
dkoviazin: git hub warnings :-)
image

sitozzz (approved):

apps/condo/admin-ui/index.js
pahaz: :)) GitHub think that we use TS

apps/condo/index.js
pahaz: Omg, thank.

pahaz (approved):

PR #1604, opened 2022-05-16
DOMA-3000 add `unitType` field to `Contact` domain

sitozzz:


sitozzz:
Also had submodule dependency https://github.com/open-condo-software/condo_call_center/pull/32

apps/condo/domains/contact/schema/Contact.js
```diff
@@ -39,6 +39,10 @@ const Contact = new GQLListSchema('Contact', {
             isRequired: false,
         },
 
+        unitType: {
```
AntonAL, line 42: Why not to assign it directly without destructiring?

apps/condo/domains/contact/schema/Contact.js
```diff
@@ -96,14 +100,15 @@ const Contact = new GQLListSchema('Contact', {
...
-            const { property, unitName, name, phone } = newItem
+            const { property, unitName, unitType, name, phone } = newItem
 
             if (operation === 'create' && !property) {
                 return addValidationError(`${ PROPERTY_REQUIRED_ERROR }] no property for contact`)
             }
             const condition = {
                 property: { id: property },
                 unitName: unitName || null,
+                unitType,
```
AntonAL, line 111: What will happen if we will try to create a contact, having another contact with ‘unitType’ with ‘null’ value, that was created before this changes in validation logic?

apps/condo/domains/contact/schema/Contact.js
sitozzz: Done

apps/condo/domains/contact/schema/Contact.js
AntonAL: > @AntonAL We won't have a case where `unitType` can be `null`, since all values will be overwritten by the default value via migration

@sitozzz, I don't see a default value setting in migration.
Take a look at:
```
ALTER TABLE "Contact" ALTER COLUMN "unitType" DROP DEFAULT;
```

sitozzz: > > @AntonAL We won't have a case where `unitType` can be `null`, since all values will be overwritten by the default value via migration
> 
> @sitozzz, I don't see a default value setting in migration. Take a look at:
> 
> ```
> ALTER TABLE "Contact" ALTER COLUMN "unitType" DROP DEFAULT;
> ```

@AntonAL plz take a look at this line of code `ALTER TABLE "Contact" ADD COLUMN "unitType" text DEFAULT 'flat' NOT NULL;`

AntonAL (approved):

dkoviazin (approved):

PR #1645, opened 2022-05-25
DOMA-3088 Ability to send commands to IFrame and send answer back

AleX83Xpert:


apps/condo/domains/common/components/IFrame/IFrame.tsx
```diff
@@ -84,6 +87,17 @@ export const IFrame: React.FC<IFrameProps> = (props) => {
         setFrameHeight(message.height)
     }, [])
 
+    const handleCommand = useCallback((message) => {
+        const iFrameReceiver = get(iFrameRef, 'current.contentWindow', null)
+        switch (message.command) {
+            case 'getOrganization':
```
sitozzz, line 93: May be add default case with something like `Not implemented error`?

sitozzz (approved):

apps/condo/domains/common/components/IFrame/IFrame.tsx
AleX83Xpert: I omitted this intentionally to reduce future refactoring. For now this is only internal code.

PR #1662, opened 2022-05-30
DOMA-3138 fixed ticket number overlapping by indicator on small screens

nomerdvadcatpyat:
Before:
изображение

After:
изображение


AntonAL:
Looks ugly in both cases :)
What about to inline indicator and text and wrap it?
Is number wrapping acceptable in this case?

AntonAL (approved):

akarjakin (approved):

dkoviazin (approved):

PR #1663, opened 2022-05-30
Feature/condo/doma 2902/send current user to iframe

AleX83Xpert:
To be sure that userLoading value passed to the context in the same time with `userData` we pass this value through `useEffect` as `userData`.
Also the new command for CondoBridge was added.

### DIFF
--- a/packages/@core.next/auth.jsx
+++ b/packages/@core.next/auth.jsx
@@ -69,10 +69,16 @@ const AuthProvider = ({ children, initialUserValue }) => {
 
     const { data: userData, loading: userLoading, error: userError, refetch } = useQuery(USER_QUERY)
 
+    const [isUserLoading, setIsUserLoading] = useState(userLoading)
+
     useEffect(() => {
         if (userData) onData(userData)
         if (userError) onError(userError)
     }, [userData, userError])
+
+    useEffect(() => {
+        setIsUserLoading(userLoading)
+    }, [userLoading])
 
     const [signin, { loading: signinLoading }] = useMutation(SIGNIN_MUTATION, {
         onCompleted: async ({ authenticateUserWithPassword: { item } = {}, error }) => {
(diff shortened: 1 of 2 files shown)


pahaz (approved):

packages/@core.next/auth.jsx
```diff
@@ -69,11 +69,17 @@ const AuthProvider = ({ children, initialUserValue }) => {
...
 
+    const [isUserLoading, setIsUserLoading] = useState(userLoading)
+
     useEffect(() => {
         if (userData) onData(userData)
         if (userError) onError(userError)
     }, [userData, userError])
 
+    useEffect(() => {
+        setIsUserLoading(userLoading)
```
pahaz, line 80: it's better to check the value before the call setState to avoid redundant rerendering. But, in that case it's not required.

packages/@core.next/auth.jsx
```diff
@@ -127,7 +133,7 @@ const AuthProvider = ({ children, initialUserValue }) => {
         <AuthContext.Provider
             value={{
                 isAuthenticated: !!user,
-                isLoading: userLoading || signinLoading || signoutLoading,
+                isLoading: isUserLoading || signinLoading || signoutLoading,
```
pahaz, line 136: maybe we need to remove `signinLoading || signoutLoading` ?

pahaz (approved):

packages/@core.next/auth.jsx
AleX83Xpert: if the `userLoading` variable is the same as current state, re-render won't occurred.
https://reactjs.org/docs/hooks-reference.html#bailing-out-of-a-state-update

packages/@core.next/auth.jsx
AleX83Xpert: These two variables are showing other loading states within this hook. I'd like to keep them here.

PR #1676, opened 2022-05-31
Parallel test execution

pahaz:


packages/@core.keystone/fields/AutoIncrementInteger/Implementation.js
```diff
@@ -31,10 +35,32 @@ class AutoIncrementIntegerKnexFieldAdapter extends Integer.adapters.knex {
...
+                    currentMaxNumber = max
+                }
+            }
+            currentMaxNumber = parseInt(currentMaxNumber || 0)
+            await taskQueue.client.set(redisMaxValueKey, currentMaxNumber + 1)
+            return currentMaxNumber + 1
+        } finally {
+            await lock.release()
+        }
+    }
```
AntonAL, line 63: Why not to use sequences in database?

pahaz: Because you can set the number by hands. And we probably will change this logic in a future. We want to use organization specific numbers. Uniq for organization

pahaz: I don't found a way to use it without too match hacks ...

toplenboren (approved):

gegius (approved):

PR #1706, opened 2022-06-06
DOMA-2729 change logo (update cc submodule)

VKislov:


AntonAL (commented):
Changed logo has different vertical position of text.
Is it still aligned with other layout elements.
[public/logoDoma.svg](https://github.com/open-condo-software/condo/pull/1706/files#diff-96c85f7a0dc3a77d845424f80a452408f916fcd90eaae91ed62bd3143e493de6)

VKislov:
According to the new layout, the text should have moved. In order to accurately convey the required position, a svg was added with a house icon and an inscription with the desired position. It seems that this is ok because svg with text, but without a house, is not used anywhere except for the logo. This moved the sun icon, which I corrected in the condo repository. The Logo component itself is taken from the condo repository. 
Logo now:
image

PR #1714, opened 2022-06-07
Fix normalisation and expiresAt check!

pahaz:


apps/condo/domains/user/schema/RegisterNewUserService.js
```diff
@@ -99,28 +100,31 @@ const RegisterNewUserService = new GQLCustomSchema('RegisterNewUserService', {
             },
             resolver: async (parent, args, context) => {
                 const { data } = args
-                const { confirmPhoneActionToken, ...restUserData } = data
+                // TODO(DOMA-3209): check dv! and make it required
+                const { confirmPhoneActionToken, dv, phone, email, ...restUserData } = data
                 const userData = {
                     ...restUserData,
+                    email: normalizeEmail(email),
+                    phone: normalizePhone(phone),
```
AntonAL, line 108: AFAIK, normalizePhone will return null in case of wrong format. Maybe to validate here explicitly?

pahaz: yes, bun not in that PR

sitozzz (approved):

PR #1725, opened 2022-06-09
HOTFIX debug why x-request-cache was not working!

toplenboren:
This PR debugs and fixes why `X-Request-ID not working!`

### DIFF
--- a/apps/condo/index.js
+++ b/apps/condo/index.js
@@ -21,7 +21,7 @@ const { formatError } = require('@condo/domains/common/utils/apolloErrorFormatte
 const { hasValidJsonStructure } = require('@condo/domains/common/utils/validation.utils')
 const { SbbolRoutes } = require('@condo/domains/organization/integrations/sbbol/routes')
 const FileAdapter = require('@condo/domains/common/utils/fileAdapter')
-const { KeystoneCacheMiddleware } = require('@core/keystone/cache')
+const { KeystoneCacheMiddleware, initCache } = require('@core/keystone/cache')
 const { expressErrorHandler } = require('@condo/domains/common/utils/expressErrorHandler')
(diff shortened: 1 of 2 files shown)


apps/condo/index.js
```diff
@@ -71,6 +71,9 @@ const keystone = new Keystone({
     },
 })
 
+const cache = {}
+initCache(keystone, {})
```
AleX83Xpert, line 75: What about resetting the cache after the request is made?

toplenboren: /**
 * Keystone-level cache.
 *
 * This package adds basic caching capabilities to keystone.
 *
 * Requirements:
 * 1. Every request should have x-request-id. If it doesn't have it -- this request will not use cache
 *
 * How it works:
 *                           -> [GQL Request] -> [DB]
 * [HTTP Request] -> [Cache] -> [GQL Request] -> [DB]
 *                           -> [GQL Request] -> [DB]
 *
 * On every incoming request module simply checks x-request-id header and operation.
 * If request is `read` - return cache item (assuming cache has this item)
 * If request has other type - remove cache item by request id
 *
 * Cache structure:
 * cache: {
 *     <x-request-id>: {
 *      	<request-args>: []
 *      	<request-args>: []
 *      	<request-args>: []
 *     }
 * }
 *
 */

packages/@core.keystone/cache.js
```diff
@@ -131,5 +128,6 @@ const initCache = (keystone, cache) => {
 
 module.exports = {
     KeystoneCacheMiddleware,
+    initCache,
```
AleX83Xpert, line 131: Looks like this export should be removed after this refactoring 5b969cb5ea9d1c4b3e4d2a1e2af33d1b85218fa4

AleX83Xpert (approved):

dkoviazin (approved):

PR #1727, opened 2022-06-10
DOMA-3213 fix validation placeClassifier&categoryClassifier on ticket…

VKislov:
… creation page. 
`setFieldsValue` does not call `onFieldsChange` or `onValuesChane`. Only user interaction can trigger the change event. To properly link category and place fields into valid pairs, the select logic is described using `setFields`. Therefore forced validation is needed after `setFields`

### DIFF
--- a/apps/condo/domains/ticket/components/TicketClassifierSelect.tsx
+++ b/apps/condo/domains/ticket/components/TicketClassifierSelect.tsx
@@ -325,6 +325,8 @@ export const useTicketThreeLevelsClassifierHook = ({ initialValues: {
             { name: 'categoryClassifier', value: ruleRef.current.category },
             { name: 'problemClassifier', value: ruleRef.current.problem },
         ])
+        // the validation call is forced because antd does not make the field touched with setFields
+        ticketForm.current.validateFields(['placeClassifier', 'categoryClassifier', 'problemClassifier'])
     }
     // We need to find out whether user is still following classifiers rules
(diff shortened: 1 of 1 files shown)


akarjakin (approved):

sitozzz (approved):

apps/condo/domains/ticket/components/TicketClassifierSelect.tsx
```diff
@@ -325,6 +325,8 @@ export const useTicketThreeLevelsClassifierHook = ({ initialValues: {
             { name: 'categoryClassifier', value: ruleRef.current.category },
             { name: 'problemClassifier', value: ruleRef.current.problem },
         ])
+        // the validation call is forced because antd does not make the field touched with setFields
```
AleX83Xpert, line 328: > ... antd does not make the field touched with setFields

Actually [it](https://ant.design/components/form/#FormInstance) [does](https://ant.design/components/form/#FieldData) :) But here we call validation programmatically according to our requirements.

AleX83Xpert (approved):

apps/condo/domains/ticket/components/TicketClassifierSelect.tsx
VKislov:
```suggestion
        // calling validation programmatically is necessary because antd does not mark the field as changed in `FieldData.touched` when using `setFields`
```

PR #1728, opened 2022-06-10
WiP Doma 2681 Adapter level cache proof of concept 

toplenboren:
> Not intended to be merged right now!


apps/condo/index.js
```diff
@@ -71,6 +73,9 @@ const keystone = new Keystone({
     },
 })
 
+const cache = {}
+initCache(keystone, cache)
```
pahaz, line 77: why you need this? you can path everything inside you app. Just use:
```
keystone.adapter.listAdapters ...
```

packages/@core.keystone/cache.js
```diff
@@ -131,5 +139,6 @@ const initCache = (keystone, cache) => {
 
 module.exports = {
     KeystoneCacheMiddleware,
+    initCache,
```
pahaz, line 142: =)

PR #1757, opened 2022-06-15
DOMA-3259 use `emailFrom` field as value for `Reply-To` header

AleX83Xpert:
HOTFIX


apps/condo/domains/common/constants/requisites.js
```diff
@@ -2,5 +2,4 @@ module.exports = {
     SUPPORT_EMAIL: 'help@doma.ai',
```
akarjakin, line 2: maybe some day we would want to move all this to .env vars

akarjakin (approved):

apps/condo/domains/common/constants/requisites.js
AleX83Xpert: ¯\_(ツ)_/¯

PR #1792, opened 2022-06-20
DOMA-3079 BillingAccount.fullName

toplenboren:
# Problem

For some special cases we need to collect fullName from billing. For example: 

SBOL report from GIS-GKH and Accounting Bank requires fullName 

Plus, collecting this data is good for analytics :-)

# Solution

Add new field `formData` for `BillingAccount` model


pahaz (approved):

toplenboren:
Если тесты пройдут, я замержу!

AntonAL (approved):

PR #1832, opened 2022-06-28
DOMA-3121 new notification when ticket is created

nomerdvadcatpyat:
изображение

### DIFF
--- a/apps/condo/domains/common/utils/mutations.utils.js
+++ b/apps/condo/domains/common/utils/mutations.utils.js
@@ -50,10 +50,10 @@ const getMessageFrom = (error) => (
  * @param form
  * @param {ErrorToFormFieldMsgMapping} ErrorToFormFieldMsgMapping - mapping of errors either in old or new format
  * @param {null|String|} [OnErrorMsg] - controls passing errors to Ant `notification` util
- * @param OnCompletedMsg
+ * @param getCompletedNotification
  * @return {*}
  */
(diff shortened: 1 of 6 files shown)


apps/condo/domains/ticket/components/TicketForm/CreateTicketForm.tsx
```diff
@@ -97,13 +103,27 @@ export const CreateTicketForm: React.FC = () => {
...
     }), [auth.user.id])
 
+    const getCompletedNotification = useCallback((data) => ({
+        message: (
+            <Typography.Text strong>
+                {intl.formatMessage({ id: 'pages.condo.ticket.notification.success.message' }, { number: data.number })}
+            </Typography.Text>
+        ),
+        description: (
+            <a style={LINK_STYLES} href={`/ticket/${data.id}`} target={'_blank'} rel="noreferrer">
```
dkoviazin, line 113: you can use Typography.Link instead of raw html tags

apps/condo/domains/common/utils/mutations.utils.js
```diff
@@ -73,15 +73,15 @@ function runMutation ({ action, mutation, variables, onCompleted, onError, onFin
     return action()
         .then(
             (data) => {
-                if (OnCompletedMsg === null) {
+                if (getCompletedNotification === null) {
```
dkoviazin, line 76: We still have some places in code where we are using OnCompletedMsg. Maybe, you can add one more check if OnCompletedMsg is a function ? It will be more safe then remove old variable and add another

AleX83Xpert (approved):

PR #1849, opened 2022-06-30
DOMA-3282 add payment order field to payment

dkoviazin:
Add payment order field to Payment schema


DanilovMaxim13 (approved):

apps/condo/domains/acquiring/schema/Payment.js
```diff
@@ -186,6 +186,12 @@ const Payment = new GQLListSchema('Payment', {
             defaultValue: PAYMENT_INIT_STATUS,
         },
 
+        order: {
+            schemaDoc: 'Payment order number',
+            type: Text,
```
AleX83Xpert, line 191: `Text`? Why not `Integer`?

apps/condo/migrations/20220630110118-0141_payment_order_paymenthistoryrecord_order.js
```diff
@@ -0,0 +1,34 @@
+// auto generated by kmigrator
```
AleX83Xpert, line 1: Please create the new migration file, cause of my [migration](https://github.com/open-condo-software/condo/pull/1835/files#diff-d88e4b725d67b36fa324ee22d110638e68743f68fcce0358782cd964c40ac1ea), sorry :)

apps/condo/domains/acquiring/schema/Payment.js
dkoviazin: Not sure if payment orders will contain special chars or not - dots, slashes .... And it's not really important to have number field here, as we are not going to make math operations on database layer with this field

apps/condo/migrations/20220630110118-0141_payment_order_paymenthistoryrecord_order.js
dkoviazin: Done

AleX83Xpert (approved):

PR #1852, opened 2022-06-30
DOMA-3375 fixed `reduceNonEmpty` function

VKislov:
Data coming into a function is not always an array. Therefore, the filter count function does not work with such cases.

In this case, the button for clearing filters will be unavailable and the counter of applied filters is empty.
image

### DIFF
--- a/apps/condo/pages/ticket/index.tsx
+++ b/apps/condo/pages/ticket/index.tsx
@@ -77,9 +77,8 @@ export const TicketsPageContent = ({
     const router = useRouter()
     const { filters } = parseQuery(router.query)
 
-    const reduceNonEmpty = (cnt, filter) => cnt + Number(Array.isArray(filters[filter]) && filters[filter].length > 0)
+    const reduceNonEmpty = (cnt, filter) => cnt + Number(filters[filter] && filters[filter].length > 0)
     const appliedFiltersCount = Object.keys(filters).reduce(reduceNonEmpty, 0)
-
     const { MultipleFiltersModal, ResetFiltersModalButton, setIsMultipleFiltersModalVisible } = useMultipleFiltersModal(filterMetas, TicketFilterTemplate)
 
     searchTicketsQuery = { ...searchTicketsQuery, ...{ deletedAt: null } }
--- a/apps/condo/pages/ticket/index.tsx
+++ b/apps/condo/pages/ticket/index.tsx
@@ -77,7 +77,7 @@ export const TicketsPageContent = ({
     const router = useRouter()
     const { filters } = parseQuery(router.query)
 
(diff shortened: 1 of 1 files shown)


apps/condo/pages/ticket/index.tsx
```diff
@@ -77,9 +77,8 @@ export const TicketsPageContent = ({
     const router = useRouter()
     const { filters } = parseQuery(router.query)
 
-    const reduceNonEmpty = (cnt, filter) => cnt + Number(Array.isArray(filters[filter]) && filters[filter].length > 0)
+    const reduceNonEmpty = (cnt, filter) => cnt + Number(filters[filter] && filters[filter].length > 0)
```
AleX83Xpert, line 80: In the case of filters[filter] = `<some Number>` you will get `0` instead of `1`. It means no button to reset filters.
image

AleX83Xpert (changes requested):

apps/condo/pages/ticket/index.tsx
AntonAL, line 80: Could you write a comment about different cases of filters

VKislov:
The `reduceNonEmpty` function is used only in tickets, and since the filter cannot be a single number, the function should work correctly. The clear button works correctly in different cases. Tested on different filters everything works. The search input is also correctly seen by the function
image

apps/condo/pages/ticket/index.tsx
VKislov: There is no case in ticket filters when the filter is a number.

akarjakin, line 80: `filters[filter] && filters[filter].length -> get(filters, 'length')`

apps/condo/pages/ticket/index.tsx
```diff
@@ -40,6 +40,7 @@ import { TablePageContent } from '@condo/domains/common/components/containers/Ba
 import { useWarrantySearch } from '@condo/domains/ticket/hooks/useWarrantySearch'
 import { useFiltersTooltipData } from '@condo/domains/ticket/hooks/useFiltersTooltipData'
 import { useReturnedSearch } from '@condo/domains/ticket/hooks/useReturnedSearch'
+import _ from 'lodash'
```
pahaz, line 43: it's better to use `isString`. or `lodash/isString` or `typeof === 'string'` to reduce the bundle size

akarjakin, line 43: `import isString from 'lodash/isString'
`import isArray from 'lodash/isArray'

AleX83Xpert (approved):

dkoviazin (approved):

PR #1885, opened 2022-07-06
DOMA-3466 add routing in metric

DanilovMaxim13:
added a new package "react-yandex-metrika"

### DIFF
--- a/apps/condo/domains/common/components/containers/YandexMetrika.tsx
+++ b/apps/condo/domains/common/components/containers/YandexMetrika.tsx
@@ -2,8 +2,7 @@ import { useRouter } from 'next/router'
 import { useEffect } from 'react'
 import getConfig from 'next/config'
 
-import { YMInitializer } from 'react-yandex-metrika'
-import ym from 'react-yandex-metrika'
+import ym, { YMInitializer } from 'react-yandex-metrika'
 
 const YandexMetrika = () => {
     const { publicRuntimeConfig } = getConfig()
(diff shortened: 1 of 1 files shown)


sitozzz (changes requested):
You need to remove event listener when component was unmounted!

apps/condo/domains/common/components/containers/YandexMetrika.tsx
```diff
@@ -1,40 +1,36 @@
-import React from 'react'
+import { useRouter } from 'next/router'
+import { useEffect } from 'react'
 import getConfig from 'next/config'
 
+import { YMInitializer } from 'react-yandex-metrika'
```
akarjakin, line 5:
```
import ym, { YMInitializer } from 'react-yandex-metrika'
```

akarjakin (approved):

apps/condo/domains/common/components/containers/YandexMetrika.tsx
```diff
@@ -1,40 +1,41 @@
...
+import ym, { YMInitializer } from 'react-yandex-metrika'
+
 const YandexMetrika = () => {
     const { publicRuntimeConfig } = getConfig()
     const { yandexMetrikaID } = publicRuntimeConfig
 
+    const router = useRouter()
+
+    const routeChangeComplete = () => {
+        ym('hit', window.location.pathname)
```
sitozzz, line 14: Please check the case on first render. In case that we use next with SSR enabled `window` object might be undefined here

sitozzz (approved):

PR #1890, opened 2022-07-07
DOMA-3163 added restriction on downloading files

Alllex202:


Alllex202:
Files with the extension html, txt, svg can currently be downloaded, but in the future, uploading and downloading files of these types (and, most likely, some others) is planned to be limited. Therefore, if files of these types cannot be downloaded, then a notification with the text of the download error will be displayed in this case.

Alllex202:
Sorry, I did not know what the name of the commit would be when resolving a conflict through the site

apps/condo/pages/ticket/[id]/index.tsx
```diff
@@ -106,6 +109,20 @@ export const TicketFileList: React.FC<ITicketFileListProps> = ({ files }) => {
         a.click()
     }
 
+    const handleFileDownload: (file: UploadFile) => void = async (file: UploadFile) => {
+        const forbiddenFileTypes = ['html', 'svg', 'txt']
```
Comment (akarjakin) [inline apps/condo/pages/ticket/[id]/index.tsx:113]: this is static contant, we use to name them in caps like ALLOWED_FILE_TYPES and move outside of functions to be able to reuse

apps/condo/pages/ticket/[id]/index.tsx
```diff
@@ -106,6 +109,20 @@ export const TicketFileList: React.FC<ITicketFileListProps> = ({ files }) => {
...
+        const fileType = file.name.split('.').reverse()[0]
+        if (forbiddenFileTypes.includes(fileType)) {
```
Comment (akarjakin) [inline apps/condo/pages/ticket/[id]/index.tsx:115]: I think you meant allowed, not forbidden.

akarjakin (changes requested):

apps/condo/pages/ticket/[id]/index.tsx
```diff
@@ -106,6 +109,20 @@ export const TicketFileList: React.FC<ITicketFileListProps> = ({ files }) => {
         a.click()
     }
 
+    const handleFileDownload: (file: UploadFile) => void = async (file: UploadFile) => {
+        const forbiddenFileTypes = ['html', 'svg', 'txt']
+        const fileType = file.name.split('.').reverse()[0]
```
Comment (akarjakin) [inline apps/condo/pages/ticket/[id]/index.tsx:114]: such Regex would work better `/.+\.(svg|pdf|txt)$/gi`

apps/condo/pages/ticket/[id]/index.tsx
Comment (Alllex202) [reply apps/condo/pages/ticket/[id]/index.tsx]: Redesigned using Regex

apps/condo/pages/ticket/[id]/index.tsx
Comment (Alllex202) [reply apps/condo/pages/ticket/[id]/index.tsx]: It's probably better to use `/.+\.(svg|pdf|txt)$/gi` rather than `/.*\.(svg|pdf|txt)$/gi` (`*` instead of `+`). I think theoretically the filename can be empty

apps/condo/pages/ticket/[id]/index.tsx
```diff
@@ -106,6 +109,18 @@ export const TicketFileList: React.FC<ITicketFileListProps> = ({ files }) => {
         a.click()
     }
 
+    const handleFileDownload: (file: UploadFile) => void = async (file: UploadFile) => {
+        if (/.*\.(svg|pdf|txt)$/gi.test(file.name)) {
```
Comment (akarjakin) [inline apps/condo/pages/ticket/[id]/index.tsx:113]: :) we use to move static regexps (that are not changed) to constants in order not to compile it on every function call

Comment (Alllex202) [reply apps/condo/pages/ticket/[id]/index.tsx]: its done

Comment (Alllex202) [reply apps/condo/pages/ticket/[id]/index.tsx]: But I did not find similar regexes in constants :)

apps/condo/pages/ticket/[id]/index.tsx
```diff
@@ -106,6 +110,18 @@ export const TicketFileList: React.FC<ITicketFileListProps> = ({ files }) => {
         a.click()
     }
 
+    const handleFileDownload: (file: UploadFile) => void = async (file: UploadFile) => {
```
Comment (nomerdvadcatpyat) [inline apps/condo/pages/ticket/[id]/index.tsx:113]: useCallback

apps/condo/pages/ticket/[id]/index.tsx
```diff
@@ -96,7 +100,7 @@ export const TicketFileList: React.FC<ITicketFileListProps> = ({ files }) => {
         return fileInList
     })
 
-    const handleFileDownload: (file: UploadFile) => void = async (file: UploadFile) => {
+    const downloadFile = async (file: UploadFile) => {
```
Comment (nomerdvadcatpyat) [inline apps/condo/pages/ticket/[id]/index.tsx:103]: useCallback

apps/condo/pages/ticket/[id]/index.tsx
Comment (Alllex202) [reply apps/condo/pages/ticket/[id]/index.tsx]: and without flag `g`

nomerdvadcatpyat (approved):

PR #1909, opened 2022-07-12
DOMA-3525 🌎 Added global B2B apps 🌎

SavelevMatthew:


sitozzz:
You also need to check that this type is not be rendered in the general list

apps/condo/domains/miniapp/schema/B2BApp.js
```diff
@@ -48,6 +50,12 @@ const B2BApp = new GQLListSchema('B2BApp', {
         connectedMessage: CONNECTED_MESSAGE_FIELD,
         appUrl: IFRAME_URL_FIELD,
         isHidden: IS_HIDDEN_FIELD,
+        isGlobal: {
+            schemaDoc: 'Indicates whether the app is global or not. If so, then the application will be opened in hidden mode and receive various notifications from the condo. It\'s also possible to trigger some condo IFrame methods via global app outside of miniapps CRM section',
+            type: Checkbox,
+            defaultValue: false,
+            isRequired: true,
+        },
```
sitozzz, line 58: May be add hook for check if `isHidden` has `!isGlobal` value?

SavelevMatthew: `isHidden` should always works for global B2B app. Since we need to turn-off any broken apps in runtime

SavelevMatthew:
> You also need to check that this type is not be rendered in the general list

Check new test in `AllMiniApps` query

toplenboren (approved):

sitozzz (approved):

dkoviazin (approved):

PR #1923, opened 2022-07-15
[DOMA-3526] 🖼️🖼️🖼️  More Iframes?) 🖼️🖼️🖼️

SavelevMatthew:
Added global IFrames for global B2BApps.
Possibilities:
- Scan for user-launched actions
- Open web-pages inside modal
- Send notifications

PR #1925, opened 2022-07-16
[DOMA-3547] add global miniapp task progress handler

sitozzz:
This PR added mechanism to manage condo ui tasks for global miniapps

telegram-cloud-photo-size-2-5381945655611735208-y

### DIFF
--- a/apps/condo/domains/common/utils/iframe.utils.ts
+++ b/apps/condo/domains/common/utils/iframe.utils.ts
@@ -1,9 +1,11 @@
 import get from 'lodash/get'
+import { TASK_STATUS } from '@condo/domains/common/components/tasks'
 
 const Ajv = require('ajv')
 const ajv = new Ajv()
 
 // CONST DECLARATION BLOCK (for checking by external observer)
+export const TASK_MESSAGE_TYPE = 'task'
 export const NOTIFICATION_MESSAGE_TYPE = 'notification'
 export const REQUIREMENT_MESSAGE_TYPE = 'requirement'
 export const LOADED_STATUS_MESSAGE_TYPE = 'loading'
@@ -13,11 +15,21 @@ export const COMMAND_MESSAGE_TYPE = 'command'
 export const REDIRECT_MESSAGE_TYPE = 'redirect'
 export const IFRAME_MODAL_ACTION_MESSAGE_TYPE = 'modal'
 
+export type TaskOperationType = 'create' | 'update'
 export type NotificationType = 'info' | 'warning' | 'error' | 'success'
 export type RequirementType = 'auth' | 'organization'
 export type LoadingStatuses = 'done'
 
 // TYPES DECLARATION BLOCK
+type TaskMessageType = {
+    type: 'task',
+    id: string,
+    title: string,
+    description: string,
+    progress: number,
+    status: TASK_STATUS
+    operation: TaskOperationType
+}
 export type NotificationMessageType = {
     type: 'notification',
     notificationType: NotificationType,
@@ -72,6 +84,7 @@ export type CloseModalMessageType = {
 
 type SystemMessageType =
     RequirementMessageType
+    | TaskMessageType
     | NotificationMessageType
     | LoadedStatusMessageType
     | ErrorMessageType
@@ -97,6 +110,7 @@ type parseMessageType = (data: any) => ParsedMessageReturnType
 
 // CONFIGURATION BLOCK
 const AvailableMessageTypes = [
+    TASK_MESSAGE_TYPE,
     NOTIFICATION_MESSAGE_TYPE,
     REQUIREMENT_MESSAGE_TYPE,
     LOADED_STATUS_MESSAGE_TYPE,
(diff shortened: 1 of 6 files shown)


sitozzz:
part of changes done here https://github.com/open-condo-software/condo/pull/1923

SavelevMatthew (commented):
You should probably register task by sending its title, description, may be external id, then receiver internal ID...And with that ID you should update its progress (may be. description as well)

apps/condo/domains/common/utils/iframe.utils.ts
```diff
@@ -190,6 +212,22 @@ export const parseMessage: parseMessageType = (data) => {
     if (systemMessageDetector(data)) {
         if (!systemMessageValidator(data)) return null
         switch (data.type) {
+            case TASK_MESSAGE_TYPE:
+                console.log('parse message task type')
```
pahaz, line 216: debug?

sitozzz: Yup, fixed this

apps/condo/domains/miniapp/components/GlobalApps/GlobalAppsContainer.tsx
```diff
@@ -64,6 +75,37 @@ export const GlobalAppsContainer: React.FC = () => {
...
+        }
+
+        if (message.taskOperation === 'create') {
+            taskRecord.id = uuidV4()
+            const createMiniAppTask = miniAppTaskUIInterface.storage.useCreateTask({}, () => {
+                addTask({
+                    ...miniAppTaskUIInterface,
+                    record: taskRecord,
+                })
+            })
```
AntonAL, line 96: Could you uplift this code on component scope because it is in a hook style.

apps/condo/domains/miniapp/components/GlobalApps/GlobalAppsContainer.tsx
```diff
@@ -64,6 +75,37 @@ export const GlobalAppsContainer: React.FC = () => {
...
+
+            createMiniAppTask(taskRecord)
+
+        } else if (message.taskOperation === 'update') {
+            const updateMiniAppTask = miniAppTaskUIInterface.storage.useUpdateTask({}, () => {
+                updateTask(taskRecord)
+            })
```
AntonAL, line 103: Could you please uplift this code to component scope, because it is in hooks format, which are not allowed to be initialized in conditions. You can take `taskRecord` from argument of `onComplete`.

apps/condo/domains/miniapp/components/GlobalApps/GlobalAppsContainer.tsx
```diff
@@ -159,6 +208,23 @@ export const GlobalAppsContainer: React.FC = () => {
...
+    useEffect(() => {
+        if (!isEmpty(objs) && !isEmpty(tasks) && iframeLoadedCount === iframeRefs.current.length) {
+            for (const iframe of iframeRefs.current) {
+                const targetOrigin = extractOrigin(iframe.src)
+                const targetWindow = get(iframe, 'contentWindow', null)
+
+                if (targetWindow) {
+                    sendMessage({
+                        type: TASK_PROCESSING_STATUS,
+                        data: { tasks: tasks.map(task => task.record).filter(task => task.status === TASK_STATUS.PROCESSING) },
```
AntonAL, line 221: Will every iframe get all tasks including tasks, not belonging to it?

apps/condo/domains/common/utils/iframe.utils.ts
```diff
@@ -131,6 +147,12 @@ const SystemMessageSchema = {
     type: 'object',
     properties: {
         type: { enum: AvailableMessageTypes },
+        taskId: { type: 'string' },
+        taskTitle: { type: 'string' },
+        taskDescription: { type: 'string' },
+        taskProgress: { type: 'number' },
+        taskStatus: { enum: ['processing', 'completed', 'error'] },
+        taskOperation: { enum: ['create', 'update'] },
```
AntonAL, line 155: We have a mess of all properties here, belonging to different parts of the system. The prefix `task…` is a workaround of this complexity.
My original idea was to incapsulate all implementation related to tasks as maximally, as possible to make it clear, where tasks handling implementation is.
That's why I've created a component `MiniAppTaksPostMessageAdapter`
Could you follow this approach?
Is it possible to move all implementation of handling tasks via postMessage into `MiniAppTaksPostMessageAdapter`, mount it into `GlobalAppsComponent` and write a little glue-code?

apps/condo/domains/miniapp/components/GlobalApps/GlobalAppsContainer.tsx
sitozzz: Good point, working on it right now

sitozzz: Suppose I fixed this issue by last changes made

SavelevMatthew (approved):

apps/condo/domains/common/utils/iframe.utils.ts
SavelevMatthew: All postMessage handling should be in single place (actually 2 places, one for global apps and one for page-apps)

dkoviazin (approved):

apps/condo/domains/miniapp/components/GlobalApps/GlobalAppsContainer.tsx
```diff
@@ -64,6 +73,60 @@ export const GlobalAppsContainer: React.FC = () => {
...
+
+                const targetWindow = get(iframe, 'contentWindow', null)
+                if (targetWindow) {
+
+                    sendMessage({
+                        type: TASK_GET_PROCESSING_STATUS,
+                        data: { tasks: [taskRecord] },
+                    }, targetWindow, targetOrigin)
+                }
+            }
```
AntonAL, line 110: It seems like this code should be executed in `onComplete` handler of `useCreateTask`

PR #1944, opened 2022-07-20
DOMA-3501 fixed field source in tickets filter (only sources: mobile …

Alllex202:
…app and calls)


Alllex202:
Before

Снимок экрана 2022-07-20 в 16 22 08

After

Снимок экрана 2022-07-20 в 16 21 52

AleX83Xpert (approved):
Looks good!
But I think it's better to create enable/disable flag within `TicketSource` model. In this case, we won't have to update the code if we will need other sources.

nomerdvadcatpyat (approved):

PR #1970, opened 2022-07-25
Contact fields validation added on meter create page

2000Ilya:
Screenshot 2022-07-25 at 12 06 59
It was necessary to make sure that it was impossible to send the form without completed contacts.
Submit form button and Adding new meter button are inactive when contacts fields are empty.
New strings added to ErrorsContainer - "Phone" and "Full name".


VKislov (approved):

AntonAL (commented):
There is no final conclusion in task description about what to do.
It states:
1. either you can submit MeterReading without contacts
2. or an error should be thrown.

The second case was implemented, right?

akarjakin (approved):

PR #1990, opened 2022-07-27
fix/condo/doma-3697/roles-sorting

AleX83Xpert:


apps/condo/domains/contact/components/contactRoles/ContactRolesSettingsContent.tsx
```diff
@@ -46,11 +45,7 @@ export const ContactRolesSettingsContent = (props) => {
 
     const searchContactRolesQuery = useMemo(() => ({
         ...filtersToWhere(filters),
```
akarjakin, line 47: filters are not being changed, I think this destructuring here is redundant.

akarjakin (approved):

PR #2005, opened 2022-07-28
💄💄  [DOMA-3719]  Front-end reload after release  💄💄

SavelevMatthew:
https://user-images.githubusercontent.com/50069872/182110814-62287332-f850-4610-b39f-e18390f55d2c.mov

# How does this work?
The whole trick is powered by 2 hooks, first one to fetch version from server and another to compare results.
### Why useSWR?
It provides caching mechanism like Apollo Client, but Apollo Client can only work on single route (remember schema stitching). swr also have mechanism of validation (load cache value first, then send request, then update value if changed)
### How does usePrevious works?
useEffect and useState are async hooks, but ref is not. And that's the base foundation of reload trick. Then new value enters `usePrevious` we're starting async useEffect operation, but return always happens first. And that's why we have state where `previous !== current` which can trigger it's own side-effects. After useEffect is done, we have `previous === current` state.


apps/condo/lang/ru/ru.json
```diff
@@ -70,6 +70,8 @@
   "example.Email": "name@example.org",
   "example.Name": "Иванов Сергей",
   "example.Phone": "+7 (900) 800 22 22",
+  "HotReload.title": "Обновление системы!",
+  "HotReload.message": "Мы обновили систему, поэтому данная страница сейчас перезагрузится",
```
toplenboren, line 74:
```suggestion
  "HotReload.message": "Вышло обновление Doma.ai, страница сейчас будет обновлена",
```

toplenboren, line 74:
```suggestion
  "HotReload.message": "Мы обновили систему, эта страница сейчас будет перезагружена",
```

toplenboren:
Why not let user to reload page by himself? I might be in the middle of writing Ticket...

apps/condo/domains/common/hooks/useCurrentBuild.tsx
```diff
@@ -0,0 +1,21 @@
+import useSWR from 'swr'
```
toplenboren, line 1: Why not let user to reload page by himself? He might be in the middle of Ticket editing...

apps/condo/domains/common/hooks/usePrevious.tsx
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useRef } from 'react'
+
+/**
+ * Monitoring some variable value
+ * After it's change useEffect will update ref
+ * But before that will be returned previous value
```
toplenboren, line 6:
```suggestion
 * Monitor some `variable` value
 * Until this `variable` is changed it's previous value would be returned
 * When this `variable` is changed `ref` will be updated using `useEffect`
```

apps/condo/domains/common/hooks/useCurrentBuild.tsx
```diff
@@ -0,0 +1,21 @@
...
+import get from 'lodash/get'
+import getConfig from 'next/config'
+
+const {
+    publicRuntimeConfig: { serverUrl },
+} = getConfig()
+
+const REQUEST_KEY = '/api/version'
```
AntonAL, line 9:
```suggestion
const REQUEST_PATH = '/api/version'
```

apps/condo/domains/common/hooks/useCurrentBuild.tsx
```diff
@@ -0,0 +1,21 @@
...
+const API_ROUTE = `${serverUrl}${REQUEST_KEY}`
```
AntonAL, line 10:
```suggestion
const API_ROUTE = `${serverUrl}${REQUEST_PATH}`
```

apps/condo/domains/common/hooks/useHotUpdate.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { notification } from 'antd'
+import Router from 'next/router'
+import { useCurrentBuild } from './useCurrentBuild'
+import { usePrevious } from './usePrevious'
+import { useIntl } from 'react-intl'
+
+const DELAY_IN_SECONDS = 5
+
+
+export function useHotUpdate (): void {
```
toplenboren, line 11: It looks really clean, but i think a quick explanation of whats going on here and how it works would help

apps/condo/domains/common/hooks/useCurrentBuild.tsx
toplenboren: Also, why do we need to have swr now? We already have fetch as HTTP client...

apps/condo/lang/en/en.json
```diff
@@ -70,6 +70,8 @@
   "example.Email": "name@example.org",
   "example.Name": "Jon Johnson",
   "example.Phone": "+1 (300) 300 22 22",
+  "HotReload.title": "System update!",
+  "HotReload.message": "We have updated the system, so this page will now reload",
```
AntonAL, line 74:
```suggestion
  "HotReload.message": "We have updated the system, this page will now be reloaded",
```

AntonAL:
This technique is called `Hot Code Reloading` and it's widespread in evergreen frameworks.
Maybe to rename `useHotUpdate` to `useHotCodeReloading` to conform the commonly known technique?

AntonAL:
User can be in the process of filling a long form and because of automatic reloading, the manual work will be lost.
I've described this case in https://doma.atlassian.net/browse/DOMA-3084

apps/condo/domains/common/hooks/usePrevious.tsx
```diff
@@ -0,0 +1,17 @@
...
+ * @param value to monitor
+ */
+export function usePrevious<T> (value: T): T {
+    const ref = useRef<T>()
```
AntonAL, line 10: First execution of this function will return `undefined` result.
Maybe to initialize it first?

```suggestion
    const ref = useRef<T>(value)
```

apps/condo/domains/common/hooks/useCurrentBuild.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+async function fetchBuildInfo () {
+    const response = await fetch(API_ROUTE)
```
AleX83Xpert, line 13: According to documentation:
```suggestion
async function fetchBuildInfo (url) {
    const response = await fetch(url)
```
Yes, it looks strange: a key becomes an url.

apps/condo/domains/common/hooks/useHotUpdate.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+    useEffect(() => {
+        if (buildId && previousBuildId && previousBuildId !== buildId) {
+            notification.warning({
+                message: NotificationTitle,
+                description: NotificationMessage,
+                duration: DELAY_IN_SECONDS,
+                closeIcon: React.Fragment,
+            })
+            setTimeout(() => {
```
AleX83Xpert, line 26: What about a countdown?
5... 4... 3... 2... 1... reload

apps/condo/domains/common/hooks/useHotUpdate.tsx
```diff
@@ -0,0 +1,31 @@
...
+                Router.reload()
```
AleX83Xpert, line 27: Agree with @toplenboren. Maybe let the user reload the page by himself?
We may show something at the top of the page.

apps/condo/domains/common/hooks/useCurrentBuild.tsx
AleX83Xpert: According to the [documentation](https://www.npmjs.com/package/swr), this is actually a `key`

apps/condo/domains/common/hooks/usePrevious.tsx
AleX83Xpert, line 10: Why not `useState(value)`?

SavelevMatthew:
> User can be in the process of filling a long form and because of automatic reloading, the manual work will be lost.
> I've described this case in https://doma.atlassian.net/browse/DOMA-3084

Yes, but it's not about API refresh. The form itself can change entirely after update. That's why we're forcing to reload page.

apps/condo/domains/common/hooks/usePrevious.tsx
SavelevMatthew: useState unlike useRef changes it's value not instantly. That's why trick with useEffect will not work :)

The trick here is to push value to update in useEffect (async) and return old value, which is not changed by that time. So we can have state where `previous !== current`. After useEffect worked in async we ll have new state where `current === current`

apps/condo/domains/common/hooks/useCurrentBuild.tsx
SavelevMatthew: Yes, it's a key to store cached data.

apps/condo/domains/common/hooks/useCurrentBuild.tsx
SavelevMatthew: swr used for cache storing. I'll add "TRICK DESCRIPTION" section 😄

apps/condo/domains/common/hooks/usePrevious.tsx
SavelevMatthew: We don't want to do it here...Since previous of first value is nothing 😄

apps/condo/lang/ru/ru.json
SavelevMatthew: Will be removed after conversation with Max

SavelevMatthew: Will be removed after conversation with Max

apps/condo/lang/en/en.json
SavelevMatthew: Will be removed after conversation with Max

apps/condo/domains/common/hooks/useHotUpdate.tsx
SavelevMatthew: New solution will be used after conversation with Max

apps/condo/domains/common/hooks/useHotUpdate.tsx
SavelevMatthew: refactored

apps/condo/domains/common/hooks/usePrevious.tsx
SavelevMatthew: changed docstring

apps/condo/domains/common/hooks/useHotUpdate.tsx
SavelevMatthew: added

AleX83Xpert (approved):

toplenboren (approved):

PR #2007, opened 2022-07-28
DOMA-3099 added `meta` to `file` field value of `TicketExportTask` for `FileAdapter` in case of `SberCloudFileAdapter`

AntonAL:


AntonAL:
Wrong task number

PR #2021, opened 2022-08-01
DOMA-3741 added the ability to open a ticket from tickets table in a …

Alllex202:
…new tab when clicking on the ticket number with the right mouse button (behavior like a link

Before
Снимок экрана 2022-08-01 в 10 46 43
)

After
Снимок экрана 2022-08-01 в 10 47 07

### DIFF
--- a/apps/condo/domains/common/components/Table/Renders.tsx
+++ b/apps/condo/domains/common/components/Table/Renders.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react'
+import React, { useEffect, useMemo } from 'react'
 import { Typography, Tooltip } from 'antd'
 import { TextProps } from 'antd/es/typography/Text'
 import dayjs from 'dayjs'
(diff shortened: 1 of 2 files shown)


apps/condo/domains/common/components/Table/Renders.tsx
```diff
@@ -92,11 +93,22 @@ export const getHighlightedContents: TGetHighlightedFN = (search, postfix, extra
     )
 }
 
+const withLink = (content: JSX.Element | string, href: string) => {
```
nomerdvadcatpyat, line 96: maybe `renderLink`?

nomerdvadcatpyat (approved):

dkoviazin (approved):

apps/condo/domains/common/components/Table/Renders.tsx
Alllex202: renamed it

PR #2034, opened 2022-08-03
Fix/doma 3099/duplicated tasks after update task while using localstorage

AntonAL:
@sitozzz, take a look please

### DIFF
--- a/apps/miniapp/pages/index.tsx
+++ b/apps/miniapp/pages/index.tsx
@@ -13,17 +13,11 @@ import { useOrganization } from '@miniapp/domains/common/utils/organization'
 
 const { publicRuntimeConfig: { condoUrl } } = getConfig()
 
-const IndexPage = () => {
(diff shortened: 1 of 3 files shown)


apps/miniapp/pages/index.tsx
```diff
@@ -13,17 +13,11 @@ import { useOrganization } from '@miniapp/domains/common/utils/organization'
 
 const { publicRuntimeConfig: { condoUrl } } = getConfig()
 
-const IndexPage = () => {
+const TasksExample = () => {
```
sitozzz, line 16: I don't think that it's part of this PR

sitozzz:
> @sitozzz, take a look please

Tested locally, now it works like it should

apps/miniapp/pages/index.tsx
AntonAL: It's a casual refactoring

PR #2042, opened 2022-08-03
Fix/condo/doma 3099/added filtering processing tasks on frontend

AntonAL:
Filtering on backend is not enough because for User with `isAdmin` flag we need to:
- return all tasks in Keystone admin section
- return only his own tasks in "public" section

## Review note
First commit is a refactoring.
The objective of this PR is in d81d9c1be7044f3cea8a0e91f7a409d5c76be160 commit

### DIFF
--- a/apps/condo/domains/common/components/tasks/index.tsx
+++ b/apps/condo/domains/common/components/tasks/index.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import { User } from '../../../../schema'
 
 // Should be used in case when there is no technical way to tell exactly the progress of the task
 export const TASK_PROGRESS_UNKNOWN = 'TASK_PROGRESS_UNKNOWN'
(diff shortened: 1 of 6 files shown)


apps/condo/domains/common/components/tasks/index.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react'
+import { User } from '../../../../schema'
```
sitozzz, line 2: absolute import please

apps/condo/domains/common/components/tasks/storage/TasksLocalStorage.tsx
```diff
@@ -23,16 +23,19 @@ const isServerSide = typeof window === 'undefined'
...
         }
+        if (!user) {
+            return { records: [] }
+        }
         const tasks = this.getAllItemsFromStorage()
         if (!status) {
             return tasks
         }
-        const records = tasks.filter(task => task.status === status)
+        const records = tasks.filter(task => task.status === status && task.user && task.user.id === user.id)
```
sitozzz, line 38:
```suggestion
        const records = tasks.filter(task => task.status === status && get(task, 'user.id') === user.id)
```

sitozzz (commented):
Need to test it locally again with miniapp flow

apps/condo/domains/common/components/tasks/index.tsx
AleX83Xpert:
```suggestion
import { User } from '@app/condo/schema'
```

AleX83Xpert (approved):

sitozzz (changes requested):
There is some issues with auth. You need to handle this cases

Screenshot 2022-08-04 at 12 10 14

Screenshot 2022-08-04 at 11 22 19

AntonAL:
Resolved the issues "Cannot read properties" and other.
They are related to wrong hooks usage.

sitozzz (approved):
Now it's ok

PR #2046, opened 2022-08-04
DOMA-3722 increased sender max length to 128 chars

akarjakin:
Current max length of 42 chars is too less for some cases.


sitozzz (approved):

akarjakin:
Decided to cut off extra data from sender

PR #2064, opened 2022-08-08
DOMA-3823 notification black list

nomerdvadcatpyat:
We need to limit the sending of various types of messages by phone number, user, email, organization or all organizations.

Added 3 models:
1) MessageUserBlackList - to prohibit sending messages by type for a specific user
2) MessageOrganizationBlackList - to prohibit sending messages by type for a specific organization or for all organizations if the organization field is not filled in
3) MessageOrganizationWhiteList - If a rule is set in the MessageOrganizationBlackList for all organizations, then adding a rule to the MessageOrganizationWhiteList will allow to send messages by type for a specific organization


apps/condo/domains/notification/schema/BlackList.js
```diff
@@ -0,0 +1,68 @@
...
+ */
+
+const { Text, Relationship, Select } = require('@keystonejs/fields')
+const { GQLListSchema } = require('@core/keystone/schema')
+const { historical, versioned, uuided, tracked, softDeleted } = require('@core/keystone/plugins')
+const { dvAndSender } = require('@condo/domains/common/schema/plugins/dvAndSender')
+const access = require('@condo/domains/notification/access/BlackList')
+const { MESSAGE_TYPES } = require('@condo/domains/notification/constants/constants')
+
+const BlackList = new GQLListSchema('BlackList', {
```
pahaz, line 12: It’s better to add some xtra cord here. It’s context less. When you see the name blackList it’s hard to understand that it’s about messages. My be something like “send message black list” or “message black list”

apps/condo/domains/notification/schema/BlackList.js
```diff
@@ -0,0 +1,68 @@
...
+            schemaDoc: 'The type of message we want to block (null means all types)',
+            type: Select,
+            options: MESSAGE_TYPES,
+            dataType: 'string',
+            isRequired: false,
+            knexOptions: { isNotNullable: false },
+            kmigratorOptions: { null: true },
+        },
+
+        description: {
```
pahaz, line 26: It should be required

apps/condo/domains/notification/schema/BlackList.js
```diff
@@ -0,0 +1,68 @@
...
+            type: Text,
+        },
+
+        email: {
+            schemaDoc: 'Email to which we want to block the sending of messages',
+            type: Text,
+        },
+
+    },
+    plugins: [uuided(), versioned(), tracked(), softDeleted(), dvAndSender(), historical()],
```
pahaz, line 56: It’s better to create some uniq constraints for the model

apps/condo/domains/notification/tasks.js
```diff
@@ -100,6 +100,27 @@ async function deliverMessage (messageId) {
...
+    }
+
+    const blackListRules = await BlackList.getAll(keystone, blackListWhere)
+    if (!isEmpty(blackListRules)) {
+        const isMessageTypeInBlackList = blackListRules.find(rule => rule.type === message.type || isEmpty(rule.type))
+
+        if (isMessageTypeInBlackList) {
+            return await Message.update(keystone, message.id, {
+                ...baseAttrs,
+                status: MESSAGE_ERROR_STATUS,
```
pahaz, line 119: We need some way to understand why the message it not sent

apps/condo/domains/notification/schema/BlackList.js
nomerdvadcatpyat: renamed to MessageBlackList

apps/condo/domains/notification/schema/MessageBlackList.test.js
```diff
@@ -0,0 +1,165 @@
...
+const { makeClientWithNewRegisteredAndLoggedInUser, makeClientWithSupportUser, createTestEmail, createTestPhone } = require('@condo/domains/user/utils/testSchema')
+const { MessageBlackList, createTestMessageBlackList, updateTestMessageBlackList, Message } = require('@condo/domains/notification/utils/testSchema')
+const { createTestBillingIntegrationLog } = require('@condo/domains/billing/utils/testSchema')
+const { makeClientWithRegisteredOrganization, inviteNewOrganizationEmployee, reInviteNewOrganizationEmployee } = require('@condo/domains/organization/utils/testSchema/Organization')
+const { DIRTY_INVITE_NEW_EMPLOYEE_MESSAGE_TYPE, MESSAGE_SENT_STATUS, EMAIL_TRANSPORT, MESSAGE_ERROR_STATUS } = require('@condo/domains/notification/constants/constants')
+
+describe('MessageBlackList', () => {
+    describe('accesses', () => {
+        describe('create', () => {
+            it('support can create MessageBlackList', async () => {
```
AntonAL, line 22: Optional: Make it readable in natural language format, like
"test support can create…"
You have now:
"it support can create"

```suggestion
            test('support can create MessageBlackList', async () => {
```

An so on for others.

apps/condo/domains/notification/tasks.js
```diff
@@ -100,6 +100,27 @@ async function deliverMessage (messageId) {
...
+    }
+
+    const blackListRules = await MessageBlackList.getAll(keystone, blackListWhere)
+    if (!isEmpty(blackListRules)) {
+        const isMessageTypeInBlackList = blackListRules.find(rule => rule.type === message.type || isEmpty(rule.type))
+
+        if (isMessageTypeInBlackList) {
+            return await Message.update(keystone, message.id, {
+                ...baseAttrs,
+                status: MESSAGE_ERROR_STATUS,
```
AntonAL, line 119: Is there a way to tell the reason of this error as presence in a black list?

AntonAL (approved):

apps/condo/domains/notification/tasks.js
nomerdvadcatpyat: added error message in processingMeta

apps/condo/domains/notification/tasks.js
nomerdvadcatpyat: added error message in processingMeta of Message

apps/condo/domains/notification/schema/MessageOrganizationBlackList.test.js
```diff
@@ -0,0 +1,186 @@
...
+        it('dont send notification if all organizaitons added in MessageOrganizationBlackList', async () => {
+            const admin = await makeLoggedInAdminClient()
+            const userAttrs = {
+                name: faker.name.firstName(),
+                email: createTestEmail(),
+                phone: createTestPhone(),
+            }
+            const client = await makeClientWithRegisteredOrganization()
+
+            await createTestMessageOrganizationBlackList(admin, {
```
akarjakin, line 152: don't you need to undo this after test? this would affect all organizations forever in other case.

akarjakin: what if some day we want to block some kind of message to be sent to all organizations, and then, after some time, we would want to allow it for some of them? we would have to delete this rule and add it for all organizations except  allowed ones. maybe there should be whitelist too in order to make these more flexible?

apps/condo/domains/notification/schema/MessageUserBlackList.js
```diff
@@ -0,0 +1,82 @@
...
+                condition: 'Q(deletedAt__isnull=True)',
+                name: 'message_user_black_list_unique_phone_and_type',
+            },
+            {
+                type: 'models.UniqueConstraint',
+                fields: ['email', 'type'],
+                condition: 'Q(deletedAt__isnull=True)',
+                name: 'message_user_black_list_unique_email_and_type',
+            },
+        ],
```
akarjakin, line 68: Please add either constraint or logic in resolver that user.id, email and phone could not be null all at the same time, one of them should have a value.

apps/condo/domains/notification/tasks.js
```diff
@@ -100,6 +101,53 @@ async function deliverMessage (messageId) {
         sender: message.sender,
     }
 
+    if (message.organization) {
+        const messageOrganizationBlackListWhere = {
```
akarjakin, line 105: looks like this logic can be moved to helper

apps/condo/domains/notification/tasks.js
```diff
@@ -100,6 +101,53 @@ async function deliverMessage (messageId) {
...
+            message.user && { user: { id: message.user.id } },
+            message.phone && { phone: message.phone },
+            message.email && { email: message.email },
+        ].filter(Boolean),
+        deletedAt: null,
+    }
+    const messageUserBlackListRules = await MessageUserBlackList.getAll(keystone, messageUserBlackListWhere)
+
+    if (!isEmpty(messageUserBlackListRules)) {
+        const isMessageTypeInUserBlackList = messageUserBlackListRules.find(rule => rule.type === message.type || isEmpty(rule.type))
```
akarjakin, line 137: looks like this logic can be moved to helper

sitozzz (approved):

apps/condo/domains/notification/tasks.js
nomerdvadcatpyat: moved to `checkMessageTypeInBlackList` helper

apps/condo/domains/notification/tasks.js
nomerdvadcatpyat: moved to `checkMessageTypeInBlackList` helper

apps/condo/domains/notification/schema/MessageUserBlackList.js
nomerdvadcatpyat: added

apps/condo/domains/notification/schema/MessageOrganizationBlackList.test.js
nomerdvadcatpyat: > don't you need to undo this after test? this would affect all organizations forever in other case.

deleted it in `beforeEach`

nomerdvadcatpyat: > what if some day we want to block some kind of message to be sent to all organizations, and then, after some time, we would want to allow it for some of them? we would have to delete this rule and add it for all organizations except allowed ones. maybe there should be whitelist too in order to make these more flexible?

Added `MessageOrganizationWhiteList` for this case. The logic of the work is as follows: if a rule with empty organizations is set in the `MessageOrganizationBlackList` (block messages by type for all organizations), then adding a rule to the `MessageOrganizationWhiteList` will allow to send messages by type for a specific organization

PR #2077, opened 2022-08-10
Hotfix request level cache

toplenboren:
This is a work in progress try to catch a wandering bug with `cache.js` module. Bug is caused (well, we think it caused) by race-condition.

### DIFF
--- a/apps/condo/domains/user/schema/User.test.js
+++ b/apps/condo/domains/user/schema/User.test.js
@@ -417,24 +417,6 @@ describe('UserHistoryRecord', () => {
 })
 
 describe('Cache tests', () => {
-    test('Clients that ask for different set of fields get different set of fields', async () => {
-        const client = await makeClientWithNewRegisteredAndLoggedInUser()
-        
-        const userId = client.user.id
-        
-        const UserModifiedGQL = generateGqlQueries('User', '{ id name }')
-        const UserModifiedGQLAPI = generateGQLTestUtils(UserModifiedGQL)
-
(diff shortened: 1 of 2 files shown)


packages/keystone/cache.js
```diff
@@ -97,7 +99,7 @@ const patchQuery = (queryContext, query, cache) => {
         }
         const listResult = await query.call(queryContext, args, context, gqlName, info, from)
 
-        if (requestId && key) { cache[requestId][key] = listResult }
+        if (requestId && key && get(cache, [requestId, key])) { cache[requestId][key] = listResult }
```
pahaz, line 102: probably it's easy to use lodash set?

apps/condo/domains/user/schema/User.test.js
```diff
@@ -415,3 +415,72 @@ describe('UserHistoryRecord', () => {
         ])
     })
 })
+
+describe('Cache tests', () => {
+    test('Clients that ask for different set of fields get different set of fields', async () => {
```
pahaz, line 420: it looks ledundant

packages/keystone/cache.js
toplenboren: Done!

apps/condo/domains/user/schema/User.test.js
toplenboren: Do we need to remove it?

pahaz (approved):

pahaz (approved):

PR #2078, opened 2022-08-11
DOMA-3884 linter rules improved

AleX83Xpert:
We've decided to use single quotes everywhere. This PR includes linter config improvements and all code fixed by linter with new rules applied.


package.json
```diff
@@ -136,6 +137,7 @@
       "react/prop-types": "off",
       "react/no-children-prop": "off",
       "react/display-name": "warn",
+      "react/jsx-curly-brace-presence": "error",
```
sitozzz, line 140: I think there would be better to add more complex rule 

```suggestion
      "react/jsx-curly-brace-presence": ["error", { "props": "never", "children": "never" }],
```

sitozzz:
With `error` level you should run 
```bash
eslint . --ext .js,.jsx,.ts,.tsx --fix
```

package.json
```diff
@@ -75,6 +75,7 @@
         }
       ],
       "quotes": "off",
+      "jsx-quotes": ["warn", "prefer-double"],
```
pahaz, line 78: prefer-single

AleX83Xpert: Single is 
image

AleX83Xpert:
> With `error` level you should run
> 
> ```shell
> eslint . --ext .js,.jsx,.ts,.tsx --fix
> ```

Yes. The "warn" level was set.

package.json
AleX83Xpert: Done

sitozzz:
> > With `error` level you should run
> > ```shell
> > eslint . --ext .js,.jsx,.ts,.tsx --fix
> > ```
> 
> Yes. The "warn" level was set.

I think it should be `error` level here. Otherwise no one will pay attention to it

image

AleX83Xpert:
> > > With `error` level you should run
> > > ```shell
> > > eslint . --ext .js,.jsx,.ts,.tsx --fix
> > > ```
> > 
> > 
> > Yes. The "warn" level was set.
> 
> I think it should be `error` level here. Otherwise no one will pay attention to it
> 
> image

ok, I'll fix warnings after decision is made

pahaz (approved):

PR #2080, opened 2022-08-12
DOMA-3775 changed all tooltips (black --> white)

Alllex202:


apps/condo/domains/meter/utils/clientSchema/Renders.tsx
```diff
@@ -2,7 +2,10 @@ import { FilterValue } from 'antd/es/table/interface'
 import { TextProps } from 'antd/es/typography/Text'
 import get from 'lodash/get'
 
-import { getTableCellRenderer } from '@condo/domains/common/components/Table/Renders'
+import { getTableCellRenderer, RenderReturnType } from '@condo/domains/common/components/Table/Renders'
+import { METER_READING_SOURCE_EXTERNAL_IMPORT_TYPE } from '../../constants/constants'
```
AleX83Xpert, line 6:
```suggestion
import { METER_READING_SOURCE_EXTERNAL_IMPORT_TYPE } from '@condo/domains/meter/constants/constants'
```

AleX83Xpert (approved):

apps/condo/domains/meter/utils/clientSchema/Renders.tsx
```diff
@@ -22,3 +25,20 @@ export const getUnitRender = (intl, search: FilterValue) => {
         return getTableCellRenderer(search, true, null, null, POSTFIX_PROPS, extraTitle)(unitName)
     }
 }
+
+export const getResourceRender = (intl, search?: FilterValue | string) => {
+    return function render (text, meterReading): RenderReturnType {
+        const value = get(meterReading, ['meter', 'resource', 'name'])
+        const isAutomatic = get(meterReading, ['meter', 'isAutomatic'], false)
+        const isExternalSource = Boolean(get(meterReading, ['source', 'type']) === METER_READING_SOURCE_EXTERNAL_IMPORT_TYPE)
+        const AutoMessage = intl.formatMessage({ id: 'pages.condo.meter.AutoPrefix' })
```
sitozzz, line 34: Translations should be at the top of function body (this is common rule)

sitozzz, line 34: May this line be moved to outer closure? Looks like it is constant value. So it may be saved to parent function context

AntonAL (approved):

apps/condo/domains/meter/utils/clientSchema/Renders.tsx
Alllex202: `intl` is passed to the outer function. All similar functions are implemented in this way. It will be strange if the implementation of this function differs from similar.

Thus, we can change the implementation of the function from one place

apps/condo/domains/common/components/containers/GlobalStyle.tsx
```diff
@@ -185,6 +185,16 @@ export default function GlobalStyle () {
                 border-radius: 2px;
               }
 
+              .ant-tooltip > .ant-tooltip-content {
```
pahaz, line 188: are you sure that there is no way to avoid adding new global?

pahaz:
Any screen?

Alllex202:
> Any screen?



Before
Снимок экрана 2022-08-17 в 12 08 22

After
Снимок экрана 2022-08-17 в 12 08 02

PR #2083, opened 2022-08-12
DOMA-3825 display only today tasks

AntonAL:
Prevents stucking with tasks, not completed for some reason, created in previous days

Open question about pivot point today/tomorrow, that arises an answer — why not to display tasks only in 24 hours.
From working schedule point of view it's better to relate on calendar rather then durations, so, an employee opens a page on the beginning of the day and he dont see stuck tasks from yesterday.

@sitozzz, take a look please, I've added `createdAt` field to `TaskRecord` type

### DIFF
--- a/apps/condo/domains/common/components/tasks/index.tsx
+++ b/apps/condo/domains/common/components/tasks/index.tsx
@@ -45,6 +45,8 @@ export type TaskRecord = {
     // For example: local id at miniapp database whom would be used to update progress
     taskId?: string
     sender?: string
+
+    createdAt: string
 }
 
 /**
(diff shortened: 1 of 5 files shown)


sitozzz (approved):

apps/condo/domains/common/components/tasks/index.tsx
```diff
@@ -71,6 +73,8 @@ type UseDeleteTaskFunction = (attrs: unknown, onComplete: OnCompleteFunc) => (at
 
 type TasksWhereCondition = {
     status: TASK_STATUS,
+    // Helps to avoid displaying tasks, not completed for some reason, created on previous days
+    today?: boolean,
```
sitozzz, line 77:
```suggestion
    isToday?: boolean,
```

toplenboren (approved):

PR #2133, opened 2022-08-19
DOMA-3964 added parking `sectionType` and `sectionName` displaying on…

VKislov:
… ticket page
resident ticket before: 
Screenshot 2022-08-19 at 20 15 17

resident ticket after: 
Screenshot 2022-08-19 at 20 14 08

### DIFF
--- a/apps/condo/domains/organization/utils/helpers.ts
+++ b/apps/condo/domains/organization/utils/helpers.ts
@@ -80,12 +80,12 @@ export const emailToQuery = (email?: string) => {
     }
 }
 
-export const searchToQuery = (search?: string, translates = null): OrganizationEmployeeWhereInput[] => {
+export const searchToQuery = (search?: string, translations = null): OrganizationEmployeeWhereInput[] => {
     if (!search) {
         return
     }
 
-    const roleQuery = roleToSearchQuery(search, translates)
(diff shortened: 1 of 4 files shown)


sitozzz (approved):

packages/keystone/schema.spec.js
```diff
@@ -80,7 +80,7 @@ test('registerSchema without preprocessors', () => {
             User: new GQLListSchema('User', {
                 fields: {
                     name: {
-                        type: Text,
+                        type: 'Text',
```
pahaz, line 83: it look like not a part of this PR

VKislov: For some unknown reason, two tests in schema.spec.js were failing in CI. When running them locally, there was an error Text is not defined. And after the error in expect.toMatchObject

nomerdvadcatpyat: > For some unknown reason, two tests in schema.spec.js were failing in CI. When running them locally, there was an error Text is not defined. And after the error in expect.toMatchObject

it's fixed in https://github.com/open-condo-software/condo/pull/2136
i think you need to remove this changes

nomerdvadcatpyat (approved):

PR #2138, opened 2022-08-22
Feat/condo/doma 3191/feat dv and sender global preprocessors

VKislov:
The global preprocessor dvAndSenderPreprocessor has been added, to which the logic of the dvAndSender plugin has been moved. Changed schema registration logic in keystone, now plugins are applied after schema preprocessors. The dvAndSender plugin has been removed from all condo schemes

### DIFF
--- a/packages/keystone/preprocessors/utils/dvAndSender.helper.js
+++ b/packages/keystone/preprocessors/utils/dvAndSender.helper.js
@@ -1,0 +1,123 @@
+const { isEmpty } = require('lodash')
+const validate = require('validate.js')
+
+const { GQLError, GQLErrorCode: { BAD_USER_INPUT } } = require('@condo/keystone/errors')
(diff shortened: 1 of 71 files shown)


packages/keystone/preprocessors/utils/dvAndSender.helper.js
```diff
@@ -0,0 +1,123 @@
+const { isEmpty } = require('lodash')
+const validate = require('validate.js')
```
SavelevMatthew, line 2: All used imports MUST be specified inside child package.json. That's `condo/keystone` in your case. There's no `validate.js` in that package. This behaviour can cause conflicts in the future...like it was with `condo/ui `module

SavelevMatthew (changes requested):

VKislov:
it doesn't seem to be needed anymore. Maybe someday we'll get back to it.

PR #2142, opened 2022-08-22
DOMA-3978 fix empty contact/property Page

VKislov:
before: 
Screenshot 2022-08-22 at 18 46 53

after: 
Screenshot 2022-08-22 at 17 49 12


pahaz:
image
You can't use random generated class names as the source of true.

sitozzz (commented):
Why not just to make button prop optional at this component?

sitozzz:
> image
> 
> You can't use random generated class names as the source of true.

Also this doesn't work, find another solution

VKislov:
> Why not just to make button prop optional at this component?

Because the entire property page is imported from the condo. Adding a prop to a nested component will require dragging in unnecessary and non-obvious logic for condos

sitozzz (approved):

nomerdvadcatpyat:
pls do not merge submodule branches into master/main until you have passed the review in condo

nomerdvadcatpyat (approved):

PR #2143, opened 2022-08-22
Fix/condo/doma 3800/error by executing export ticket task in batch mode

AntonAL:


apps/condo/domains/common/utils/serverSchema/index.js
```diff
@@ -78,12 +78,6 @@ class GqlWithKnexLoadList {
             knexQuery.leftJoin(...join(idx))
         })
         knexQuery.whereIn('mainModel.id', ids)
-        if (offset) {
-            knexQuery.offset(offset)
-        }
-        if (limit) {
```
sitozzz, line 84: Why limit was removed?

sitozzz (approved):

apps/condo/domains/common/utils/serverSchema/index.js
AntonAL: Offset and limit was set for related tables, in case of loading non-first chunk caused to return no data for them.

PR #2150, opened 2022-08-23
DOMA-3770 added redirect to ticket/{id} from ticket create page.

VKislov:
After canceling the changes made, there is a redirect to the page of the edited ticket.

### DIFF
--- a/apps/condo/domains/ticket/components/TicketForm/UpdateTicketForm.tsx
+++ b/apps/condo/domains/ticket/components/TicketForm/UpdateTicketForm.tsx
@@ -1,6 +1,6 @@
 import { useRouter } from 'next/router'
 import { get, isEmpty } from 'lodash'
-import React, { useEffect, useMemo } from 'react'
+import React, { useCallback, useEffect, useMemo } from 'react'
 import { Form, Typography, Space } from 'antd'
 
 import { useOrganization } from '@condo/next/organization'
@@ -11,7 +11,6 @@ import { Ticket, TicketFile } from '@condo/domains/ticket/utils/clientSchema'
 import ActionBar from '@condo/domains/common/components/ActionBar'
 import { Loader } from '@condo/domains/common/components/Loader'
 import { REQUIRED_TICKET_FIELDS, TICKET_SOURCE_TYPES } from '@condo/domains/ticket/constants/common'
-import { FormResetButton } from '@condo/domains/common/components/FormResetButton'
 
 import { BaseTicketForm } from '../BaseTicketForm'
 import { ErrorsContainer } from '../BaseTicketForm/ErrorsContainer'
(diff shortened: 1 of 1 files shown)


apps/condo/domains/common/components/FormResetButton.tsx
```diff
@@ -1,19 +1,26 @@
...
 import { Button, CustomButtonProps } from './Button'
 
 export const FormResetButton: React.FC<CustomButtonProps> = (props) => {
     const intl = useIntl()
     const CancelMessage = intl.formatMessage({ id: 'Cancel' })
 
+    const { push, query: { id } } = useRouter()
+    const onCancel = useCallback((resetFields) => {
+        resetFields()
+        push(`/ticket/${id}`)
```
nomerdvadcatpyat, line 14: it's a component from common domain and it's used in many other forms

nomerdvadcatpyat (changes requested):

apps/condo/domains/common/components/FormResetButton.tsx
VKislov: oh you are right

apps/condo/domains/ticket/components/TicketForm/UpdateTicketForm.tsx
```diff
@@ -11,28 +11,39 @@ import { Ticket, TicketFile } from '@condo/domains/ticket/utils/clientSchema'
...
 
     return (
         <Form.Item noStyle shouldUpdate>
             {
-                ({ getFieldsValue }) => {
+                ({ getFieldsValue, resetFields, isFieldsTouched }) => {
                     const { property, details, placeClassifier, categoryClassifier, deadline } = getFieldsValue(REQUIRED_TICKET_FIELDS)
                     const disabledCondition = !property || !details || !placeClassifier || !categoryClassifier || !deadline
-
+                    const isNotTouched = !isFieldsTouched()
```
nomerdvadcatpyat, line 35: i think it's redundant

nomerdvadcatpyat (approved):

apps/condo/domains/ticket/components/TicketForm/UpdateTicketForm.tsx
```diff
@@ -11,28 +11,39 @@ import { Ticket, TicketFile } from '@condo/domains/ticket/utils/clientSchema'
...
                     return (
                         <ActionBar isFormActionBar>
-                            <FormResetButton
+                            <Button
+                                key='cancel'
+                                onClick={() => onCancel(resetFields)}
```
sitozzz, line 40: Why not to store `resetFields` with closure pattern? It would be look like this
```suggestion
                                onClick={onCancel(resetFields)}
```

sitozzz (approved):

apps/condo/domains/ticket/components/TicketForm/UpdateTicketForm.tsx
VKislov: I kept the logic. Previously, if there were no changes in the form, then the button was disabled and you could only exit by pressing back. Do you think it's worth giving the ability to redirect the user back by clicking the cancel button?

apps/condo/domains/ticket/components/TicketForm/UpdateTicketForm.tsx
VKislov: Screenshot 2022-08-24 at 13 44 07

PR #2165, opened 2022-08-25
Fix/condo/doma 4011/added alert canceling edit ticket

VKislov:
Removed the `resetField` and now, after `route.push`, in the case when something is changed, a `Prompt` appears that allows you to either cancel everything and leave, or save the changes.
And fix `Prompt` message


apps/condo/domains/ticket/components/TicketForm/UpdateTicketForm.tsx
```diff
@@ -36,7 +35,7 @@ export const ApplyChangesActionBar = ({ handleSave, isLoading }) => {
                         <ActionBar isFormActionBar>
```
2000Ilya, line 35: unused `resetFields` var

2000Ilya (approved):

SavelevMatthew (approved):

PR #2193, opened 2022-08-30
build: added figma-tokens parse pipeline

SavelevMatthew:


packages/ui/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "repository": "https://github.com/open-condo-software/condo/blob/master/packages/ui",
+  "author": {
+    "name": "Matthew Savelev",
+    "email": "msavelyev@doma.ai",
+    "url": "https://github.com/SavelevMatthew"
+  },
+  "license": "MIT",
+  "scripts": {
+    "prepare:tokens": "rm -rf src/tokens/sets && node bin/prepare-tokens.js",
+    "build:tokens": "find src/tokens -name 'variables.*' -delete && node bin/build-vars.js",
```
pahaz, line 14: why do you need to delete `bin/build-vars.js`?

pahaz, line 14: it looks like a part of `prepare:tokens` command

packages/ui/sd.config.js
```diff
@@ -0,0 +1,24 @@
+const path = require('path')
+
+module.exports = {
+    source: [path.join(__dirname, 'src/tokens/sets/*.json')],
+    platforms: {
```
pahaz, line 5: What is the `sd.config` file?

packages/ui/package.json
SavelevMatthew: Sometimes we don't want to generate tokens all the times. But maybe you're right...

packages/ui/sd.config.js
SavelevMatthew: config for `style-dictionary`

packages/ui/bin/prepare-tokens.js
```diff
@@ -0,0 +1,62 @@
...
+                }, null, 2))
+                console.log(`✓ Set ${setName} is processed`)
+            } else {
+                throw Error(`✖ Error occurred while processing ${setName} set`)
+            }
+        })
+    }
+}
+
+prepareTokens(...process.argv.slice(2))
```
sitozzz, line 61: Check for args -> throw readable error

sitozzz (approved):

packages/ui/bin/prepare-tokens.js
SavelevMatthew: There's only 1 arg...and it's not necessary...So I guess it will be redundant check

PR #2200, opened 2022-08-31
DOMA-4067 added delete contact duplicates script

nomerdvadcatpyat:
Required before setting the unique constraint `property + unitName + unitType + phone` for `Contact` model.

Script finds duplicates (property + unitName + unitType + phone) of contacts and:
1. If found contacts do not have tickets, then deletes all except the newest one
2. If found contacts have tickets, then deletes all duplicates except the newest contact with tickets and relinks tickets to this contact


AntonAL (commented):
It's hard to make sure it works properly
My suggestion would be to implement it as a function and place it into utils, write tests for it and in this script just call it.

dkoviazin (approved):

pahaz (approved):

PR #2224, opened 2022-09-02
DOMA-4066 update contacts phone

nomerdvadcatpyat:
There are a large number of contacts that have the same number (for example, +79999999999). This number is filled in when the contact number is unknown and only because the contact's phone field is required.
It is necessary to bring all such existing numbers to the same form


dkoviazin (commented):
Strange logic. Maybe, we can make this field not required? 
And, please, remove EPS update from PR

PR #2234, opened 2022-09-05
DOMA-3964 fix parking label for resident ticket

VKislov:
In `property->map->parking` `sectionType` is stored as `'section'` . When creating a ticket, the `sectionType` field is filled in as `parking`. Therefore, when using `res.sectionType = section.type` , the value will always be `'section'` and will be displayed as a `section`. This logic is only called when creating a ticket as a resident
linked pr: https://github.com/open-condo-software/condo/pull/2133
before: 
Screenshot 2022-09-05 at 20 42 26
After:
Screenshot 2022-09-05 at 20 43 14

### DIFF
--- a/apps/condo/domains/ticket/utils/unit.js
+++ b/apps/condo/domains/ticket/utils/unit.js
@@ -1,4 +1,5 @@
 const { get } = require('lodash')
+const { SECTION_SECTION_TYPE, PARKING_UNIT_TYPE } = require('@condo/domains/property/constants/common')
 
 function getSectionAndFloorByUnitName (property, unitName, unitType) {
     const res = {
@@ -7,13 +8,13 @@ function getSectionAndFloorByUnitName (property, unitName, unitType) {
         floorName: null,
     }
     if (unitName && unitType) {
-        const sections = get(property, ['map', `${unitType !== 'parking' ? 'sections' : 'parking'}`], [])
+        const sections = get(property, ['map', `${unitType !== PARKING_UNIT_TYPE ? SECTION_SECTION_TYPE : PARKING_UNIT_TYPE}`], [])
         for (const section of sections) {
             for (const floor of section.floors) {
                 for (const unit of floor.units) {
                     if (unit.label === unitName) {
                         res.sectionName = section.name
-                        res.sectionType = section.type
+                        res.sectionType = unitType !== PARKING_UNIT_TYPE ? 'sections' : 'parking'
                         res.floorName = floor.name
                         break
                     }


apps/condo/domains/ticket/utils/unit.js
```diff
@@ -13,7 +13,7 @@ function getSectionAndFloorByUnitName (property, unitName, unitType) {
                 for (const unit of floor.units) {
                     if (unit.label === unitName) {
                         res.sectionName = section.name
-                        res.sectionType = section.type
+                        res.sectionType = unitType !== 'parking' ? section.type : 'parking'
```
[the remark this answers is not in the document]
akarjakin: and in row 10 too

pahaz, line 16: It looks strange. Probably you need to talk with @sitozzz .
Probably, we need to fix all map data for that case.

sitozzz: @pahaz that's because `section.unitType` field is not required at `Ticket` domain:
image

So this lines of code is used for determining section attributes by unit

apps/condo/domains/ticket/utils/unit.js
```diff
@@ -1,19 +1,21 @@
 const { get } = require('lodash')
 
+const parkingType = 'parking'
```
sitozzz, line 3: Please use `PARKING_UNIT_TYPE` from `@condo/domains/property/constants/common`

apps/condo/domains/ticket/utils/unit.js
```diff
@@ -1,19 +1,21 @@
...
+
 function getSectionAndFloorByUnitName (property, unitName, unitType) {
     const res = {
         sectionName: null,
         sectionType: null,
         floorName: null,
     }
     if (unitName && unitType) {
-        const sections = get(property, ['map', `${unitType !== 'parking' ? 'sections' : 'parking'}`], [])
+        const sections = get(property, ['map', `${unitType !== parkingType ? 'sections' : parkingType}`], [])
```
sitozzz, line 12: `property.map` is not actually unit type. It is another block and should be get from another constant (in case of it will be changed in the future)

sitozzz: You can use `BuildingMap` type `@app/condo/schema` for that

apps/condo/domains/ticket/utils/unit.js
```diff
@@ -1,19 +1,21 @@
...
         for (const section of sections) {
             for (const floor of section.floors) {
                 for (const unit of floor.units) {
                     if (unit.label === unitName) {
                         res.sectionName = section.name
-                        res.sectionType = section.type
+                        res.sectionType = unitType !== parkingType ? section.type : parkingType
```
sitozzz, line 18: Look's a bit strange. May be better to do it more clear like this
```suggestion
                        res.sectionType = unitType !== PARKING_UNIT_TYPE ? SECTION_SECTION_TYPE : PARKING_SECTION_TYPE
```

sitozzz (commented):
Please use constants from `property` domain

pahaz (approved):

sitozzz (approved):

PR #2237, opened 2022-09-06
DOMA-4122 fix display payments indicator

nomerdvadcatpyat:
1. Changed the periods of requested receipts
2. Added read rights to BillingReceipt for related organizations
3. Wrote and used the `useCount` hook to get the number of receipts

### DIFF
--- a/apps/condo/domains/common/utils/codegeneration/generate.hooks.ts
+++ b/apps/condo/domains/common/utils/codegeneration/generate.hooks.ts
@@ -16,6 +16,9 @@ type IUseObjectsQueryReturnType<GQLObject> = {
     objs?: GQLObject[]
     meta?: { count?: number }
 }
+type IUseCountQueryReturnType = {
+    meta?: { count?: number }
+}
 type IRefetchType<GQLObject, QueryVariables> = (variables?: Partial<QueryVariables>) => Promise<ApolloQueryResult<IUseObjectsQueryReturnType<GQLObject>>>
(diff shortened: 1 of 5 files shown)


dkoviazin (approved):

apps/condo/domains/common/utils/codegeneration/generate.hooks.ts
```diff
@@ -165,6 +174,39 @@ export function generateReactHooks<
         }, [rowAction, onComplete])
     }
 
+    function useCount (variables: QueryVariables, options?: QueryHookOptions<IUseCountQueryReturnType, QueryVariables>) {
```
pahaz, line 177: Why it’s not the same as useObjects types

apps/condo/domains/common/utils/codegeneration/generate.hooks.ts
```diff
@@ -165,6 +174,39 @@ export function generateReactHooks<
...
+        const intl = useIntl()
+        const AccessDeniedError = intl.formatMessage({ id: 'AccessError' })
+        const ServerError = intl.formatMessage({ id: 'ServerErrorPleaseTryAgainLater' })
+
+        const { data, error, loading, refetch, fetchMore, stopPolling } = useQuery<IUseObjectsQueryReturnType<GQLObject>, QueryVariables>(gql.GET_COUNT_OBJS_QUERY, {
+            variables,
+            notifyOnNetworkStatusChange: true,
```
pahaz, line 184: Should we add it to others heeks

apps/condo/domains/common/utils/codegeneration/generate.hooks.ts
nomerdvadcatpyat: Because different return types, `useCount` does not return objs

apps/condo/domains/common/utils/codegeneration/generate.hooks.ts
nomerdvadcatpyat: This is already used in other hooks in generate.hooks

apps/condo/domains/common/utils/codegeneration/generate.hooks.ts
AleX83Xpert, line 177: Is it able to avoid code duplication? Looks like `useCount` should wrap `useObjects` and return only what is necessary.

AleX83Xpert (approved):

PR #2241, opened 2022-09-07
DOMA-4133 Added clear filter function for tooltip filters and fix it …

VKislov:
…in getTextFilterDropdown
For some reason, `onChange` inside filter components stopped working when using the default antd `clearFilters` cleanup function. Added an explicit clear call to `handleClear` and used it with `clearFilters`

### DIFF
--- a/apps/condo/domains/common/components/Table/Filters.tsx
+++ b/apps/condo/domains/common/components/Table/Filters.tsx
@@ -31,6 +31,8 @@ export const getTextFilterDropdown = (placeholder: string, containerStyles?: CSS
         const handleClear = useCallback(() => {
             setValue('')
             isFunction(clearFilters) && clearFilters()
+            setSelectedKeys('')
+            confirm({ closeDropdown: false })
         }, [clearFilters])
 
         return (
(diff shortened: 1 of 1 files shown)


AntonAL (commented):
I don't see in `getSelectFilterDropdown` how `clearFilters` triggers `confirm({ closeDropdown: false })`, that is called in `Select.onChange` in `apps/condo/domains/common/components/Table/Filters.tsx` module.
Maybe to make changed there?

AleX83Xpert (approved):

VKislov:
> I don't see in `getSelectFilterDropdown` how `clearFilters` triggers `confirm({ closeDropdown: false })`, that is called in `Select.onChange`

When `value` changes the component should re-render and execute `onChange`. But for some reason, this logic is broken. When a change comes through `clearFilters` `onChange` is not called and 
```
setSelectedKeys(e)
confirm({ closeDropdown: false })
```
is not executed. Therefore, `selectedKeys` holds the state before it was cleared and the filters are not updated.
`clearFilters` didn't call this logic because it should happen in `onChange` as described above had to call `setSelectedKeys` along with `clearFilters` and `confirm()` to apply the changes

PR #2252, opened 2022-09-08
DOMA-3960 changed buttons and title in `TasksProgress` component

AntonAL:
image


apps/condo/domains/common/components/tasks/TaskProgress.tsx
```diff
@@ -201,17 +203,28 @@ export const TaskProgressTracker: React.FC<ITaskProgressTrackerProps> = ({ task
...
+  transform: rotate(180deg);
+`
+
 const infoIconStyles = css`
   color: ${colors.infoIconColor};
   font-size: 20px;
 `
 
+const TasksProgressTitleStyle = css`
+  font-size: 16px;
```
sitozzz, line 223: please use font sizes from constants

sitozzz (approved):

sitozzz (commented):
Any screenshoots "before->after"

apps/condo/domains/common/components/containers/GlobalStyle.tsx
```diff
@@ -295,9 +295,21 @@ export default function GlobalStyle () {
               ${cardCSS}
               
               .ant-notification-notice.tasks {
+                padding: 17px;
```
pahaz, line 298: Please avoid to use global

pahaz (approved):

apps/condo/domains/common/components/containers/GlobalStyle.tsx
AntonAL: It's not a global, It's only for notification with `tasks` CSS-class.
Since there is an Ant notification component, not a component from `TaskProgress.tsx` module, here is an only place to declare it.

apps/condo/domains/common/components/tasks/TaskProgress.tsx
AntonAL: It is a custom component, It does not uses something global.
No need to extract it into constant, because we have only one usage place of that.

PR #2256, opened 2022-09-09
DOMA-2883 added condition on missing `OnBoarding` for redirect

AntonAL:
In some cases user may not have a connected `OnBoarding` record, like after invite with creating new `User` in `InviteNewOrganizationEmployeeService`
So, when no `OnBoarding` record is connected, redirect from this page

### DIFF
--- a/apps/condo/pages/onboarding.tsx
+++ b/apps/condo/pages/onboarding.tsx
@@ -45,10 +45,12 @@ const OnBoardingPage: IOnBoardingIndexPage = () => {
     }, [refetchOnBoarding])
 
     useEffect(() => {
-        if (get(onBoarding, 'completed', false)) {
-            router.push('/')
+        // In some cases user may not have a connected `OnBoarding` record, like after invite with creating new `User` in `InviteNewOrganizationEmployeeService`
+        // So, when no `OnBoarding` record is connected, redirect from this page
+        if (!isLoading && (!onBoarding || onBoarding.completed)) {
+            router.push('/reports')
         }
-    }, [onBoarding])
+    }, [onBoarding, isLoading])
 
     const sortedOnBoardingSteps = useMemo(() => {
         return onBoardingSteps.sort((leftStep, rightStep) => {


apps/condo/pages/onboarding.tsx
```diff
@@ -45,8 +45,10 @@ const OnBoardingPage: IOnBoardingIndexPage = () => {
     }, [refetchOnBoarding])
 
     useEffect(() => {
-        if (get(onBoarding, 'completed', false)) {
-            router.push('/')
+        // In some cases user may not have a connected `OnBoarding` record, like after invite with creating new `User` in `InviteNewOrganizationEmployeeService`
+        // So, when no `OnBoarding` record is connected, redirect from this page
+        if (!onBoarding || onBoarding.completed) {
+            router.push('/report')
```
AleX83Xpert, line 51: Why `/report`?

apps/condo/pages/onboarding.tsx
```diff
@@ -45,8 +45,10 @@ const OnBoardingPage: IOnBoardingIndexPage = () => {
     }, [refetchOnBoarding])
 
     useEffect(() => {
-        if (get(onBoarding, 'completed', false)) {
-            router.push('/')
+        // In some cases user may not have a connected `OnBoarding` record, like after invite with creating new `User` in `InviteNewOrganizationEmployeeService`
+        // So, when no `OnBoarding` record is connected, redirect from this page
+        if (!onBoarding || onBoarding.completed) {
```
AleX83Xpert, line 50: During the first rendering `onBoarding` will be null. What about checking the `isLoading` variable?

apps/condo/pages/onboarding.tsx
VKislov: `report` this seems to be our main page

VKislov: But it seems to be `/reports`

AntonAL: It was a typo

VKislov (approved):

apps/condo/pages/onboarding.tsx
AntonAL: Fixed

AleX83Xpert (approved):

PR #2266, opened 2022-09-12
DOMA-3363 fix getIntegerPartOfReading fn

VKislov:
after: 
Screenshot 2022-10-17 at 13 19 00
before: 
Screenshot 2022-09-12 at 16 09 07

### DIFF
--- a/apps/condo/domains/common/components/Table/Renders.tsx
+++ b/apps/condo/domains/common/components/Table/Renders.tsx
@@ -6,6 +6,7 @@ import dayjs from 'dayjs'
 import get from 'lodash/get'
 import isBoolean from 'lodash/isBoolean'
 import isString from 'lodash/isString'
+import isNull from 'lodash/isNull'
 import { FilterValue } from 'antd/es/table/interface'
 import { EllipsisConfig } from 'antd/es/typography/Base'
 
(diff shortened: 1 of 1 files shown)


sitozzz (approved):

Alllex202 (approved):

nomerdvadcatpyat (approved):

apps/condo/domains/common/components/Table/Renders.tsx
```diff
@@ -221,7 +221,7 @@ export const getTextRender = (search?: FilterValue | string) => {
     }
 }
 
-const getIntegerPartOfReading = (value: string) => value?.split(DEFAULT_CURRENCY_SEPARATOR)[0]
+const getIntegerPartOfReading = (value: string) => isString(value) && value.split(DEFAULT_CURRENCY_SEPARATOR)[1].includes('0000') ? value[0] : value
```
dkoviazin, line 224: You cannot be sure that
```
value.split(DEFAULT_CURRENCY_SEPARATOR)
```
will always have [1] element

pahaz, line 224: it looks like a hack. Probably, you should check the meter `units of measurement`

pahaz (changes requested):
It looks like a dirty hack >_< 
If you want, you can convert one units of measurement to another. 

But you can't just drop the fractional part

apps/condo/domains/common/components/Table/Renders.tsx
VKislov: The database stores readings in the decimal type and it will always have a separator character and numbers after it. If the number is an integer, then it will be stored as x.0000 so [1] will always be there if some value comes

VKislov:
> It looks like a dirty hack >_< If you want, you can convert one units of measurement to another.

I don't need to change the units of measurement, I just need to not display zeros if the number is an integer. If the number has a fractional part I will display it unchanged. And as an example, 123.0000 === 123, so I don't see any problem with that. The design confirmed that such a display should have been originally

apps/condo/domains/common/components/Table/Renders.tsx
```diff
@@ -221,20 +221,25 @@ export const getTextRender = (search?: FilterValue | string) => {
     }
 }
 
-const getIntegerPartOfReading = (value: string) => value?.split(DEFAULT_CURRENCY_SEPARATOR)[0]
+const getPrettyValueOfReading = (value: string) => {
+    const splitedValue = isString(value) && value.split(DEFAULT_CURRENCY_SEPARATOR)
```
pahaz, line 225: why did you just delete an integer part of the value?

pahaz (approved):

PR #2275, opened 2022-09-13
fix(condo): DOMA-4134 fixed sort tickets by unitName

Alllex202:
Before: sort unitName as string
Снимок экрана 2022-09-13 в 17 01 07

After: sort unitName as number
Снимок экрана 2022-09-13 в 16 49 07

### DIFF
--- a/apps/condo/domains/common/utils/tables.utils.ts
+++ b/apps/condo/domains/common/utils/tables.utils.ts
@@ -14,6 +14,8 @@ import { preciseFloor } from './helpers'
 import { FilterDropdownProps } from 'antd/es/table/interface'
 import dayjs from 'dayjs'
 
+import { Ticket } from '@app/condo/schema'
+
 export type DataIndexType = string | Array<string>
 export type QueryArgType = string | Array<string>
 export type FiltersFromQueryType = { [key: string]: QueryArgType }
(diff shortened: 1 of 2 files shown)


apps/condo/domains/common/utils/tables.utils.ts
```diff
@@ -380,3 +382,12 @@ export const convertColumns = (
 export const getSorterMap: SorterMapType = (sorters) => {
     return Object.assign({}, ...sorters.map((sorter) => ({ [sorter.columnKey]: sorter.order })))
 }
+
+export const sorterUnitName: (a: Ticket, b: Ticket) => number = (a: Ticket, b: Ticket) => {
```
akarjakin, line 386: bad function name. `sortUnitName`

apps/condo/domains/common/utils/tables.utils.ts
```diff
@@ -380,3 +382,12 @@ export const convertColumns = (
...
+    const firstUnitName = get(a, 'unitName', null)
+    const secondUnitName = get(b, 'unitName', null)
+    if (!firstUnitName && !secondUnitName) return 0
+    if (!secondUnitName) return -1
+    if (!firstUnitName) return 1
+    return Number(firstUnitName) - Number(secondUnitName)
```
akarjakin, line 392: there could be case with NaN, I think you should handle it too. And please add tests for this sorter function.

akarjakin (changes requested):

apps/condo/domains/common/utils/tables.utils.ts
Alllex202: first 3 conditions also check NaN

apps/condo/domains/common/utils/tables.utils.ts
sitozzz, line 386: Please add tests for this util function

AleX83Xpert: This is actually a sorter.
`unitNameSorter`

AleX83Xpert (approved):

Alllex202:
unitName can be a string, not just a number. And the sorting happens on the server, not on the client. It's better to leave it as is.

PR #2294, opened 2022-09-15
fix(condo): DOMA-3886 fix unitType display

dkoviazin:
Before - always FLAT
image

After - Depending on unitType
image

### DIFF
--- a/apps/condo/domains/billing/components/ServicesModal.tsx
+++ b/apps/condo/domains/billing/components/ServicesModal.tsx
@@ -99,7 +99,8 @@ export const ServicesModal: React.FC<IServicesModalProps> = ({
     const address = get(receipt, ['property', 'address'])
     const unitName = get(receipt, ['account', 'unitName'])
     const unitType = get(receipt, ['account', 'unitType'])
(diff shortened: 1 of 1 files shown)


sitozzz (approved):

apps/condo/domains/billing/components/ServicesModal.tsx
```diff
@@ -99,17 +98,20 @@ export const ServicesModal: React.FC<IServicesModalProps> = ({
     const accountNumber = get(receipt, ['account', 'number'])
     const address = get(receipt, ['property', 'address'])
     const unitName = get(receipt, ['account', 'unitName'])
+    const unitType = get(receipt, ['account', 'unitType'])
+
+    const UnitTypePrefix = intl.formatMessage({ id: `field.UnitType.prefix.${unitType}` }).toLocaleLowerCase()
```
pahaz, line 103: Do you really need to use to lowe?

dkoviazin: We have the same way in address display from TicketChange.  So, this was done to keep consistency
```js
const addressChangePostfix += `, ${intl.formatMessage({ id: `field.UnitType.prefix.${unitType}` }).toLowerCase()}. ${unitName}`
```

PR #2306, opened 2022-09-20
fix(condo): DOMA-4253 Cookies settings were tuned

AleX83Xpert:
In order to make session cookie available we set sameSite='none' for keystone's session cookie. The same was performed for short-term oidc cookies.

### DIFF
--- a/apps/condo/domains/user/oidc/configuration.js
+++ b/apps/condo/domains/user/oidc/configuration.js
@@ -52,6 +52,9 @@ module.exports = function createConfiguration (context) {
         cookies: {
             // TODO(pahaz): take it from .env!
             keys: ['some secret key', 'and also the old rotated away some time ago', 'and one more'],
+            short: {
+                sameSite: 'None',
+            },
         },
         claims: {
             // TODO(pahaz): SCOPES think about it!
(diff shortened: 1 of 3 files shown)


SavelevMatthew (approved):

dkoviazin (approved):

sitozzz (approved):

apps/condo/domains/user/oidc/configuration.js
```diff
@@ -52,6 +52,9 @@ module.exports = function createConfiguration (context) {
         cookies: {
             // TODO(pahaz): take it from .env!
             keys: ['some secret key', 'and also the old rotated away some time ago', 'and one more'],
+            short: {
+                sameSite: 'None',
```
pahaz, line 56: You also need to set secure attribute. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#none

AleX83Xpert: This option doesn't exist for oidc-provider

https://github.com/panva/node-oidc-provider/blob/main/docs/README.md#cookiesshort

PR #2324, opened 2022-09-27
DOMA-3529 fix import after changing organization

VKislov:
When you change the organization, the import occurs in the previous one. This is because the organizationId is not being used from the function scope, causing the organizationId to not be updated. Solution: use useRef so that the function always has access to the current organization id

### DIFF
--- a/apps/condo/domains/contact/hooks/useImporterFunctions.tsx
+++ b/apps/condo/domains/contact/hooks/useImporterFunctions.tsx
@@ -13,6 +13,7 @@ import { useApolloClient } from '@condo/next/apollo'
 import { useIntl } from '@condo/next/intl'
 import { useOrganization } from '@condo/next/organization'
 import get from 'lodash/get'
+import { useEffect, useRef } from 'react'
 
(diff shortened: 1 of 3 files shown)


AntonAL (approved):

akarjakin (approved):

apps/condo/domains/contact/hooks/useImporterFunctions.tsx
```diff
@@ -210,7 +215,7 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
             }
 
             const contactData = {
-                organization: { connect: { id: String(userOrganizationId) } },
+                organization: { connect: { id: String(userOrganizationIdRef.current) } },
```
sitozzz, line 218: Why here you need to wrap organizationId with `String()`?

VKislov: it doesn't seem to affect anything. 
if it matters i can remove it

PR #2327, opened 2022-09-27
fix(condo): DOMA-4014 trim role names before save

lapakota:
trim role names to avoid duplication of names


nomerdvadcatpyat (approved):
I think you should also add normalization on the server side

apps/condo/domains/contact/components/contactRoles/BaseContactRoleForm.tsx
```diff
@@ -47,13 +46,12 @@ export const BaseContactRoleForm: React.FC<BaseTicketPropertyHintFormProps> = ({
...
     const router = useRouter()
 
-    const createContactRoleAction = ContactRole.useCreate({})
-    const softDeleteContactRoleAction = ContactRole.useSoftDelete()
-
     const validationRules = useContactRoleValidationRules()
 
     const handleFormSubmit = useCallback(async (values) => {
         const initialContactRoleId = get(initialValues, 'id')
+        const normalizedName = values.name && values.name.trim()
```
sitozzz, line 53: Name can be undefined here?

lapakota: It seems not, because field must be filled

fixed

sitozzz (approved):

PR #2328, opened 2022-09-27
Fix/condo/doma 4224/fix stead support

Kanol:
Problem:
Some Properties have address with "участок/уч." instead of "дом/д.". It cause visual problems on Properties page. Also we can't add property like this
Solution:
Add stead support

### DIFF
--- a/apps/condo/domains/common/utils/helpers.ts
+++ b/apps/condo/domains/common/utils/helpers.ts
@@ -47,6 +47,9 @@ export const getAddressDetails = (property: ObjectWithAddressInfo) => {
     const houseType = get(addressMeta, 'house_type')
     const houseName = get(addressMeta, 'house')
 
+    const steadType = get(addressMeta, 'stead_type')
+    const steadName = get(addressMeta, 'stead')
+
     const blockType = get(addressMeta, 'block_type')
     const blockName = get(addressMeta, 'block')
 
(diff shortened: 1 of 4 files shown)


toplenboren (commented):
Whats our product position on this?

apps/condo/domains/property/components/BasePropertyForm/index.tsx
```diff
@@ -164,7 +164,7 @@ const BasePropertyForm: React.FC<IPropertyFormProps> = (props) => {
                                             setAddressValidatorError={setAddressValidatorError}
                                             onSelect={(_, option) => {
                                                 const address = JSON.parse(option.key as string) as AddressMetaField
-                                                if (!validHouseTypes.includes(address.data.house_type_full)) {
+                                                if (!validHouseTypes.includes(address.data.house_type_full) && address.data.stead_type_full == null) {
```
toplenboren, line 167: Why not use get? 

Also i'd move logic with stead to different place, like with !validHouseTypes

toplenboren:
how is this going to be used thorough the system?

sitozzz:
Could this problem been resolved when we switch to "address service"?

apps/condo/domains/common/utils/helpers.ts
```diff
@@ -47,6 +47,9 @@ export const getAddressDetails = (property: ObjectWithAddressInfo) => {
     const houseType = get(addressMeta, 'house_type')
     const houseName = get(addressMeta, 'house')
 
+    const steadType = get(addressMeta, 'stead_type')
```
AntonAL, line 50: Could you please add comments, what this is used for?

toplenboren:
?

apps/condo/domains/property/components/BasePropertyForm/index.tsx
pahaz: Looks something wrong here

PR #2356, opened 2022-10-03
fix(condo): DOMA-4332 fixed template filter saving with unitName

Alllex202:
Problems 1: users cannot save filter template if it have `unitName`. 
Solution 1: 'unitName' field should be string array, but was just string. 

Problem 2: if press `Enter` in dropdown filter, then will change sort by for this column. 
Solution 2: stop propagation `onKeyDown` for these dropdown filters containers

### DIFF
--- a/apps/condo/domains/common/components/TableFilter.tsx
+++ b/apps/condo/domains/common/components/TableFilter.tsx
@@ -21,13 +21,15 @@ export interface IFilterContainerProps {
 
 const FILTER_CONTAINER_STYLES: CSSProperties = { padding: 16 }
 
+const handleStopPropagation = (e) => e.stopPropagation()
+
 export const FilterContainer: React.FC<IFilterContainerProps> = (props) => {
     const { showClearButton, clearFilters, children } = props
     const intl = useIntl()
     const ResetLabel = intl.formatMessage({ id: 'filters.Reset' })
 
     return (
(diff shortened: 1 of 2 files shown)


dkoviazin (approved):

DanilovMaxim13 (approved):

apps/condo/domains/common/components/TableFilter.tsx
```diff
@@ -21,13 +21,15 @@ export interface IFilterContainerProps {
...
+const handleStopPropagation = (e) => e.stopPropagation()
+
 export const FilterContainer: React.FC<IFilterContainerProps> = (props) => {
     const { showClearButton, clearFilters, children } = props
     const intl = useIntl()
     const ResetLabel = intl.formatMessage({ id: 'filters.Reset' })
 
     return (
-        <div style={FILTER_CONTAINER_STYLES}>
+        <div style={FILTER_CONTAINER_STYLES} onKeyDown={handleStopPropagation}>
```
AntonAL, line 32: Does this fix actual for "callcenter" app only.
Is it presented in "condo" app?

Alllex202: it was fixed in all tables, who use these containers. And "condo", and "callcenter"

PR #2357, opened 2022-10-04
DOMA-4288 added use desk widget

VKislov:
Screenshot 2022-10-04 at 09 54 02
This widget is visible in the lower right corner. For now, there will be two widgets on `condo.d`. After the release, through the settings of the `jivoSite` widget, it will be disabled and will not be displayed on v1, after which the `jivoSite` component will be deleted


apps/condo/domains/common/components/UseDeskWidget.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import getConfig from 'next/config'
+
+const UseDeskWidget: React.FC = () => {
+
+    const { publicRuntimeConfig:{ UseDeskWidgetId } } = getConfig()
+
+    return UseDeskWidgetId ?
+        <script async src={`https://lib.usedesk.ru/secure.usedesk.ru/${UseDeskWidgetId}.js`}></script> : null
```
AleX83Xpert, line 9: Maybe it's better to use the current schema?
```suggestion
        <script async src={`//lib.usedesk.ru/secure.usedesk.ru/${UseDeskWidgetId}.js`}></script> : null
```

AleX83Xpert (approved):

apps/condo/domains/common/components/UseDeskWidget.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import getConfig from 'next/config'
+
+const UseDeskWidget: React.FC = () => {
+
+    const { publicRuntimeConfig:{ UseDeskWidgetId } } = getConfig()
```
sitozzz, line 6: Why is the config access method located in render section? May be move this line out of component body?

sitozzz (approved):

PR #2358, opened 2022-10-04
DOMA-4330 extend unit form validation to consider with unitType changes

sitozzz:


AleX83Xpert (approved):

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -960,16 +960,21 @@ const UnitForm: React.FC<IPropertyMapModalForm> = ({ builder, refresh }) => {
...
+            const labelChanged = selectedUnit.label !== label
+            const labelValidation = labelChanged
                 ? builder.validateUniqueUnitLabel(label, 'section')
-                : true
+                : false
 
-            isUnitLabelUnique = unitPlacementChanged && labelValidation
+            isUnitLabelUnique = unitPlacementChanged
+                ? unitPlacementChanged && (!labelChanged || labelValidation )
+                : labelValidation
```
Alllex202, line 973: Why use `unitPlacementChanged` twice?

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -1438,11 +1443,14 @@ const ParkingUnitForm: React.FC<IPropertyMapModalForm> = ({ builder, refresh })
...
+            const labelChanged = selectedUnit.label !== label
+            const labelValidation = labelChanged
                 ? builder.validateUniqueUnitLabel(label, 'parking')
-                : true
+                : false
 
-            isUnitLabelUnique = unitPlacementChanged && labelValidation
+            isUnitLabelUnique = unitPlacementChanged
+                ? unitPlacementChanged && (!labelChanged || labelValidation)
+                : labelValidation
```
Alllex202, line 1453: Why use unitPlacementChanged twice?

VKislov (approved):

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
sitozzz: @Alllex202 It results from the unit uniqueness validation logic. 
So we have already created map and we want to edit unit. First of all we should check may any fields of unit has been changed. There it is - unitType, section or floor. Second part is check that label of the unit is unique by all of the sections.
Otherwise we only check label uniqueness of the edited unit

PR #2362, opened 2022-10-04
fix(condo): DOMA-4222 fixed selected template in modal filter. 

Alllex202:
Problem: selected template was reset because MultipleFiltersModal was re-renders. 

Solution: use useMemo instead of useCallback

Some changes: 
1. Also wrong template order - `New template` should be first.
2. Was selected first loaded template, should `New Template` or applied template. 
3. Was added label and placeholder in filed for changing template name
4. Change layout for TabPane

Before:
Снимок экрана 2022-10-04 в 19 46 00

After: 
Снимок экрана 2022-10-04 в 19 47 38

### DIFF
--- a/apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
+++ b/apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
@@ -322,7 +322,7 @@ const PaymentsTable: React.FC<IPaymentsTableProps> = ({ billingContext, contexts
                 </Typography.Text>
             </Modal>
 
-            <MultipleFiltersModal/>
+            {MultipleFiltersModal}
         </>
     )
 }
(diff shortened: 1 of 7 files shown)


toplenboren (approved):

akarjakin (approved):

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
```diff
@@ -322,7 +322,7 @@ const PaymentsTable: React.FC<IPaymentsTableProps> = ({ billingContext, contexts
                 </Typography.Text>
             </Modal>
 
-            <MultipleFiltersModal/>
+            {MultipleFiltersModal}
```
SavelevMatthew, line 325: Why passing component like this?

Alllex202: if call component, then component rerender and reset selected template. 
If use just jsx, then this doesn't happen

Alllex202: I tried passing the `key` for MultipleFiltersModal like `<MultipleFiltersModal key='modal' />`, but that didn't help either.

dkoviazin (approved):

PR #2367, opened 2022-10-05
fix(condo): DOMA-4059 fixed data loading in GraphQlSearchInput

Alllex202:
Problem 1: GraphQlSearchInput show id while not loaded data. 
Solution 1: show loading label while not loaded data. 

Problem 2: not loading initial options because `initialValue` is empty. 
Solution 2: if `initialValue` is empty get values from `props.value`. 

Also was fixed infinity loading stopping if all data loaded

Before: 
Запись экрана 2022-10-05 в 17 00 51

After: 
Запись экрана 2022-10-05 в 17 02 52


apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -63,17 +64,32 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
+    const LoadingMessage = intl.formatMessage({ id: 'Loading' })
+
     const client = useApolloClient()
     const [selected, setSelected] = useState('')
     const [isLoading, setLoading] = useState(false)
     const [data, setData] = useState([])
     const [isAllDataLoaded, setIsAllDataLoaded] = useState(false)
     const [value, setValue] = useState('')
 
+    const _placeholder = useMemo(() => (isLoading && !data.length) ? LoadingMessage : placeholder,
```
AleX83Xpert, line 84: Looks like in the case of no data (data.length === 0) we'll still see loading message.

AleX83Xpert (approved):

apps/condo/domains/common/components/GraphQlSearchInput.tsx
Alllex202: only if loading

AleX83Xpert: Hmm, you right! My bad :)

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -63,17 +64,32 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
+        [LoadingMessage, data.length, isLoading, placeholder])
+
+    const _selectedValue = useMemo(() => (isLoading && !data.length) ? [] : selectedValues,
```
akarjakin, line 87: you're calculating `isLoading && !data.length` several times, it's code duplication. better to calculate it once, store to a variable and reuse it.

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -101,18 +117,19 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
         : data.map((option, index) => renderOption(option, index))
 
     const loadInitialOptions = useCallback(async () => {
-        const initialValueQuery = isFunction(getInitialValueQuery) ? getInitialValueQuery(initialValue) : { id_in: initialValue }
+        const values = initialValue || selectedValues
+        const initialValueQuery = isFunction(getInitialValueQuery) ? getInitialValueQuery(values) : { id_in: values }
 
-        if (Array.isArray(initialValue) && initialValue.length) {
+        if (Array.isArray(values) && values.length) {
```
akarjakin, line 123: if (!isEmpty(values))

apps/condo/domains/common/components/GraphQlSearchInput.tsx
SavelevMatthew, line 84: Seems a bit strange to mix `normalNotation` with `_privateNotation`. And what's the reason behind using private notation here?

Alllex202:
have some bugs

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -63,16 +65,35 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
+    const [isInitialLoading, seInitialLoading] = useState(false)
     const [isAllDataLoaded, setIsAllDataLoaded] = useState(false)
-    const [value, setValue] = useState('')
+    const [initialData, setInitialData] = useState([])
+    const [allData, setAllData] = useState([])
+    const [showingData, setShowingData] = useState([])
+    const [searchData, setSearchData] = useState([])
+    const [searchValue, setSearchValue] = useState('')
+
+    const isShowLoader = useMemo(() => isInitialLoading || ((isSearchLoading || isMoreLoading) && isEmpty(allData)),
```
nomerdvadcatpyat, line 89:
```suggestion
    const loading = useMemo(() => isInitialLoading || ((isSearchLoading || isMoreLoading) && isEmpty(allData)),
```

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -63,16 +65,35 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
-    const [isLoading, setLoading] = useState(false)
-    const [data, setData] = useState([])
+    const [isSearchLoading, setSearchLoading] = useState(false)
+    const [isMoreLoading, setMoreLoading] = useState(false)
+    const [isInitialLoading, seInitialLoading] = useState(false)
     const [isAllDataLoaded, setIsAllDataLoaded] = useState(false)
-    const [value, setValue] = useState('')
+    const [initialData, setInitialData] = useState([])
+    const [allData, setAllData] = useState([])
+    const [showingData, setShowingData] = useState([])
```
nomerdvadcatpyat, line 85:
```suggestion
    const [options, setOptions] = useState([])
```

apps/condo/domains/common/components/GraphQlSearchInput.tsx
```diff
@@ -96,53 +117,34 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
         )
     }
 
-    const options = renderOptions
-        ? renderOptions(data, renderOption)
-        : data.map((option, index) => renderOption(option, index))
+    const options = useMemo(() => renderOptions
```
nomerdvadcatpyat, line 120:
```suggestion
    const renderedOptions = useMemo(() => renderOptions
```

nomerdvadcatpyat (approved):

dkoviazin (approved):

Alllex202:
> have some bugs
with initial props work bad

it fixed

Also add loader and debounce for searching. 
Now every time you press a key, no request is sent. Only one request is sent when printing stops

pahaz:
ScreenShot looks nice!

Alllex202:
The problem with the synchronization of dropdown filters and the filter in the modal was also fixed.

PR #2377, opened 2022-10-06
fix(condo): DOMA-4335 fix ticket source edit field 

Alllex202:
1) Added loader for TicketSourceSelect. 
2) Added source 'mobile app' for editing. (Problem: was showed only id for this value)

Before: 
Снимок экрана 2022-10-06 в 19 46 46

After: 
Снимок экрана 2022-10-06 в 19 48 25


apps/condo/domains/ticket/components/BaseTicketForm/index.tsx
```diff
@@ -276,15 +277,16 @@ export const TicketSourceSelect: React.FC = () => {
...
             data-cy='ticket__source-item'
             initialValue={DEFAULT_TICKET_SOURCE_CALL_ID}
         >
             <Select
+                loading={loading}
                 style={TICKET_SOURCE_SELECT_STYLE}
                 options={sourceOptions}
-                defaultValue={DEFAULT_TICKET_SOURCE_CALL_ID}
                 disabled={loading}
+                value={loading ? LoadingMessage : ''}
```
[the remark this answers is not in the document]
Alllex202: I tried. placeholder don't showing. Value for this field will get from form. value from prop will get only if name is null

apps/condo/domains/ticket/components/BaseTicketForm/index.tsx
```diff
@@ -276,15 +277,16 @@ export const TicketSourceSelect: React.FC = () => {
         <Form.Item
             label={TicketSourceLabel}
             required
-            name='source'
+            name={loading ? null : 'source'}
```
AntonAL, line 280: It looks strange.
What the reason of doing this?

Alllex202: if item have name, then it use value from form. But we need set custom value in `Input` component if loading data that's why name set null

apps/condo/domains/ticket/components/BaseTicketForm/index.tsx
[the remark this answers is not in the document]
Alllex202: It's been refactored

apps/condo/domains/ticket/components/BaseTicketForm/index.tsx
Alllex202: It's been refactored

AntonAL (approved):

apps/condo/domains/ticket/components/BaseTicketForm/index.tsx
```diff
@@ -266,12 +266,28 @@ const DEFAULT_TICKET_SOURCE_CALL_ID = '779d7bb6-b194-4d2c-a967-1f7321b2787f'
...
     const sourceOptions = convertToOptions<TicketSourceType>(sources, 'name', 'id')
 
+    const LoadingSelect = useMemo(() => (
+        <Form.Item
+            label={TicketSourceLabel}
+            required
+        >
+            <Select
+                style={TICKET_SOURCE_SELECT_STYLE}
+                disabled={true}
```
akarjakin, line 283: why don't you put these to original select below ?

akarjakin (approved):

PR #2383, opened 2022-10-07
DOMA-4288 send tin and user info with user message

VKislov:
sending user info with user messages in useDesk for support staff


nomerdvadcatpyat (approved):

apps/condo/domains/common/components/UseDeskWidget.tsx
```diff
@@ -1,9 +1,33 @@
...
+        // @ts-ignore
+        if (UseDeskWidgetId) window.usedeskMessenger.userIdentify(
+            {
+                name: get(link, 'name', null),
+                email: get(link, 'email', null),
+                phone: get(link, 'phone', '').slice(1),
+                additional_fields:
+                    [
+                        {
+                            id: 20560, value: get(link, ['organization', 'tin'], null),
```
sitozzz, line 21: May be move id to constants or config?

apps/condo/domains/common/components/UseDeskWidget.tsx
```diff
@@ -1,9 +1,33 @@
...
+import React, { useCallback, useEffect } from 'react'
 import getConfig from 'next/config'
+import get from 'lodash/get'
 
 const { publicRuntimeConfig:{ UseDeskWidgetId } } = getConfig()
 
 const UseDeskWidget: React.FC = () => {
+    const { link } = useOrganization()
+
+    const userIdentify = useCallback(() => {
```
sitozzz, line 11: I think useCallback here is unnecessary

sitozzz (approved):

VKislov:
Removed `JivoSiteWidget` from the code. Initially, it was planned to disable it through the `JivoSite` admin panel, but they have a bug in the admin panel and so far it is impossible to disable the widget remotely. Therefore, in order to avoid overlapping widgets, I removed it from the code. I remember about the config in Helm, I'll drink it in one of the following

PR #2384, opened 2022-10-10
DOMA-4288 added organization and role name to user info for UseDeskWidget

VKislov:
added 2 more fields, role name and organization name

### DIFF
--- a/apps/condo/domains/common/components/UseDeskWidget.tsx
+++ b/apps/condo/domains/common/components/UseDeskWidget.tsx
@@ -4,7 +4,11 @@ import getConfig from 'next/config'
 import get from 'lodash/get'
 
 const { publicRuntimeConfig:{ UseDeskWidgetId } } = getConfig()
-const fieldId = 20560
+const useDeskFieldsIdsMap = {
+    tin: 20560,
+    role: 20574,
+    organizationName: 20572,
+}
 
 const UseDeskWidget: React.FC = () => {
     const { link } = useOrganization()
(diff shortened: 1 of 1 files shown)


apps/condo/domains/common/components/UseDeskWidget.tsx
```diff
@@ -4,7 +4,11 @@ import getConfig from 'next/config'
 import get from 'lodash/get'
 
 const { publicRuntimeConfig:{ UseDeskWidgetId } } = getConfig()
-const fieldId = 20560
+const fieldId = {
```
AleX83Xpert, line 7: What about the more understandable name?
```suggestion
const useDeskFieldsIdsMap = {
```

AleX83Xpert (approved):

DanilovMaxim13 (approved):

apps/condo/domains/common/components/UseDeskWidget.tsx
VKislov: yes you are right, I forgot to change :)

apps/condo/domains/common/components/UseDeskWidget.tsx
```diff
@@ -4,7 +4,11 @@ import getConfig from 'next/config'
 import get from 'lodash/get'
 
 const { publicRuntimeConfig:{ UseDeskWidgetId } } = getConfig()
-const fieldId = 20560
+const useDeskFieldsIdsMap = {
+    tin: 20560,
```
akarjakin, line 8: Better to put such data to .helm, in order to be able to change values without release process.

VKislov: okay, I'll move it to helm when I cut out the jivosite id from there in the next release

PR #2389, opened 2022-10-10
fix(condo): DOMA-4288 added get request for userIdentify

DanilovMaxim13:


apps/condo/domains/common/components/UseDeskWidget.tsx
```diff
@@ -1,7 +1,7 @@
 import { useOrganization } from '@condo/next/organization'
 import React, { useEffect } from 'react'
 import getConfig from 'next/config'
-import get from 'lodash/get'
+import { get, isFunction } from 'lodash'
```
VKislov, line 4: it is better to import only to the required function. 
import get from 'lodash/get'
import get from 'lodash/isFunction'

VKislov (approved):

Alllex202 (approved):

PR #2398, opened 2022-10-12
fix(condo): DOMA-4388 fixed parse empty response with feature flags

nomerdvadcatpyat:
An error occurs in `FeatureFlagsContext` when parsing a response in json, if this response is `undefined`

изображение

**Solution:** return an empty object in cases when we can't get feature flags

### DIFF
--- a/packages/featureflags/featureToggleManager.js
+++ b/packages/featureflags/featureToggleManager.js
@@ -12,6 +12,8 @@ let featureToggleApiKey
 const REDIS_FEATURES_KEY = 'features'
 const FEATURES_EXPIRED_IN_SECONDS = 60
 
+const WRONG_FEATURE_TOGGLE_CONFIG_ERROR = 'Wrong FEATURE_TOGGLE_CONFIG config!'
+
 class FeatureToggleManager {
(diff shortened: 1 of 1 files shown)


Alllex202 (approved):

packages/featureflags/featureToggleManager.js
```diff
@@ -40,6 +40,8 @@ class FeatureToggleManager {
                 redisClient.set(REDIS_FEATURES_KEY, JSON.stringify(fetchedFeatureFlags), 'EX', FEATURES_EXPIRED_IN_SECONDS)
 
                 return fetchedFeatureFlags
+            } else {
+                return {}
```
AleX83Xpert, line 44: Now the empty object means no configuration data. It looks strange :)
What about throwing an error in the case of no configuration data?
```
if (!featureToggleApiUrl || !featureToggleApiKey) {
    throw new Error('no config message')
}
```

Also, it may be helpful to return an empty object from the catch-branch.

nomerdvadcatpyat: done

PR #2400, opened 2022-10-12
fix(condo): DOMA-4389 removed ticket import from callcenter

Alllex202:
`callcenter` should not be able to import tickets


nomerdvadcatpyat (approved):

AleX83Xpert (approved):
Looks like this is not "removing ticket import from call center", but "allowing ticket import in condo" :)

PR #2402, opened 2022-10-12
fix(keystone): DOMA-4391 add task usage logic guard

pahaz:


packages/keystone/tasks.js
```diff
@@ -157,6 +157,8 @@ function createTaskWrapper (name, fn, defaultTaskOptions = {}) {
         throw new Error('This function is converted to a task, and you need to use fn.delay(...) to call the tasks')
     }
```
akarjakin, line 159:
```
function createTaskWrapper (name, fnRaw, defaultTaskOptions = {}) {
  function fn(...args) {
    return fnRaw(...args)
  }
...
}
```

akarjakin (approved):

packages/keystone/tasks.spec.js
```diff
@@ -1,8 +1,10 @@
 const { createTask, createWorker, taskQueue } = require('./tasks')
 
-async function asyncAddTask (a, b) {
-    if (b === 0) throw new Error('DivBy0')
-    return Promise.resolve(a / b)
+function createTaskFactory () {
+    return async function asyncAddTask (a, b) {
```
Dimitreee, line 4:
```suggestion
    return async function addAsyncTask (a, b) {
```

sitozzz (approved):

PR #2420, opened 2022-10-14
fix(condo): DOMA-4231 added baseQueryLoading prop to TicketPageContent

nomerdvadcatpyat:
In the call center, we need to wait for the `relatedOrganizations` request, which are forwarded to the `baseQuery`.
If this is not done, then `ticketsWithoutFilters` will initially be empty and a stub will be displayed.

### DIFF
--- a/apps/condo/pages/ticket/index.tsx
+++ b/apps/condo/pages/ticket/index.tsx
@@ -135,6 +135,7 @@ export const TicketsPageContent = ({
     sortableProperties,
     useTableColumns,
     showImport = false,
+    baseQueryLoading = false,
 }): JSX.Element => {
     const intl = useIntl()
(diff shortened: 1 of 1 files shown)


VKislov (approved):

apps/condo/pages/ticket/index.tsx
```diff
@@ -135,6 +135,7 @@ export const TicketsPageContent = ({
     sortableProperties,
     useTableColumns,
     showImport = false,
+    baseQueryLoading = false,
```
sitozzz, line 138: I think "base query" is not clear name for that context. Maybe add little bit more context to this prop? For example `relatedOrganizationsLoading`?

nomerdvadcatpyat: No, just here it should be without the context of the cc.
This can also be used if we need additional requests in the condo to load tickets.

sitozzz (approved):

PR #2422, opened 2022-10-14
Feature/condo/doma 4064/address-service-main-code

AleX83Xpert:
Address service was integrated into the `Property` and `Resident` models using a newly created plugin


apps/address-service/domains/address/schema/Address.js
```diff
@@ -0,0 +1,54 @@
...
+    fields: {
+
+        source: {
+            schemaDoc: 'The origin which used to detect the address',
+            type: Text,
+            isRequired: true,
+        },
+
+        address: {
+            schemaDoc: 'The address itself in one string',
```
pahaz, line 23: normalized?

apps/condo/domains/property/schema/Property.js
```diff
@@ -29,15 +29,45 @@ const MapSchemaJSON = require('@condo/domains/property/components/panels/Builder
...
+                ...addressSearch,
+                organization: {
+                    id: organizationId,
+                },
+                deletedAt: null,
+            }
+            const sameAddressProperties = await PropertyAPI.getAll(context, where, { first: 1 })
+
+            if (!isEmpty(sameAddressProperties)) {
+                addFieldValidationError(`${UNIQUE_ALREADY_EXISTS_ERROR}${fieldPath}] Property with same address exist in current organization`)
```
pahaz, line 65: it's bettr to throw GQLError

packages/keystone/plugins/addressService.js
```diff
@@ -0,0 +1,159 @@
...
+ * @param {string?} addressFieldName
+ * @param {Object} fieldsHooks
+ */
+const addressService = (addressFieldName = 'address', fieldsHooks = {}) => plugin(({
+    fields = {},
+    hooks = {},
+    ...rest
+}, schemaObj) => {
+    if (!addressFieldName) {
+        return { fields, hooks, ...rest }
```
pahaz, line 96: it's better to throw an error

packages/keystone/plugins/utils/addressMetaDefinition.js
```diff
@@ -0,0 +1,197 @@
...
+    schemaDoc: 'Property address components',
+    type: Json,
+    extendGraphQLTypes: [ADDRESS_META_FIELD_GRAPHQL_TYPES],
+    graphQLReturnType: 'AddressMetaField',
+    graphQLAdminFragment: `{ ${ADDRESS_META_SUBFIELDS_QUERY_LIST} }`,
+    isRequired: true,
+    kmigratorOptions: { null: false },
+    hooks: {
+        validateInput: ({ resolvedData, fieldPath, addFieldValidationError }) => {
+            if (!resolvedData.hasOwnProperty(fieldPath)) return addFieldValidationError(`${fieldPath} value is required`)
```
pahaz, line 180: it looks like old style validation

apps/condo/domains/property/schema/Property.js
```diff
@@ -283,7 +282,7 @@ const Property = new GQLListSchema('Property', {
             },
         },
     },
-    plugins: [uuided(), versioned(), tracked(), softDeleted(), dvAndSender(), historical()],
+    plugins: [uuided(), addressService('address', { address: addressFieldHooks }), versioned(), tracked(), softDeleted(), dvAndSender(), historical()],
```
pahaz, line 285: why we need to owerride the field name? it looks like we use `address` everywhere

apps/address-service/domains/address/schema/Address.js
AleX83Xpert: Yes

apps/condo/domains/property/schema/Property.js
AleX83Xpert: I thought about `Ticket.propertyAddress` and `Ticket.propertyAddressMeta`.
But now I think this overriding may not need. If yes, I'll remove the code for overriding.

apps/condo/domains/property/schema/Property.js
AleX83Xpert: Done

packages/keystone/plugins/addressService.js
AleX83Xpert: Done

packages/keystone/plugins/utils/addressMetaDefinition.js
AleX83Xpert: Fixed

apps/address-service/domains/address/utils/testSchema/index.js
```diff
@@ -0,0 +1,58 @@
...
+    const obj = await Address.create(client, attrs)
+    return [obj, attrs]
+}
+
+async function updateTestAddress (client, id, extraAttrs = {}) {
+    if (!client) throw new Error('no client')
+    if (!id) throw new Error('no id')
+    const sender = { dv: 1, fingerprint: faker.random.alphaNumeric(8) }
+
+    // TODO(codegen): check the updateTestAddress logic for generate fields
```
pahaz, line 42: ! please fix TODO(codegen)

apps/condo/domains/property/schema/Property.js
pahaz: propertyAddress and propertyAddressMeta are not the same as addressService plugin fields. It's just cache fields with their own logic. You can't set it directly. And you will not use your plugin for that fields. It's better to remove this redundant extensibility.

apps/condo/domains/property/schema/Property.test.js
```diff
@@ -177,13 +177,14 @@ describe('Property', () => {
 
     test('user: can not create Property when same address exist in user organization', async () => {
         const client = await makeClientWithRegisteredOrganization()
-        const [, attrs] = await createTestProperty(client, client.organization)
+        const [property, attrs] = await createTestProperty(client, client.organization)
         const { address } = attrs
         await catchErrorFrom(async () => {
             await createTestProperty(client, client.organization, { address })
-        }, ({ errors }) => {
+        }, ({ errors, data }) => {
```
pahaz, line 184: it's better to use catchGQLError

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -316,6 +316,7 @@ const Ticket = new GQLListSchema('Ticket', {
                 delete: false,
             },
         },
+        //TODO(nas): Create an ability to override some fields for addressMeta field
```
pahaz, line 319: this fields has own logic and you don't need to override this fields by plugin

packages/address-service-client/README.md
```diff
@@ -0,0 +1,12 @@
...
+==
+
+The client for condo address service
+
+***
+
+## How to use:
+
+```
+const client = new AddressServiceClient(...)
```
pahaz, line 11: it's better to have some inner logic that will decide what type of client we need Fake or NonFake

packages/address-service-client/index.js
```diff
@@ -0,0 +1,26 @@
...
+
+    return instance
+}
+
+function createTestInstance (existingItem = null) {
+    // In the case of testing, we must return a new instance every time because all tests have a unique context.
+    return new MockedAddressServiceClient(existingItem)
+}
+
+module.exports = { createInstance, createTestInstance }
```
pahaz, line 26: it's better to export just one client that will decide is it test environment or not

apps/condo/domains/property/schema/Property.js
```diff
@@ -51,38 +91,6 @@ const Property = new GQLListSchema('Property', {
             isRequired: false,
         },
 
-        address: {
```
pahaz, line 54: What about other places? Resident?

packages/address-service-client/index.js
```diff
@@ -0,0 +1,26 @@
+const conf = require('@condo/config')
```
pahaz, line 1: This package is used only inside the plugin. It's better to move all this logic inside the plugin. And not to create a new package.

apps/address-service/domains/address/utils/testSchema/index.js
AleX83Xpert: done

apps/condo/domains/ticket/schema/Ticket.js
AleX83Xpert: Ok, I've deleted the comment.

packages/address-service-client/README.md
AleX83Xpert: I've fix this example

packages/address-service-client/index.js
AleX83Xpert: These clients have different constructor signatures.
The mocked client doesn't need any service URL but only some property data (required to keep tests working)

apps/condo/domains/property/schema/Property.js
AleX83Xpert: The plugin may be applied to any model. Of course, I'll add it to the `Resident` model if needed.

packages/address-service-client/index.js
AleX83Xpert: Done. Address service client was moved into plugin.

apps/address-service/domains/common/utils/services/suggest/providers/DadataSuggestionProvider.js
```diff
@@ -171,10 +171,23 @@ class DadataSuggestionProvider extends AbstractSuggestionProvider {
...
         )
 
+        //TODO?(nas): add some calls counter (maybe datadog)
+
         const status = result.status
         if (status === 200) {
             const response = await result.json()
             return get(response, 'suggestions', [])
+        } else if (status === 403) {
+            /**
```
sitozzz, line 181: Maybe return error here?

packages/keystone/plugins/utils/address-service-client/AddressServiceClient.js
```diff
@@ -0,0 +1,102 @@
...
+            context,
+        }
+    }
+
+    /**
+     * @param {AddressServiceParams} params
+     * @returns {string}
+     * @private
+     */
+    urlifyParams (params) {
```
sitozzz, line 40: Is that same with native `URLSearchParams`?
https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

apps/address-service/domains/common/utils/services/suggest/providers/DadataSuggestionProvider.js
AleX83Xpert: This may be an error about the limit being exceeded. I want to create the rotation of API keys here.

apps/address-service/domains/address/access/Address.js
```diff
@@ -0,0 +1,30 @@
...
+    if (user.isAdmin || user.isSupport) return {}
+
+    return false
+}
+
+async function canManageAddresses ({ authentication: { item: user }, originalInput, operation, itemId }) {
+    if (!user) return throwAuthenticationError()
+    if (user.deletedAt) return false
+
+    return !!user.isAdmin
```
sitozzz, line 20: What about supports?

AleX83Xpert: These models (`Address` and `AddressSource`) will be created with external data, so we restrict editing them.

apps/address-service/domains/common/utils/services/search/SearchProviderDetector.js
```diff
@@ -0,0 +1,32 @@
+const {
+    DadataSearchProvider,
+} = require('@address-service/domains/common/utils/services/search/providers')
+const { DADATA_PROVIDER, GOOGLE_PROVIDER } = require('@address-service/domains/common/constants/providers')
+
+class SearchProviderDetector {
+
+    /**
+     * @param {string} geo
```
sitozzz, line 9: Any examples for `geo` param

apps/address-service/domains/common/utils/services/search/plugins/SearchByInjectionId.js
```diff
@@ -0,0 +1,76 @@
+const { AbstractSearchPlugin } = require('@address-service/domains/common/utils/services/search/AbstractSearchPlugin')
+const { Address } = require('@address-service/domains/address/utils/serverSchema')
+const { InjectionsSeeker } = require('@address-service/domains/common/utils/services/InjectionsSeeker')
+const { AddressInjection } = require('@address-service/domains/address/utils/serverSchema')
+const get = require('lodash/get')
+const { generateAddressKey } = require('@address-service/domains/common/utils/addressKeyUtils')
+const { INJECTIONS_PROVIDER } = require('@address-service/domains/common/constants/providers')
+
+const SEPARATOR = ':'
+const UUID_REGEX = /^[0-9a-fA-F]{8}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{4}\b-[0-9a-fA-F]{12}$/
```
sitozzz, line 10: I suppose you can use `import { validate } from 'uuid'` for that

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
```diff
@@ -0,0 +1,75 @@
+const { AbstractSearchPlugin } = require('@address-service/domains/common/utils/services/search/AbstractSearchPlugin')
+const { Address } = require('@address-service/domains/address/utils/serverSchema')
+const { generateAddressKey } = require('@address-service/domains/common/utils/addressKeyUtils')
+const { SearchProviderDetector } = require('@address-service/domains/common/utils/services/search/SearchProviderDetector')
+const get = require('lodash/get')
+
+class SearchByProvider extends AbstractSearchPlugin {
+
+    /**
+     * @param s
```
sitozzz, line 10: Please provide type of this param if possible

apps/address-service/domains/common/utils/services/search/SearchProviderDetector.js
AleX83Xpert: Done.
Actually this variable influences directly the choice of the data provider. The choice may depend on geolocation, or we may force some provider by name, or use some provider for a particular city.
I don't know how to name it ¯\\\_(ツ)_/¯

apps/address-service/domains/common/utils/services/search/plugins/SearchByInjectionId.js
AleX83Xpert: Refactored. Thanks for the suggestion :)

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
AleX83Xpert: Fixed.
Anyway, jsdoc is inherited from the parent class.

sitozzz (approved):

.github/workflows/nodejs.address-service.test.yml
```diff
@@ -0,0 +1,113 @@
...
+    steps:
+      - name: Checkout code with submodules
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          submodules: recursive
+          ssh-key: ${{ secrets.SSH_DOCK_SERVER_PRIVATE_KEY }}
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 14
```
SavelevMatthew, line 38: pass node from matrix

AleX83Xpert: Done

AntonAL (commented):
Could you please add a README.md with a basic overview of this service

apps/address-service/admin-ui/index.js
```diff
@@ -1,5 +1,10 @@
 import Logo from '@app/address-service/admin-ui/logo'
+import React from 'react'
 
 export default {
     logo: Logo,
+    pages: () => {
+        window.React = React
```
AntonAL, line 7: Where it's been used?

apps/address-service/domains/address/schema/AddressSource.js
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+const { Text, Relationship } = require('@keystonejs/fields')
+const { GQLListSchema } = require('@open-condo/keystone/schema')
+const { historical, versioned, uuided, tracked, softDeleted, dvAndSender } = require('@open-condo/keystone/plugins')
+const access = require('@address-service/domains/address/access/AddressSource')
+
+
+const AddressSource = new GQLListSchema('AddressSource', {
+    schemaDoc: 'The source for the address',
```
AntonAL, line 12: It's unclear from this comment what the source is. Could you please write more detailed comment.

apps/address-service/domains/address/schema/Address.js
```diff
@@ -0,0 +1,57 @@
...
+
+const { Text, Relationship } = require('@keystonejs/fields')
+const { Json } = require('@open-condo/keystone/fields')
+const { GQLListSchema } = require('@open-condo/keystone/schema')
+const { historical, versioned, uuided, tracked, softDeleted } = require('@open-condo/keystone/plugins')
+const access = require('@address-service/domains/address/access/Address')
+const { dvAndSender } = require('@open-condo/keystone/plugins/dvAndSender')
+
+const Address = new GQLListSchema('Address', {
+    schemaDoc: 'Some specific building',
```
AntonAL, line 13: Could you bring more details into description

apps/address-service/domains/common/utils/addressKeyUtils.spec.js
```diff
@@ -0,0 +1,40 @@
...
+                    region: 'far \r  \n\t    region',
+                    area: '\t',
+                    city: 'The \t N\nCity',
+                    city_district: 'District 19',
+                    settlement: '',
+                    street: '  Straight\tstreet',
+                    house: 42,
+                    block: ' ',
+                },
+                `some${SPACE_REPLACER}country${JOINER}far${SPACE_REPLACER}region${JOINER}the${SPACE_REPLACER}n${SPACE_REPLACER}city${JOINER}district${SPACE_REPLACER}19${JOINER}straight${SPACE_REPLACER}street${JOINER}42`,
```
AntonAL, line 22: Array and join will look simpler

apps/address-service/domains/address/schema/AddressSource.js
```diff
@@ -0,0 +1,44 @@
...
+    labelResolver: ({ source }) => source,
+    fields: {
+
+        source: {
+            schemaDoc: 'Source string itself',
```
AntonAL, line 17: If I understood it correctly…

```suggestion
            schemaDoc: 'Search string, that was used to find an address string, stored in "address" field',
```

apps/address-service/domains/address/schema/AddressSource.test.js
```diff
@@ -0,0 +1,199 @@
...
+    makeClientWithSupportUser,
+} = require('@address-service/domains/user/utils/testSchema')
+
+const {
+    AddressSource,
+    createTestAddressSource,
+    updateTestAddressSource,
+} = require('@address-service/domains/address/utils/testSchema')
+
+describe('AddressSource', () => {
```
AntonAL, line 25: Could you please write tests on unique constraints

apps/address-service/domains/common/utils/addressKeyUtils.js
```diff
@@ -0,0 +1,46 @@
...
+        get(data, 'street'),
+        get(data, 'house'),
+        get(data, 'block'),
+    ]
+
+    return parts
+        // Remove empty parts
+        .filter(Boolean)
+        // Keep single space between words
+        .map(
```
AntonAL, line 31: What about filtering some punctuation characters and other, that is not belonging to words?

apps/address-service/domains/common/utils/codegeneration/generate.hooks.ts
```diff
@@ -0,0 +1,220 @@
+import { useCallback } from 'react'
```
AntonAL, line 1: Can you use `@open-condo/codegen/generate.hooks`?

apps/address-service/domains/common/utils/services/InjectionsSeeker.js
```diff
@@ -0,0 +1,190 @@
+const { AddressInjection } = require('@address-service/domains/address/utils/serverSchema')
+const get = require('lodash/get')
+const { INJECTIONS_PROVIDER } = require('@address-service/domains/common/constants/providers')
+
+const SPECIAL_SYMBOLS_REGEX = /[!@#$%^&*)(+=.\-_:;"'`[\]]/g
+
+class InjectionsSeeker {
```
AntonAL, line 7: Could you please drop a few lines about what this class does

apps/address-service/domains/common/utils/services/InjectionsSeeker.spec.js
```diff
@@ -0,0 +1,11 @@
+const { InjectionsSeeker } = require('@address-service/domains/common/utils/services/InjectionsSeeker')
+
+describe('Injections seeker', () => {
+    test('Search parts extracted correctly', async () => {
+        const injectionsSeeker = new InjectionsSeeker('Some _- +== string with a comma: , ...  Hello world!')
```
AntonAL, line 5: Why the class is named `InjectionsSeeker`, if it just extracts words from provided string?

apps/address-service/admin-ui/index.js
AleX83Xpert: This is neessary for make toast messages work within admin inteface.

apps/address-service/domains/address/schema/AddressSource.js
AleX83Xpert: Done

apps/address-service/domains/address/schema/Address.js
AleX83Xpert: Done

apps/address-service/domains/common/utils/addressKeyUtils.spec.js
AleX83Xpert: Refactored

apps/address-service/domains/address/schema/AddressSource.js
AleX83Xpert: Yes, almost right! Not address string, but address model (relation).
schemaDoc updated

apps/address-service/domains/common/utils/addressKeyUtils.js
AleX83Xpert: Done

apps/address-service/domains/common/utils/codegeneration/generate.hooks.ts
AleX83Xpert: Done

apps/address-service/domains/common/utils/services/InjectionsSeeker.js
AleX83Xpert: Done

apps/address-service/domains/common/utils/services/InjectionsSeeker.spec.js
AleX83Xpert: `InjectionsSeeker` searches for injections to inject them into the suggestions array.
This test ensures the search words are extracted correctly from the search string.
There are some tests for correct injecting (see AddressInjection.test.js)

apps/address-service/domains/address/schema/AddressSource.test.js
AleX83Xpert: Done

apps/address-service/README.md
```diff
@@ -1,3 +1,25 @@
...
+## Endpoints
+
+### /suggest
+Used to get a list of suggestions while the user trying to add a new property. There is an ability to use different suggestions providers. If some property still not exists in external providers, we may inject proper data before the answer is sent to the response. It may be reached by adding new AddressInjection model via admin interface.
+
+### /search
+Used to search addresses. This endpoint utilizes search plugins under the hood. The first not-empty plugin's result will send to the response.
+
+### /add
+Used to create new property during creating the same property in the condo database.
```
pahaz, line 20: ? why it's here

apps/address-service/domains/address/schema/AddressSource.js
```diff
@@ -0,0 +1,73 @@
...
+    },
+}
+
+const AddressSource = new GQLListSchema('AddressSource', {
+    schemaDoc: 'A model containing data on the particular building\'s address origin',
+    labelResolver: ({ source }) => source,
+    fields: {
+
+        source: {
+            schemaDoc: 'The string the address was found by (address origin)',
```
pahaz, line 29: It's not clear. is it a request query source?

apps/address-service/domains/address/schema/AddressSource.js
```diff
@@ -0,0 +1,73 @@
...
+                    const isCreate = operation === 'create'
+                    const isUpdateAddress = operation === 'update' && resolvedData.source !== existingItem.source
+
+                    if (isCreate || isUpdateAddress) {
+                        const sourceSearch = isCreate ? { source_i: value } : { source: value }
+                        const where = { ...sourceSearch, deletedAt: null }
+                        const sameAddressSourceRows = await AddressSourceApi.getAll(context, where, { first: 1 })
+
+                        if (!isEmpty(sameAddressSourceRows)) {
+                            throw new GQLError(errors.SAME_SOURCE, context)
```
pahaz, line 45: looks like you need to add a constrint

apps/address-service/domains/common/utils/services/InjectionsSeeker.spec.js
pahaz: may we you just need a function?

apps/address-service/domains/common/utils/services/search/SearchProviderDetector.js
```diff
@@ -0,0 +1,32 @@
...
+} = require('@address-service/domains/common/utils/services/search/providers')
+const { DADATA_PROVIDER, GOOGLE_PROVIDER } = require('@address-service/domains/common/constants/providers')
+
+class SearchProviderDetector {
+
+    /**
+     * @param {string} geo 'ru'|'en'|'dadata'|'google'|'<some city name>'|...
+     * @returns {AbstractSearchProvider}
+     */
+    getProvider (geo) {
```
pahaz, line 12: why not just a function?

apps/address-service/package.json
```diff
@@ -4,15 +4,15 @@
   "private": true,
   "scripts": {
     "test": "jest",
-    "dev": "keystone dev --port 3001",
-    "debug": "node inspect node_modules/.bin/keystone dev --port 3001",
+    "dev": "keystone dev",
```
pahaz, line 7: please move this service to another port

apps/condo/domains/property/schema/Property.js
```diff
@@ -283,7 +293,7 @@ const Property = new GQLListSchema('Property', {
             },
         },
     },
-    plugins: [uuided(), versioned(), tracked(), softDeleted(), dvAndSender(), historical()],
+    plugins: [uuided(), addressService('address', { address: addressFieldHooks }), versioned(), tracked(), softDeleted(), dvAndSender(), historical()],
```
pahaz, line 296: it's better to leave the 'address' field as default value

apps/condo/domains/resident/schema/Resident.js
```diff
@@ -187,6 +185,7 @@ const Resident = new GQLListSchema('Resident', {
     plugins: [
         addOrganizationFieldPlugin({ fromField: 'property' }),
         uuided(),
+        addressService('address', { address: addressFieldHooks }),
```
pahaz, line 188: we always use 'address' as field name

apps/address-service/domains/common/utils/services/InjectionsSeeker.spec.js
AleX83Xpert: Not a single function but several ones. I've decided to encapsulate them in class. It is convenient for testing too.
But yes, it's also possible to collect these functions within some module.

apps/condo/migrations/20221118163155-0189_property_addresskey_property_addresssource_and_more.js
```diff
@@ -0,0 +1,82 @@
...
+
+exports.up = async (knex) => {
+    await knex.raw(`
+    BEGIN;
+--
+-- Add field addressKey to property
+--
+ALTER TABLE "Property" ADD COLUMN "addressKey" text NULL;
```
pahaz, line 10: You need to initiate the value for that field. And add not null constraint

apps/condo/migrations/20221118163155-0189_property_addresskey_property_addresssource_and_more.js
```diff
@@ -0,0 +1,82 @@
...
+--
+-- Add field addressSource to property
+--
+ALTER TABLE "Property" ADD COLUMN "addressSource" text NULL;
```
pahaz, line 14: initial value?

apps/address-service/README.md
AleX83Xpert: I trying to make the review easier and more attractive a bit. It's to make an overall understanding of how the service works :)

packages/keystone/fields/AddressPartWithType/Implementation.js
```diff
@@ -0,0 +1,16 @@
+const { JsonImplementation } = require('../Json/Implementation')
+
+class AddressPartWithTypeImplementation extends JsonImplementation {
```
pahaz, line 3: it will be great to move this field inside the plugin. You don't need to use this field without the plugin contex

packages/keystone/plugins/addressService.js
```diff
@@ -0,0 +1,114 @@
...
+            || (operation === 'update' && !!resolvedData[addressFieldName] && existingItem[addressFieldName] !== resolvedData[addressFieldName])
+        ) {
+            const client = conf.NODE_ENV === 'test' || get(conf, 'ADDRESS_SERVICE_CLIENT_MODE') === 'fake'
+                ? createTestAddressServiceClientInstance({ ...existingItem, ...resolvedData })
+                : createAddressServiceClientInstance(get(conf, 'ADDRESS_SERVICE_URL'))
+
+            const result = await client.add({
+                source: resolvedData[addressFieldName],
+                value: resolvedData[getMetaFieldName(addressFieldName)],
+                token: 'TODO(AleX83Xpert)', // todo(AleX83Xpert)
```
pahaz, line 99: ?

apps/address-service/domains/address/schema/AddressSource.js
AleX83Xpert: Yes, this is the string sent to the external provider to search address.

packages/keystone/plugins/addressService.js
```diff
@@ -0,0 +1,114 @@
...
+    },
+    [getMetaFieldName(mainField)]: ADDRESS_META_FIELD,
+})
+
+/**
+ * Plugin to extend address field with address service
+ * @param {string?} addressFieldName
+ * @param {Object} fieldsHooks
+ */
+const addressService = (addressFieldName = 'address', fieldsHooks = {}) => plugin(({
```
pahaz, line 56: use the same argument pattern as in others plugins. Just one argument with all options

apps/address-service/domains/address/schema/AddressSource.js
AleX83Xpert: Isn't it enough?
image

packages/keystone/plugins/addressService.js
```diff
@@ -0,0 +1,114 @@
...
+    })
+
+    //
+    // Modify hooks
+    //
+    const newResolveInput = async ({ resolvedData, operation, existingItem, originalInput, context }) => {
+        // We will call to address service in the following cases:
+        if (
+            // 1. In the case of new property/resident creation
+            (operation === 'create' && !resolvedData[getKeyFieldName(addressFieldName)])
```
pahaz, line 84: it's better to read if you extract all these names in variables and will use it like `addressFieldName`

packages/keystone/plugins/addressService.js
```diff
@@ -0,0 +1,114 @@
...
+        }
+    })
+
+    //
+    // Modify hooks
+    //
+    const newResolveInput = async ({ resolvedData, operation, existingItem, originalInput, context }) => {
+        // We will call to address service in the following cases:
+        if (
+            // 1. In the case of new property/resident creation
```
pahaz, line 83: I think it should be just one case with an address field!

Then you call the address suggestion service search and save the result inside others fields.

1) createProperty({ address: `some string` }) 
2) call address suggestion service and get `addressKey`, `addressMeta`, `normalizedAddress`
3) save `some string` -> `addressSource`, `addressKey` -> `addressKey`, `addressMeta` -> `addressMeta`, `normalizedAddress` -> `address`

apps/address-service/domains/common/utils/services/search/SearchProviderDetector.js
AleX83Xpert: You right. Refactored.

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
```diff
@@ -0,0 +1,75 @@
...
+
+class SearchByProvider extends AbstractSearchPlugin {
+
+    /**
+     * @param {String} s
+     * @returns {Promise<Object[]>}
+     */
+    async search (s) {
+        const searchDetector = new SearchProviderDetector()
+        const searchProvider = searchDetector.getProvider(this.geo)
```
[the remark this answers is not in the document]
AleX83Xpert: Yep! Refactored.

apps/address-service/package.json
AleX83Xpert: I'd prefer not to set port here. Each developer may wanna use his own port number.
So, he should use `yarn dev --port <portNumber>`.

apps/condo/domains/property/schema/Property.js
AleX83Xpert: Yes, already done.
image

But I have to pass this parameter here because of 2nd parameter exists.

apps/address-service/domains/address/schema/AddressSource.test.js
```diff
@@ -0,0 +1,217 @@
...
+                })
+            })
+
+            test('anonymous can\'t', async () => {
+                await expectToThrowAuthenticationErrorToObj(async () => {
+                    await createTestAddressSource(anonymousClient)
+                })
+            })
+
+            test('Can not create two sources with the same address', async () => {
```
AntonAL, line 79: Could you please write a similar test for 'update' operation

apps/address-service/domains/common/utils/services/InjectionsSeeker.js
```diff
@@ -0,0 +1,194 @@
+const { AddressInjection } = require('@address-service/domains/address/utils/serverSchema')
+const get = require('lodash/get')
+const { INJECTIONS_PROVIDER } = require('@address-service/domains/common/constants/providers')
+
+const SPECIAL_SYMBOLS_REGEX = /[!@#$%^&*)(+=.,\-_:;"'`[\]]/g
+
+/**
+ * A class used to search injections within database.
```
AntonAL, line 8:
```suggestion
 * Searches injections within database.
```
```suggestion
 * A class used to search injections within database.
```

apps/address-service/domains/common/utils/services/InjectionsSeeker.js
```diff
@@ -0,0 +1,194 @@
...
+ * The injections are normalizing and including to the suggested results {@see SuggestionKeystoneApp}
```
AntonAL, line 9:
```suggestion
 * Injections will be included into suggested results in normalized form {@see SuggestionKeystoneApp}
```

packages/keystone/plugins/addressService.js
AleX83Xpert: Removed

apps/address-service/README.md
AleX83Xpert: The `/add` endpoint was eliminated

apps/address-service/domains/common/utils/services/InjectionsSeeker.spec.js
AntonAL: Classes nightmare is in many places of the project ;)
If a class instance lives in scope of a short function call, — it's not a use-case the class was designed for…

apps/condo/migrations/20221118163155-0189_property_addresskey_property_addresssource_and_more.js
AleX83Xpert: If we restrict `null` we will have to make some migrations.
Without restrictions, we will update these fields on demand during saving each property model. See the big triple condition at `addressService.js` plugin.

apps/condo/migrations/20221118163155-0189_property_addresskey_property_addresssource_and_more.js
AleX83Xpert: Not necessary. This field will be filled after the first saving. Anyway, we have the ability to force the saving of all models.

apps/address-service/domains/address/schema/AddressSource.js
AleX83Xpert: It's a string used to find the address

packages/keystone/plugins/addressService.js
AleX83Xpert: Done

packages/keystone/plugins/addressService.js
AleX83Xpert: Refactored

apps/address-service/domains/address/schema/AddressSource.js
```diff
@@ -0,0 +1,73 @@
...
+const { GQLError, GQLErrorCode: { BAD_USER_INPUT } } = require('@open-condo/keystone/errors')
+const { AddressSource: AddressSourceApi } = require('@address-service/domains/address/utils/serverSchema')
+const isEmpty = require('lodash/isEmpty')
+
+const errors = {
+    SAME_SOURCE: {
+        mutation: 'createAddressSource',
+        code: BAD_USER_INPUT,
+        type: SOURCE_ALREADY_EXISTS_ERROR,
+        message: 'Source with the same address already exists',
```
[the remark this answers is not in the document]
AleX83Xpert: Address service is the backend service. No translations are needed.

packages/keystone/plugins/addressService.js
AleX83Xpert: The first case is the most frequent. The others (no address key and new address) will help us update the address key field manually.

apps/address-service/domains/common/utils/services/InjectionsSeeker.js
AleX83Xpert: Class is not searching itself :) This is just a set of methods convenient to use and for testing.

apps/address-service/domains/common/utils/services/InjectionsSeeker.js
AleX83Xpert: Catch a new GitHub achievement :)

apps/address-service/domains/common/utils/services/InjectionsSeeker.spec.js
AleX83Xpert: Some classes were refactored to functions. But data providers weren't, because we use an abstraction within code.

apps/address-service/domains/address/schema/AddressSource.test.js
AleX83Xpert: Done

apps/address-service/domains/address/schema/AddressInjection.js
```diff
@@ -0,0 +1,109 @@
...
+
+        block: {
+            schemaDoc: 'Some part of the building',
+            type: AddressPartWithType,
+        },
+
+        keywords: {
+            schemaDoc: 'The autogenerated keywords for searching',
+            type: Text,
+            access: ({ operation }) => operation === 'read',
```
pahaz, line 73: use dict style. We don't need this field inside CreatInput / UpdateInput

apps/address-service/domains/address/access/AddressSource.js
```diff
@@ -0,0 +1,30 @@
...
+    if (user.isAdmin || user.isSupport) return {}
+
+    return false
+}
+
+async function canManageAddressSources ({ authentication: { item: user }, originalInput, operation, itemId }) {
+    if (!user) return throwAuthenticationError()
+    if (user.deletedAt) return false
+
+    return !!user.isAdmin
```
sitozzz, line 20: It is already boolean type. So maybe it can be done like this:
```suggestion
    return user.isAdmin
```

packages/keystone/plugins/addressService.js
```diff
@@ -0,0 +1,109 @@
...
+            (operation === 'create' && !resolvedData['addressKey'])
+
+            // 2. In the case of property/resident update and updating property/resident has no address key for some reason
+            || (operation === 'update' && !existingItem['addressKey'])
```
pahaz, line 79: ?

apps/address-service/domains/common/utils/services/search/AbstractSearchProvider.js
```diff
@@ -0,0 +1,76 @@
...
+     * Returns the context object
+     * @param {?string} context
+     * @returns {Object}
+     * @protected
+     */
+    getContext (context = null) {
+        return {
+            ...get(searchContexts, ['default', this.getProviderName()], {}),
+            ...(
+                context
```
sitozzz, line 45: May this be replaced with only lodash get expression?
Like `get(searchContexts, [context, this.getProviderName()], {})`

apps/address-service/package.json
sitozzz, line 7: Is that service become required to be up for correct work of main application? If this is true I think main service should also start address-service before itself

apps/address-service/domains/address/schema/Address.js
```diff
@@ -0,0 +1,57 @@
...
+const Address = new GQLListSchema('Address', {
+    schemaDoc: 'A model containing data on the particular building\'s address',
+    labelResolver: ({ address }) => address,
+    fields: {
+
+        sources: {
+            schemaDoc: 'The origin which used to detect the address',
+            type: Relationship,
+            ref: 'AddressSource.address',
+            many: true,
```
sitozzz, line 21: It may slowdown db queries a lot

pahaz, line 21: ! allAddress {
   source{ name id}
 }

Query complexity: `O(n * m)`

pahaz: it's better to use AddressSource (where: {...})

packages/keystone/plugins/addressService.js
```diff
@@ -0,0 +1,109 @@
...
+    })
+
+    //
+    // Modify hooks
+    //
+    const newResolveInput = async ({ resolvedData, operation, existingItem, originalInput, context }) => {
+        // We will call to address service in the following cases:
+        if (
+            // 1. In the case of new property/resident creation
+            (operation === 'create' && !resolvedData['addressKey'])
```
pahaz, line 76: looks like a bug

packages/keystone/plugins/addressService.js
```diff
@@ -0,0 +1,109 @@
...
+            // 2. In the case of property/resident update and updating property/resident has no address key for some reason
+            || (operation === 'update' && !existingItem['addressKey'])
+
+            // 3. In the case of update and if property/resident address has been changed
+            || (operation === 'update' && !!resolvedData['address'] && existingItem['address'] !== resolvedData['address'])
+        ) {
+            const client = conf.NODE_ENV === 'test' || get(conf, 'ADDRESS_SERVICE_CLIENT_MODE') === 'fake'
+                ? createTestAddressServiceClientInstance({ ...existingItem, ...resolvedData })
+                : createAddressServiceClientInstance(get(conf, 'ADDRESS_SERVICE_URL'))
+
```
pahaz, line 87: add todo to remove rawValue

apps/address-service/README.md
```diff
@@ -1,3 +1,21 @@
+# Description
+It's a separate service for address processing.
```
AntonAL, line 2: It's still not clear what it does without digging into code and what the difference between `/suggest` and `/search` API methods. Please describe for a person, that sees this first time and not read the code.

What I'm understood (correct me if something is wrong):

```suggestion
Returns detailed information about property by address search string. Uses remote data sources and local. Remote data sources are official opened APIs. Local data source is used in case when required property is not found by remote data sources or it needs to be corrected.
```

apps/address-service/README.md
```diff
@@ -1,3 +1,21 @@
...
+
+## Models
+- **Address**. A model containing data on the particular building's address.
+- **AddressSource**. A model containing data on the particular building's address origin.
+- **AddressInjection**. Addresses that do not exist in external providers.
+
+## Endpoints
+
+### /suggest
+Used to get a list of suggestions while the user trying to add a new property. There is an ability to use different suggestions providers. If some property still not exists in external providers, we may inject proper data before the answer is sent to the response. It may be reached by adding new AddressInjection model via admin interface.
```
AntonAL, line 14:
```suggestion
Used to get a list of suggestions while the user trying to add a new property. There is an ability to use different suggestions providers. If some property still not exists in external providers, we may inject proper data before the answer is sent to the response. It may be reached by adding new AddressInjection record via admin interface.
```

apps/address-service/README.md
```diff
@@ -1,3 +1,21 @@
...
+
+### /search
+Used to search addresses. This endpoint utilizes search plugins under the hood. The first not-empty plugin's result will send to the response.
```
AntonAL, line 17: What the difference between `/suggest`. Where search is performed?

apps/address-service/domains/common/utils/services/providerDetectors.js
```diff
@@ -0,0 +1,57 @@
...
+} = require('@address-service/domains/common/utils/services/suggest/providers')
+
+/**
+ * @param {string} geo 'ru'|'en'|'dadata'|'google'|'<some city name>'|...
+ * @returns {AbstractSearchProvider}
+ */
+function getSearchProvider (geo) {
+    /** @type {AbstractSearchProvider} */
+    let searchProvider
+    switch (geo) {
```
AntonAL, line 17: It's a set of provider names and countries values in variable, that by it's name determines location.

packages/keystone/errors.js
```diff
@@ -100,16 +100,16 @@ const GQLErrorCode = {
  * Error object, that can be thrown in a custom GraphQL mutation or query
  *
  * @typedef GQLError
- * @property {String} [mutation] - name of mutation where the error has been occured
- * @property {String} [query] - name of query where the error has been occured
- * @property {Array.<String>} [variable] - path to mutation or query argument, that is a subject of an error
+ * @property {String?} [mutation] - name of mutation where the error has been occured
+ * @property {String?} [query] - name of query where the error has been occured
+ * @property {Array.<String>?} [variable] - path to mutation or query argument, that is a subject of an error
```
AntonAL, line 105: It's a wrong syntax for JSDoc
Optional params are in square brackets.
Take a look at https://jsdoc.app/tags-param.html
```suggestion
 * @property {String} [mutation] - name of mutation where the error has been occured
 * @property {String} [query] - name of query where the error has been occured
 * @property {Array.<String>} [variable] - path to mutation or query argument, that is a subject of an error
```

apps/address-service/domains/common/utils/services/search/AbstractSearchPlugin.js
```diff
@@ -0,0 +1,40 @@
+/**
```
AntonAL, line 1: Could you drop a few lines about what lib/framework this plugin is made for?

packages/keystone/fields/AddressPartWithType/views/Cell.js
```diff
@@ -0,0 +1,13 @@
+/** @jsx jsx */
+import { jsx } from '@emotion/react'
+
+export default function AddressPartWithTypeCell ({ data }) {
+    const { name = '', typeShort = '', typeFull = '' } = JSON.parse(data)
+
+    return (
+        <div>
+            <div>{typeFull} · {typeShort}</div>
```
AntonAL, line 9: Do we need this extra div if both parts are empty?

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
```diff
@@ -0,0 +1,78 @@
+const { AbstractSearchPlugin } = require('@address-service/domains/common/utils/services/search/AbstractSearchPlugin')
+const { Address } = require('@address-service/domains/address/utils/serverSchema')
+const { generateAddressKey } = require('@address-service/domains/common/utils/addressKeyUtils')
+const { getSearchProvider } = require('@address-service/domains/common/utils/services/providerDetectors')
+const get = require('lodash/get')
+
+class SearchByProvider extends AbstractSearchPlugin {
```
AntonAL, line 7: Why provider extends plugin?

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @param {String} s
+     * @returns {Promise<?Object>}
+     */
+    async search (s) {
+        const searchProvider = getSearchProvider(this.geo)
+        const godContext = this.keystoneContext.sudo()
```
AntonAL, line 15: Could you please make names consistent.
```suggestion
        const sudoContext = this.keystoneContext.sudo()
```

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
```diff
@@ -0,0 +1,78 @@
...
+
+        let addressItem
+        if (addressFoundByKey) {
+            addressItem = addressFoundByKey
+            if (!addressFoundByKey.sources.map(({ source }) => source).includes(s)) {
+                addressItem = await Address.update(
+                    godContext,
+                    addressFoundByKey.id,
+                    {
+                        sources: { create: { source: s, ...dvSender } },
```
AntonAL, line 43: Don't you need to provide value for `address` field?

apps/address-service/domains/common/constants/providers.js
```diff
@@ -2,8 +2,12 @@ const DADATA_PROVIDER = 'dadata'
 const GOOGLE_PROVIDER = 'google'
 const YANDEX_PROVIDER = 'yandex'
 
+const INJECTIONS_PROVIDER = 'injections'
```
AntonAL, line 5: Looks strange…
What do you mean here by this name of injections provider?

```suggestion
const INJECTIONS_PROVIDER = 'injections'
```

apps/address-service/domains/common/utils/services/suggest/SuggestionKeystoneApp.js
```diff
@@ -56,15 +56,19 @@ class SuggestionKeystoneApp {
             }
 
             // 1. Detect the suggestion provider
-            const suggestionProvider = providerDetector.getProvider(geo)
+            const suggestionProvider = getSuggestionsProvider(geo)
```
AntonAL, line 60: Could you please also move JSDoc comments for local variables to header of the function and describe an expected shape of `req` variable. It's easier to understand what this method expects without digging into it's implementation.

AntonAL (commented):
Could you please move `AbstractSuggestionProvider` into `./providers` folder. It will make structure simpler and in provider implementations you will not need to make a step back for importing.

apps/address-service/domains/address/schema/AddressInjection.js
AleX83Xpert: Done

apps/address-service/domains/address/access/AddressSource.js
AleX83Xpert: I'd like to return `boolean` despite of is `isAdmin` undefined or not.

packages/keystone/plugins/addressService.js
```diff
@@ -0,0 +1,109 @@
...
+            const result = await client.search(get(resolvedData, ['addressMeta', 'rawValue'], resolvedData['address']))
```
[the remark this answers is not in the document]
AleX83Xpert: Yes. Done.

packages/keystone/plugins/addressService.js
AleX83Xpert: In the case of update, we check if the existing item has no `addressKey`. Looks good.

apps/address-service/domains/common/utils/services/search/AbstractSearchProvider.js
AleX83Xpert: Yes, looks good. Refactored.

apps/address-service/domains/address/schema/Address.js
AleX83Xpert: Removed

AleX83Xpert: Refactored

packages/keystone/plugins/addressService.js
AleX83Xpert: Hmmm... Fixed, yes.

packages/keystone/plugins/addressService.js
AleX83Xpert: `rawValue` was removed from the keystone plugin. But I keep sending `rawValue` within suggestions to make injections work because I need an `injectionId` as the `address` field value.

apps/address-service/README.md
AleX83Xpert: `/suggest` returns a list of suggestions got from external providers + internal injections.
`/search` should be used to get info about a single address by source (it's a cache) or some uuid.

apps/address-service/domains/common/utils/services/providerDetectors.js
AleX83Xpert: It's some value to detect the provider.

packages/keystone/errors.js
AleX83Xpert: IDE supports both types. But fixed.

apps/address-service/domains/address/schema/Address.js
```diff
@@ -0,0 +1,57 @@
...
+        },
+
+        address: {
+            schemaDoc: 'The normalized address itself in one string',
+            type: Text,
+            isRequired: true,
+        },
+
+        key: {
+            schemaDoc: 'The unique key of the address',
```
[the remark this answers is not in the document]
AleX83Xpert: It's generated by `generateAddressKey()`

apps/address-service/domains/common/utils/services/search/AbstractSearchPlugin.js
AleX83Xpert: Done

packages/keystone/fields/AddressPartWithType/views/Cell.js
AleX83Xpert: Yes, why not
image

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
AleX83Xpert: This is not a provider. This is a plugin that uses a provider.

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
AleX83Xpert: I'd like to show that this is a dangerous context. Any other developer will pay attention to it.

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
AleX83Xpert: This code do not needed anymore

apps/address-service/domains/common/constants/providers.js
AleX83Xpert: "injections provider" is an internal provider. The application inserts its data into the data got from external providers.

apps/address-service/domains/common/utils/services/suggest/SuggestionKeystoneApp.js
AleX83Xpert: Done

apps/address-service/admin-ui/index.js
AntonAL: Could you please write comments, because this is implicit logic, not directly trackable

apps/address-service/README.md
AntonAL: Could you add this into comments:)

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
AntonAL: From name it is not obvious

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
AntonAL: Maybe `sudoContext` to avoid creating many one-off words…

apps/address-service/domains/common/constants/providers.js
AntonAL: Maybe to reflect it in its name, like 'localInjections`?

sitozzz (approved):

apps/address-service/package.json
AleX83Xpert: I'll add a command if there are troubles.

apps/address-service/README.md
AleX83Xpert: ok

AntonAL (approved):

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
AleX83Xpert: I looked at the naming of keystone plugins while created these ones. Also, all plugins are situated in a separated directory. Also, base class called AbstractSearch**Plugin**

apps/address-service/domains/common/utils/services/search/plugins/SearchByProvider.js
AleX83Xpert: Windows users won't understand :)

apps/address-service/domains/common/constants/providers.js
AleX83Xpert: According to architecture `'injections'` is an internal feature.
Actually, the provider name is needed to detect where data is received from.

PR #2430, opened 2022-10-18
fix(condo): DOMA-4059 fixed GraphQlSearchInput

Alllex202:
Problem 1: GraphQlSearchInput show id while not loaded data.
Solution 1: show loading label while not loaded data.

Problem 2: not loading initial options because initialValue is empty.
Solution 2: if initialValue is empty get values from props.value.

Also:
- was fixed infinity loading stopping if all data loaded.
- the number of sent requests has been optimized. (previously, every time the search string changed, a request was sent. now the request is sent only after the user has stopped typing)

Before:
Запись экрана 2022-10-05 в 17 00 51

After:
Запись экрана 2022-10-05 в 17 02 52

### DIFF
--- a/apps/condo/domains/common/components/GraphQlSearchInput/GraphQlSearchInput.tsx
+++ b/apps/condo/domains/common/components/GraphQlSearchInput/GraphQlSearchInput.tsx
@@ -5,10 +5,13 @@ import { Select, SelectProps, Typography } from 'antd'
 import get from 'lodash/get'
 import throttle from 'lodash/throttle'
 import isFunction from 'lodash/isFunction'
+import isEmpty from 'lodash/isEmpty'
 import uniqBy from 'lodash/uniqBy'
+import debounce from 'lodash/debounce'
 
 import { ApolloClient } from '@apollo/client'
 import { useApolloClient } from '@condo/next/apollo'
+import { useIntl } from '@condo/next/intl'
 
 import {
     useTracking,
@@ -16,6 +19,7 @@ import {
     TrackingEventType,
 } from '@condo/domains/common/components/TrackingContext'
 
+import { Loader } from '../Loader'
 import { WhereType } from '../../utils/tables.utils'
 import { isNeedToLoadNewElements } from '../../utils/select.utils'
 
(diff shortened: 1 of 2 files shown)


pahaz (commented):
Is it possible to simpify the componen state? it looks to complex

Alllex202:
> Is it possible to simpify the componen state? it looks to complex

in this case, the actions will be able to interrupt each other. and that's not good

Alllex202:
Having one loading state or one state for storing data will cause conflicts because they can change from different places. By separating states, we have more control

VKislov (approved):

apps/condo/domains/common/components/GraphQlSearchInput/GraphQlSearchInput.tsx
```diff
@@ -160,40 +184,38 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
-                10,
-                skip
-            )
-            setLoading(false)
+            setMoreLoading(true)
+            const data = await search(client, value, null, 10, skip)
 
             if (data.length > 0) {
-                setData(prevData => [...prevData, ...data])
+                if (searchValue) {
```
akarjakin, line 201: This way it would be a bit readable and skip code duplication

```
const uniqData = prevData => uniqBy([...prevData, ...data], 'value')
const dataHandler = searchValue ? setSearchData : setAllData

dataHandler(uniqData)
```

apps/condo/domains/common/components/GraphQlSearchInput/GraphQlSearchInput.tsx
```diff
@@ -202,29 +224,45 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
-        if (isNeedToLoadNewElements(scrollEvent, isLoading)) {
-            await throttledSearchMore(value, data.length)
+        if (isNeedToLoadNewElements(scrollEvent, isSearchLoading)) {
+            await throttledSearchMore(searchValue, searchValue ? searchData.length : allData.length)
+        }
+    }, [isSearchLoading, throttledSearchMore, searchValue, searchData.length, allData.length])
+
+    useEffect(() => {
+        searchMoreSuggestions('', 0)
+        loadInitialOptions()
```
akarjakin, line 234: await ?

apps/condo/domains/common/components/GraphQlSearchInput/GraphQlSearchInput.tsx
```diff
@@ -202,29 +224,45 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
+            .catch(err => console.error('failed to load initial options', err))
+    }, [])
+
+    useEffect(() => {
+        if (searchValue) {
```
akarjakin, line 239:
```
const uniqData = searchValue ? uniqBy([...searchData], 'value') : uniqBy([...allData, ...initialData], 'value')

setOptions(uniqData)
```

apps/condo/domains/common/components/GraphQlSearchInput/GraphQlSearchInput.tsx
Alllex202: Effect callbacks are synchronous to prevent race conditions.

You can do something similar, but is it really necessary?
Снимок экрана 2022-10-21 в 15 14 18

Alllex202:
squash commits

apps/condo/domains/common/components/GraphQlSearchInput/GraphQlSearchInput.tsx
```diff
@@ -60,28 +64,51 @@ export const GraphQlSearchInput: React.FC<ISearchInputProps> = (props) => {
...
+        [value, needShowLoadingMessage])
+    const isDisabled = disabled || needShowLoadingMessage
+
+    const notFoundContent = useMemo(() => isInitialLoading || isSearchLoading || isLoadingMore
+        ? <Loader size='small' delay={0} fill />
+        : NotFoundMessage,
+    [NotFoundMessage, isInitialLoading, isLoadingMore, isSearchLoading])
+
+    const renderOption = useCallback((option, index?) => {
         let optionLabel = option.text
```
akarjakin, line 112:
```
const optionLabel = isFunction(formatLabel) ? formatLabel(option) : option.text
```

akarjakin (approved):

apps/condo/domains/common/components/GraphQlSearchInput/GraphQlSearchInput.tsx
Alllex202: this function is specified in the interface

PR #2434, opened 2022-10-19
DOMA-4406 access check for side menu links

VKislov:
Added access check for links in the sidebar. If an organization is not selected - all links will be visible, but all will be inactive until the user selects an organization. Some accesses are not described in the role. For example, `canReadB2BApps`. If there is no access in the role, the link will be displayed
P.S.: git displayed the diff a little awkwardly

### DIFF
--- a/apps/condo/pages/_app.tsx
+++ b/apps/condo/pages/_app.tsx
@@ -1,5 +1,6 @@
 import '@condo/domains/common/components/wdyr'
 import UseDeskWidget from '@condo/domains/common/components/UseDeskWidget'
+import get from 'lodash/get'
 
 import React, { useMemo } from 'react'
 import { ConfigProvider } from 'antd'
@@ -66,6 +67,7 @@ interface IMenuItemData {
     path: string,
     icon: React.FC,
     label: string,
+    access?: () => boolean,
 }
 
 const ANT_DEFAULT_LOCALE = enUS
@@ -77,44 +79,50 @@ const MenuItems: React.FC = () => {
     const disabled = !link || (hasSubscriptionFeature && isExpired)
     const { isCollapsed } = useLayoutContext()
     const { wrapElementIntoNoOrganizationToolTip } = useNoOrganizationToolTip()
+    const role = get(link, 'role', {})
 
-    const menuItemsData: IMenuItemData[] = [{
-        path: 'reports',
-        icon: BarChartIconNew,
-        label: 'menu.Analytics',
-    }, {
-        path: 'ticket',
-        icon: BarTicketIcon,
-        label: 'menu.ControlRoom',
-    }, {
-        path: 'property',
-        icon: BarPropertyIcon,
-        label: 'menu.Property',
-    }, {
-        path: 'contact',
-        icon: BarUserIcon,
-        label: 'menu.Contacts',
-    }, {
-        path: 'employee',
-        icon: BarEmployeeIcon,
-        label: 'menu.Employees',
-    }, {
-        path: 'billing',
-        icon: BarBillingIcon,
-        label: 'menu.Billing',
-    }, {
-        path: 'payments',
-        icon: BarPaymentsIcon,
-        label: 'menu.Payments',
-    }, {
-        path: 'meter',
-        icon: BarMeterIcon,
-        label: 'menu.Meters',
-    }, {
-        path: 'miniapps',
-        icon: BarMiniAppsIcon,
-        label: 'menu.MiniApps',
-    }]
+    const menuItemsData: IMenuItemData[] = useMemo(() => {
+        const itemsConfigs = [{
+            path: 'reports',
+            icon: BarChartIconNew,
+            label: 'menu.Analytics',
+        }, {
+            path: 'ticket',
+            icon: BarTicketIcon,
+            label: 'menu.ControlRoom',
+        }, {
+            path: 'property',
+            icon: BarPropertyIcon,
+            label: 'menu.Property',
+        }, {
+            path: 'contact',
+            icon: BarUserIcon,
+            label: 'menu.Contacts',
+        }, {
+            path: 'employee',
+            icon: BarEmployeeIcon,
+            label: 'menu.Employees',
+        }, {
+            path: 'billing',
+            icon: BarBillingIcon,
+            label: 'menu.Billing',
+            access: () => get(role, 'canReadBillingReceipts', false ),
+        }, {
+            path: 'payments',
+            icon: BarPaymentsIcon,
+            label: 'menu.Payments',
+            access: () => get(role, 'canReadPayments', false ),
+        }, {
+            path: 'meter',
+            icon: BarMeterIcon,
+            label: 'menu.Meters',
+        }, {
+            path: 'miniapps',
+            icon: BarMiniAppsIcon,
+            label: 'menu.MiniApps',
+        }]
+        return itemsConfigs.filter((item) => get(item, 'access', () => true)())
+    }, [link])
 
     return (
         <>


apps/condo/pages/_app.tsx
```diff
@@ -114,7 +100,46 @@ const MenuItems: React.FC = () => {
         path: 'miniapps',
         icon: BarMiniAppsIcon,
         label: 'menu.MiniApps',
-    }]
+    },
+]
+const accessConfig = [
```
AleX83Xpert, line 105: Am I understanding right that this array's elements order must match to items of the `listConfig` array?
Why didn't you add an `access` field to `listConfig`?

apps/condo/pages/_app.tsx
```diff
@@ -114,7 +100,46 @@ const MenuItems: React.FC = () => {
...
+    const disabled = !link || (hasSubscriptionFeature && isExpired)
+    const { isCollapsed } = useLayoutContext()
+    const { wrapElementIntoNoOrganizationToolTip } = useNoOrganizationToolTip()
+    const role = get(link, 'role', {})
+
+    const memoMenuItemsData: IMenuItemData[] = useMemo(() => {
+        const menuItemsData = []
+        let accessIndex = 0
+        for (const link of linksConfigs) {
+            get(role, `${accessConfig[accessIndex]}`, true) ? menuItemsData.push(link) : null
```
akarjakin, line 138: using ternary operator this way isn't obvious, please use `if` instead

apps/condo/pages/_app.tsx
```diff
@@ -114,7 +100,46 @@ const MenuItems: React.FC = () => {
...
+const MenuItems: React.FC = () => {
+    const { link } = useOrganization()
+    const { isExpired } = useServiceSubscriptionContext()
+    const hasSubscriptionFeature = hasFeature('subscription')
+    const disabled = !link || (hasSubscriptionFeature && isExpired)
+    const { isCollapsed } = useLayoutContext()
+    const { wrapElementIntoNoOrganizationToolTip } = useNoOrganizationToolTip()
+    const role = get(link, 'role', {})
+
+    const memoMenuItemsData: IMenuItemData[] = useMemo(() => {
```
akarjakin, line 134: memo prefix isn't needed here

apps/condo/pages/_app.tsx
```diff
@@ -114,7 +100,46 @@ const MenuItems: React.FC = () => {
...
+        const menuItemsData = []
+        let accessIndex = 0
+        for (const link of linksConfigs) {
+            get(role, `${accessConfig[accessIndex]}`, true) ? menuItemsData.push(link) : null
+            accessIndex++
+        }
+        return menuItemsData
```
akarjakin, line 141: please avoid using `let` if you can
```
return linksConfgis.reduce((link, idx) => { 
  if (get(role, `${accessConfig[accessIndex]}`, true)) result.push(link)
  
  return result
}, [])  
```

apps/condo/pages/_app.tsx
akarjakin: relying on array indexes is not safe at all

apps/condo/pages/_app.tsx
```diff
@@ -62,59 +63,78 @@ const ANT_LOCALES = {
...
-    const disabled = !link || (hasSubscriptionFeature && isExpired)
-    const { isCollapsed } = useLayoutContext()
-    const { wrapElementIntoNoOrganizationToolTip } = useNoOrganizationToolTip()
-
-    const menuItemsData: IMenuItemData[] = [{
+const itemsConfigs = {
+    canGetExternalReportIframeUrl: {
         path: 'reports',
         icon: BarChartIconNew,
         label: 'menu.Analytics',
```
AleX83Xpert, line 70: OMG O_O
I meant that you may add a new field with access to each menu item. What if we decide to use two or more "rules"?
```suggestion
    {
        path: 'reports',
        icon: BarChartIconNew,
        label: 'menu.Analytics',
        access: ['canGetExternalReportIframeUrl'],
```

apps/condo/pages/_app.tsx
```diff
@@ -62,59 +63,78 @@ const ANT_LOCALES = {
...
+    const { isExpired } = useServiceSubscriptionContext()
+    const hasSubscriptionFeature = hasFeature('subscription')
+    const disabled = !link || (hasSubscriptionFeature && isExpired)
+    const { isCollapsed } = useLayoutContext()
+    const { wrapElementIntoNoOrganizationToolTip } = useNoOrganizationToolTip()
+    const role = get(link, 'role', {})
+
+    const menuItemsData: IMenuItemData[] = useMemo(() => {
+        const result = []
+        for (const [access, itemConfig] of Object.entries(itemsConfigs)) {
```
AleX83Xpert, line 133:
```suggestion
        itemsConfigs.forEach((itemConfig) => {...})
```

apps/condo/pages/_app.tsx
```diff
@@ -62,59 +63,78 @@ const ANT_LOCALES = {
...
+            if (get(role, `${access}`, true)) result.push(itemConfig)
```
AleX83Xpert, line 134:
```suggestion
            if (get(role, access, true)) result.push(itemConfig)
```

apps/condo/pages/_app.tsx
```diff
@@ -62,59 +63,80 @@ const ANT_LOCALES = {
...
+    const { isCollapsed } = useLayoutContext()
+    const { wrapElementIntoNoOrganizationToolTip } = useNoOrganizationToolTip()
+    const role = get(link, 'role', {})
+
+    const menuItemsData: IMenuItemData[] = useMemo(() => {
+        const result = []
+        itemsConfigs.forEach((item) => {
+            if (get(role, `${get(item, 'access')}`, true)) result.push(item)
+        })
+        return result
```
AleX83Xpert, line 138: The template string isn't needed here :)
And looks like you forgot `item.access` is an array now.

Maybe something like this:
```suggestion
        return itemsConfigs.filter((item) => {
            const itemAccesses = get(item, 'access', [])
            return itemAccesses.length === 0 || itemAccesses.some((itemAccess) => !!role[itemAccess])
        })
```

AleX83Xpert (changes requested):

apps/condo/pages/_app.tsx
```diff
@@ -62,59 +63,79 @@ const ANT_LOCALES = {
...
     }, {
         path: 'payments',
         icon: BarPaymentsIcon,
         label: 'menu.Payments',
+        access: ['canReadPayments'],
     }, {
         path: 'meter',
         icon: BarMeterIcon,
         label: 'menu.Meters',
+        access: ['canReadMeters'],
```
pahaz, line 106: we don't have can Read access for models. It's not so easy. it's not a boolean

AleX83Xpert (approved):

apps/condo/pages/_app.tsx
pahaz: In general it should be a function.

AleX83Xpert: It's not about models. It's about roles and which of them can see menu items.

pahaz: We don't have such access in roles. And it's not enough to check true/false

apps/condo/pages/_app.tsx
```diff
@@ -77,44 +79,55 @@ const MenuItems: React.FC = () => {
...
+            path: 'miniapps',
+            icon: BarMiniAppsIcon,
+            label: 'menu.MiniApps',
+        }]
+        const result = []
+        itemsConfigs.filter((item) => {
+            const itemAccesses = get(item, 'access', () => true)
+            return itemAccesses() && result.push(item)
+        })
+        return result
```
AleX83Xpert, line 129: You are already filtering menu items. It's not needed to build the same array again.
```suggestion

        return itemsConfigs.filter((item) => get(item, 'access', () => true)())
```

apps/condo/pages/_app.tsx
```diff
@@ -77,44 +79,50 @@ const MenuItems: React.FC = () => {
...
+            path: 'meter',
+            icon: BarMeterIcon,
+            label: 'menu.Meters',
+        }, {
+            path: 'miniapps',
+            icon: BarMiniAppsIcon,
+            label: 'menu.MiniApps',
+        }]
+        return itemsConfigs.filter((item) => get(item, 'access', () => true)())
+    }, [link])
```
SavelevMatthew, line 125: Why link instead of role in deps?

SavelevMatthew (approved):

PR #2441, opened 2022-10-20
feat(global): DOMA-4408 added webhooks utils

SavelevMatthew:


packages/webhooks/model-validator/validator.js
```diff
@@ -0,0 +1,208 @@
...
+    registerModel (modelName) {
+        if (this.models.includes(modelName)) return
+
+        this.models.push(modelName)
+
+        const modelFilterSchema = this.generator.getSchemaForSymbol(`${modelName}WhereInput`)
+        const ajv = new Ajv({ strict: true, allErrors: true })
+        ajv.validateSchema(modelFilterSchema)
+        if (ajv.errors) {
+            throw new Error(`Schema for class ${modelName}WhereInput was not valid!`)
```
AleX83Xpert, line 41:
```suggestion
            throw new Error(`Schema for class ${modelName}WhereInput is not valid!`)
```

packages/webhooks/model-validator/validator.js
```diff
@@ -0,0 +1,208 @@
...
+     * @param {string} modelName
+     * @param {JSON} filters
+     * @returns {{isValid: boolean, errors: *[]}}
+     */
+    validateFilters (modelName, filters) {
+        if (!this.models.includes(modelName)) {
+            throw new Error(`Unregistered model name: ${modelName}`)
+        }
+        const validator = this.filterValidators[modelName]
+        const isValid = validator(filters)
```
AleX83Xpert, line 69: In the case `this.filterValidators[modelName]` is undefined or not callable, we get an exception

packages/webhooks/model-validator/validator.js
```diff
@@ -0,0 +1,208 @@
...
+     * @param modelName
+     * @param {string} modelName
+     * @param {string} fieldString
+     * @returns {{isValid: boolean, errors: string[]}|{isValid, errors}|{isValid, errors: *[]}}
+     */
+    validateFields (modelName, fieldString) {
+        if (!this.models.includes(modelName)) {
+            throw new Error(`Unregistered model name: ${modelName}`)
+        }
+        const validator = this.fieldsValidators[modelName]
```
AleX83Xpert, line 88:
```suggestion
        const validator = this.fieldsValidators[modelName]
        if (!validator) {
            throw new Error('aaaaaaa')
        }
```

packages/webhooks/model-validator/validator.js
```diff
@@ -0,0 +1,208 @@
...
+        if (error) {
+            return { isValid: false, errors: [error] }
+        } else {
+            const isValid = validator(data)
+            if (isValid) {
+                return { isValid, errors: [] }
+            } else {
+                return { isValid, errors: validator.errors }
+            }
+        }
```
sitozzz, line 99:
```js
const isValid = !error && validator(data)

if (isValid) {
    return { isValid, errors: [] }
} else {
    ...
}

```

packages/webhooks/model-validator/validator.js
```diff
@@ -0,0 +1,208 @@
...
+     * Wraps field-string in brackets if needed and also fixes spacing issues to make
+     * @param {string} fields
+     * @returns {string}
+     */
+    static normalizeFieldsString (fields) {
+        let trimmedString = fields.trim()
+        if (!trimmedString) {
+            trimmedString = '{}'
+        } else if (!trimmedString.startsWith('{')) {
+            trimmedString = '{' + trimmedString + '}'
```
AleX83Xpert, line 112:
```suggestion
            trimmedString = `{${trimmedString}}`
```

packages/webhooks/model-validator/test.schema.ts
```diff
@@ -0,0 +1,88 @@
...
+    Float: number;
+    JSON: any;
+    Upload: any;
+}
+export type File = {
+    __typename?: 'File';
+    id?: Maybe<Scalars['ID']>;
+    publicUrl?: Maybe<Scalars['String']>;
+}
+export type B2BAppAccessRight = {
```
sitozzz, line 16: Is that types hard coded here?

packages/webhooks/model-validator/test.schema.ts
```diff
@@ -0,0 +1,88 @@
...
+export type SenderField = {
+    __typename?: 'SenderField';
+    dv: Scalars['Int'];
+    fingerprint: Scalars['String'];
+}
+export enum B2BAppGlobalFeature {
+    PropertyMapGeneration = 'PropertyMapGeneration',
+}
+
+export type B2BApp = {
```
sitozzz, line 29: Why here is only B2BApp related types? Webhooks will work only for this schema?

packages/webhooks/model-validator/validator.js
```diff
@@ -0,0 +1,208 @@
...
+        if (typeof schemaPath !== 'string' || schemaPath.split('.').pop() !== 'ts') {
+            throw new Error('Schema file type was not supported. For now only .ts files are supported.')
+        } else if (!fs.existsSync(schemaPath)) {
+            throw new Error('Schema file does not exist!')
+        }
+        const program = tjs.getProgramFromFiles([schemaPath])
+        const settings = {
+            noExtraProps: true,
+        }
+        this.generator = tjs.buildGenerator(program, settings)
```
sitozzz, line 21:
```suggestion
        this.generator = tjs.buildGenerator(program, { noExtraProps: true })
```

packages/webhooks/model-validator/validator.js
```diff
@@ -0,0 +1,208 @@
...
+     * @returns {any}
+     * @private
+     */
+    static _maskSchemaTypes (schema) {
+        if (typeof schema !== 'object') {
+            return schema
+        }
+
+        // NOTE: Deep cloning JSON object
+        const newSchema = JSON.parse(JSON.stringify(schema))
```
AleX83Xpert, line 140: This manner of cloning keeps only native js values (String, Number, Boolean, ...). Is it ok?

packages/webhooks/model-validator/validator.js
```diff
@@ -0,0 +1,208 @@
...
+
+                return result
+            }
+        })
+
+        try {
+            const sample = JSON.parse(mappedParts.join(' '))
+            return { data: sample, error: null }
+        } catch (error) {
+            return { data: null, error: 'Cannot convert fields string to JSON sample object.' }
```
AleX83Xpert, line 201: The error message may be helpful.
```suggestion
            return { data: null, error: `Cannot convert fields string to JSON sample object: ${error.message}` }
```

packages/webhooks/model-validator/test.schema.ts
SavelevMatthew: It's "very-very-minified" schema for test purposes only....I've just took random model from condo schema.... 😄

SavelevMatthew: I've also manually removed some fields, since it's not necessary....(why to keep all string fields, if logic is same for all of them)

packages/webhooks/model-validator/test.schema.ts
SavelevMatthew: Nope...I've just created my own independent local schema for test purposes

packages/webhooks/model-validator/validator.js
SavelevMatthew: We assume that function input are not falsey (already preprocessed by validateInput or something like that)

packages/webhooks/model-validator/validator.js
SavelevMatthew: Validator will always be here...since it occurs on model registration

packages/webhooks/model-validator/validator.js
SavelevMatthew: This function operates on parts of JSON-schema....so it will cloned without any problems

pahaz (approved):

packages/webhooks/model-validator/test.schema.ts
SavelevMatthew: Made test schema more abstract :)

packages/webhooks/model-validator/validator.js
SavelevMatthew: Throwing error explicitly is more preferable then explicitly feeding `null` value to validator

packages/webhooks/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "keywords": [
+    "condo",
+    "open-condo-software",
+    "keystone",
+    "webhooks"
+  ],
+  "version": "0.0.0-development",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/open-condo-software/condo.git"
```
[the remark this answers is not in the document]
SavelevMatthew: Took it from ui package where it was autogenerated :)

So I guess this is fine

AleX83Xpert (approved):

sitozzz (approved):

PR #2446, opened 2022-10-21
[WIP] DOMA-4473 Initialize feature verifying contact

VKislov:
### At the moment, the feature is that the organization itself verifies contacts according to its flow and marks them manually in the user card

- [x] Added a status (verified/not verified) to the contact card as a tag in the card header
- [x] In the table, in the status column, I display whether the contact is verified or not with the same tag as in the card
- [x] There is filtering directly from the table in the drop-down filter box. Filtering is possible both only verified and only unverified
- [x] Added a checkbox for quick access, which filters only verified clients, in the header with a search input and an import button
- [x] Added the column "Verified" in the export
- [ ] Added the column "Verified" in the contact import


apps/condo/domains/contact/constants/constants.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * This is colors of ant's tags
+ * @link https://ant.design/components/tag/
+ */
+const CONTACT_IS_VERIFIED_TAGS_COLORS = {
```
pahaz, line 5: it's redundant

VKislov:
it's redundant

PR #2474, opened 2022-10-26
DOMA-4339 dont update units/parking number when delete sections/parking

VKislov:
the task is not to recalculate room numbers when removing a parking lot or an entrance. Recalculation should be carried out only in case of changes inside one of the parking / entrance


apps/condo/domains/property/components/panels/Builder/MapConstructor.ts
```diff
@@ -1461,12 +1442,6 @@ class MapEdit extends MapView {
                     firstSectionUnit.index = '1'
```
Alllex202, line 1442: Why is this part left? In units you removed it
       
        if (typeof this.map.parking[removedIndex] !== 'undefined') {
            if (removedIndex === 0) {
                // Rename from first unit
                const firstSectionUnit = get(last(this.map.parking[removedIndex].floors.filter(floor => floor.index > 0)), 'units.0')
                if (firstSectionUnit) {
                    firstSectionUnit.label = '1'
                    firstSectionUnit.index = '1'
                    this.updateParkingUnitNumbers(firstSectionUnit)
                }

Alllex202 (approved):

AleX83Xpert (approved):

PR #2478, opened 2022-10-26
hotfix(condo): DOMA-4392 fix deploy to develop

VKislov:


werf.yaml
```diff
@@ -62,6 +62,8 @@ shell:
   - "echo 'NODE_ENV=production' >> /app/.env"
   - "echo 'FILE_FIELD_ADAPTER=local' >> /app/.env"
   - "echo 'FAKE_ADDRESS_SUGGESTIONS=true' >> /app/.env"
+  - >
```
pahaz, line 65: it's better to fix it inside the code not here

Alllex202 (approved):

werf.yaml
VKislov: OK, I'll fix it this way for now, then I'll think about how I can fix it differently. It's not hard to remove

PR #2501, opened 2022-10-28
hotfix(condo): DOMA-4392 Heal condo apps

VKislov:
replace throw error on console.error


dkoviazin (approved):

dkoviazin:
The problem was that another apps now are using user schema from condo app and they do not need that env.
If someone want to throw Errors on missing envs - you need to wait until EPS and RB refactoring

Alllex202 (approved):

PR #2503, opened 2022-10-28
fix(condo): DOMA-4522 fixed timezone in tickets and meters export

Alllex202:
Problem: When exporting tickets, the time zone is not transmitted. 

Solution: `timeZoneFromUser ` --> `timeZone `

Also, callcenter app, the time in exported tickets and meters must be in the MSK time zone

### DIFF
--- a/apps/condo/domains/ticket/schema/TicketExportTask.js
+++ b/apps/condo/domains/ticket/schema/TicketExportTask.js
@@ -143,8 +143,8 @@ const TicketExportTask = new GQLListSchema('TicketExportTask', {
             isRequired: true,
             hooks: {
                 resolveInput: async ({ resolvedData }) => {
-                    const { timeZoneFromUser } = resolvedData
-                    return normalizeTimeZone(timeZoneFromUser) || DEFAULT_ORGANIZATION_TIMEZONE
+                    const { timeZone } = resolvedData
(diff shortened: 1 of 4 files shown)


apps/condo/domains/ticket/schema/TicketExportTask.js
```diff
@@ -143,8 +143,8 @@ const TicketExportTask = new GQLListSchema('TicketExportTask', {
             isRequired: true,
             hooks: {
                 resolveInput: async ({ resolvedData }) => {
-                    const { timeZoneFromUser } = resolvedData
-                    return normalizeTimeZone(timeZoneFromUser) || DEFAULT_ORGANIZATION_TIMEZONE
+                    const { timeZone } = resolvedData
+                    return normalizeTimeZone(timeZone) || DEFAULT_ORGANIZATION_TIMEZONE
```
pahaz, line 147: it's better to use the timezone and lang from the organization schema. browser timezones are not reall

Alllex202: where are we going to get it from? Organizations do not currently have such data.

Alllex202: perhaps it is worth thinking about it within the framework of another task

nomerdvadcatpyat (approved):

VKislov (approved):

PR #2507, opened 2022-10-31
DOMA-4392 fix updating number section/parking on rename

VKislov:
before: entrance/parking lot renaming took the number and set it starting from the first element.
after: The entrance/parking lot is renamed starting from the specified element and for all the following


pahaz (approved):

apps/condo/domains/property/components/panels/Builder/MapConstructor.ts
```diff
@@ -836,37 +836,39 @@ class MapEdit extends MapView {
         this.notifyUpdater()
     }
 
-    public updateSection (section: BuildingSection): void {
+    public updateSection (section: BuildingSection, renameNextUnits = true): void {
```
pahaz, line 839: why it's not a default behavior? do we really need it with `renameNextUnits = false` ?

VKislov: I don’t know :) It’s written in other parts of the map constructor, I didn’t break this rule

PR #2514, opened 2022-11-01
DOMA-4438 fix display division properties and executors

nomerdvadcatpyat:
1. Displaying the same properties and executors in divisions table (all as in the last created one)
**Problem:** Requests for m2m fields (`properties`, `executors`) are executed without `args` and get into the same value in the cache.
**Solution:** Added a parent object id (ex `Division`) to request key, thus such values will fall into different cache values.

2. Tickets are not filtered by properties in divisions.
**Solution:** Added `keyFiled` to `GQLSearchInput` so that options with the same values are not removed from the select by the `uniqBy` condition

### DIFF
--- a/apps/condo/domains/ticket/hooks/useTicketTableFilters.tsx
+++ b/apps/condo/domains/ticket/hooks/useTicketTableFilters.tsx
@@ -32,7 +32,11 @@ import {
     searchOrganizationDivision,
     searchOrganizationProperty,
 } from '../utils/clientSchema/search'
-import { getIsResidentContactFilter, getTicketAttributesFilter } from '../utils/tables.utils'
+import {
+    divisionFilterQueryToWhereProcessor,
+    getIsResidentContactFilter,
+    getTicketAttributesFilter,
+} from '../utils/tables.utils'
 import {
     FilterModalCategoryClassifierSelect,
     FilterModalPlaceClassifierSelect,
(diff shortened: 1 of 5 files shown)


VKislov (approved):

Alllex202 (approved):

apps/condo/domains/ticket/hooks/useTicketTableFilters.tsx
```diff
@@ -228,9 +228,11 @@ export function useTicketTableFilters (): Array<FiltersMeta<TicketWhereInput>>
                 keyword: 'division',
                 filters: [filterProperty],
                 queryToWhereProcessor: (queryDivisions) => {
-                    // We define single division in the browser query as "propertyId1, propertyId2" ->
-                    // in GQLWhere we need ["propertyId1", "propertyId2"]
-                    return queryDivisions?.map(queryDivision => queryDivision.split(',')).flat(1)
+                    if (queryDivisions) {
+                        return queryDivisions
+                            .map(option => JSON.parse(option).value)
```
sitozzz, line 233: What is option type here? Looks like there is a potential runtime error. May be add check for option type?

nomerdvadcatpyat: In `searchOrganizationDivision` returned option value type is arrays of string.
The behavior of the selector is such that if the option value is not a string or a number, then the entire option object is cast to a string using `JSON.Stringify`

Added a try/catch here

sitozzz: Try catch isn't a good solution) I would try to define the types more explicitly)

apps/condo/domains/ticket/hooks/useTicketTableFilters.tsx
```diff
@@ -32,7 +32,11 @@ import {
     searchOrganizationDivision,
     searchOrganizationProperty,
 } from '../utils/clientSchema/search'
-import { getIsResidentContactFilter, getTicketAttributesFilter } from '../utils/tables.utils'
+import {
+    divisionFilterQueryToWhereProcessor,
+    getIsResidentContactFilter,
+    getTicketAttributesFilter,
+} from '../utils/tables.utils'
```
sitozzz, line 39: Absolute import)

sitozzz (approved):

pahaz:
Why there is no review from @toplenboren ?

PR #2518, opened 2022-11-02
DOMA-4557 fix cyclic function call

VKislov:
**Problem**: When set `editMod` to `null` inside `removeSection`, the function `removePreviewSection` was called inside which `removeSection` was called and thus the functions called each other endlessly.
**Solution**: changed set `editMode` to set `mode`

### DIFF
--- a/apps/condo/domains/property/components/panels/Builder/MapConstructor.ts
+++ b/apps/condo/domains/property/components/panels/Builder/MapConstructor.ts
@@ -852,7 +852,8 @@ class MapEdit extends MapView {
         const removeSectionIndex = (sectionIndex - 1) >= 0 ? sectionIndex - 1 : sectionIndex
         if (renameNextUnits) this.updateSectionNumbers(removeSectionIndex, renameNextUnits)
 
-        this.editMode = null
+        this.selectedSection = null
+        this.mode = null
         this.notifyUpdater()
     }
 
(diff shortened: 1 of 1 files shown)


Alllex202 (approved):

nomerdvadcatpyat (approved):

sitozzz (approved):

VKislov:
changed the task number to the correct one

Alllex202:
> changed the task number to the correct one

Please change the pr title too

apps/condo/domains/property/components/panels/Builder/MapConstructor.ts
```diff
@@ -852,7 +852,8 @@ class MapEdit extends MapView {
         const removeSectionIndex = (sectionIndex - 1) >= 0 ? sectionIndex - 1 : sectionIndex
         if (renameNextUnits) this.updateSectionNumbers(removeSectionIndex, renameNextUnits)
 
-        this.editMode = null
```
pahaz, line 855: could you please write some tests

VKislov: apps/condo/domains/property/components/panels/Builder/MapConstructor.spec.ts:466
There are tests for this, but for some reason they do not receive this error.
This function is called in them and works correctly when setting `editMode` to `null`, and not in the test, this gives an error of the depth of the call stack

PR #2522, opened 2022-11-02
HOTFIX added number of migration into start manual migration file

VKislov:
needed so that the next migrations are generated with the correct number


AleX83Xpert (approved):

sitozzz (approved):

VKislov:
Unnecessary, since the migrator will not take the desired state. You can only change this in an autogenerated migration

Alllex202 (approved):

AntonAL (approved):

PR #2531, opened 2022-11-06
feat(condo): DOMA-4412 Webhooks tasks

SavelevMatthew:


packages/webhooks/constants.js
```diff
@@ -1,5 +1,8 @@
 const DEFAULT_MAX_PACK_SIZE = 100
+// TODO(DOMA-4570) Correct this value based on received stats, when actual skip-send logic will be implemented
+const UNAVAILABILITY_THRESHOLD = 10
```
sitozzz, line 3: Why not to get this value from environment? And if is not provided -> get this constant. It could be easier to change it in the future

SavelevMatthew: Since it's not dynamic story...it's a const. I've added mechanism of blocking hooks right now. But to set const we need to get real-usage data...to give fair const value based on real scenarios

packages/webhooks/tasks/sendWebhook.js
```diff
@@ -0,0 +1,144 @@
...
+
+        await WebhookSubscription.update(keystone, subscriptionId, {
+            syncedAt: lastSyncTime,
+            dv: 1,
+            sender: { dv: 1, fingerprint: 'sendWebhook' },
+        })
+
+        return { status: OK_STATUS }
+
+    } finally {
```
sitozzz, line 135: Is here any error when webhook is at "processing status" and lock wasn't released yet?

packages/webhooks/tasks/sendWebhook.spec.js
```diff
@@ -0,0 +1,143 @@
...
+        } else {
+            return Promise.resolve({ ok: false, status: 500 })
+        }
+    } else {
+        return Promise.resolve({ ok: false, status: 523 })
+    }
+})
+
+const { sendWebhook } = require('@condo/webhooks/tasks/sendWebhook')
+const { updateTestWebhookSubscription } = require('../schema/utils/testSchema')
```
sitozzz, line 37: absolute import)

packages/webhooks/constants.js
sitozzz: I mean not at this place) May be at place of this constant usage?

dkoviazin (approved):

packages/webhooks/tasks/sendWebhook.js
```diff
@@ -0,0 +1,144 @@
...
+                    logger.error({ message: 'Data was sent, but server response was not ok', status: response.status, subscriptionId })
+                    return { status: BAD_RESPONSE_STATUS }
+                }
+
+            } else {
+                totalLoaded += lastLoaded
+                logger.info({ message: 'Data batch was sent', subscriptionId, batchSize: lastLoaded, total: totalLoaded, syncedAt })
+            }
+
+            // Step 2: Update subscription state if full batch received. Else condition will lead to final update
```
Alllex202, line 117: Step 3?

SavelevMatthew: Fixed 👍

packages/webhooks/tasks/sendWebhook.spec.js
SavelevMatthew: removed

packages/webhooks/tasks/sendWebhook.js
SavelevMatthew: As far as I know....if 30 sec will pass, and hook is not finished, then it will throw an error. Also fetch (before I wrapped it) was throwable...So I guess this construction is safe enough

packages/webhooks/constants.js
SavelevMatthew: Added optional parameter to config

sitozzz (approved):

PR #2533, opened 2022-11-07
hotfix(condo): HOTFIX added await to return value in getClientSecret

nomerdvadcatpyat:
a promise was returned instead of the clientSecret value


VKislov (approved):

sitozzz (approved):

Alllex202 (approved):

AntonAL:
It does not fixes anything because it's used with `await` everywhere like this:

```js
await sbbolSecretStorage.getClientSecret(),
```

AntonAL (approved):

PR #2538, opened 2022-11-07
fix(condo): DOMA-4585 fixed tracking by Select

Alllex202:
Problem: All selected items in Select are sent to Amplitude. Only one item should be sent when you click on it


apps/condo/domains/common/components/antd/Select.tsx
```diff
@@ -15,26 +12,20 @@ export interface CustomSelectProps<T> extends SelectProps<T> {
...
-                selectedValue = get(option, 'title')
-            }
-
-            if (!isEmpty(selectedValue)) {
-                componentProperties['component'] = { value: selectedValue }
+            const componentValue = get(option, 'title')
 
+            if (componentValue) {
                 const componentId = get(restProps, 'id')
+                componentProperties['component'] = { value: componentValue }
```
sitozzz, line 27: This will not work correctly with multiple select. For example, at the ticket filters

Alllex202: Why? If we leave it as it was, then we will have incorrect statistics on clicks on items

**Example**:

1 click: [`option_1`]
2 click: [`option_1`, `option_2`]
3 click: [`option_1`, `option_2`, `option_3`]

If we make a selection only for `option_1`, then we will have 3 actions for this item. But in fact there was only one click on `option_1`. And the stats are wrong.

Alllex202: What useful information can we extract from these arrays?

sitozzz: > What useful information can we extract from these arrays?

Useful information here is what actually user see at the component, i.e. whatever is selected, not the instant event.
 
We do not pass an input event to the text field about each character. For this, input delay is used. 
Here we want to see the set of values that was in the value of this element at the time the event was sent.

Alllex202: Why do we need to know intermediate values?

Alllex202: This way we don't know which item was selected.

PR #2554, opened 2022-11-09
feat(condo): DOMA-4585 added analytics to MultipleFiltersModal

Alllex202:
Added analytics on the use of filters. Data is sent only when filtering is applied. 
You can specify the fields for which you need detailed statistics. (no default)

You can find out which filters are used most often.
Снимок экрана 2022-11-09 в 20 01 25

Also for some filters there are statistics on the selected values. Only those filters are taken, the values of which are unchanged (except for classifiers)
Снимок экрана 2022-11-09 в 20 01 33

### DIFF
--- a/apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
+++ b/apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
@@ -1,7 +1,7 @@
 import { SizeType } from 'antd/lib/config-provider/SizeContext'
 import React, { createContext, CSSProperties, useCallback, useContext, useMemo, useRef, useState } from 'react'
 import Form from 'antd/lib/form'
-import { Col, FormInstance, Row, Tabs, Typography, ModalProps } from 'antd'
+import { Col, FormInstance, ModalProps, Row, Tabs, Typography } from 'antd'
 import { FormItemProps } from 'antd/es'
 import { Gutter } from 'antd/es/grid/row'
 import { CloseOutlined } from '@ant-design/icons'
@@ -26,6 +26,7 @@ import Checkbox from '@condo/domains/common/components/antd/Checkbox'
 import { useOrganization } from '@condo/next/organization'
 import { OptionType, parseQuery, QueryArgType } from '@condo/domains/common/utils/tables.utils'
 import { IFilters } from '@condo/domains/ticket/utils/helpers'
+import { TrackingEventType, useTracking } from '@condo/domains/common/components/TrackingContext'
 
 import { useLayoutContext } from '../components/LayoutContext'
 import DatePicker from '../components/Pickers/DatePicker'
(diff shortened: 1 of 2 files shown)


apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
```diff
@@ -413,11 +420,15 @@ const Modal: React.FC<MultipleFiltersModalProps> = ({
...
 
-    const { selectedFiltersTemplate, setSelectedFiltersTemplate } = useMultipleFilterContext()
+    const { selectedFiltersTemplate, setSelectedFiltersTemplate, loggerConverters } = useMultipleFilterContext()
     const [openedFiltersTemplate, setOpenedFiltersTemplate] = useState(
         isEqualSelectedFiltersTemplateAndFilters(selectedFiltersTemplate, filters) ? selectedFiltersTemplate : null
     )
 
+    const { logEvent } = useTracking()
+
+    const eventName = eventNamePrefix ? `${eventNamePrefix}FilterModalClickSubmit` : ''
```
sitozzz, line 430: By default if you don't specify custom eventName, you should get auto generated one from `useTracking` hook. Method name is `getEventName`

Alllex202: It is necessary?
If this is done, then this event will have some common name and will possibly be duplicated

Alllex202: Overused default name `TicketClickIndex` (or something similar)

Alllex202: Now the logic is such that if a prefix is not specified, then there is no `eventName` and statistics will not be sent

apps/condo/domains/ticket/hooks/useTicketMultipleFilterLoggerConverters.ts
```diff
@@ -0,0 +1,82 @@
...
+import { VISIBLE_TICKET_SOURCE_TYPES } from '@condo/domains/ticket/constants/common'
+import { REVIEW_VALUES } from '@condo/domains/ticket/constants'
+import { TicketSource, TicketStatus } from '@condo/domains/ticket/utils/clientSchema'
+
+
+const getOptionLabelsByValue = (options: OptionType[], keyLabel = 'label', keyValue = 'value') => {
+    return (values: string[]) => options.filter(item => values.includes(item[keyValue])).map(item => item[keyLabel])
+}
+
+export const useTicketMultipleFilterLoggerConverters = () => {
```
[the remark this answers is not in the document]
Alllex202: I can remove the hardcoded values and send the elements as they are. But if they send ids, then when creating reports it will be difficult to find out what they mean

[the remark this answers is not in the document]
Alllex202: removed this hook

apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
Alllex202: added `getEventName`. And added event name in payload

nomerdvadcatpyat (approved):

PR #2568, opened 2022-11-14
DOMA-4632 ignore preview sections on save

sitozzz:
When you have unsaved changes the submit button will be disabled

image

### DIFF
--- a/apps/condo/domains/property/components/panels/Builder/MapConstructor.ts
+++ b/apps/condo/domains/property/components/panels/Builder/MapConstructor.ts
@@ -515,15 +515,31 @@ class MapView extends Map {
 
 
 class MapEdit extends MapView {
-    public previewSectionId: string | null = null
-    public previewParkingId: string | null = null
-    public previewSectionFloor: number | null = null
-    public previewUnitId: string | null = null
-    public previewParkingUnitId: string | null = null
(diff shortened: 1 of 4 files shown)


apps/condo/domains/property/components/panels/Builder/MapConstructor.ts
```diff
@@ -633,6 +634,14 @@ class MapEdit extends MapView {
         this.mode = mode
     }
 
+    get hasPreviewComponents (): boolean {
+        return !isNull(this.previewSectionId)
```
nomerdvadcatpyat, line 638: can `preview*Id` be undefined? Mb use isNil?

SavelevMatthew (approved):

apps/condo/domains/property/components/panels/Builder/MapConstructor.ts
sitozzz: No it can not. It initial value is null and when it become cleared it also sets to null

sitozzz: @nomerdvadcatpyat make this fields private with public getters. So it can be modified only from `MapConstructor` class methods

nomerdvadcatpyat (approved):

PR #2573, opened 2022-11-15
feat(condo): DOMA-4545 add kmigrator index support

pahaz:


bin/kmigrator.py
```diff
@@ -49,8 +49,10 @@
 NAME = '_django_model_generator'
 MODELS_TPL = """
 from django.db import models
```
sitozzz, line 51: That import doesn't seem to be needed. You have 53 line, where you are imported all from `models` package. So you can replace `models.Model` to simply `Model` call.

sitozzz (approved):

akarjakin (approved):

toplenboren (approved):

bin/kmigrator.py
pahaz: It’s required for backward compatibility

PR #2574, opened 2022-11-16
fix(condo): DOMA-4517 Fix display payments indicator logic

nomerdvadcatpyat:
Trying to find receipts at the address from the application that were created no earlier than 60 days ago.
If we find, then payments are available, if not, then the old logic is used

### DIFF
--- a/apps/condo/domains/ticket/components/TicketId/TicketResidentFeatures.tsx
+++ b/apps/condo/domains/ticket/components/TicketId/TicketResidentFeatures.tsx
@@ -1,6 +1,6 @@
 import { BillingReceipt } from '@app/condo/domains/billing/utils/clientSchema'
 import { BankCardIcon } from '@app/condo/domains/common/components/icons/BankCardIcon'
-import { Ticket } from '@app/condo/schema'
+import { SortBillingReceiptsBy, Ticket } from '@app/condo/schema'
 import { MobileIcon } from '@condo/domains/common/components/icons/MobileIcon'
 import { Loader } from '@condo/domains/common/components/Loader'
 
@@ -10,7 +10,7 @@ import { Col, Row } from 'antd'
 import { Gutter } from 'antd/es/grid/row'
 import dayjs from 'dayjs'
 import get from 'lodash/get'
-import React from 'react'
+import React, { useMemo } from 'react'
 
 interface MobileAppInstalledIndicatorProps {
     isContactHasMobileApp: boolean
@@ -30,6 +30,7 @@ const MobileAppInstalledIndicator: React.FC<MobileAppInstalledIndicatorProps> = 
 
 interface PaymentsAvailableIndicatorProps {
     ticketOrganizationId: string
+    propertyAddress: string
 }
 
 /**
@@ -50,7 +51,9 @@ const getIsPaymentsInMobileAppAvailable = (receiptsInCurrentPeriod, receiptsInPr
     return false
 }
 
-const PaymentsAvailableIndicator: React.FC<PaymentsAvailableIndicatorProps> = ({ ticketOrganizationId }) => {
+const TWO_MONTHS_AGO = dayjs().startOf('day').subtract(2, 'months').toISOString()
+
+const PaymentsAvailableIndicator: React.FC<PaymentsAvailableIndicatorProps> = ({ ticketOrganizationId, propertyAddress }) => {
     const intl = useIntl()
     const PaymentsAvailableMessage = intl.formatMessage({ id: 'pages.condo.ticket.PaymentsAvailable' })
     const PaymentsNotAvailableMessage = intl.formatMessage({ id: 'pages.condo.ticket.PaymentsNotAvailable' })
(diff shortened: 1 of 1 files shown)


apps/condo/domains/ticket/components/TicketId/TicketResidentFeatures.tsx
```diff
@@ -50,37 +51,52 @@ const getIsPaymentsInMobileAppAvailable = (receiptsInCurrentPeriod, receiptsInPr
     return false
 }
 
-const PaymentsAvailableIndicator: React.FC<PaymentsAvailableIndicatorProps> = ({ ticketOrganizationId }) => {
+const TWO_MONTHS_AGO = dayjs().startOf('day').subtract(60, 'days').toISOString()
```
sitozzz, line 54: Why not to `dayjs().startOf('day').subtract(2, 'months').toISOString()` ?

apps/condo/domains/ticket/components/TicketId/TicketResidentFeatures.tsx
```diff
@@ -50,37 +51,52 @@ const getIsPaymentsInMobileAppAvailable = (receiptsInCurrentPeriod, receiptsInPr
...
         },
     })
 
+    const { count: receiptsByProperty, loading: receiptsByPropertyLoading } = BillingReceipt.useCount({
+        where: {
+            property: {
+                OR: [
+                    {
+                        address: propertyAddress,
+                        normalizedAddress: propertyAddress,
```
sitozzz, line 86: Why there are two same values used for search by different values? Can it be done only by one of provided fields?

nomerdvadcatpyat: As long as there is no address service, the normalizedAddress field is empty.
After the address normalization logic appears, it will be possible to remove the filter by the address field and fallback to the old logic

apps/condo/domains/ticket/components/TicketId/TicketResidentFeatures.tsx
nomerdvadcatpyat: changed

apps/condo/domains/ticket/components/TicketId/TicketResidentFeatures.tsx
sitozzz: But is that work correct? Maybe it should be done like this:
```js
...
OR: [
    { address: propertyAddress, },
    { normalizedAddress: propertyAddress, },
]

```

apps/condo/domains/ticket/components/TicketId/TicketResidentFeatures.tsx
```diff
@@ -50,37 +51,52 @@ const getIsPaymentsInMobileAppAvailable = (receiptsInCurrentPeriod, receiptsInPr
...
+        },
+        sortBy: [SortBillingReceiptsBy.CreatedAtDesc],
+    })
+
     const isPaymentsAvailable = getIsPaymentsInMobileAppAvailable(receiptsInCurrentPeriod, receiptsInPreviousPeriod)
 
     return (
-        <Tooltip title={isPaymentsAvailable ? PaymentsAvailableMessage : PaymentsNotAvailableMessage}>
-            <BankCardIcon active={isPaymentsAvailable} loading={currentPeriodLoading || previousPeriodLoading} />
+        <Tooltip title={receiptsByProperty || isPaymentsAvailable ? PaymentsAvailableMessage : PaymentsNotAvailableMessage}>
```
akarjakin, line 98: could you please move calculations out of markup

apps/condo/domains/ticket/components/TicketId/TicketResidentFeatures.tsx
```diff
@@ -50,37 +51,52 @@ const getIsPaymentsInMobileAppAvailable = (receiptsInCurrentPeriod, receiptsInPr
...
+            <BankCardIcon active={isPaymentsAvailable} loading={currentPeriodLoading || previousPeriodLoading || receiptsByPropertyLoading} />
```
akarjakin, line 99: same here

apps/condo/domains/ticket/components/TicketId/TicketResidentFeatures.tsx
nomerdvadcatpyat: done

apps/condo/domains/ticket/components/TicketId/TicketResidentFeatures.tsx
```diff
@@ -50,37 +51,53 @@ const getIsPaymentsInMobileAppAvailable = (receiptsInCurrentPeriod, receiptsInPr
...
+                    { address: propertyAddress },
+                    { normalizedAddress: propertyAddress },
+                ],
+            },
+            createdAt_gte: TWO_MONTHS_AGO,
+        },
+        sortBy: [SortBillingReceiptsBy.CreatedAtDesc],
+    })
+
+    const isPaymentsAvailable = useMemo(() => getIsPaymentsInMobileAppAvailable(receiptsInCurrentPeriod, receiptsInPreviousPeriod),
```
pahaz, line 93:
```suggestion
    const isPaymentsAvailable = useMemo(
        () => getIsPaymentsInMobileAppAvailable(receiptsInCurrentPeriod, receiptsInPreviousPeriod),
```

sitozzz (approved):

Alllex202 (approved):

PR #2575, opened 2022-11-16
chore(condo): exclude cypress chai types from tsconfig

sitozzz:
Problem: cypress v10^ comes with chai types, that conflicts with jest `assert` module
Solution: exclude cypress config from app

Ref: https://github.com/cypress-io/cypress/issues/22059

### DIFF
--- a/apps/condo/tsconfig.json
+++ b/apps/condo/tsconfig.json
@@ -28,6 +28,7 @@
     "node_modules",
     ".next",
     "out",
-    "cypress/cypress.config.ts"
+    "cypress/cypress.config.ts",
+    "cypress/**/*.ts"
   ]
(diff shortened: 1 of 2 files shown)


apps/condo/tsconfig.json
```diff
@@ -27,6 +27,7 @@
   "exclude": [
     "node_modules",
     ".next",
-    "out"
+    "out",
+    "cypress/cypress.config.ts"
```
AntonAL, line 31: you are actually included this, but PR title is about to exclude

sitozzz: Fixed that

pahaz (approved):

PR #2576, opened 2022-11-16
DOMA-4665 feat custom redirect after SBBOL authComplete

VKislov:
After successful authorization through SBBOL, there will be a redirect to the tab page of our site passed in the parameters from SBBOL

### DIFF
--- a/apps/condo/domains/organization/integrations/sbbol/routes.js
+++ b/apps/condo/domains/organization/integrations/sbbol/routes.js
@@ -1,5 +1,5 @@
 const express = require('express')
-const { isObject } = require('lodash')
+const { isObject, get } = require('lodash')
 const { generators } = require('openid-client') // certified openid client will all checks
 
 const { getSchemaCtx } = require('@open-condo/keystone/schema')
@@ -10,6 +10,7 @@ const { SBBOL_SESSION_KEY } = require('./constants')
 const sync = require('./sync')
 const { getOnBoardingStatus } = require('./sync/getOnBoadringStatus')
 const { initializeSbbolAuthApi } = require('./utils')
+const { isSafeUrl } = require('@condo/domains/common/utils/url.utils')
 
 const logger = getLogger('sbbol/routes')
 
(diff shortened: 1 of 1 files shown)


VKislov:
But then such a feature will be available only for RIAS. If I pass some kind of redirectUrl as a string, then SBBOL will return it to us in the callback, but if I want to pass a similar URL for another integration, then it will have to be called something else. And it seems to be inconsistent. Perhaps it was worth taking this into account right away and waiting for the SBBOL type parameter of the type [integration_name]RedirectUrl

apps/condo/domains/organization/integrations/sbbol/routes.js
```diff
@@ -20,7 +21,7 @@ class SbbolRoutes {
 
         // nonce: to prevent several callbacks from same request
         // state: to validate user browser on callback
-        const checks = { nonce: generators.nonce(), state: generators.state() }
+        const checks = { nonce: generators.nonce(), state: generators.state(), redirectUrl: get(req, 'query.redirectUrl') }
```
pahaz, line 24: You need to check it before the save

pahaz (approved):

apps/condo/domains/organization/integrations/sbbol/routes.js
AntonAL: It should not be in `checks`, because checks are for request verification.

Use following structure:

```
req.session[SBBOL_SESSION_KEY] = { checks, redirectUrl }
```

SavelevMatthew (approved):

apps/condo/domains/organization/integrations/sbbol/routes.js
```diff
@@ -81,7 +84,7 @@ class SbbolRoutes {
             const { finished } = await getOnBoardingStatus(user)
 
             logger.info({ msg: 'SBBOL OK Authenticated', userId: user.id, organizationId: organization.id, employeeId: organizationEmployee.id, data: { finished } })
-
+            if (redirectUrl && isSafeUrl(redirectUrl)) return res.redirect(redirectUrl)
```
AntonAL, line 87: Check for valid `redirectUrl` should be performed in `startAuth` to avoid saving wrong value to session and to provide feedback as earlier, as possible.

apps/condo/domains/organization/integrations/sbbol/routes.js
```diff
@@ -21,7 +22,8 @@ class SbbolRoutes {
         // nonce: to prevent several callbacks from same request
         // state: to validate user browser on callback
         const checks = { nonce: generators.nonce(), state: generators.state() }
-        req.session[SBBOL_SESSION_KEY] = checks
+        const redirectUrl = get(req, 'query.redirectUrl')
+        req.session[SBBOL_SESSION_KEY] = { checks, redirectUrl: isSafeUrl(redirectUrl) ? redirectUrl : undefined }
```
AntonAL, line 26: It's better to throw exception if `redirectUrl` is incorrect.
Otherwise, user will have no clue that something is wrong.

VKislov:
`completeAuth` expects that its second parameter will be `checks` and they were in the session. With the addition of `redirectUrl` separately from `checks`, it was necessary to take into account the changed nesting level for `checks` and fix it in the `completeAuth` call parameters

AntonAL (approved):

PR #2577, opened 2022-11-16
fix(condo): DOMA-4679 fixed ticket deadline autoset. Removed unused code

Alllex202:
Automatically set deadline must be calculated from date ticket was created


pahaz (approved):

apps/condo/domains/ticket/components/BaseTicketForm/TicketDeadlineField.tsx
```diff
@@ -86,12 +86,14 @@ export const TicketDeadlineField = ({ initialValues, form }) => {
         } else {
             form.setFields([{ name: 'deadline', value: autoDeadlineValue }])
         }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
AntonAL, line 89: Can this rule disabling be avoided?
For example to add `autoDeadlineValue` to dependencies or by another way?

Alllex202: No, we should leave dependencies unchanged.
This logic should only be done to set initial values

nomerdvadcatpyat (approved):

PR #2579, opened 2022-11-16
feat(ui): Typography

SavelevMatthew:


packages/ui/src/components/Typography/paragraph.tsx
```diff
@@ -0,0 +1,49 @@
...
+    type?: typeof PARAGRAPH_TYPES[number]
+    size?: typeof PARAGRAPH_SIZES[number]
+}
+
+const TYPOGRAPHY_CLASS_PREFIX = 'condo-typography'
+
+const Paragraph = React.forwardRef<HTMLSpanElement, TypographyParagraphProps>((props, ref) => {
+    const { type, size = 'lg', ...rest } = props
+    const className = classNames(
+        // TODO(DOMA-4681): Remove TYPOGRAPHY_CLASS_PREFIX after next antd release including this patch: https://github.com/ant-design/ant-design/pull/38586
```
pahaz, line 26: 🥇

packages/ui/src/components/Typography/text.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import type { TextProps as DefaultTextProps } from 'antd/lib/typography/Text'
+import Base from 'antd/lib/typography/Base'
+import classNames from 'classnames'
+
+declare const TEXT_TYPES: ['secondary', 'inverted', 'danger', 'warning', 'info', 'success']
+declare const TEXT_SIZES: ['lg', 'md', 'sm']
+export type TypographyTextProps = Omit<DefaultTextProps,
```
pahaz, line 8: Some style problem

packages/ui/src/components/Typography/link.tsx
```diff
@@ -0,0 +1,44 @@
...
+    disabled?: boolean
+    ellipsis?: boolean
+    size?: typeof TEXT_SIZES[number]
+    id?: string
+    children?: React.ReactNode
+    ['aria-label']?: string
+}
+
+
+const TYPOGRAPHY_CLASS_PREFIX = 'condo-typography'
```
AntonAL, line 17: Can this constant be shared across all components, related to typography?

packages/ui/src/components/Typography/text.tsx
SavelevMatthew: Why? That's how eslint auto format

packages/ui/src/components/Typography/paragraph.tsx
SavelevMatthew: 👉 👈

packages/ui/src/components/style/core/global.less
```diff
@@ -1 +1,4 @@
+// Main font
 @import "https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,400;0,600;0,700;1,400;1,600;1,700&display=swap";
+// Monospace font is used for code blocks
+@import "https://fonts.googleapis.com/css2?family=Noto+Sans+Mono&display=swap";
```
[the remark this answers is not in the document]
SavelevMatthew: Both fonts supports cyrillic
image

packages/ui/src/components/Typography/link.tsx
SavelevMatthew: Refactored 👍

packages/ui/src/components/Typography/style.less
```diff
@@ -0,0 +1,164 @@
...
+  &-md {
+    .condo-typography-p2();
+  }
+
+  &-sm {
+    .condo-typography-p3();
+  }
+
+  &-secondary {
+    color: @condo-global-color-gray-7;
```
pahaz, line 25: Should we change the color naming to a `primary` color? to avoid text hardcode

SavelevMatthew: No, we shouldn't. We already discussed it with design team through multiple iterations. What we should do is to create component-based tokens. But for now design team does not have time for this

packages/ui/src/components/Typography/paragraph.tsx
```diff
@@ -0,0 +1,46 @@
...
+            [`${TYPOGRAPHY_CLASS_PREFIX}-${type}`]: type,
+            [`${TYPOGRAPHY_CLASS_PREFIX}-${size}`]: size,
+        }
+    )
+
+    return (
+        <Base
+            prefixCls={TYPOGRAPHY_CLASS_PREFIX}
+            ref={ref}
+            component='div'
```
Alllex202, line 35: Why does a paragraph have a `div` component and not a `p` component?

packages/ui/src/components/Typography/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const TYPOGRAPHY_CLASS_PREFIX = 'condo-typography' as const
+export const TEXT_SIZES = ['lg', 'md', 'sm'] as const
+export const TEXT_TYPES = ['secondary', 'inverted', 'danger', 'warning', 'info', 'success'] as const
```
AntonAL, line 3: What the reason to use `export … as const`, since you already declared a const?

packages/ui/src/components/Typography/link.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import Base from 'antd/lib/typography/Base'
+import classNames from 'classnames'
+import { TYPOGRAPHY_CLASS_PREFIX, TEXT_SIZES } from './constants'
+
+export type TypographyLinkProps =  Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, 'type'> & {
```
AntonAL, line 6:
```suggestion
export type TypographyLinkProps = Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, 'type'> & {
```

packages/ui/src/components/Typography/title.tsx
```diff
@@ -0,0 +1,41 @@
...
+import classNames from 'classnames'
+
+declare const TITLE_LEVELS: [1, 2, 3, 4, 5, 6]
+declare const TITLE_TYPES: ['inverted']
+export type TypographyTitleProps = Pick<DefaultTitleProps, 'ellipsis' | 'onClick' | 'title' | 'children'> & {
+    type?: typeof TITLE_TYPES[number]
+    level?: typeof TITLE_LEVELS[number]
+}
+
+const TYPOGRAPHY_CLASS_PREFIX = 'condo-typography'
```
Alllex202, line 13: The same constant is in the file `packages/ui/src/components/Typography/constants.ts`

packages/ui/src/components/Typography/constants.ts
SavelevMatthew: It's the way to tell typescript that's it is "true" const, so it can safely generate types on top of it

AntonAL: I don't see any errors from TypeScript

Tested in following way:

module 1:

```ts
export const TEST_ARRAY = [1, 2]
```

module 2:

```ts
TEST_ARRAY.push(3)
```

packages/ui/src/components/Typography/paragraph.tsx
SavelevMatthew: It's from antd source code :) 
Decided not to change it here, since it can lead to something unpredictable later

packages/ui/src/components/Typography/constants.ts
Alllex202: `as const` makes the constant read-only. 
Typescript will say that we cannot change it (including the elements in the array)

SavelevMatthew: Yep... @Alllex202 is right...And we can use read-only constants in type definitions

packages/ui/src/components/Typography/link.tsx
SavelevMatthew: Fixed 👍

Alllex202 (approved):

PR #2590, opened 2022-11-18
feat(ui): DOMA-4649 Added Markdown basic typography

SavelevMatthew:
To be done in another PR:

- [ ] tables styles
- [ ] alerts
- [ ] checkboxes

capture


packages/ui/src/components/Markdown/index.ts
```diff
@@ -0,0 +1,7 @@
+import { Markdown } from './markdown'
+import './style.less'
+
+export type { MarkdownProps } from './markdown'
```
sitozzz, line 4: I think you should remove `level` prop from this type.
For example:
```suggestion
Omit<MarkdownProps, 'level'>
```

AleX83Xpert (approved):

packages/ui/src/components/Markdown/markdown.tsx
```diff
@@ -0,0 +1,38 @@
...
+                h1: (props) => <Typography.Title {...props} level={1}/>,
+                h2: (props) => <Typography.Title {...props} level={2}/>,
+                h3: (props) => <Typography.Title {...props} level={3}/>,
+                h4: (props) => <Typography.Title {...props} level={4}/>,
+                h5: (props) => <Typography.Title {...props} level={5}/>,
+                h6: (props) => <Typography.Title {...props} level={6}/>,
+                p: (props) => <Typography.Paragraph {...props} type='secondary' />,
+                a: (props) => <Typography.Link {...props} target='_blank'/>,
+                code: (props) => <Typography.Text {...props} code/>,
+                li: ({ children, ...rest }) => <li {...rest}><Typography.Text type='secondary'>{children}</Typography.Text></li>,
```
sitozzz, line 31:
```suggestion
                li: ({ children, ...restProps }) => <li {...restProps}><Typography.Text type='secondary'>{children}</Typography.Text></li>,
```
Or just `props`. `rest` is not a good name here imho

pahaz (approved):

packages/ui/src/components/Markdown/markdown.tsx
SavelevMatthew: fixed 👍

PR #2592, opened 2022-11-19
DOMA-4632 check preview prop at property-map form validation

sitozzz:
Remove duplicated check for label uniqueness inside antd form and replaced it with unsaved changes validation


apps/condo/domains/property/components/BasePropertyMapForm/index.tsx
```diff
@@ -2,6 +2,8 @@ import React, { useState, useCallback } from 'react'
 import { Typography, Form } from 'antd'
 import Input from '@condo/domains/common/components/antd/Input'
```
akarjakin, line 3: let's sort imports

nomerdvadcatpyat (approved):

dkoviazin (approved):

PR #2598, opened 2022-11-21
fix(condo): DOMA-4531 change validation error output

Kekmus:
when user enter empty string in username field, they can see several errors at the same time
Screenshot 2022-11-21 at 17 10 02

now they will see only one error if they don't enter a name
Screenshot 2022-11-21 at 17 10 14


AntonAL:
Please, make a rebase on master

AntonAL (approved):

PR #2599, opened 2022-11-21
feat(ui): DOMA-4692 Added Modal

Alllex202:
Снимок экрана 2022-11-21 в 17 23 29

### DIFF
--- a/packages/ui/src/components/Modal/cross.tsx
+++ b/packages/ui/src/components/Modal/cross.tsx
@@ -1,0 +1,13 @@
+import React from 'react'
+
+export const Cross: React.FC<React.HtmlHTMLAttributes<HTMLDivElement>> = (props) => {
+    return (
+        <div {...props}>
+            {/*TODO (DOMA-4666): Move to icons pack*/}
+            <svg xmlns='http://www.w3.org/2000/svg' width='24' height='24' viewBox='0 0 24 24' fill='none'>
+                <path d='M8 8L16 16' stroke='#222222' strokeWidth='2' strokeLinecap='round' strokeLinejoin='round'/>
+                <path d='M16 8L8 16' stroke='#222222' strokeWidth='2' strokeLinecap='round' strokeLinejoin='round'/>
+            </svg>
+        </div>
+    )
+}
--- a/packages/ui/src/components/Modal/modal.tsx
+++ b/packages/ui/src/components/Modal/modal.tsx
@@ -1,0 +1,79 @@
+import React, { useEffect, useRef, useState } from 'react'
+import { Modal as DefaultModal, ModalProps as DefaultModalProps } from 'antd'
+import classNames from 'classnames'
+import { Typography } from '../Typography'
+import { Cross } from './cross'
+
+type CondoModalWidthType = 'small' | 'big' | 'fit-content'
+
+type CondoModalProps = {
+    title: string
+    open: boolean
+    width?: CondoModalWidthType
+}
+
+export type ModalProps = Pick<DefaultModalProps,
+'children'
+| 'footer'
+| 'className'
+| 'afterClose'
+| 'maskClosable'
+| 'destroyOnClose'
+| 'getContainer'
+| 'zIndex'
+> & CondoModalProps
+
+const MODAL_CLASS_PREFIX = 'condo-modal'
+const CONDO_MODAL_WIDTH: Readonly<Record<CondoModalWidthType, number | string>> = {
+    'small': 570,
+    'big': 900,
+    'fit-content': 'fit-content',
+}
+
+const Modal: React.FC<ModalProps> = (props) => {
+    const { children, className, title, width = 'small', footer, ...rest } = props
+    const contentChildrenRef = useRef<HTMLDivElement>(null)
+    const [isScrolling, setIsScrolling] = useState<boolean>(false)
+
+    const classes = classNames(
+        { [`${MODAL_CLASS_PREFIX}-scrolling`]: isScrolling },
+        className,
+    )
+
+    useEffect(() => {
+        // NOTE This logic allows you to find out if the content scrolls inside the Modal
+        if (!contentChildrenRef.current) return
+
+        const targetElement = contentChildrenRef.current.parentElement
+        const handleScroll = () => {
+            const scrollTop = targetElement?.scrollTop
+            setIsScrolling(Boolean(scrollTop))
+        }
+        if (contentChildrenRef.current) {
+            targetElement?.addEventListener('scroll', handleScroll)
+        }
+        return () => {
+            targetElement?.removeEventListener('scroll', handleScroll)
+        }
+    }, [])
+
+    return (
+        <DefaultModal
+            {...rest}
+            prefixCls={MODAL_CLASS_PREFIX}
+            centered
+            closable
+            className={classes}
+            closeIcon={<Cross />}
+            footer={footer || null}
+            title={<Typography.Title level={3} ellipsis={{ rows: 2 }} children={title}/>}
+            width={CONDO_MODAL_WIDTH[width]}
+            // NOTE this hack need for forwarding ref
+            children={<div ref={contentChildrenRef} children={children} className={`${MODAL_CLASS_PREFIX}-wrapper`}/>}
+        />
+    )
+}
+
+export {
+    Modal,
+}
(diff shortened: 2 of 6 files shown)


packages/ui/.storybook/main.ts
```diff
@@ -14,7 +14,12 @@ const config: StorybookConfig = {
     ],
     'addons': [
         '@storybook/addon-links',
-        '@storybook/addon-essentials',
+        {
+            name: '@storybook/addon-essentials',
+            options: {
+                docs: false,
```
SavelevMatthew, line 20: You should hide it on story level, not globally

packages/ui/src/components/Modal/cross.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+export const Cross: React.FC<React.HtmlHTMLAttributes<HTMLDivElement>> = (props) => {
+    return (
+        <div {...props}>
+            <svg xmlns='http://www.w3.org/2000/svg' width='24' height='24' viewBox='0 0 24 24' fill='none'>
+                <path d='M8 8L16 16' stroke='#222222' strokeWidth='2' strokeLinecap='round' strokeLinejoin='round'/>
```
SavelevMatthew, line 7: Left DOMA-4666 todo here

packages/ui/src/components/Modal/modal.tsx
```diff
@@ -0,0 +1,77 @@
...
+| 'getContainer'
+| 'zIndex'
+> & CondoModalProps
+
+const MODAL_CLASS_PREFIX = 'condo-modal'
+
+const Modal: React.FC<ModalProps> = (props) => {
+    // NOTE The popup should have two widths, but so far only the size of the smallest width is known
+    // Therefore, for now, change the width to any. But by default we set 570
+    const { children, className, title, width = 570, footer, ...rest } = props
```
SavelevMatthew, line 35: There should be 3 sizes as far as I know...Small, big, and adaptive...Adaptive is used in billing for example

packages/ui/src/components/Modal/modal.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const classes = classNames(
+        { [`${MODAL_CLASS_PREFIX}-scrolling`]: isScrolling },
+        className,
+    )
+
+    useEffect(() => {
+        if (!contentChildrenRef.current) return
+
+        const targetElement = contentChildrenRef.current.parentElement
+        const handler = () => {
```
SavelevMatthew, line 48: What's this?

packages/ui/src/stories/Modal.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+                <Button type='primary' children='OK' />
+            </>
+        }
+        return footer
+    }, [customFooter])
+
+    return (
+        <div>
+            <Typography.Paragraph>
+                Lorem ipsum dolor sit amet, consectetur adipisicing elit. Blanditiis consectetur deserunt dicta, dignissimos ducimus esse harum, molestias odit officia optio quaerat qui quos rerum sapiente tempora tenetur unde voluptates. Nam.
```
SavelevMatthew, line 57: Maybe add some value to text...Since it will be sent as a story snapshot

packages/ui/src/components/Modal/modal.tsx
Alllex202: This logic allows you to find out if the content scrolls inside the Modal. 
Variable name changed to `handleScroll`

packages/ui/src/components/Modal/style.less
```diff
@@ -0,0 +1,97 @@
...
+      .condo-modal-wrapper {
+        width: fit-content;
+      }
+    }
+
+    .condo-modal-footer {
+      display: flex;
+      flex-shrink: 0;
+      flex-wrap: wrap-reverse;
+      gap: @condo-modal-footer-buttons-gap;
```
SavelevMatthew, line 87: Gap is relatively new css property. ([Safari for example started supporting it only in 2021](https://caniuse.com/flexbox-gap)). So may be use margin with `:not(:last-child)`

packages/ui/src/components/Modal/modal.tsx
```diff
@@ -0,0 +1,87 @@
...
+    big = 900,
+    auto = 'auto',
+}
+
+type CondoModalProps = {
+    title: string
+    open: boolean
+    okText?: string
+    cancelText?: string
+    width?: CondoModalWidth | number | string
```
SavelevMatthew, line 19: May be limit this to only enumed value?

packages/ui/src/components/Modal/modal.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { Modal as DefaultModal, ModalProps as DefaultModalProps } from 'antd'
+import classNames from 'classnames'
+import { Typography } from '../Typography'
+import { ModalFooter } from './footer'
+import { Cross } from './cross'
+
+enum CondoModalWidth {
+    small = 570,
+    big = 900,
+    auto = 'auto',
```
SavelevMatthew, line 11: IMO something like "fit-content" or "min-content" will be more informative...

SavelevMatthew (commented):
LGTM, but I guess you should replace gap with margins if possible...

And may be add story for demo dinamic-width modal

packages/ui/src/components/Modal/footer.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { Button } from '../Button'
+import { ModalProps } from './modal'
+
+type CondoFooterProps = Pick<ModalProps, 'footer' | 'okText' | 'cancelText' | 'onOk' | 'onCancel' | 'confirmLoading'>
+
+const ModalFooter: React.FC<CondoFooterProps> = (props) => {
+    const { footer, okText, cancelText, onOk, onCancel, confirmLoading } = props
+
+    if (footer === undefined) {
```
pahaz, line 10: it looks strange. If you have a custom footer why need to use this component?

packages/ui/src/components/Modal/style.less
```diff
@@ -0,0 +1,97 @@
...
+@condo-modal-footer-buttons-gap: @condo-global-spacing-20;
+
+.condo-modal-root {
+  .condo-modal-mask {
+    background-color: @condo-global-color-gray-7;
+    opacity: 0.5;
+  }
+
+  .condo-modal {
+    max-width: calc(100vw - @condo-modal-margin-horizontal * 2);
```
pahaz, line 25: why we need to use too many new CSS features?

pahaz: Probably you need to use ANT styles and then wverride some of them

Alllex202: This is the style override.
Only we replace the prefix "ant" with the prefix "condo"

packages/ui/src/components/Modal/footer.tsx
Alllex202: Removed `ModalFooter` component. Now there is no footer by default, and you can add a custom one

packages/ui/src/components/Modal/style.less
SavelevMatthew: `calc` is old AF and `@var * x` is default less feature, so nothing criminal here 👍

packages/ui/src/components/Modal/style.less
Alllex202: Removed this css property. I left this implementation to the discretion of the developers who will add footers

packages/ui/src/components/Modal/modal.tsx
```diff
@@ -0,0 +1,81 @@
...
+| 'destroyOnClose'
+| 'getContainer'
+| 'zIndex'
+> & CondoModalProps
+
+const MODAL_CLASS_PREFIX = 'condo-modal'
+
+const Modal: React.FC<ModalProps> = (props) => {
+    // NOTE The popup has three widths ('small', 'big', 'auto')
+    // But it also remains possible to set a custom value
```
SavelevMatthew, line 34: It should not be possible. Let's just set it to only 3 values

Alllex202: oops, this note should be removed)

SavelevMatthew (approved):

PR #2606, opened 2022-11-22
DOMA-4668 add popupsmart script to main app

sitozzz:
I will add configs at another PR

### DIFF
--- a/apps/condo/domains/common/components/PopupSmart.tsx
+++ b/apps/condo/domains/common/components/PopupSmart.tsx
@@ -1,0 +1,27 @@
+import React, { useEffect } from 'react'
+import get from 'lodash/get'
+import cookie from 'js-cookie'
+import getConfig from 'next/config'
+import { useOrganization } from '@open-condo/next/organization'
+
+const PopupSmart = (): React.ReactElement | null => {
+    const { link } = useOrganization()
+    const role = get(link, 'role.nameNonLocalized')
+
+    const { publicRuntimeConfig: { popupSmartUrl } } = getConfig()
+
+    // watch for user role update. This is the way to tell popupsmart local user context
+    useEffect(() => {
+        if (role) {
+            cookie.set('roleName', role)
+        } else {
+            cookie.remove('roleName')
+        }
+    }, [role])
+
+    return popupSmartUrl
+        ? <script type='text/javascript' src={popupSmartUrl} async></script>
+        : null
+}
+
+export default PopupSmart
--- a/apps/condo/domains/common/components/PopupSmart.tsx
+++ b/apps/condo/domains/common/components/PopupSmart.tsx
@@ -19,24 +19,26 @@ const PopupSmart = (): null => {
 
     // watch for user role update. This is the way to tell popupsmart local user context
     useEffect(() => {
-        // We don't want to share support or admin user's
-        if (role && userIsNotStaff && popupSmartUrl) {
-            cookie.set('roleName', role)
+        if (typeof window !== 'undefined') {
+            // We don't want to share support or admin user's
+            if (role && userIsNotStaff && popupSmartUrl) {
+                cookie.set('roleName', role)
 
-            // If script is not loaded yet -> add it to the body section
-            // Placing script in render section not fire it's loading. That was fixed at nextjs v11 with next/script
-            if (isNull(scriptRef.current)) {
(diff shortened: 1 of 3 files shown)


pahaz (approved):

apps/condo/domains/common/components/PopupSmart.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const PopupSmart = (): React.ReactElement | null => {
+    const { link } = useOrganization()
+    const role = get(link, 'role.nameNonLocalized')
+
+    const { publicRuntimeConfig: { popupSmartUrl } } = getConfig()
+
+    // watch for user role update. This is the way to tell popupsmart local user context
+    useEffect(() => {
+        if (role) {
```
pahaz, line 15: Maybe there is a race condition. it may be already loaded

apps/condo/domains/common/components/PopupSmart.tsx
```diff
@@ -0,0 +1,27 @@
...
+            cookie.set('roleName', role)
+        } else {
+            cookie.remove('roleName')
+        }
+    }, [role])
+
+    return popupSmartUrl
```
pahaz, line 22: Could you please return null for isSupport and isAdmin users for security reason

sitozzz: Done this

apps/condo/domains/common/components/PopupSmart.tsx
sitozzz: I don't quite understand what is the race condition about?

pahaz: The script can be loaded before you set the cookies. But you already fixit

Alllex202 (approved):

apps/condo/domains/common/components/PopupSmart.tsx
```diff
@@ -0,0 +1,46 @@
...
+    // watch for user role update. This is the way to tell popupsmart local user context
+    useEffect(() => {
+        // We don't want to share support or admin user's
+        if (role && userIsNotStaff && popupSmartUrl) {
+            cookie.set('roleName', role)
+
+            // If script is not loaded yet -> add it to the body section
+            // Placing script in render section not fire it's loading. That was fixed at nextjs v11 with next/script
+            if (isNull(scriptRef.current)) {
+                const scriptElement = document.createElement('script')
```
SavelevMatthew, line 29: Please check `document` for availability. Seems like not important, because useEffect is not launched in SSR for now (but who knows 😃  )

apps/condo/domains/common/components/PopupSmart.tsx
```diff
@@ -0,0 +1,46 @@
...
+                scriptElement.async = false
+                scriptElement.src = popupSmartUrl
```
SavelevMatthew, line 31: Does `next@9` have this: https://nextjs.org/docs/api-reference/next/script 
If so, may be we can avoid manual document modification

sitozzz: No, that included from version > 11

apps/condo/domains/common/components/PopupSmart.tsx
sitozzz: Done this

apps/condo/domains/common/components/PopupSmart.tsx
```diff
@@ -0,0 +1,48 @@
...
+                // If script is not loaded yet -> add it to the body section
+                // Placing script in render section not fire it's loading. That was fixed at nextjs v11 with next/script
+                if (isNull(scriptRef.current)) {
+                    const scriptElement = document.createElement('script')
+                    scriptElement.async = false
+                    scriptElement.src = popupSmartUrl
+
+                    document.querySelector('body').appendChild(scriptElement)
+
+                    scriptRef.current = scriptElement
```
AntonAL, line 36: Do you need to delete script and remove PopupSmart objects if `PopupSmart` React-component will be unmounted?

sitozzz: No, because if we have that scenario:
User logged in and choose organization -> script added to the document and loaded locally.
Next user wants to change account without closing tab
So if we remove script -> we need to load it again. I think that is not really good

PR #2610, opened 2022-11-24
fix(condo): DOMA-4703 fix ticket close task

Alllex202:
Problem: Mobile apps can take a long time to download a large number of updated tickets.

Solution: Therefore, you should limit updating tickets for each organization at one time if there are a lot of them.
With the help of the feature flag, you can manage this restriction.

Also:  added the ability to get the value of the feature flag

### DIFF
--- a/apps/condo/domains/ticket/tasks/closeCompletedTickets.js
+++ b/apps/condo/domains/ticket/tasks/closeCompletedTickets.js
@@ -8,27 +8,35 @@ const { featureToggleManager } = require('@open-condo/featureflags/featureToggle
 
 const { STATUS_IDS } = require('@condo/domains/ticket/constants/statusTransitions')
 const { Ticket } = require('@condo/domains/ticket/utils/serverSchema')
-const { MAX_COUNT_COMPLETED_TICKET_TO_CLOSE_TASK } = require('@condo/domains/common/constants/featureflags')
+const { MAX_COUNT_COMPLETED_TICKET_TO_CLOSE_FOR_ORGANIZATION_TASK } = require('@condo/domains/common/constants/featureflags')
 
 const CHUNK_SIZE = 50
 
 const appLogger = getLogger('condo')
 const taskLogger = appLogger.child({ module: 'closeCompletedTickets' })
 
-// NOTE Mobile apps can take a long time to download a large number of updated tickets.
-// Therefore, you should limit updating tickets for each organization at one time if there are a lot of them.
-// With the help of the feature flag, you can manage this restriction.
-const closeCompletedTicketsWithLimitByOrganizations = async (limit = 100) => {
-    if (!isNumber(limit) || limit < 1) {
+/**
+ * Closes tickets that are in the "completed" status for 7 days
+ */
+const closeCompletedTickets = async (defaultLimit = 100) => {
+    const { keystone } = await getSchemaCtx('Ticket')
+    const context = await keystone.createContext()
+
+    // NOTE Mobile apps can take a long time to download a large number of updated tickets.
+    // Therefore, you should limit updating tickets for each organization at one time if there are a lot of them.
+    // With the help of the feature flag, you can manage this restriction.
+    const limitByOrganization = await featureToggleManager.getFeatureValue(
+        context, MAX_COUNT_COMPLETED_TICKET_TO_CLOSE_FOR_ORGANIZATION_TASK, defaultLimit
+    )
+
+    if (!isNumber(limitByOrganization) || limitByOrganization < 1) {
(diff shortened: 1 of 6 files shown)


packages/featureflags/featureToggleManager.js
```diff
@@ -72,6 +75,22 @@ class FeatureToggleManager {
 
         return growthbook.isOn(featureName)
     }
+
+    async getFeatureValue (keystoneContext, featureName, defaultValue, featuresContext) {
```
sitozzz, line 79: Why this method was marked with `async`? There is no any `await` keyword used

packages/featureflags/featureToggleManager.js
```diff
@@ -53,11 +53,14 @@ class FeatureToggleManager {
...
-    async isFeatureEnabled (keystoneContext, featureName, featuresContext) {
-        const context = get(keystoneContext, 'req', null)
+    async #getContext (keystoneContext) {
+        return get(keystoneContext, 'req', null)
             ? keystoneContext
             : { req: { features: await this.fetchFeatures() } }
+    }
 
+    async isFeatureEnabled (keystoneContext, featureName, featuresContext) {
+        const context = this.#getContext(keystoneContext)
```
sitozzz, line 63: You change `getContext` method to async. So you need to add await here
```suggestion
        const context = await this.#getContext(keystoneContext)
```

packages/featureflags/featureToggleManager.js
```diff
@@ -72,6 +75,22 @@ class FeatureToggleManager {
 
         return growthbook.isOn(featureName)
     }
+
+    async getFeatureValue (keystoneContext, featureName, defaultValue, featuresContext) {
+        const context = this.#getContext(keystoneContext)
```
sitozzz, line 80: As far as I known this should be also done with await
```suggestion
        const context = await this.#getContext(keystoneContext)
```

packages/featureflags/FeatureFlagsContext.tsx
```diff
@@ -10,6 +10,7 @@ const FEATURES_RE_FETCH_INTERVAL = 10 * 1000
 
 interface IFeatureFlagsContext {
     useFlag: (name: string) => boolean,
+    useFlagValue: (name: string) => any,
```
sitozzz, line 13: Maybe provide all available types from growthbook docs? I think there are `number`, `string` and `boolean` and json (for typescript it could be `Record<string, unknown>`

sitozzz:
```suggestion
    useFlagValue: (name: string) => number | string | boolean | Record<string, unknown>,
```

Alllex202: I specified the same type as in the `growthbook` library
```suggestion
    useFlagValue: <T extends JSONValue = any>(name: string) => T | null,
```

nomerdvadcatpyat (approved):

sitozzz (approved):

apps/condo/domains/ticket/tasks/closeCompletedTickets.js
```diff
@@ -1,46 +1,103 @@
...
+    const context = await keystone.createContext()
+
+    // NOTE Mobile apps can take a long time to download a large number of updated tickets.
+    // Therefore, you should limit updating tickets for each organization at one time if there are a lot of them.
+    // With the help of the feature flag, you can manage this restriction.
+    const limitByOrganization = await featureToggleManager.getFeatureValue(
+        context, MAX_COUNT_COMPLETED_TICKET_TO_CLOSE_FOR_ORGANIZATION_TASK, defaultLimit
+    )
+
+    if (!isNumber(limitByOrganization) || limitByOrganization < 1) {
```
pahaz, line 32: you should rase an error in that case!

apps/condo/domains/ticket/tasks/closeCompletedTickets.js
```diff
@@ -1,46 +1,103 @@
...
+                    error,
+                })
+            }
         }
     }
 }
 
 module.exports = {
-    closeCompletedTickets: createCronTask('closeCompletedTickets', '0 1 * * *', closeCompletedTickets),
+    closeCompletedTicketsCron: createCronTask('closeCompletedTickets', '0 1 * * *', closeCompletedTickets),
```
pahaz, line 101: You can call this task not only from corn. You don't need to rename it

Alllex202: but we call `createCronTask`...

We do separate `closeCompletedTickets ` function import to test the it.

PR #2612, opened 2022-11-24
DOMA-4762 PostMessage analytics

SavelevMatthew:
<details>
<summary> <h3>Analytic event from main application</h3> </summary>

image

</details>
<details>
<summary> <h3>Analytic event from child application</h3> </summary>

image

</details>

### DIFF
--- a/apps/condo/domains/common/components/TrackingContext.tsx
+++ b/apps/condo/domains/common/components/TrackingContext.tsx
@@ -47,6 +47,8 @@ const useTrackingContext = (): ITrackingContext => useContext<ITrackingContext>(
 
 const DETAIL_PAGE_NAMES = ['create', 'update', 'hint', 'detail', 'forgot', 'signin', 'register', 'change-password']
 
+const POST_MESSAGE_HANDLER_NAME = 'CondoWebSendAnalyticsEvent'
+
 export enum TrackingEventType {
     Visit = 'Visit',
     Click = 'Click',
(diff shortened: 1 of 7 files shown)


packages/ui/src/components/Button/button.tsx
```diff
@@ -28,13 +30,24 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>((props, ref) =>
         ? <span className={`${BUTTON_CLASS_PREFIX}-icon`}>{icon}</span>
         : null
 
+    const handleClick = useCallback((event) => {
+        if (isString(children)) {
+            sendAnalyticsClickEvent('Button', { value: children })
```
sitozzz, line 35: What about buttons with icons and text? Or buttons with just icon as a children?

SavelevMatthew: Icon are passed with Icon property

SavelevMatthew: And for button without a text, I don't really know, how to generate name dynamically

apps/condo/domains/common/components/TrackingContext.tsx
```diff
@@ -47,6 +47,8 @@ const useTrackingContext = (): ITrackingContext => useContext<ITrackingContext>(
 
 const DETAIL_PAGE_NAMES = ['create', 'update', 'hint', 'detail', 'forgot', 'signin', 'register', 'change-password']
 
+const POST_MESSAGE_HANDLER_NAME = 'CondoWebSendAnalyticsEvent'
```
sitozzz, line 50:
```suggestion
const POST_MESSAGE_HANDLER_NAME = 'CondoWebSendTrackingEvent'
```

apps/condo/domains/common/components/TrackingContext.tsx
```diff
@@ -76,9 +116,8 @@ interface IUseTracking {
 }
 
 const useTracking: IUseTracking = () => {
-    const router = useRouter()
-    const route = get(router, 'route', '')
     const { trackerInstances, eventProperties, userProperties } = useTrackingContext()
+    const { route } = useRouter()
```
sitozzz, line 120: Is `route` be defined on SSR? If not it could be runtime error with splitting `undefined`

SavelevMatthew: image

So, next typing says, that route is always defined :)
And it's a bit strange to expect it to be null/undefined in SSR from SSR framework 😄

sitozzz (approved):

AleX83Xpert (approved):

PR #2631, opened 2022-11-29
DOMA-4067 fix contact duplicates

nomerdvadcatpyat:
It is necessary to make it so that it is impossible to create contacts with the same phone numbers on one unitName.
Made a constraint on the uniqueness of property, unitName, unitType and phone.
Wrote a migration that removes duplicate contacts and relinks contacts in tickets and meterReadings


apps/condo/domains/ticket/schema/CreateResidentTicketService.test.js
```diff
@@ -110,25 +110,6 @@ describe('CreateResidentTicketService', () => {
         expect(contact.id).toMatch(UUID_RE)
     })
 
-    test('resident: do not create contact when create resident ticket with contact with same data', async () => {
```
pahaz, line 113: why you want to remove this test?

apps/condo/domains/contact/schema/Contact.js
```diff
@@ -159,6 +158,16 @@ const Contact = new GQLListSchema('Contact', {
             }
         },
     },
+    kmigratorOptions: {
+        constraints: [
+            {
+                type: 'models.UniqueConstraint',
+                fields: ['property', 'unitName', 'unitType', 'phone'],
```
pahaz, line 165: probably, we need to add some test for that case

pahaz (approved):

apps/condo/domains/ticket/schema/CreateResidentTicketService.test.js
nomerdvadcatpyat: `CreateResidentTicket` is a deprecated service, I removed it in another pr

apps/condo/domains/contact/schema/Contact.js
nomerdvadcatpyat: we have such tests in `Contact.test.js` in `unique constraint` describe block

akarjakin (approved):

PR #2633, opened 2022-11-29
fix(condo): DOMA-3203 fixed default filters on second click to /payme…

abshnko:
…nts link in side bar
Screenshot 2022-11-29 at 20 31 45

### DIFF
--- a/apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
+++ b/apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
@@ -1,4 +1,8 @@
+import React, { useEffect, useState, useCallback } from 'react'
 import { FilterFilled } from '@ant-design/icons'
+import { Col, Row, Space, Typography } from 'antd'
+import { useIntl } from '@open-condo/next/intl'
+import { useOrganization } from '@open-condo/next/organization'
 import { useQuery } from '@open-condo/next/apollo'
 import { BillingIntegrationOrganizationContext, SortPaymentsBy } from '@app/condo/schema'
 import { PAYMENT_DONE_STATUS, PAYMENT_WITHDRAWN_STATUS } from '@condo/domains/acquiring/constants/payment'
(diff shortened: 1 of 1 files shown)


apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
```diff
@@ -232,7 +246,7 @@ const PaymentsTableContent: React.FC<IPaymentsTableProps> = ({ billingContext, c
                                 <Row justify='end' align='middle'>
                                     {
                                         appliedFiltersCount > 0 && (
-                                            <Col>
+                                            <Col onClick={() => filtersAreReset = true }>
```
AleX83Xpert, line 249: It's better to use the click handler of the reset button. You may try to wrap `handleResetSearch` into your own function.

akarjakin, line 249: useCallback

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
```diff
@@ -200,6 +201,19 @@ const PaymentsTableContent: React.FC<IPaymentsTableProps> = ({ billingContext, c
...
+    **/
+    useEffect(()=>{
+        if (hasFilters){
+            filtersAreReset = false
+        }
+        if (!hasFilters && !filtersAreReset) {
+            isDefaultFilterApplied = true
+            setDateRange(DEFAULT_DATE_RANGE)
+        }
+    }, [hasFilters])
```
akarjakin, line 215: filtersAreReset

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
```diff
@@ -200,6 +201,19 @@ const PaymentsTableContent: React.FC<IPaymentsTableProps> = ({ billingContext, c
         }
     }, [])
 
+    /**
+    * set default filters only when filters are empty and reset button hasn't been pressed
+    **/
+    useEffect(()=>{
+        if (hasFilters){
+            filtersAreReset = false
```
akarjakin, line 209: using variables of outer scoupe can cause hard to find misbehaviour if you use several instances of this component - they will affect each other. please use this variable inside using setState

akarjakin (changes requested):

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
sitozzz, line 209: If you don't want to trigger rerender here you may use react ref as a value storage

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
abshnko: tried it, but because `handleResetSearch` is awaited, `useEffect` triggers first with no filters and still thinking that `filtersAreReset = false`. so it applies default filters, which shouldn’t be the case

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
abshnko: I only want this `useEffect` to run when `hasFilters` changed

apps/condo/domains/acquiring/components/payments/PaymentsTable.tsx
abshnko: thank you, fixed

AleX83Xpert: Ok, I've understood. Appropriate as a quick decision ✅.

AleX83Xpert (approved):

akarjakin (approved):

PR #2637, opened 2022-11-30
feat(condo): DOMA-4798 implemented `get` command for sbbol/credentials.js util

AntonAL:
Used for inspection of stored values in case when there is no direct access to Redis


apps/condo/bin/sbbol/credentials.js
```diff
@@ -51,6 +55,12 @@ const workerJob = async () => {
         await changeClientSecret({ clientId, currentClientSecret, newClientSecret })
     }
 
+    if (command === COMMAND.GET) {
+        const sbbolSecretStorage = getSbbolSecretStorage()
+        const values = await sbbolSecretStorage.getRawKeyValues()
+        console.log('SbbolSecretStorage values: ', JSON.stringify(values, null, 2))
```
akarjakin, line 61: console.log ?

akarjakin (approved):

apps/condo/bin/sbbol/credentials.js
sitozzz: If that logs are important you can change this to `logger.info`

sitozzz (approved):

apps/condo/bin/sbbol/credentials.js
AntonAL: This util will be invoked manually from console.

PR #2641, opened 2022-11-30
[MINIAPPS: pt1] Fields migration

SavelevMatthew:


apps/condo/bin/generate-miniapps.js
```diff
@@ -0,0 +1,89 @@
+const path = require('path')
```
toplenboren, line 1: Could you please add documentation and usage examples?

toplenboren (approved):

apps/condo/bin/generate-miniapps.js
SavelevMatthew: I'll do...but a bit later

PR #2649, opened 2022-12-02
feat(condo): DOMA-4708 Added alerts to ui kit

Kekmus:
Screenshot 2022-12-09 at 14 01 14

### DIFF
--- a/packages/ui/src/components/Alert/alert.tsx
+++ b/packages/ui/src/components/Alert/alert.tsx
@@ -1,0 +1,35 @@
+import React from 'react'
+import { Alert as DefaultAlert, AlertProps as DefaultAlertProps } from 'antd'
+import classNames from 'classnames'
+
+type CondoAlertProps = {
+    type: 'success' | 'info' | 'warning' | 'error';
+}
+
+const ALERT_CLASS_PREFIX = 'condo-alert'
+
+export type AlertProps = Omit<DefaultAlertProps, 'type' | 'prefix' | 'prefixCls' | 'banner' | 'afterClose' | 'closable' | 'closeText' | 'closeIcon' | 'onClose'> & CondoAlertProps
+
+const Alert: React.FC<AlertProps> = (props) => {
+    const { type, className, ...rest } = props
+
+    const classes = classNames(
+        {
+            [`${ALERT_CLASS_PREFIX}-${type}`]: type,
+        },
+        className,
+    )
+
+    return (
+        <DefaultAlert
+            {...rest}
+            type = {type}
+            prefixCls={ALERT_CLASS_PREFIX}
+            className={classes}
+        />
+    )
+}
+
+export {
+    Alert,
+}
--- a/packages/ui/src/components/Alert/alert.tsx
+++ b/packages/ui/src/components/Alert/alert.tsx
@@ -7,7 +7,7 @@ type CondoAlertProps = {
 
 const ALERT_CLASS_PREFIX = 'condo-alert'
 
-export type AlertProps = Pick<DefaultAlertProps, 'message' | 'showIcon' | 'description' | 'action'> & CondoAlertProps
+export type AlertProps = Pick<DefaultAlertProps, 'message' | 'showIcon' | 'description'> & CondoAlertProps
 
 const Alert: React.FC<AlertProps> = (props) => {
     return (
--- a/packages/ui/src/components/Alert/alert.tsx
+++ b/packages/ui/src/components/Alert/alert.tsx
@@ -1,6 +1,5 @@
 import React from 'react'
 import { Alert as DefaultAlert, AlertProps as DefaultAlertProps } from 'antd'
-import classNames from 'classnames'
 
 type CondoAlertProps = {
     type: 'success' | 'info' | 'warning' | 'error';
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/Alert/alert.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Alert as DefaultAlert, AlertProps as DefaultAlertProps } from 'antd'
+import classNames from 'classnames'
+
+type CondoAlertProps = {
+    type: 'success' | 'info' | 'warning' | 'error';
+}
+
+const ALERT_CLASS_PREFIX = 'condo-alert'
+
+export type AlertProps = Omit<DefaultAlertProps, 'type' | 'prefix' | 'prefixCls' | 'banner' | 'afterClose' | 'closable' | 'closeText' | 'closeIcon' | 'onClose'> & CondoAlertProps
```
SavelevMatthew, line 11: How about showIcon and icon?

packages/ui/src/components/Alert/alert.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const Alert: React.FC<AlertProps> = (props) => {
+    const { type, className, ...rest } = props
```
SavelevMatthew, line 14: pleas omit className also. We do not want anybody to be able to modify it

packages/ui/src/components/Alert/alert.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+    const classes = classNames(
+        {
+            [`${ALERT_CLASS_PREFIX}-${type}`]: type,
```
SavelevMatthew, line 18: Antd have this classes by default, so no need to override it here
https://github.com/ant-design/ant-design/blob/4.x-stable/components/alert/index.tsx#L153

packages/ui/src/components/Alert/alert.tsx
SavelevMatthew: you should also omit style attribute

SavelevMatthew: I guess it's easier to Pick needed props here rather the omit rest

packages/ui/src/components/Alert/alert.tsx
```diff
@@ -0,0 +1,35 @@
...
+        },
+        className,
+    )
+
+    return (
+        <DefaultAlert
+            {...rest}
+            type = {type}
+            prefixCls={ALERT_CLASS_PREFIX}
+            className={classes}
```
SavelevMatthew, line 28: And what about icons? seems like now we're using custom icons here. Why not it's included here? 

P.S. Seems strange to see that almost all alerts are using info icon. You better to discuss it with Anton

packages/ui/src/components/Alert/style.less
```diff
@@ -0,0 +1,42 @@
+@import "antd/lib/alert/style/index.less";
+@import (reference) "@open-condo/ui/src/tokens/variables.less";
+@import (reference) "@open-condo/ui/src/components/style/mixins/typography";
+@import "./mixins";
+
+.condo-alert {
```
SavelevMatthew, line 6: Did you tried just to set this vars?) 

https://github.com/ant-design/ant-design/blob/4.x-stable/components/style/themes/default.less#L951

packages/ui/src/stories/Alert.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+        showIcon: true,
+        description: 'Description',
+    },
+    argTypes: {
+        type: { control: 'select' },
+        icon: {
+            options: Object.keys(icons),
+            mapping: icons,
+            control: {
+                type: 'select',
```
SavelevMatthew, line 26: I guess icon should only depends on alert type and not be user-selectable

SavelevMatthew (commented):
Make sure you added "Request UI-kit review" tag as soon as you're ready to receive review from our design team :)

packages/ui/src/components/Alert/alert.tsx
pahaz: Yes, we probably want to expose only the required for us and supported by our team component API

packages/ui/src/components/Alert/alert.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import { Alert as DefaultAlert, AlertProps as DefaultAlertProps } from 'antd'
+
+type CondoAlertProps = {
+    type: 'success' | 'info' | 'warning' | 'error';
```
SavelevMatthew, line 5: Why do you need to override type with same values?)

packages/ui/src/components/Alert/alert.tsx
```diff
@@ -0,0 +1,26 @@
...
+const ALERT_CLASS_PREFIX = 'condo-alert'
+
+export type AlertProps = Pick<DefaultAlertProps, 'message' | 'showIcon' | 'description' | 'action'> & CondoAlertProps
+
+const Alert: React.FC<AlertProps> = (props) => {
+    const { type, ...rest } = props
+
+    return (
+        <DefaultAlert
+            {...rest}
```
SavelevMatthew, line 17: just do `{...props}`

packages/ui/src/components/Alert/alert.tsx
Kekmus: because in DefaultAlertProps the type property is optional, and it seemed to me that it would be better if in the storybook this property was specified as mandatory:
Screenshot 2022-12-06 at 13 37 29

or can it be done differently without override type with same values?)

SavelevMatthew: Oh....I see 👍 

Then your solution is ok...
Also solution with `type: DefaultAlertProps['type']` can be used to override field obligation

packages/ui/src/stories/Alert.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+    component: Alert,
+    args: {
+        type: 'success',
+        message: 'Message',
+        showIcon: true,
+        description: 'Description',
+    },
+    argTypes: {
+        type: { control: 'select' },
+        icon: {
```
SavelevMatthew: I guess you should delete it)

SavelevMatthew (approved):

PR #2673, opened 2022-12-08
DOMA-4871 made employee role is required

nomerdvadcatpyat:


Kanol (approved):

apps/condo/migrations/20221208142802-0198_alter_organizationemployee_role.js
```diff
@@ -0,0 +1,54 @@
+// auto generated by kmigrator
```
sitozzz, line 1: You should recreate migration due to conflicting number with origin branch

Alllex202 (approved):

PR #2677, opened 2022-12-08
fix(address-service): DOMA-4883 restrict some resident's fields to be…

AleX83Xpert:
… updated


packages/keystone/plugins/addressService.js
```diff
@@ -108,16 +109,22 @@ const addressService = (fieldsHooks = {}) => plugin(({
             const result = await client.search(get({ ...existingItem, ...resolvedData }, ['address']))
 
             if (result) {
-                resolvedData['address'] = get(result, 'address')
+                /**
+                 * We have restriction for updating some fields for existing resident
+                 * @see apps/condo/domains/resident/schema/RegisterResidentService.js:65
+                 * @see apps/condo/domains/resident/schema/Resident.js:213
+                 */
+                if (!(operation === 'update' && listKey === 'Resident')) {
```
SavelevMatthew, line 117: Resident is an app-specific model...I don't think this logic must be stored inside `packages`, which goal is to be open to use across all apps

AleX83Xpert: Done. This logic moved to the particular model.

SavelevMatthew (approved):

Alllex202 (approved):

PR #2681, opened 2022-12-09
fix(address-service): DOMA-4880 Address service client + POST = ❤️

AleX83Xpert:
Address service client was simplified
Call POST method for suggests
tin (inn) was added to jsdoc

### DIFF
--- a/packages/keystone/plugins/utils/address-service-client/AddressServiceClient.js
+++ b/packages/keystone/plugins/utils/address-service-client/AddressServiceClient.js
@@ -1,6 +1,11 @@
 const fetch = require('node-fetch')
 
 /**
+ * @typedef {Object} SuggestionHelpersType
+ * @property {string} tin The organization's tin (inn)
+ */
+
+/**
  * @typedef {Object} AddressServiceParams
  * @property {string?} geo
  * @property {number?} count Number of results to return (20 by default)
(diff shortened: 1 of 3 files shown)


packages/keystone/plugins/utils/address-service-client/AddressServiceClient.js
```diff
@@ -102,9 +96,9 @@ class AddressServiceClient {
             throw new Error('The `s` parameter is mandatory')
         }
 
-        const urlParams = [`s=${s}`, this.urlifyParams(params)].filter(Boolean)
+        const urlParams = new URLSearchParams({ s, ...params }).toString()
```
dkoviazin, line 99: URLSearchParams - do not work recursive
image

dkoviazin (approved):

packages/keystone/plugins/utils/address-service-client/AddressServiceClient.js
AleX83Xpert: 🤦‍♂️
will fix soon

AleX83Xpert: Recursivity is not needed here, because we pass a one-level object with strings (according to jsdoc).

akarjakin (approved):

PR #2683, opened 2022-12-09
fix(ui): DOMA-4653 fix borders style of markdown table

Kekmus:
Screenshot 2022-12-09 at 18 56 24

### DIFF
--- a/packages/ui/src/components/style/mixins/table.less
+++ b/packages/ui/src/components/style/mixins/table.less
@@ -6,6 +6,7 @@
   text-align: left;
   border: @condo-global-border-width-default solid @condo-global-color-gray-3;
   border-radius: @condo-global-border-radius-medium;
+  border-collapse: separate !important;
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/style/mixins/table.less
```diff
@@ -6,6 +6,7 @@
   text-align: left;
   border: @condo-global-border-width-default solid @condo-global-color-gray-3;
   border-radius: @condo-global-border-radius-medium;
+  border-collapse: separate !important;
```
akarjakin, line 9: using !important is antypattern. instead you can increase specificity of the selector

AntonAL (commented):
Why `separate` and not `collapse`? Are there some use-cases to have individual border styling for each cell?

akarjakin (approved):

SavelevMatthew (approved):

PR #2686, opened 2022-12-11
Adr15/lang refactor

sitozzz:
- [x] Move page related translations to a new place {lang}.pages.json at condo
- [ ] Do the same as above at all related submodules
- [x] Change i18 lint logic
- [x] Fix default values for `LocalizedText` with a new custom migration
- [ ] Check for possible duplicates


packages/codegen/templates/app00/lang/en.json
```diff
@@ -1,5 +1,5 @@
 {
-  "pages.index.PageTitle": "demo",
-  "menu.Users": "Users",
-  "WeAreStillDevelopingThisPage": "We are still **developing** this page, it is not **yet** available"
+    "index.PageTitle": "demo",
+    "menu.Users": "Users",
+    "WeAreStillDevelopingThisPage": "We are still **developing** this page, it is not **yet** available"
```
AntonAL, line 4: Should not key be in camelCase as it states in ADR?

PR #2704, opened 2022-12-15
feat(condo): DOMA-4604 new miniapps 💻💄

SavelevMatthew:


apps/condo/domains/common/components/containers/GlobalStyle.tsx
```diff
@@ -463,3 +451,19 @@ const cardCSS = css`
...
+const previewCSS = css`
+  .ant-image-preview-switch-left,
+  .ant-image-preview-switch-right {
+    width: 40px;
+    height: 40px;
+    background: transparent;
+  }
+  .ant-image-preview-img {
+    width: auto;
+    //border-radius: 12px;
```
sitozzz, line 464: That's unnecessary
```suggestion
```

apps/condo/domains/miniapp/components/AppDescription/ConnectModal.tsx
```diff
@@ -0,0 +1,48 @@
...
+            {ModalButtonLabel}
+        </Button>
+    )
+
+    return (
+        <Modal
+            title={ModalTitle}
+            open={open}
+            onCancel={closeModal}
+            destroyOnClose={true}
```
sitozzz, line 40: Maybe it can be done like that?
```suggestion
            destroyOnClose
```

apps/condo/domains/miniapp/components/AppDescription/PageContent.tsx
```diff
@@ -0,0 +1,190 @@
...
+import { useOrganization } from '@open-condo/next/organization'
+import { useLazyQuery } from '@open-condo/next/apollo'
+import { MiniAppOutput } from '@app/condo/schema'
+import { ALL_MINI_APPS_QUERY } from '@condo/domains/miniapp/gql'
+import { getClientSideSenderInfo } from '@condo/domains/common/utils/userid.utils'
+import { useContainerSize } from '@condo/domains/common/hooks/useContainerSize'
+
+import { TopCard } from './TopCard'
+import { DeveloperCard } from './DeveloperCard'
+import { AppCard, MIN_CARD_WIDTH } from '../AppCard'
```
sitozzz, line 16: Absolute import)

apps/condo/domains/miniapp/components/Catalog/PageContent.tsx
```diff
@@ -0,0 +1,188 @@
...
+                            {promoBlocks.map(promoBlock => (
+                                <Banner
+                                    key={promoBlock.id}
+                                    title={promoBlock.title}
+                                    subtitle={promoBlock.subtitle}
+                                    backgroundColor={promoBlock.backgroundColor}
+                                    actionText={BannerMoreMessage}
+                                    imgUrl={get(promoBlock, ['backgroundImage', 'publicUrl'])}
+                                    {...PROMO_BLOCK_TEXT_VARIANTS_TO_PROPS[promoBlock.textVariant]}
+                                    onClick={() => handleBannerClick(promoBlock.targetUrl, promoBlock.external)}
```
sitozzz, line 150: Maybe rewrite this with closure?

nomerdvadcatpyat:
Do you really need to update eps and registry submodules?
There doesn't seem to be any commits for this task.

sitozzz:
Why there are two migration files?

apps/condo/bin/generate-billings.js
```diff
@@ -52,9 +51,8 @@ const SuccessfulBilling = {
     sender: SENDER,
     name: 'Интеграция через загрузку вашего реестра (DONE STATUS)',
     shortDescription: 'Поддерживаемые форматы: 1С, СБ Бизнес Онлайн 8_2 и 9_1',
-    instruction: 'Выберите формат, в котором хотите загружать ваши реестры в Домá.\n' +
+    detailedDescription: 'Выберите формат, в котором хотите загружать ваши реестры в Домá.\n' +
```
Alllex202, line 54:
```suggestion
    detailedDescription: 'Выберите формат, в котором хотите загружать ваши реестры в «Домá».\n' +
```

apps/condo/domains/miniapp/components/Catalog/CardGrid.tsx
```diff
@@ -0,0 +1,188 @@
...
+const HIDE_SEARCH_CSS = css`
+  .tabs-hide-first {
+    .ant-tabs-tab:first-of-type {
+      display: none;
+    }
+    
+    .ant-tabs-tab:nth-of-type(2) {
+      margin-left: 0 !important;
+    }
+    .ant-tabs-nav-operations { display: none !important; }
```
Alllex202, line 51:
```suggestion
    .ant-tabs-nav-operations {
      display: none !important; 
    }
```

apps/condo/pages/miniapps/[id]/about.tsx
```diff
@@ -1,45 +1,48 @@
...
+        if (type === BILLING_APP_TYPE) return <BillingAppPage id={id}/>
+        if (type === ACQUIRING_APP_TYPE) return <AcquiringAppPage id={id}/>
+        return <B2BAppPage id={id}/>
     }, [id, type])
 
     if (!canManageIntegrations) {
         return <LoadingOrErrorPage title={PageTitle} error={NoPermissionsMessage}/>
     }
 
     return (
```
Comment (Alllex202) [inline apps/condo/pages/miniapps/[id]/about.tsx:39]:
```suggestion
    return pageContent
```

apps/condo/pages/miniapps/[id]/index.tsx
```diff
@@ -1,43 +1,47 @@
...
+        if (type === ACQUIRING_APP_TYPE) return <AcquiringAppPage id={id}/>
+        return <B2BAppPage id={id}/>
     }, [id, type])
 
+
     if (!canManageIntegrations) {
         return <LoadingOrErrorPage title={PageTitle} error={NoPermissionsMessage}/>
     }
 
     return (
```
Comment (Alllex202) [inline apps/condo/pages/miniapps/[id]/index.tsx:41]:
```suggestion
    return pageContent
```

apps/condo/domains/miniapp/components/AppDescription/AcquiringAppPage.tsx
```diff
@@ -0,0 +1,91 @@
...
+                        publishedAt={app.createdAt}
+                        partnerUrl={app.partnerUrl}
+                        price={app.price}
+                        gallery={app.gallery}
+                        contextStatus={get(context, 'status', null)}
+                        appUrl={app.appUrl}
+                        connectAction={createContextAction}
+                    />
+                    <ConnectModal
+                        miniappHasFrame={!!app.appUrl}
```
Alllex202, line 82:
```suggestion
                        miniappHasFrame={Boolean(app.appUrl)}
```

apps/condo/domains/miniapp/components/AppDescription/B2BAppPage.tsx
```diff
@@ -0,0 +1,88 @@
...
+                        publishedAt={app.createdAt}
+                        partnerUrl={app.partnerUrl}
+                        price={app.price}
+                        gallery={app.gallery}
+                        contextStatus={get(context, 'status', null)}
+                        appUrl={app.appUrl}
+                        connectAction={createContextAction}
+                    />
+                    <ConnectModal
+                        miniappHasFrame={!!app.appUrl}
```
Alllex202, line 79:
```suggestion
                        miniappHasFrame={Boolean(app.appUrl)}
```

apps/condo/domains/miniapp/components/AppDescription/BillingAppPage.tsx
```diff
@@ -0,0 +1,91 @@
...
+                        publishedAt={app.createdAt}
+                        partnerUrl={app.partnerUrl}
+                        price={app.price}
+                        gallery={app.gallery}
+                        contextStatus={get(context, 'status', null)}
+                        appUrl={app.appUrl}
+                        connectAction={createContextAction}
+                    />
+                    <ConnectModal
+                        miniappHasFrame={!!app.appUrl}
```
Alllex202, line 82:
```suggestion
                        miniappHasFrame={Boolean(app.appUrl)}
```

apps/condo/domains/miniapp/components/AppDescription/PageContent.tsx
SavelevMatthew: Our current code-style agreement allows us to use relative imports inside component Group

apps/condo/domains/miniapp/components/AppDescription/TopCard.tsx
```diff
@@ -1,169 +1,238 @@
...
+        setCurrentPreview(curr => (curr + 1) % imagesAmount)
+    }, [imagesAmount])
+
+    const handlePrevPreview = useCallback(() => {
+        setCurrentPreview(curr => (curr + imagesAmount - 1) % imagesAmount)
+    }, [imagesAmount])
+
+    const sectionSpan = isWide ? HALF_COL_SPAN : FULL_COL_SPAN
+    const buttonSpacing = isWide ? SPACED_BUTTON_SPACING : SHRINKED_BUTTON_SPACING
+    const rowStyles = isWide ? undefined : SPACED_ROW_STYLES
```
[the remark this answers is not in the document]
SavelevMatthew: This syntax can break Antd API, since `&&` return type will be `Boolean | {...}`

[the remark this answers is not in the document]
SavelevMatthew: May be empty object instead of undefined would be better choice right here...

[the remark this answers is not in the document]
SavelevMatthew: image

apps/condo/domains/miniapp/components/AppDescription/PageContent.tsx
```diff
@@ -0,0 +1,190 @@
...
+                        <DeveloperCard
+                            developer={developer}
+                            publishedAt={publishedAt}
+                            partnerUrl={partnerUrl}
+                            display={isDevCardWide ? 'row' : 'col'}
+                        />
+                    </Col>
+                </Row>
+            </Col>
+            {Boolean(moreApps.length) && (
```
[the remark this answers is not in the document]
SavelevMatthew: Why? isEmpty is useful only if object is null / undefined. Using util just for util is increasing bundle size

apps/condo/domains/miniapp/components/AppDescription/TopCard.tsx
```diff
@@ -1,169 +1,238 @@
...
 }) => {
     const intl = useIntl()
-    const DeveloperMessage = intl.formatMessage({ id: 'Developer' })
-    const PublishedMessage = intl.formatMessage({ id: 'Published' })
-    const PartnerSiteMessage = intl.formatMessage({ id: 'miniapps.PartnerSite' })
-
-    const { width } = useWindowSize()
-    const isSmallLayout = Boolean(width && width < TITLE_VERTICAL_MARK)
+    const CategoryMessage = intl.formatMessage({ id: `miniapps.categories.${category}.name` })
+    const router = useRouter()
```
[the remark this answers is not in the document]
SavelevMatthew: Make sure this rule is included in code-style guide

apps/condo/domains/miniapp/components/AppDescription/AcquiringAppPage.tsx
SavelevMatthew: Will change to `Boolean` to match over usages inside current PR, but there's not rule in our code-style about this

apps/condo/migrations/20221215183610-0201_fill_acquiring_context_status.js
```diff
@@ -0,0 +1,18 @@
+// auto generated by kmigrator
+// KMIGRATOR:20221215183610-0201_fill_acquiring_context_status
+
+exports.up = async (knex) => {
+    await knex.raw(`
+        BEGIN;
+        UPDATE "AcquiringIntegrationContext" SET "status" = 'Finished' WHERE True;
```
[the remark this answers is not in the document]
SavelevMatthew: We need to modify all objects in DB. Using update without where is linter warning (probably you forgot to add a filter)...Setting where explicitly to true confirms your intention to modify all elements of the table.

SavelevMatthew:
> Why there are two migration files?

One is auto-generated migration for adding new fields / remove old fields and one for manually setting current current DB   state (since stats field is pseudo-required. (Not required, so migrator doesn't ask default values, but auto-resolvable)

SavelevMatthew:
> Do you really need to update eps and registry submodules?
> There doesn't seem to be any commits for this task.

There was,  but I've moved them outside to separate PRs, so now I'm just pointing this submodules to current master, since submodule cherry-pick-rebase is quite problematic

apps/condo/domains/miniapp/components/Catalog/CardGrid.tsx
```diff
@@ -0,0 +1,188 @@
...
+    return (
+        <section ref={setRef}>
+            <Global styles={HIDE_SEARCH_CSS}/>
+            <TabComponent
+                tabPosition={sideTabs ? 'right' : 'top'}
+                type='card'
+                tabBarGutter={TAB_GUTTER}
+                defaultActiveKey={tabFromQuery}
+                activeKey={selectedTab}
+                tabBarStyle={sideTabs ? SIDE_TAB_SIDE_BAR_STYLES : undefined}
```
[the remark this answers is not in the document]
SavelevMatthew: See comment above

apps/condo/domains/miniapp/components/Catalog/CardGrid.tsx
```diff
@@ -0,0 +1,188 @@
...
+    const categories = tabs.map(tab => tab.category)
+    const tabFromQuery = (tab && !Array.isArray(tab) && categories.includes(tab.toUpperCase())) ? tab.toUpperCase() : ALL_APPS_CATEGORY
+    const selectedTab = search ? SEARCH_TAB_KEY : tabFromQuery
+
+    const handleTabChange = useCallback((newKey) => {
+        const newRoute = `${router.route}?tab=${newKey}`
+        return router.push(newRoute).then(resetSearch)
+    }, [router, resetSearch])
+
+    const searchTab: TabContent = {
```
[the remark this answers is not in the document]
SavelevMatthew: It's useless here, since useMemo does not use deep-comparison under the hood.
And also we're pre-rendering rest of tabs as well, so why need to memoize just a single tab?)

apps/condo/domains/miniapp/components/AppDescription/PageContent.tsx
sitozzz: No you don't) You're allowed to use relative inside module, as you say. So 14 and 15 lines are representing this rule. It's a little bit struggling, that you first of all should look at files tree before you actually can understand destination of imported file

SavelevMatthew: Let's imagine I have a structure like this:
```
.
├── ResuablePart.tsx
├── ComplexComponent.tsx
└── EvenMoreComplexComponent/
    ├── index.ts
    └── SomeSpecificPartOfComponent
```

So why then I'm allowed to call `./ReusablePart` from `ComplexComponent` and not allowed to use it in `EvenMoreComplexComponent `.

What I'm trying to explain, that some of components are more than 1 file, but they're in 99.9% flat file list in some folder. So if I put `EvenMoreComplexComponent` files in top level, this will be `./` local import same as in `ComplexComponent, which is currently allowed.

So, here we should either ban ANY relative import, or deal with this cases specifically.

nomerdvadcatpyat (approved):

apps/condo/domains/miniapp/components/AppDescription/TopCard.tsx
```diff
@@ -1,169 +1,238 @@
...
+    const [{ width }, setRef] = useContainerSize()
+
+    const buttonProps = useMemo<ButtonProps>(() => {
+        const btnProps: ButtonProps = { type: 'primary' }
+        if (!contextStatus) {
+            btnProps.children = intl.formatMessage({ id: 'miniapps.addDescription.action.connect' })
```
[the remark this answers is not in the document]
SavelevMatthew: This solution was more clear for me, rather than expand memo with pack of translation deps

apps/condo/domains/miniapp/components/AppDescription/PageContent.tsx
sitozzz: Yes, that's how it is. But what if we had the following structure:
```
.
├── ResuableDomainSpecificPart.tsx
├── ComplexComponent.tsx
└── EvenMoreComplexComponent/
    ├── index.ts
    └── SomeSpecificPartOfComponent
    └── SuperSpecificComponent/
        ├── index.ts
        └── SuperSpecificPartOfComponent
```

So based on your logic, we allowed to use `../../ReusableDomainSpecificPart` inside `SuperSpecificComponent`. So the main idea, as far as I can tell was place local reusable part's of the project near to the place of usage. But if you need to split this grouping by another level it's become not so clear to understand what does it refer to.

Anyway your position make sense, but I would say that such nesting in the structure of components should be avoided

apps/condo/migrations/20221215183610-0201_fill_acquiring_context_status.js
```diff
@@ -0,0 +1,18 @@
+// auto generated by kmigrator
+// KMIGRATOR:20221215183610-0201_fill_acquiring_context_status
+
+exports.up = async (knex) => {
+    await knex.raw(`
+        BEGIN;
```
sitozzz, line 6: Any comment about "why we doing so"?

Alllex202 (approved):

sitozzz (approved):

apps/condo/domains/common/components/containers/AccessDeniedPage.tsx
```diff
@@ -27,7 +27,7 @@ interface IAccessDeniedPageProps {
 export const AccessDeniedPage: React.FC<IAccessDeniedPageProps> = ({ title }) => {
     const intl = useIntl()
     const ContactTheAdministratorMessage = intl.formatMessage({ id: 'ContactTheAdministrator' })
-    const NoPermissionMessage = intl.formatMessage({ id: 'NoPermissionToPage' })
+    const NoPermissionMessage = intl.formatMessage({ id: 'global.noPageViewPermission' })
```
AntonAL, line 30: Maybe just `accessDenied`?

SavelevMatthew: `accessDenied` is too abstract to understand it's usage

PR #2710, opened 2022-12-19
Feature/condo/doma 4545/ticket search

Alllex202:


apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -541,6 +561,15 @@ const Ticket = new GQLListSchema('Ticket', {
         delete: false,
         auth: true,
     },
+    kmigratorOptions: {
+        indexes: [
+            {
+                type: 'Index', // or BTreeIndex
```
pahaz, line 567: it looks lika part of another PR

apps/condo/domains/user/schema/User.js
```diff
@@ -214,6 +215,7 @@ const User = new GQLListSchema('User', {
         dvAndSender(),
         historical(),
         webHooked(),
+        searchUpdateByTrigger([{ targetSchema: 'Ticket', triggerPathsToFields: ['assignee.name', 'executor.name'], schemaGql: Ticket }]),
```
pahaz, line 218: it looks like a bad code design. Probably this should work automatically inside the `searchBy` plugin

packages/keystone/plugins/searchBy.js
```diff
@@ -0,0 +1,189 @@
...
+    kmigratorOptions: { null: false },
+    defaultValue: SEARCH_FIELD_DEFAULT_VALUE,
+    access: {
+        create: false,
+        read: true,
+        update: true,
+    },
+    knexOptions: { defaultTo: () => SEARCH_FIELD_DEFAULT_VALUE },
+}
+const SEARCH_VERSION_OPTIONS = {
```
pahaz, line 23: how is it supposed to be used ?

packages/keystone/plugins/searchBy.js
```diff
@@ -0,0 +1,189 @@
...
+    schemaDoc: 'Service field for versioning the search field by different fields',
+    type: 'Integer',
+    isRequired: true,
+    defaultValue: 1,
+    access: {
+        create: false,
+        read: true,
+        update: true,
```
pahaz, line 31: anyone can update this field?

packages/keystone/plugins/searchBy.js
```diff
@@ -0,0 +1,189 @@
...
+
+const SEARCH_OPTIONS = {
+    schemaDoc: 'Service field for searching by different fields',
+    type: 'Text',
+    kmigratorOptions: { null: false },
+    defaultValue: SEARCH_FIELD_DEFAULT_VALUE,
+    access: {
+        create: false,
+        read: true,
+        update: true,
```
pahaz, line 19: anyone can update this field

Alllex202: no one

packages/keystone/plugins/searchBy.js
Alllex202: only server utils

apps/condo/domains/user/schema/User.js
Alllex202: moved to global preprocessors

Alllex202: Moved to global plugins

packages/keystone/plugins/searchBy.js
Alllex202: The "searchVersion" field is used as a trigger.
This allows you to control the values of the "search" field from the main model. Linked models can only report that they have updated without interacting with the "search" field directly.

Tracked field updated --> update "searchVersion" field --> need to update "search" field.

Alllex202: But it is better to use not a number, but some random value. For example, uuid

Alllex202:
Issue: updating related models can trigger a bulk ticket update

PR #2716, opened 2022-12-20
DOMA-4526 feat map constructor validation

VKislov:
During the creation of units, if there is a duplicate in a large chessboard, it is difficult to find where this unit is already represented. 

**How it highlights:**
Screenshot 2022-12-20 at 19 03 34

### DIFF
--- a/apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
+++ b/apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
@@ -240,6 +240,7 @@ export const BuildingPanelEdit: React.FC<IBuildingPanelEditProps> = (props) => {
     }, [mapEdit])
 
     const changeMode = useCallback((mode) => {
+        setDuplicatedUnitIds([])
         mapEdit.editMode = mode
         refresh()
     }, [mapEdit, refresh])
--- a/apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
+++ b/apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
@@ -669,6 +669,10 @@ interface IPropertyMapUnitProps {
 }
 
 const PropertyMapUnit: React.FC<IPropertyMapUnitProps> = ({ builder, refresh, unit }) => {
+    console.log('duplicatedUnitIds', builder.duplicatedUnits)
+    const duplicatedUnitIds = useMemo(() => {
+        return builder.duplicatedUnits
+    }, [builder, refresh])
     const selectUnit = useCallback(() => {
         if (unit.unitType !== BuildingUnitSubType.Parking) {
             builder.removePreviewUnit()
@@ -688,6 +692,7 @@ const PropertyMapUnit: React.FC<IPropertyMapUnitProps> = ({ builder, refresh, un
         <UnitButton
             onClick={selectUnit}
             disabled={unit.preview}
+            duplicatedUnitIds={duplicatedUnitIds}
             preview={unit.preview}
             selected={isUnitSelected}
             unitType={unit.unitType}
--- a/apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
+++ b/apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
@@ -539,7 +539,6 @@ const ChessBoard: React.FC<IChessBoardProps> = (props) => {
                                                 section={section}
                                                 builder={builder}
                                                 refresh={refresh}
-                                                duplicatedUnitIds={duplicatedUnitIds}
                                             >
                                                 <PropertyMapFloorContainer
                                                     builder={builder}
@@ -557,7 +556,6 @@ const ChessBoard: React.FC<IChessBoardProps> = (props) => {
                                                 builder={builder}
                                                 refresh={refresh}
                                                 isParkingSection
-                                                duplicatedUnitIds={duplicatedUnitIds}
                                             >
                                                 <PropertyMapFloorContainer
                                                     builder={builder}
@@ -597,7 +595,7 @@ const FULL_SIZE_UNIT_STYLE: React.CSSProperties = { width: '100%', marginTop: '8
 const SECTION_UNIT_STYLE: React.CSSProperties = { ...FULL_SIZE_UNIT_STYLE, zIndex: 2 }
 
 const PropertyMapSection: React.FC<IPropertyMapSectionProps> = (props) => {
-    const { section, children, builder, refresh, isParkingSection = false, duplicatedUnitIds } = props
+    const { section, children, builder, refresh, isParkingSection = false } = props
     const intl = useIntl()
     const SectionTitle = isParkingSection
         ? `${intl.formatMessage({ id: 'pages.condo.property.select.option.parking' })} ${section.name}`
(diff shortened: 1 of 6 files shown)


apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -218,6 +219,8 @@ export const BuildingPanelEdit: React.FC<IBuildingPanelEditProps> = (props) => {
     const sections = mapEdit.sections
     const address = get(property, 'address')
 
+    const [duplicatedUnitIds, setDuplicatedUnitIds] = useState<string[]>([])
```
sitozzz, line 222: Why you need a state to handle duplicates? That's too heavy solution, that could reduce performance.
Is it possible to do this only with special method of `MapConstructor`?

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -273,16 +276,16 @@ export const BuildingPanelEdit: React.FC<IBuildingPanelEditProps> = (props) => {
     }, [mapEdit, refresh])
 
     const menuContent = useMemo(() => ({
-        addSection: <AddSectionForm builder={mapEdit} refresh={refresh}/>,
-        addUnit: <UnitForm builder={mapEdit} refresh={refresh}/>,
-        editSection: <EditSectionForm builder={mapEdit} refresh={refresh}/>,
-        editUnit: <UnitForm builder={mapEdit} refresh={refresh}/>,
+        addSection: <AddSectionForm builder={mapEdit} refresh={refresh} />,
+        addUnit: <UnitForm builder={mapEdit} refresh={refresh} setDuplicatedUnitIds={setDuplicatedUnitIds} />,
```
sitozzz, line 280: Also you need to pass this state to all of the components as a dependency. But they already contain `MapConstructor` instance

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
VKislov: The state is needed so that after validation, duplicates can be highlighted. I didn't see any major changes in performance.

pahaz:
It looks to heavy. Probably you really need to find a better code-design solution. 
It doesn't look like the implementation doesn't take into account the current code architecture design.

apps/condo/domains/property/components/panels/Builder/UnitButton.tsx
```diff
@@ -124,6 +124,9 @@ const noninteractiveCss = css`
         border-color:  #F5F5F5;
     }
 `
+const duplicatedUnit = css`
```
AntonAL, line 127:
```suggestion
const duplicatedUnitStyle = css`
```

apps/condo/domains/property/components/panels/Builder/MapConstructor.ts
```diff
@@ -150,24 +150,30 @@ class Map {
...
     }
-
-    public validateUniqueUnitLabel (renamedUnit: string = null, destination: keyof typeof MapViewMode = null): boolean {
-        const unitLabels = this.map.sections
+    get getUnitLabelsWithoutPreview (): string[] {
+        return this.map.sections
             ?.map((section) => section.floors
                 ?.map(floor => floor.units
-                    ?.map(unit => unit.label)
+                    ?.map(unit => !unit.preview && unit.label)
```
AntonAL, line 157: Don't you need to filter for 'false' values in case of `unit.preview` will be `true`?

VKislov: Only units without `preview` status are needed here. `map()` will push a unit into the array if it is not a `preview`

AntonAL: I mean following:

```
const a = [
    {name: 'Bob', preview: true},
    {name: 'Alice', preview: false}
  ]
  const b = a.map(p => !p.preview && p.name)
  console.debug('b', b)
```

You will get `[false, "Alice"]`.

Do you really need `false` item?

VKislov: ah, got it :) Actually, this will not affect the work, but I think it's better to fix it

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -534,11 +539,13 @@ const ChessBoard: React.FC<IChessBoardProps> = (props) => {
                                                 section={section}
                                                 builder={builder}
                                                 refresh={refresh}
+                                                duplicatedUnitIds={duplicatedUnitIds}
```
sitozzz, line 542: Why you need to pass id's to the section component? That's not make sense. Anyway you are not used this prop
```suggestion
```

apps/condo/domains/property/components/panels/Builder/forms/UnitForm.tsx
```diff
@@ -95,41 +95,35 @@ const UnitForm: React.FC<IPropertyMapModalForm> = ({ builder, refresh }) => {
...
-            const labelChanged = selectedUnit.label !== label
-            const labelValidation = labelChanged
-                ? builder.validateUniqueUnitLabel(label, 'section')
-                : false
-
-            isUnitLabelUnique = unitPlacementChanged
-                ? (!labelChanged || labelValidation)
-                : labelValidation
+            isUnitLabelUnique = builder.validateInputUnitLabel(selectedUnit, label)
+            setDuplicatedUnitIds(builder.duplicatedUnits)
```
sitozzz, line 107: Maybe just add special class to all of the found units with native javascript api? That's not trigger rerender cycle at all

VKislov: So, on the contrary, a rerender cycle is needed, otherwise the information in the UnitButton will simply not be relevant

VKislov: I'm not sure it's a good idea to do this bypassing react logic. At a minimum, this breaks the code consistency and looks like a hack.

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -539,6 +544,7 @@ const ChessBoard: React.FC<IChessBoardProps> = (props) => {
                                                     builder={builder}
                                                     section={section}
                                                     refresh={refresh}
+                                                    duplicatedUnitIds={duplicatedUnitIds}
```
sitozzz, line 547:
```suggestion
```

apps/condo/domains/property/components/panels/Builder/BuildingPanelEdit.tsx
```diff
@@ -550,12 +556,14 @@ const ChessBoard: React.FC<IChessBoardProps> = (props) => {
                                                 builder={builder}
                                                 refresh={refresh}
                                                 isParkingSection
+                                                duplicatedUnitIds={duplicatedUnitIds}
```
sitozzz, line 559:
```suggestion
```

sitozzz (approved):

PR #2726, opened 2022-12-22
feat(condo): DOMA-4709 added support for alerts in the markdown

Kekmus:
Screenshot 2022-12-22 at 16 55 41
Screenshot 2022-12-22 at 16 55 56

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -125,6 +125,8 @@
     "antd": "^4.24.3",
     "classnames": "^2.3.2",
     "react-markdown": "^6",
-    "remark-gfm": "^1"
+    "remark-directive": "^1",
+    "remark-gfm": "^1",
+    "unist-util-visit": "^4.1.2"
   }
 }
--- a/packages/ui/src/components/Markdown/markdown.tsx
+++ b/packages/ui/src/components/Markdown/markdown.tsx
@@ -1,13 +1,18 @@
 import omit from 'lodash/omit'
 import React from 'react'
 import ReactMarkdown from 'react-markdown'
+import directive from 'remark-directive'
 import remarkGfm from 'remark-gfm'
 
 import { CodeWrapper } from './codeWrapper'
+import customDirectives from './customDirectives'
 
+import { Alert } from '../Alert'
 import { Typography } from '../Typography'
 
 const REMARK_PLUGINS = [
+    directive,
+    customDirectives,
     remarkGfm,
 ]
 
@@ -33,6 +38,9 @@ export const Markdown: React.FC<MarkdownProps> = ({ children }) => {
                 a: (props) => <Typography.Link {...omit(props, 'ref')} target='_blank'/>,
                 li: ({ children, ...restProps }) => <li {...restProps}><Typography.Text type='secondary'>{children}</Typography.Text></li>,
                 pre: (props) => <CodeWrapper {...props}/>,
+                section: ({ children, title, ...restProps }) => {
+                    return <Alert description={children} type='success' showIcon={true} message={title} {...restProps}/>
+                },
             }}
         >
             {children}
--- a/apps/dev-portal/package.json
+++ b/apps/dev-portal/package.json
@@ -32,6 +32,7 @@
     "react-intl": "6.2.7",
     "rehype-prism-plus": "1.5.1",
     "remark-custom-heading-id": "1.0.0",
+    "remark-directive": "^1",
     "remark-gfm": "3.0.1",
     "title": "3.5.3",
     "typescript": "4.9.4"
--- a/packages/ui/src/components/Markdown/customDirectives.ts
+++ b/packages/ui/src/components/Markdown/customDirectives.ts
@@ -1,0 +1,21 @@
+import { Node } from 'unist'
+import { visit } from 'unist-util-visit'
+
+
+export default function customDirectives () {
+    return transform
+
+    function transform (tree: Node) {
+        return visit(tree, ['textDirective', 'leafDirective', 'containerDirective'], ondirective)
+    }
+
+    function ondirective (node: any) {
+        const { name, attributes, children } = node
+        const text = children[0]?.value ?? ''
+
+        const data = node.data || (node.data = {})
+        data.hName = 'section'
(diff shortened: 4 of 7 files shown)


packages/ui/package.json
```diff
@@ -116,6 +116,7 @@
     "antd": "^4.24.3",
     "classnames": "^2.3.2",
     "react-markdown": "^6",
+    "remark-directive": "^1.0.1",
```
pahaz, line 119: It doesn't looks like a maintainable package! 
It less then 50 LOC. And less then 1k stars. 
Do we really need it?

Kekmus: We use the package [react-markdown](https://www.npmjs.com/package/react-markdown) to convert markdown to react elements. To add custom syntax to markdown for elements such as tables or alerts, react-markdown suggests using plugins written for it. react-markdown mentions in its documentation a [list of plugins](https://github.com/remarkjs/remark/blob/main/doc/plugins.md#list-of-plugins) that can be used with it.

We already use the [remark-gfm](https://www.npmjs.com/package/remark-gfm) plugin for custom table syntax which also has less than 1k stars. And [remark-directive](https://www.npmjs.com/package/remark-directive) is the most popular among all plugins which fit the right functionality for the task and which the react-markdown documentation refers to.

packages/ui/src/components/Markdown/customDirectives.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+
+export default function customDirectives () {
+    return transform
+
+    function transform (tree: Node) {
+        return visit(tree, ['textDirective', 'leafDirective', 'containerDirective'], ondirective)
+    }
+
+    function ondirective (node: any): VisitorResult {
```
AleX83Xpert, line 11: Is there a way not to use `any` type?
Maybe union or something else.

packages/ui/src/components/Markdown/markdown.tsx
```diff
@@ -30,6 +35,13 @@ export const Markdown: React.FC<MarkdownProps> = ({ children }) => {
                 a: (props) => <Typography.Link {...omit(props, 'ref')} target='_blank'/>,
                 code: (props) => <Typography.Text {...omit(props, 'ref')} code/>,
                 li: ({ children, ...restProps }) => <li {...restProps}><Typography.Text type='secondary'>{children}</Typography.Text></li>,
+                section: ({ children, title, ...restProps }) => {
+                    return <Alert description={children} type='success' showIcon={true} message={title} {...restProps}/>
+                },
+                // success: ({ children, title, ...restProps }: AlertProps) => <Alert description={children} type='success' showIcon={true} message={title} {...restProps}/>,
+                // info: ({ children, title, ...restProps }) => <Alert description={children} type='info' showIcon={true} message={title} {...restProps}/>,
+                // error: ({ children, title, ...restProps }) => <Alert description={children} type='warning' showIcon={true} message={title} {...restProps}/>,
+                // warning: ({ children, title, ...restProps }) => <Alert description={children} type='error' showIcon={true} message={title} {...restProps}/>,
```
AleX83Xpert, line 44: Are these comments mandatory?

packages/ui/src/components/Markdown/markdown.tsx
```diff
@@ -1,11 +1,16 @@
 import React from 'react'
 import ReactMarkdown from 'react-markdown'
 import remarkGfm from 'remark-gfm'
+import directive from 'remark-directive'
+import customDirectives from './customDirectives'
 import omit from 'lodash/omit'
 import { Typography } from '../Typography'
+import { Alert } from '../Alert'
```
AleX83Xpert, line 8: `@open-condo/ui` instead of relative paths.

packages/ui/src/components/Markdown/customDirectives.tsx
```diff
@@ -0,0 +1,22 @@
+import { visit, Node, VisitorResult } from 'unist-util-visit'
```
AleX83Xpert, line 1: What is `unist-util-visit`? Why do we use it here but there is nothing about this package in package.json?

packages/ui/src/components/Markdown/customDirectives.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+    function ondirective (node: any): VisitorResult {
+        // console.log('node', node)
+        const { name, attributes, children } = node
+        const text = children[0]?.value ?? ''
+
+        const data = node.data || (node.data = {})
+        data.hName = 'section'
+        data.hProperties = { ...attributes, text, type: name }
+        console.log('data', data)
```
AleX83Xpert, line 19: Please remove debug outputs like this.
or use
```suggestion
        console.debug('data', data)
```
if it is really necessary.

packages/ui/src/components/Markdown/customDirectives.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export default function customDirectives () {
+    return transform
+
+    function transform (tree: Node) {
+        return visit(tree, ['textDirective', 'leafDirective', 'containerDirective'], ondirective)
+    }
+
+    function ondirective (node: any): VisitorResult {
+        // console.log('node', node)
```
AleX83Xpert, line 12:
```suggestion
```

AleX83Xpert (commented):
Please rename `customDirectives.tsx` to `.ts`, this is not component, but just set of functions.

packages/ui/package.json
```diff
@@ -125,6 +125,8 @@
     "antd": "^4.24.3",
     "classnames": "^2.3.2",
     "react-markdown": "^6",
-    "remark-gfm": "^1"
+    "remark-directive": "^1",
+    "remark-gfm": "^1",
```
AleX83Xpert, line 129: These packages have newer versions. Why we are using two-year-old ones?

Kekmus: we are using react-markdown package version 6, we can't update it at this time because if we build our apps the "document is not defined error" and react-markdown version 6 is only compatible with older versions of these packages

akarjakin (commented):
Some tests for this feature would be nice

PR #2734, opened 2022-12-23
Feat/condo/doma 5011/new schema bank integration

AntonAL:
We've decided to make data schema for "Account per Property" have the same "…Integration" and "…IntegrationContext" entities like in "billing" and "acquiring" domains.
`BankIntegration` means available data source for transactions. Currently we've developing obtaining transactions from 2 data sources:
1. SBBOL Fintech API
2. Exported file in `1CClientBankExchange` format
For each data source a corresponding record was inserted into `BankIntegration` table, which is not supposed to be changed by any user, — they are hardcoded.

`BankIntegrationContext` means usage of specific data source by specific organization with various settings.


apps/condo/domains/banking/constants.js
```diff
@@ -2,5 +2,11 @@
  * Generated by `createschema banking.BankCategory 'name:Text;'`
  */
 
+const BANK_INTEGRATION_IDS = {
+    SBBOL: 'd94743b0-e5d5-4d06-a244-ea4b2edb8633',
```
sitozzz, line 6: Why not from environment? What is that constants about?

apps/condo/domains/banking/utils/clientSchema/BankIntegration.ts
```diff
@@ -0,0 +1,30 @@
...
+import {
+    BankIntegration,
+    BankIntegrationCreateInput,
+    BankIntegrationUpdateInput,
+    QueryAllBankIntegrationsArgs,
+} from '@app/condo/schema'
+import { generateReactHooks } from '@open-condo/codegen/generate.hooks'
+import { BankIntegration as BankIntegrationGQL } from '@condo/domains/banking/gql'
+
+// TODO(codegen): write utils like convertToFormState and formValuesProcessor if needed, otherwise delete this TODO
```
sitozzz, line 14: Plz remove TODO line)
```suggestion
```

sitozzz (commented):
What if we need to add another `BankIntegration`? Based on the logic in this PR, should we create a new migration?

apps/condo/domains/banking/constants.js
AntonAL: `BankIntegration` means available data source for transactions. Currently we've developing obtaining transactions from 2 data sources:
1. SBBOL Fintech API
2. Exported file in `1CClientBankExchange` format
For each data source a corresponding record was inserted into `BankIntegration` table, which is not supposed to be changed by any user, — they are hardcoded.

sitozzz (approved):

PR #2735, opened 2022-12-23
ci(global): properly check for keystone build successful

sitozzz:
That's handle case, when keystone build ends with 0 code, but actually not create a `dist` directory with result static files
More information about this case you can find [here](https://github.com/keystonejs/keystone/issues/4339)


SavelevMatthew (approved):
LGTM

PR #2736, opened 2022-12-23
chore(condo): DOMA-5036 added eslint rule about deprecated "jspdf"

Alllex202:


package.json
```diff
@@ -178,7 +178,13 @@
         {
           "int32Hint": false
         }
-      ]
+      ],
+      "no-restricted-imports": ["warn", {
```
SavelevMatthew, line 182: How about error? Will it break current code-base?

SavelevMatthew (approved):
🏅

package.json
Alllex202: Now this library is still used in the project. Therefore, I think a warning will suffice.

PR #2745, opened 2022-12-28
DOMA-3960 fix layout task progress bar

VKislov:
Some fixes to match layouts. Refactoring of icon changing logic. Found strange behavior when using gradients in css or the `@emotion` bug. The icon switch logic has been rewritten from css styles to use a react state and a chain of condition checks.
**After:**
Screenshot 2022-12-28 at 11 53 49

**Before:** 
image

### DIFF
--- a/apps/condo/domains/common/components/tasks/TaskProgress.tsx
+++ b/apps/condo/domains/common/components/tasks/TaskProgress.tsx
@@ -64,7 +64,7 @@ export const CircularProgress = ({ progress }: ICircularProgressProps) => (
     />
 )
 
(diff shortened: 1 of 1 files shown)


Alllex202:
Perhaps you should talk to designers and make this icon one-color, no gradient. Then it can only be solved using css. 
Since the problem is in the gradient. 
Gradient is a sore subject for browsers :)

VKislov:
Yes, gradients are often problematic. However, our entire design system is full of gradients.

AntonAL (approved):
I don't see code related to gradients, as you mentioned in description.
LGTM, overall.

PR #2751, opened 2022-12-29
feat(condo): DOMA-4950 init telephony integration

nomerdvadcatpyat:
Added telephony mini-app and websocket service (integration between telephony server and telephony mini-app).

- Websocket service connects to the telephony server via AMI and listens for events (for example, the start of a call to a certain member). When an event is received, it returns a response to the telephony miniapp.
- The telephony miniapp includes 2 miniapps: **local** and **global**.
- A **local** miniapp is needed to connect a miniapp. In it, the administrator specifies the websocket url of the telephony service (`OrganizationTelephony` schema) and the internal sip numbers of employees (`TelephonyMember` schema).
- A **global miniapp** is needed for a specific employee. It connects via websocket to the telephony service (by the employee's sip number from `TelephonyMember`), receives events from telephony service and interacts with the condo front-end via postMessage.
- Currently, one `call` event has been implemented, which redirects the operator to the client's card.

изображение


packages/bridge/src/types/methods/index.ts
```diff
@@ -12,18 +12,19 @@ export type { ResizeWindowParams, ResizeWindowData } from './ResizeWindow'
...
  */
 export type RequestMethodsParamsMap = {
     CondoWebAppGetActiveProgressBars: GetActiveProgressBarsParams
-
     CondoWebAppGetLaunchParams: GetLaunchParamsParams
     CondoWebAppResizeWindow: ResizeWindowParams
     CondoWebAppShowNotification: ShowNotificationParams
     CondoWebAppShowProgressBar: ShowProgressBarParams
     CondoWebAppUpdateProgressBar: UpdateProgressBarParams
+    CondoWebSendCallEvent: CondoWebSendCallEventParams
```
SavelevMatthew, line 27: `CondoWebApp<Event-prefix>`

apps/condo/domains/common/components/TelephonyContext.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+        window.open(`/phone/+${phone}`, '_blank')
+    }, [])
+
+    useEffect(() => {
+        addEventHandler('CondoWebAppCall', '*', handleMessage)
+    }, [addEventHandler, handleMessage])
+
+    return (
+        <TelephonyContext.Provider value={{}}>
```
Alllex202, line 22: Why is context needed here?

apps/condo/domains/common/components/TelephonyContext.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const TelephonyContext = React.createContext({})
+
+export const TelephonyContextProvider = ({ children }) => {
+    const { addEventHandler } = usePostMessageContext()
+    
+    const handleMessage = useCallback<RequestHandler<'CondoWebAppCall'>>(async ({ phone }) => {
+        if (typeof window === 'undefined') return null
+
+        window.open(`/phone/+${phone}`, '_blank')
```
Alllex202, line 14: If we are on another tab, then this new window will in focus?

apps/condo/domains/common/components/TelephonyContext.tsx
nomerdvadcatpyat: Rename it to `TelephonyEventsHandler`, remove context

apps/condo/domains/common/components/TelephonyContext.tsx
nomerdvadcatpyat: Yes

apps/condo/domains/common/components/PostMessageProvider/types.ts
```diff
@@ -15,13 +15,15 @@ export const COMMON_ERROR_PREFIX = 'CondoWebAppCommonError' as const
 export type RequestParamsMap = {
     [Method in BridgeRequestMethodsName]: BridgeRequestParams<Method>
 } & {
-    CondoWebSendAnalyticsEvent: AnalyticsParams
+    CondoWebSendAnalyticsEvent: AnalyticsParams,
+    CondoWebAppCall: { phone: string },
```
SavelevMatthew, line 19: `...CallEvent`

apps/condo/domains/common/components/PostMessageProvider/validators.ts
```diff
@@ -84,4 +93,5 @@ export const validators: ValidatorsType = {
     CondoWebAppShowNotification: ajv.compile(CondoWebAppShowNotificationParamsSchema),
     CondoWebAppShowProgressBar: ajv.compile(CondoWebAppShowProgressBarParamsSchema),
     CondoWebAppUpdateProgressBar: ajv.compile(CondoWebAppUpdateProgressBarParamsSchema),
+    CondoWebAppCall: ajv.compile(CondoWebAppCallParamsSchema),
```
SavelevMatthew, line 96: Put it closer to analytics please...Since events are grouped by source + ordered alphabetically for better reading

Alllex202 (approved):

apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
```diff
@@ -10,7 +10,7 @@ import {
     useLaunchParamsHandler,
     useShowProgressBarHandler,
     useGetActiveProgressBarsHandler,
-    useUpdateProgressBarHandler,
+    useUpdateProgressBarHandler, useRedirectHandler,
```
SavelevMatthew, line 13: newline

apps/condo/domains/common/components/PostMessageProvider/globalHandlers.ts
```diff
@@ -138,4 +140,20 @@ export const useUpdateProgressBarHandler: () => RequestHandler<'CondoWebAppUpdat
...
+export const useRedirectHandler: () => RequestHandler<'CondoWebAppRedirect'> = () => {
+    const router = useRouter()
+
+    return useCallback(({ url, target }) => {
+        if (target === '_blank') {
+            if (typeof window !== 'undefined' && isSafeUrl(url)) {
+                window.open(url, target)
+            }
+        } else {
+            router.push(url)
```
SavelevMatthew, line 154: I guess checking for safe url should be here too. And also it's seems weird to allow redirect in same window on any provided url....

apps/condo/domains/common/components/PostMessageProvider/validators.ts
```diff
@@ -11,6 +11,16 @@ const NoParamsSchema = {
 }
 const NoParamsValidator = ajv.compile(NoParamsSchema)
 
+const CondoWebAppRedirectParamsSchema = {
+    type: 'object',
+    properties: {
+        url: { type: 'string' },
+        target: { type: 'string' },
```
SavelevMatthew, line 18: `{ "const": "_blank" }`?

apps/condo/domains/common/components/PostMessageProvider/globalHandlers.ts
```diff
@@ -138,4 +140,23 @@ export const useUpdateProgressBarHandler: () => RequestHandler<'CondoWebAppUpdat
...
         // TODO(DOMA-5171): Adding miniAppTaskUIInterface in deps causing rerender hell!
     }, [userId])
+}
+
+export const useRedirectHandler: () => RequestHandler<'CondoWebAppRedirect'> = () => {
+    const router = useRouter()
+
+    return useCallback(({ url, target }) => {
+        let success = false
+        if (isSafeUrl(url)) {
```
SavelevMatthew, line 150: if not safe need to throw exeption

apps/condo/domains/common/components/PostMessageProvider/globalHandlers.ts
```diff
@@ -138,4 +140,23 @@ export const useUpdateProgressBarHandler: () => RequestHandler<'CondoWebAppUpdat
         return { updated: true }
         // TODO(DOMA-5171): Adding miniAppTaskUIInterface in deps causing rerender hell!
     }, [userId])
+}
+
+export const useRedirectHandler: () => RequestHandler<'CondoWebAppRedirect'> = () => {
+    const router = useRouter()
+
+    return useCallback(({ url, target }) => {
```
SavelevMatthew, line 148: default target must be '_self'...possible undefined here

apps/condo/domains/common/components/PostMessageProvider/globalHandlers.ts
```diff
@@ -138,4 +140,23 @@ export const useUpdateProgressBarHandler: () => RequestHandler<'CondoWebAppUpdat
...
+        let success = false
+        if (isSafeUrl(url)) {
+            if (target === '_blank' && typeof window !== 'undefined') {
+                window.open(url, target)
+                success = true
+            } else if (target === '_self') {
+                router.push(url)
```
SavelevMatthew, line 155: Can we check here, that origin is same as current window?
So only condo / v1 redirects will be allowed in same window....

apps/condo/domains/common/components/PostMessageProvider/globalHandlers.ts
```diff
@@ -138,4 +141,34 @@ export const useUpdateProgressBarHandler: () => RequestHandler<'CondoWebAppUpdat
         return { updated: true }
         // TODO(DOMA-5171): Adding miniAppTaskUIInterface in deps causing rerender hell!
     }, [userId])
+}
+
+export const useRedirectHandler: () => RequestHandler<'CondoWebAppRedirect'> = () => {
+    const router = useRouter()
+
+    return useCallback(({ url, target = '_self' }) => {
+        let success = false
```
SavelevMatthew, line 150: redundant variable...Just return `success: true` or throw exception with message

apps/condo/domains/common/components/PostMessageProvider/globalHandlers.ts
```diff
@@ -138,4 +141,34 @@ export const useUpdateProgressBarHandler: () => RequestHandler<'CondoWebAppUpdat
...
+        if (!isSafeUrl(url)) {
+            throw new Error('The URL is not safe')
+        }
+
+        if (typeof window !== 'undefined') {
+            if (target === '_blank') {
+                window.open(url, target)
+                success = true
+            } else if (target === '_self') {
+                const urlOrigin = extractOrigin(url)
```
SavelevMatthew, line 160: what if I'll pass something like `/contacts`, will it work?

nomerdvadcatpyat: Yes, it's return `null`

SavelevMatthew: But it's allowed url, isn't it?

apps/condo/domains/common/components/PostMessageProvider/globalHandlers.ts
```diff
@@ -138,4 +141,31 @@ export const useUpdateProgressBarHandler: () => RequestHandler<'CondoWebAppUpdat
         return { updated: true }
         // TODO(DOMA-5171): Adding miniAppTaskUIInterface in deps causing rerender hell!
     }, [userId])
+}
+
+export const useRedirectHandler: () => RequestHandler<'CondoWebAppRedirect'> = () => {
+    const router = useRouter()
+
+    return useCallback(({ url, target = '_self' }) => {
+        if (!isSafeUrl(url)) {
```
SavelevMatthew, line 150: Lets it be more linear:
```
if (not window):
    throw ("No window object")
if (not safe)
    throw("Requested url is forbidden. ....")

if (blank)
    // logic
else if (not self url)
     throw("Requested url is forbidden. Only self same url allowed for in-window redirects")
else
     // logic

return success
```

apps/condo/domains/common/components/PostMessageProvider/globalHandlers.ts
nomerdvadcatpyat: Yes, in this case no error will be thrown

SavelevMatthew (approved):

PR #2753, opened 2022-12-29
DOMA-5056 add List component to ui kit

sitozzz:
image

### DIFF
--- a/packages/ui/src/components/List/list.tsx
+++ b/packages/ui/src/components/List/list.tsx
@@ -1,13 +1,14 @@
 import React, { useMemo } from 'react'
-import { List as DefaultList, ListProps as DefaultListProps } from 'antd'
-import classNames from 'classnames'
-import { Typography, TypographyTextProps } from '@open-condo/ui/src'
-import { ListItemProps as DefaultListItemProps } from 'antd/es/list'
+import { List as DefaultList } from 'antd'
+import type { ListProps as DefaultListProps } from 'antd'
+import { Typography } from '@open-condo/ui/src'
+import type { TypographyTextProps } from '@open-condo/ui/src'
+import type { ListItemProps as DefaultListItemProps } from 'antd/lib/list'
 
 const LIST_CLASS_PREFIX = 'condo-list'
 const ELLIPSIS_WORD_LENGTH = 80
 
-type ListDataSource = { label: string, value: string | number, valueType?: TypographyTextProps['type'] }
+type ListDataSource = { label: string, value: string, valueType?: TypographyTextProps['type'] }
 
 type CondoListProps = {
     title: string
@@ -33,7 +34,7 @@ const ListItem: IListItem = (props) => {
         return label.length > ELLIPSIS_WORD_LENGTH ? { tooltip: { children: label } } : false
     }, [label])
     const valueEllipsis = useMemo(() => {
-        return String(value).length > ELLIPSIS_WORD_LENGTH ? { tooltip: { children: value } } : false
+        return value.length > ELLIPSIS_WORD_LENGTH ? { tooltip: { children: value } } : false
     }, [value])
 
     return (
@@ -52,9 +53,7 @@ const ListItem: IListItem = (props) => {
 }
 
 const List: React.FC<ListProps> = (props) => {
-    const { children, className, title, ...restProps } = props
-
-    const classes = classNames({}, className)
+    const { children, title, ...restProps } = props
 
     return (
         <DefaultList
@@ -62,7 +61,6 @@ const List: React.FC<ListProps> = (props) => {
             header={<Typography.Title level={3}>{title}</Typography.Title>}
             prefixCls={LIST_CLASS_PREFIX}
             bordered={false}
-            className={classes}
             renderItem={(item, key) => (
                 <ListItem key={`${LIST_CLASS_PREFIX}-item-${key}`} item={item} />
             )}
--- a/packages/ui/src/components/List/list.tsx
+++ b/packages/ui/src/components/List/list.tsx
@@ -8,7 +8,7 @@ import type { ListItemProps as DefaultListItemProps } from 'antd/lib/list'
 const LIST_CLASS_PREFIX = 'condo-list'
 const ELLIPSIS_WORD_LENGTH = 80
 
-type ListDataSource = { label: string, value: string, valueTypographyType?: TypographyTextProps['type'] }
+type ListDataSource = { label: string, value: string, valueTextType?: TypographyTextProps['type'] }
 
 type CondoListProps = {
     title: string
@@ -28,7 +28,7 @@ interface IListItem {
 }
 
 const ListItem: IListItem = (props) => {
-    const { item: { label, value, valueTypographyType }, ...restProps } = props
+    const { item: { label, value, valueTextType }, ...restProps } = props
 
     const labelEllipsis = useMemo(() => {
         return label.length > ELLIPSIS_WORD_LENGTH ? { tooltip: { children: label } } : false
@@ -44,7 +44,7 @@ const ListItem: IListItem = (props) => {
                     {label}
                 </Typography.Text>
                 <div className='condo-list-item-divider'></div>
-                <Typography.Text ellipsis={valueEllipsis} type={valueTypographyType}>
+                <Typography.Text ellipsis={valueEllipsis} type={valueTextType}>
                     {value}
                 </Typography.Text>
             </>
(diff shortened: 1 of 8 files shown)


packages/ui/src/components/List/list.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useMemo } from 'react'
+import { List as DefaultList, ListProps as DefaultListProps } from 'antd'
+import classNames from 'classnames'
+import { Typography, TypographyTextProps } from '@open-condo/ui/src'
+import { ListItemProps as DefaultListItemProps } from 'antd/es/list'
```
SavelevMatthew, line 5: third-party import should be above first party...
and why from `/es` instead of `/lib`

packages/ui/src/components/List/list.tsx
```diff
@@ -0,0 +1,79 @@
+import React, { useMemo } from 'react'
+import { List as DefaultList, ListProps as DefaultListProps } from 'antd'
```
SavelevMatthew, line 2: please use `import type` syntax

packages/ui/src/components/List/list.tsx
```diff
@@ -0,0 +1,79 @@
...
+import classNames from 'classnames'
+import { Typography, TypographyTextProps } from '@open-condo/ui/src'
+import { ListItemProps as DefaultListItemProps } from 'antd/es/list'
+
+const LIST_CLASS_PREFIX = 'condo-list'
+const ELLIPSIS_WORD_LENGTH = 80
+
+type ListDataSource = { label: string, value: string | number, valueType?: TypographyTextProps['type'] }
```
SavelevMatthew, line 10: I would expect valueType to be number, string and so on...How about more clear name for specifying a specific text type?

SavelevMatthew, line 10: `string | number` seems strange. If you want to display any toString-able content here, just make it `string` or `React.ReactNode`...Probably first option, since you're casting it to string below

packages/ui/src/components/List/list.tsx
```diff
@@ -0,0 +1,79 @@
...
+                </Typography.Text>
+            </>
+        </DefaultListItem>
+    )
+}
+
+const List: React.FC<ListProps> = (props) => {
+    const { children, className, title, ...restProps } = props
+
+    const classes = classNames({}, className)
```
SavelevMatthew, line 57: just pass className directly, why merging empty state?)

packages/ui/src/components/List/style.less
```diff
@@ -0,0 +1,32 @@
+@import "antd/lib/list/style/index.less";
```
SavelevMatthew, line 1: seems like styles are unordered, you can turn on `stylelint` and enable `.less` files check in your IDE. Valid config can be found in monorepo root

packages/ui/src/components/List/style.less
```diff
@@ -0,0 +1,32 @@
...
+@import "antd/lib/tooltip/style/index.less";
+@import (reference) "@open-condo/ui/src/tokens/variables.less";
+
+.condo-list {
+  .condo-list-header {
+    border-bottom: none;
+    padding-bottom: 24px;
```
SavelevMatthew, line 8: `@condo-global-spacing-24` :)

packages/ui/src/components/List/style.less
```diff
@@ -0,0 +1,32 @@
...
+    max-width: 40%;
+  }
+
+  .condo-list-item .condo-typography:last-of-type {
+    text-align: right;
+  }
+
+  .condo-list-item-divider {
+    flex: 1;
+    border-bottom: 1px solid @condo-global-color-gray-3;
```
SavelevMatthew, line 29: `@condo-global-border-width-default`

packages/ui/src/components/List/style.less
```diff
@@ -0,0 +1,32 @@
...
+    margin: 0 8px;
```
SavelevMatthew, line 30: same about spacing

packages/ui/src/components/Typography/style.less
```diff
@@ -186,4 +187,22 @@
...
-}
\ No newline at end of file
+}
+
+.condo-tooltip > .condo-tooltip-content {
+  .condo-tooltip-inner {
+    font-family: @condo-global-font-family-default;
+    font-size: @condo-global-font-size-text-s;
+    line-height: @condo-global-typography-text-s-line-height;
+    padding: 12px;
```
SavelevMatthew, line 197: use token here please

packages/ui/src/components/Typography/style.less
```diff
@@ -186,4 +187,22 @@
...
+    background-color: @condo-global-color-white;
+    color: @condo-global-color-black;
+    filter: drop-shadow(0px 20px 65px rgba(208, 216, 225, 0.24)) drop-shadow(0px 4px 15px rgba(208, 216, 225, 0.28)) drop-shadow(0px 2px 6px rgba(208, 216, 225, 0.16));
```
SavelevMatthew, line 200: Where did this value came from?

packages/ui/src/components/Typography/text.tsx
```diff
@@ -18,19 +21,27 @@ export type TypographyTextProps = Omit<DefaultTextProps,
...
         [`${TYPOGRAPHY_CLASS_PREFIX}-${size}`]: size,
     })
     // NOTE: Used wrapper destructuring to explicitly pass component props, which is marked as internal in antd
     const componentProps = { component: 'span' }
+    // NOTE: We should pass default config anyway if any prop type was provided because prefixCls should exist at result prop
+    let ellipsisProp: TypographyTextProps['ellipsis'] = false
+    if (typeof ellipsis === 'boolean' && ellipsis) {
+        ellipsisProp = DEFAULT_ELLIPSIS_CONFIG
+    } else if (typeof ellipsis === 'object') {
+        ellipsisProp = merge(DEFAULT_ELLIPSIS_CONFIG, ellipsis)
```
SavelevMatthew, line 36: Why util function for something JS can do natively?

packages/ui/src/index.ts
```diff
@@ -28,6 +28,10 @@ export { Tag } from './components/Tag'
 export type { TagProps } from './components/Tag'
 
 export { Typography } from './components/Typography'
+
+export { List } from './components/List'
```
SavelevMatthew, line 32: Please place it in Alphabetical order, it's a common practice for an entry files...

packages/ui/src/components/List/list.tsx
sitozzz: Yup, that's artefact after refactoring

packages/ui/src/components/Typography/style.less
sitozzz: From figma I guess...

packages/ui/src/components/Typography/text.tsx
sitozzz: That's not a shallow copy of object... It's a deep copy.
Natively copy object with this case would work like this:
```js
const shallowCopy = { 
    ...{ item: { nestedValue: { nestedProp: 1 } } }, 
    ...{ item: { nestedValue: { anotherProp: 2 } } } 
}
console.log(shallowCopy)
{
    "item": {
        "nestedValue": {
            "anotherProp": 2
        }
    }
}
```

And in the case of using the utility, we will get a deep copy

```js
const deepCopy = _.merge({ item: { nestedValue: { nestedProp: 1 } } }, { item: { nestedValue: { anotherProp: 2 } } })

console.log(deepCopy)
{
    "nestedValue": {
        "nestedProp": 1,
        "anotherProp": 2
    }
}

```

packages/ui/src/index.ts
sitozzz: That's not consistent with rule of import order in component file...
In component we should order imports another way.
Can it be done with combination of eslint rules?

packages/ui/src/components/Typography/text.tsx
SavelevMatthew: Ok, reasonable....but then DEFAULT_ELLIPSIS_CONFIG must go second, so it cannot be overwritten

packages/ui/src/index.ts
SavelevMatthew: I guess index.ts will be autogenerated later...but for now it's done by that to be easy to orient...

You can Ctrl+Click on lib to jump to it's type declarations, so UI-kit have only 1 big entry file...That's why it's important to be able to scan it for needed component easily...And that's why component-level imports is not that important (at least for now)

packages/ui/src/components/List/list.tsx
```diff
@@ -28,7 +28,7 @@ interface IListItem {
 }
 
 const ListItem: IListItem = (props) => {
-    const { item: { label, value, valueType }, ...restProps } = props
+    const { item: { label, value, valueTypographyType }, ...restProps } = props
```
SavelevMatthew, line 31: how about `valueTextType `?

SavelevMatthew (approved):
🥇

toplenboren (approved):

PR #2754, opened 2022-12-30
Feat/condo/doma 4429/new schema bank transaction

AntonAL:


apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -18,6 +18,8 @@ async function checkOrganizationPermission (userId, organizationId, permission)
         organization: { id: organizationId },
     })
 
+    console.debug('>>>>> employeeRole', employeeRole)
```
VKislov, line 21: Seems forgotten

apps/condo/domains/banking/schema/BankTransaction.test.js
```diff
@@ -0,0 +1,557 @@
...
+
+                await expectToThrowAuthenticationErrorToObjects(async () => {
+                    await BankTransaction.getAll(anonymous, { id: objCreated.id }, { sortBy: ['updatedAt_DESC'] })
+                })
+            })
+        })
+    })
+
+    describe('Validation tests', () => {
+        test('Should have correct dv field (=== 1)', async () => {
```
VKislov, line 509: It seems to be redundant. The `DvAndSender` plugin has a validator for this. If you try to create in a test with a different `Dv` value, an error will be thrown. You can see here `packages/keystone/plugins/dvAndSender.js:128`

apps/condo/domains/banking/schema/BankTransaction.test.js
```diff
@@ -0,0 +1,557 @@
...
+                })
+                const [contractorAccount] = await createTestBankContractorAccount(support, organization)
+
+                await expectToThrowAccessDeniedErrorToObj(async () => {
+                    await createTestBankTransaction(support, account, contractorAccount, integrationContext, organization)
+                })
+            })
+
+            test('user can if it is an employee of organization with "canManageBankTransactions" permission', async () => {
+                const userClient = await makeClientWithNewRegisteredAndLoggedInUser()
```
VKislov, line 86: Perhaps it's better to put it in `beforeAll`. Each test creates a new `organization` and `employee` instance. It seems that in this case there will be no collisions in different tests when using the same user and different organizations

VKislov: Same for tests below

apps/condo/domains/banking/access/BankTransaction.js
```diff
@@ -0,0 +1,47 @@
...
+const { checkPermissionInUserOrganizationOrRelatedOrganization, queryOrganizationEmployeeFor,
+    queryOrganizationEmployeeFromRelatedOrganizationFor,
+} = require('@condo/domains/organization/utils/accessSchema')
+const { getById } = require('@open-condo/keystone/schema')
+
+async function canReadBankTransactions ({ authentication: { item: user } }) {
+    if (!user) return throwAuthenticationError()
+    if (user.deletedAt) return false
+
+    if (user.isAdmin) return {}
```
sitozzz, line 16: Why not simply boolean value?

sitozzz (approved):

VKislov (approved):

apps/condo/domains/banking/schema/BankTransaction.test.js
VKislov: Yes, it is better to create an organization and other entities related to it clean. I am looking at userClient. It seems that userClient can be the same in different organizations.

nomerdvadcatpyat (approved):

PR #2764, opened 2023-01-11
[PROOF OF WORK ⚒️ ] feat: DOMA-4654 Condo bridge client

SavelevMatthew:
https://user-images.githubusercontent.com/50069872/211710144-d0303531-b13e-4c68-a95d-4f0e96c207fe.mov

image

image


packages/bridge/src/bridge.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const ALL_METHODS = [
+    ...(IS_BROWSER_CLIENT ? WEB_SPECIFIC_METHODS : []),
+]
+
+const webBridge: WebBridge | undefined = IS_BROWSER_CLIENT
+    ? parent
+    : undefined
+
+export function createCondoBridge (): CondoBridge {
```
dkoviazin, line 28: Can we extend CondoBridge from EventEmitter?

packages/bridge/src/promisify.ts
```diff
@@ -0,0 +1,85 @@
...
+    RequestIdType,
+    AnyRequestMethodName,
+    RequestId,
+    RequestParams,
+    CondoBridgeSubscriptionListener,
+    AnyResponseMethodName,
+    ResultResponseData,
+} from './types/bridge'
+
+type PromiseController = {
```
dkoviazin, line 11: How it differs from type Promise ?

dkoviazin (approved):

packages/bridge/src/browser.ts
```diff
@@ -0,0 +1,5 @@
+import bridge from './index'
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
sitozzz, line 4: You can define property of window with brackets syntax. In this case you don't need to add ts-ignore comment line

packages/bridge/src/browser.ts
```diff
@@ -0,0 +1,5 @@
...
+window.condoBridge = bridge
```
sitozzz, line 5:
```suggestion
window['condoBridge'] = bridge
```

AleX83Xpert: The safest way: https://lodash.com/docs/4.17.15#set

SavelevMatthew: It'a bit weird to increase bundledize with lodash for this purpose

packages/bridge/src/promisify.ts
```diff
@@ -0,0 +1,85 @@
...
+        current: 0,
+        next () {
+            return ++this.current
+        },
+    }
+}
+
+function createRequestResolver () {
+    const requestIdCounter = createCounter()
+    const controllers: Record<RequestIdType, PromiseController | null> = {}
```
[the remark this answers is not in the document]
SavelevMatthew: Yep...null is not necessary here anymore :)

packages/bridge/src/promisify.ts
SavelevMatthew: Promise itself is similar to async function....and inside a controller we extract `resolve` and `reject` functions and pass it `requestResolver`

SavelevMatthew: So I need not type of Promise, but its `resolve` and `reject` functions...Did not found clear solution of doing it :(

packages/bridge/src/browser.ts
sitozzz: Another way is declaring namespace with custom window object. But the simplest way is not use dot syntax here

packages/bridge/src/promisify.ts
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }
+    })
+
+    return function promisifiedSend<Method extends AnyRequestMethodName> (
+        method: Method,
+        params: RequestParams<Method> & RequestId = {} as RequestParams<Method> & RequestId
+    ): Promise<Method extends AnyResponseMethodName ? ResultResponseData<Method> : void> {
+        return new Promise((resolve, reject) => {
+            const requestId = requestResolver.add({ resolve, reject }, params.requestId)
```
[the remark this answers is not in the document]
SavelevMatthew: Yep...thought about it, that's why it's just a proof of work for now 😄 
Added 1 sec timeout...Must be enough for now for all client-client processes

packages/bridge/src/promisify.ts
```diff
@@ -0,0 +1,85 @@
...
+    AnyResponseMethodName,
+    ResultResponseData,
+} from './types/bridge'
+
+type PromiseController = {
+    resolve: (value: any) => unknown
+    reject: (reason: unknown) => unknown
+}
+
+function createCounter () {
```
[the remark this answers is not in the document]
SavelevMatthew: Generator functions has return type of `T | void`, so if I'll use generator like:
```js
function* gen() {
    let counter = 0
    while(true) {
        yield counter++
    } 
}
```
I need to do one of the following:
1. Cast explicitly `int | void` to `int`
2. Check `gen.next()` for done each time, which also looks not clean enough
3. Suppress it with `@ts-ignore`

What's why used syntax with this

packages/bridge/src/bridge.ts
SavelevMatthew: Emitters are useful, when you're emit and subscribe to events from different part of code...In our cases this parts are on separate clients...And we cannot share emitter instance across that parts...So it's partlial emitter :)

packages/bridge/src/browser.ts
SavelevMatthew: image

SavelevMatthew: Found elegant way to extend window interface

packages/bridge/src/promisify.ts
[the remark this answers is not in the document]
AleX83Xpert: Cool! Thanks!

AleX83Xpert (approved):

packages/bridge/src/browser.ts
```diff
@@ -0,0 +1,8 @@
+import type { CondoBridge } from './types/bridge'
+import bridge from './index'
+
+declare global {
```
sitozzz, line 4: 🥇

packages/bridge/src/promisify.ts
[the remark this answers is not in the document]
sitozzz: I suppose you can use IterableIterator<number> type for that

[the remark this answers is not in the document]
sitozzz: function* gen(): IterableIterator<number> {
    let counter = 0
    while(true) {
        yield counter++
    } 
}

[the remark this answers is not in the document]
sitozzz: And you don't need to check generator 'done' state because it's a infinite loop inside. And it won't done never

sitozzz (approved):

PR #2768, opened 2023-01-11
feat(condo): DOMA-4545 added improved search by addresses for tickets

Alllex202:
Problem: filtering by address through the search field only works if the query is entered with an accuracy of spaces / punctuation marks

Resolve: for filter by address, split the search line into components (words / numbers) and clear unnecessary words (abbreviations, short or common words). And filter by each remaining word


apps/condo/domains/ticket/utils/tables.utils.ts
```diff
@@ -67,4 +70,43 @@ export const getPropertyScopeFilter = () => {
...
+    return function getWhereQuery (search) {
+        if (!search || !isString(search)) return
+
+        const addresses = uniq(search
+            .replace(/[^а-яa-z0-9\s/]/ig, ' ')
+            .split(' ')
+            .filter((item) => {
+                if (!item || ADDRESS_SEARCH_STOP_WORDS.includes(item.toLowerCase())) return false
+                const startWithNumber = !Number.isNaN(Number.parseInt(item))
+                return !(!startWithNumber && item.length < 3)
```
nomerdvadcatpyat, line 101:
```suggestion
                return startWithNumber || item.length >= 3
```

apps/condo/domains/common/constants/stopWords.js
```diff
@@ -0,0 +1,40 @@
...
+    'территория',
+    'респ',
+    'республика',
+    'р-н',
+    'р-он',
+    'район',
+]
+
+// NOTE: If you really need, you can add stop words for other languages
+const ADDRESS_SEARCH_STOP_WORDS = [
```
SavelevMatthew, line 34: How about moving this in aggregate `index` file?

apps/condo/domains/ticket/utils/tables.utils.ts
```diff
@@ -67,4 +70,43 @@ export const getPropertyScopeFilter = () => {
...
+ *          ],
+ *      },
+ *  }
+ */
+export const getFilterAddressForSearch = (): FilterType => {
+    return function getWhereQuery (search) {
+        if (!search || !isString(search)) return
+
+        const addresses = uniq(search
+            .replace(/[^а-яa-z0-9\s/]/ig, ' ')
```
SavelevMatthew, line 96: image

`\w` - any alphanumeric char
`u` flag - unicode chars

apps/condo/domains/ticket/utils/tables.utils.ts
SavelevMatthew: De Morgan be like: 🤘😎

apps/condo/domains/common/constants/stopWords.js
Alllex202: is it really worth splitting the contents of this file?

apps/condo/domains/ticket/utils/tables.utils.ts
Alllex202: In js it works differently

Снимок экрана 2023-01-20 в 12 25 02

apps/condo/domains/common/constants/stopWords.js
SavelevMatthew: For now it's optional and ok. But it can become unreadable on scale

apps/condo/domains/ticket/utils/tables.utils.ts
SavelevMatthew: Then use more complex solution:
image
Source:
https://stackoverflow.com/questions/74611855/unicode-flag-not-working-for-regex-in-javascript

apps/condo/domains/ticket/utils/tables.utils.ts
```diff
@@ -67,4 +70,43 @@ export const getPropertyScopeFilter = () => {
...
+ *          ],
+ *      },
+ *  }
+ */
+export const getFilterAddressForSearch = (): FilterType => {
+    return function getWhereQuery (search) {
+        if (!search || !isString(search)) return
+
+        const addresses = uniq(search
+            .replace(/[^\p{Alphabetic}\p{Decimal_Number}\s/]/igu, ' ')
```
SavelevMatthew, line 96: Moving this to const can be more efficient in terms of future fixes / refactoring

apps/condo/domains/ticket/utils/tables.utils.ts
```diff
@@ -67,4 +70,43 @@ export const getPropertyScopeFilter = () => {
...
+            .split(' ')
```
SavelevMatthew, line 97: You can also split by regex (\s), source: https://bobbyhadz.com/blog/javascript-split-string-by-regex

SavelevMatthew (approved):

nomerdvadcatpyat (approved):

PR #2770, opened 2023-01-13
feat(DOMA-5106): unified postMessage provider

SavelevMatthew:
## What the hell is this thing?
`PostMessageProvider` is unified hub, which goal is to process all incoming messages from different sources and manage different handlers in single  place.
## Ok, what sources and messages?
- Analytics events from UI-components inside condo
- Analytics events from miniapps sent via UI-kit
- All bridge events
## And what about possibilities?
- By default allowedOrigins is only current `serverUrl`, you can add any trusted source by using `registerOrigin` on component (iFrame) mounting, and removing it `resetOrigin` on unmounting
- You can register local handlers for this origin....For example events, affecting state such as Window resizing
- You can also register global handlers by specifying `"*"`origin... Example: processing notifications somewhere else, or processing analytic events inside analytics context
## Other benefits?
No need to stack all handlers inside Global or usual Iframe. Only need for now is local state-changing handlers

---
image

### DIFF
--- a/apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
+++ b/apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
@@ -6,7 +6,6 @@ import omit from 'lodash/omit'
 import get from 'lodash/get'
 import type {
     AllRequestMethods,
-    RequestParams,
     RequestHandler,
     ClientErrorResponse,
     RequestId,
@@ -29,7 +28,6 @@ type OriginHandlers = Partial<{ [Method in AllRequestMethods]: RequestHandler<Me
 type RegisterHandler = <Method extends AllRequestMethods>(
     event: Method,
     origin: '*' | string,
-    paramsSchema: JSONSchemaType<RequestParams<Method>>,
     handler: RequestHandler<Method>
 ) => void
 
@@ -113,7 +111,7 @@ export const PostMessageProvider: React.FC = ({ children }) => {
     const [registeredHandlers, setRegisteredHandlers] = useState<Record<HandlerOrigin, OriginHandlers>>({})
     const isOnClient = typeof window !== 'undefined'
 
-    const registerHandler: RegisterHandler = useCallback((event, origin, paramsSchema, handler) => {
+    const registerHandler: RegisterHandler = useCallback((event, origin, handler) => {
         setRegisteredHandlers(prev => {
             return {
                 ...prev,
@@ -184,10 +182,11 @@ export const PostMessageProvider: React.FC = ({ children }) => {
                             ...result,
                             ...(typeof requestId !== 'undefined' ? { requestId } : null),
                         },
-                    })
+                    }, event.origin)
                 } catch (err) {
                     return event.source.postMessage(
-                        getClientErrorMessage('HANDLER_ERROR', err.message, requestId, method)
+                        getClientErrorMessage('HANDLER_ERROR', err.message, requestId, method),
+                        event.origin
                     )
                 }
             } else {
--- a/apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
+++ b/apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
@@ -36,10 +36,10 @@ type RegisterHandler = <Method extends AllRequestMethods>(
  */
 type IPostMessageContext = {
     allowedOrigins: Readonly<Array<string>>
-    registerOrigin: (origin: string) => void
-    resetOrigin: (origin: string) => void
+    addOrigin: (origin: string) => void
+    removeOrigin: (origin: string) => void
     handlers: Readonly<Record<HandlerOrigin, OriginHandlers>>
-    registerHandler: RegisterHandler
+    addEventHandler: RegisterHandler
     validators: Readonly<ValidatorsType>
 }
 
@@ -48,10 +48,10 @@ type IPostMessageContext = {
  */
 const PostMessageContext = createContext<IPostMessageContext>({
     allowedOrigins: [serverUrl],
-    registerOrigin: () => ({}),
-    resetOrigin: () => ({}),
+    addOrigin: () => ({}),
+    removeOrigin: () => ({}),
     handlers: {},
-    registerHandler: () => ({}),
+    addEventHandler: () => ({}),
     validators,
 })
 /**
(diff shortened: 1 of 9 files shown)


apps/condo/domains/common/components/PostMessageProvider/errors.ts
```diff
@@ -0,0 +1,20 @@
...
+export const UNKNOWN_ERROR = 'UNKNOWN_ERROR'
+export const UNKNOWN_METHOD = 'UNKNOWN_METHOD'
+export const INVALID_PARAMETERS = 'INVALID_PARAMETERS'
+export const HANDLER_ERROR = 'HANDLER_ERROR'
+export const TIMEOUT_REACHED = 'TIMEOUT_REACHED'
+
+export const ERROR_REASONS = [ACCESS_DENIED, UNKNOWN_ERROR, UNKNOWN_METHOD, INVALID_PARAMETERS, HANDLER_ERROR, TIMEOUT_REACHED] as const
+export type ErrorReason = typeof ERROR_REASONS[number]
+export const ERROR_CODES: { [reason in ErrorReason]: number } = {
+    ACCESS_DENIED:0,
```
AleX83Xpert, line 12:
```suggestion
    ACCESS_DENIED: 0,
```

apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
```diff
@@ -0,0 +1,223 @@
...
+    const resetOrigin = useCallback((origin: string) => {
+        setAllowedOrigins((prev) => prev.filter(element => element !== origin))
+        setRegisteredHandlers((prev) => omit(prev, origin))
+    }, [])
+
+    const handleMessage = useCallback((event: MessageEvent) => {
+        if (!event.isTrusted || !event.source ||
+            event.source instanceof ServiceWorker || event.source instanceof MessagePort) {
+            return
+        }
```
AleX83Xpert, line 139:
```suggestion
        if (
          !event.isTrusted
          || !event.source
          || event.source instanceof ServiceWorker
          || event.source instanceof MessagePort
        ) {
            return
        }
```

apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+        // If not valid message interface - skip. It can be system events, like webpack or some other legacy messages
+        if (validateMessage(event.data)) {
+            const message = event.data
+            const { requestId, ...params } = message.params
```
AleX83Xpert, line 144: Is the `requestId` guaranteed?
Maybe add an error or warning in the case of no `requestId`?
```suggestion
            const { requestId = null, ...params } = message.params
            
            if (!requestId) {
              // ...err
            }
```

apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
```diff
@@ -0,0 +1,223 @@
...
+    const [registeredHandlers, setRegisteredHandlers] = useState<Record<HandlerOrigin, OriginHandlers>>({})
+    const isOnClient = typeof window !== 'undefined'
+
+    const registerHandler: RegisterHandler = useCallback((event, origin, handler) => {
+        setRegisteredHandlers(prev => {
+            return {
+                ...prev,
+                [origin]: {
+                    ...prev[origin],
+                    [event]: handler,
```
AleX83Xpert, line 120: There is an ability to set only one handler for each event. Is it ok?
What if we'll want to process the same event differently in different parts of the code?

AleX83Xpert (approved):

apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
```diff
@@ -0,0 +1,223 @@
...
+import get from 'lodash/get'
+import type {
+    AllRequestMethods,
+    RequestHandler,
+    ClientErrorResponse,
+    RequestId,
+    RequestIdType,
+} from './types'
+import type { ValidatorsType } from './validators'
+import type{ ErrorReason } from './errors'
```
Alllex202, line 15:
```suggestion
import type { ErrorReason } from './errors'
```

apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
SavelevMatthew: Then it's not clear which one should be used as a response...

apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
```diff
@@ -0,0 +1,225 @@
...
+ * 1. Making sure messages are trusted and coming from allowed Windows
+ * 2. Making sure message can be handled and passing parameters are valid
+ * 3. Handling errors thrown by individual handlers and converting it to corresponding postMessage response
+ */
+export const PostMessageProvider: React.FC = ({ children }) => {
+    const [allowedOrigins, setAllowedOrigins] = useState<Array<string>>([serverUrl])
+    const [registeredHandlers, setRegisteredHandlers] = useState<Record<HandlerOrigin, OriginHandlers>>({})
+    const isOnClient = typeof window !== 'undefined'
+
+    const registerHandler: RegisterHandler = useCallback((event, origin, handler) => {
```
AntonAL, line 114: What about to make naming consistent with DOM, like `add…`/`…remove` instead of `register…`/`reset…`?
For example: `addEventHandler`, `removeEventHandler`, `addOrigin`, `removeOrigin`.

apps/condo/domains/common/components/PostMessageProvider/validators.ts
```diff
@@ -0,0 +1,19 @@
+import type { RequestParamValidator, AllRequestMethods } from './types'
+import Ajv from 'ajv'
+
+const ajv = new Ajv()
+
+const CondoWebAppResizeWindowParamsSchema = {
+    type: 'object',
+    properties: {
+        height: { type: 'number' },
```
AntonAL, line 9: What about width?

AntonAL (approved):

apps/condo/domains/common/components/PostMessageProvider/validators.ts
SavelevMatthew: Width for iframes is always 100% for now

PR #2772, opened 2023-01-13
feat(condo):DOMA-5083 add invalidServicesError field to BillingReceipt

abshnko:
Before: 
services validity in a BillingReceipt wasn't checked

Now:
1. added a new field `invalidServicesError`
2. if the sum of all services doesn't add up to the total of the receipt (field toPay) --> `invalidServicesError = 'error...'`
3. if the sum adds up correctly --> `invalidServicesError = null`

Also, AllResidentBillingReceiptsService now will have `services=null` if mentioned error is not null (so that mobile app users won't see invalid services)


apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -115,6 +121,22 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
         auth: true,
     },
     hooks: {
+        resolveInput: async ({ resolvedData, existingItem }) => {
+            const newItem = { ...existingItem, ...resolvedData }
+            let servicesTotal = 0
+
+            for (let service of newItem.services){
+                servicesTotal += parseFloat(service.toPay)
```
toplenboren, line 129: Please use Big.js here, usage of parsefloat can lead to rounding errors

apps/condo/domains/billing/schema/BillingReceipt.test.js
```diff
@@ -40,6 +40,7 @@ const {
 } = require('@open-condo/keystone/test.utils')
 const { makeClient } = require('@open-condo/keystone/test.utils')
 const { WRONG_TEXT_FORMAT, UNEQUAL_CONTEXT_ERROR } = require('@condo/domains/common/constants/errors')
+const { services } = require('@condo/domains/billing/utils/testData/servicesData')
```
toplenboren, line 43: Do we have anything like this in other domains? 

I'd move it simly to testutils

apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -105,6 +105,12 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
             kmigratorOptions: { null: true, on_delete: 'models.CASCADE' },
         },
 
+        invalidServicesError:  {
```
toplenboren, line 108: Maybe we can use virtual field for that?

apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -115,6 +121,22 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
         auth: true,
     },
     hooks: {
+        resolveInput: async ({ resolvedData, existingItem }) => {
+            const newItem = { ...existingItem, ...resolvedData }
```
toplenboren, line 125: Better to move it to some function

toplenboren (changes requested):

apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -106,6 +106,27 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
...
+            resolver: async (item) => {
+                let servicesTotal = 0
+
+                if (get(item, 'services')){
+                    for (let service of get(item, 'services')){
+                        servicesTotal = Big(get(service, 'toPay')).plus(Big(servicesTotal))
+                    }
+                }
+
+                const isServicesValid = Big(servicesTotal).cmp(Big(get(item, 'toPay'))) === 0
```
toplenboren, line 122: servicesTotal is already a `Big` type value, so you might not need to cast it to `Big` here

apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -106,6 +106,27 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
             kmigratorOptions: { null: true, on_delete: 'models.CASCADE' },
         },
 
+        invalidServicesError:  {
+            schemaDoc: 'Indicates if services are valid and add up to total sum toPay.',
+            type: Virtual,
+            isRequired: false,
+            resolver: async (item) => {
+                let servicesTotal = 0
```
toplenboren, line 114: I think its better to initialize it as `Big(0)` to indicate, that you are going to do some calculations here

apps/condo/domains/billing/schema/BillingReceipt.js
toplenboren: Also toPay might be initialized just after the `servicesTotal` initialization line. So it would look like this:

```
let servicesTotal = Big(0)
const toPay = Big(get(item, 'toPay')

...

// not sure if this will work, but i think it looks better this way
const servicesAreValid = servicesTotal.cmp(toPay) === 0

```

apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -106,6 +106,27 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
             kmigratorOptions: { null: true, on_delete: 'models.CASCADE' },
         },
 
+        invalidServicesError:  {
+            schemaDoc: 'Indicates if services are valid and add up to total sum toPay.',
+            type: Virtual,
+            isRequired: false,
+            resolver: async (item) => {
```
toplenboren, line 113: Also, lets rename item to receipt, i think its more readable this way

apps/condo/domains/billing/utils/testSchema/index.js
```diff
@@ -873,6 +851,24 @@ async function makeResidentClientWithOwnReceipt(existingResidentClient) {
     }
 }
 
+function generateServicesData(count=3, toPay=''){
```
toplenboren, line 854: Its best to add a docstring that tells you how to use this function

apps/condo/migrations/20230214175344-0221_billingreceipthistoryrecord_invalidserviceserror.js
```diff
@@ -0,0 +1,26 @@
+// auto generated by kmigrator
+// KMIGRATOR:0221_billingreceipthistoryrecord_invalidserviceserror:IyBHZW5lcmF0ZWQgYnkgRGphbmdvIDQuMS4zIG9uIDIwMjMtMDItMTQgMTI6NTMKCmZyb20gZGphbmdvLmRiIGltcG9ydCBtaWdyYXRpb25zLCBtb2RlbHMKCgpjbGFzcyBNaWdyYXRpb24obWlncmF0aW9ucy5NaWdyYXRpb24pOgoKICAgIGRlcGVuZGVuY2llcyA9IFsKICAgICAgICAoJ19kamFuZ29fc2NoZW1hJywgJzAyMjBfYXV0b18yMDIzMDIxM18xMDU1JyksCiAgICBdCgogICAgb3BlcmF0aW9ucyA9IFsKICAgICAgICBtaWdyYXRpb25zLkFkZEZpZWxkKAogICAgICAgICAgICBtb2RlbF9uYW1lPSdiaWxsaW5ncmVjZWlwdGhpc3RvcnlyZWNvcmQnLAogICAgICAgICAgICBuYW1lPSdpbnZhbGlkU2VydmljZXNFcnJvcicsCiAgICAgICAgICAgIGZpZWxkPW1vZGVscy5KU09ORmllbGQoYmxhbms9VHJ1ZSwgbnVsbD1UcnVlKSwKICAgICAgICApLAogICAgXQo=
+
+exports.up = async (knex) => {
```
toplenboren, line 4: BillingReceipt is the (perhaps?) biggest table in all condo. So this migration would fail on production. 

The reason for failing would be migration time exceeding the time limit set by the postgres on SQL operations (i think its 10 seconds or so....) 

You need to add a bigger time limit in the beginiing of migration and set it back to 10 seconds in the end of migration. 

Same applies to exports.down migration as well...

Please check out this migration for example: apps/condo/migrations/20220727181205-0156_ticket_lastcommentat_ticketchange_lastcommentatfrom_and_more.js

apps/condo/migrations/20230214175344-0221_billingreceipthistoryrecord_invalidserviceserror.js
```diff
@@ -0,0 +1,26 @@
...
+    await knex.raw(`
+    BEGIN;
+--
+-- Add field invalidServicesError to billingreceipthistoryrecord
+--
+ALTER TABLE "BillingReceiptHistoryRecord" ADD COLUMN "invalidServicesError" jsonb NULL;
```
toplenboren, line 10: Why it is jsonb ?

toplenboren (changes requested):

toplenboren (commented):
Looking good. Please do two more things:

1. Write a test on AllResidentBillingReceipts (do not show services if they are invalid)
2. Make branch up to date and regenerate migrations

toplenboren (approved):

apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -106,6 +106,28 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
...
+            resolver: async (receipt) => {
+                let servicesTotal = Big(0)
+                const toPay = Big(get(receipt, 'toPay'))
+
+                if (get(receipt, 'services')){
+                    for (let service of get(receipt, 'services')){
+                        servicesTotal = Big(get(service, 'toPay')).plus(Big(servicesTotal))
+                    }
+                }
+
```
dkoviazin, line 122: You can do it using array.reduce
```
const services = get(receipt, 'services', [])
const servicesTotal = services.reduce((total, { toPay = 0 }) => Big(total).add(Big(toPay)), Big(0))
```

apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -106,6 +106,28 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
...
+                const servicesAreValid = servicesTotal.cmp(toPay) === 0
+
+                if (!servicesAreValid) return `Services sum (${servicesTotal}) does not add up to the toPay (${toPay}) amount correctly`
```
dkoviazin, line 125: is it ok to have servicesAreValid = false if there are no services?

apps/condo/domains/billing/schema/BillingReceipt.test.js
```diff
@@ -915,6 +883,18 @@ describe('BillingReceipt', () => {
                 expect(receipt).toHaveProperty(['recipient', 'tin'], recipient.tin)
                 expect(receipt).toHaveProperty(['receiver', 'id'], recipient.id)
             })
+            test('Should validate that services add up to the total sum (toPay)', async () => {
```
dkoviazin, line 886: Please, add positive test and test when receipt has no services

apps/condo/domains/billing/schema/BillingReceipt.js
abshnko: thank you, fixed it so now if `services` are an empty array `invalidServicesError` will be null

apps/condo/domains/billing/schema/BillingReceipt.js
abshnko: reduce is very cool 😎 thanks!

PR #2791, opened 2023-01-20
feat(condo): DOMA-3534 Bridge integration

SavelevMatthew:
https://user-images.githubusercontent.com/50069872/213642206-a6c0a523-d09d-447e-970a-4aa575c6660c.mov


apps/condo/domains/miniapp/components/AppIndex/B2BAppPage.tsx
```diff
@@ -8,11 +8,11 @@ import { useOrganization } from '@open-condo/next/organization'
 import { Typography } from '@open-condo/ui'
 import LoadingOrErrorPage from '@condo/domains/common/components/containers/LoadingOrErrorPage'
 import { PageContent, PageWrapper, PageHeader } from '@condo/domains/common/components/containers/BaseLayout'
-import IFrame from '@condo/domains/common/components/IFrame'
+// import IFrame from '@condo/domains/common/components/IFrame'
```
nomerdvadcatpyat, line 11:
```suggestion
```

apps/condo/domains/miniapp/components/GlobalApps/GlobalAppsContainer.tsx
```diff
@@ -324,22 +100,14 @@ export const GlobalAppsContainer: React.FC = () => {
...
         <>
             {appUrls.map((url, index) => (
-                <GlobalIframe
-                    key={`${get(user, 'id', null)}-${url}`}
-                    pageUrl={url}
-                    ref={el => iframeRefs.current[index] = el}
+                <IFrame
+                    key={url}
+                    src={url}
+                    reloadScope='user'
```
nomerdvadcatpyat, line 106: what if I need to reload the global miniapp when changing organization?

apps/condo/domains/common/components/PostMessageProvider/globalHandlers.ts
```diff
@@ -0,0 +1,138 @@
...
+import { useOrganization } from '@open-condo/next/organization'
+import { STAFF } from '@condo/domains/user/constants/common'
+import { useMiniappTaskUIInterface } from '@condo/domains/common/hooks/useMiniappTaskUIInterface'
+import { TASK_STATUS, TasksContext } from '@condo/domains/common/components/tasks'
+import type { RequestHandler } from './types'
+
+export const handleNotification: RequestHandler<'CondoWebAppShowNotification'> = (params) => {
+    const { type, ...restParams } = params
+    notification[type](restParams)
+    return { success: true }
```
AleX83Xpert, line 18: Always successful result? Is there an ability to fail? What if `notification[type]` is not defined?

apps/condo/domains/common/components/PostMessageProvider/globalHandlers.ts
```diff
@@ -0,0 +1,138 @@
...
+export const useLaunchParamsHandler: () => RequestHandler<'CondoWebAppGetLaunchParams'> = () => {
+    const { locale } = useIntl()
+    const { user } = useAuth()
+    const { organization } = useOrganization()
+    const userId = get(user, 'id', null)
+    const organizationId = get(organization, 'id', null)
+    return useCallback(() => {
+        return {
+            condoUserId: userId,
+            condoUserType: STAFF,
```
AleX83Xpert, line 30: Why not `user.type`? For now, it's `stuff`, yes. But in future its possible place for the bug.

apps/condo/domains/miniapp/components/AppFrameWrapper.tsx
```diff
@@ -0,0 +1,29 @@
...
+  body {
+    height: auto;
+  }
+`
+
+export const AppFrameWrapper: React.FC = ({ children }) => {
+    useEffect(() => {
+        const observer = new ResizeObserver((entries) => {
+            if (entries && entries.length) {
+                bridge.send('CondoWebAppResizeWindow', { height: entries[0].target.clientHeight })
```
AleX83Xpert, line 15: `height` only? What about width?

packages/bridge/src/types/methods/index.ts
```diff
@@ -0,0 +1,48 @@
...
+import type { GetLaunchParamsParams, GetLaunchParamsData } from './GetLaunchParams'
+export type { GetLaunchParamsParams, GetLaunchParamsData } from './GetLaunchParams'
+import type { ResizeWindowParams, ResizeWindowData } from './ResizeWindow'
+export type { ResizeWindowParams, ResizeWindowData } from './ResizeWindow'
+import type { ShowNotificationParams, ShowNotificationData } from './ShowNotification'
+export type { ShowNotificationParams, ShowNotificationData } from './ShowNotification'
+import type { ShowProgressBarParams, ShowProgressBarData } from './ShowProgressBar'
+export type { ShowProgressBarParams, ShowProgressBarData } from './ShowProgressBar'
+import type { UpdateProgressBarParams, UpdateProgressBarData } from './UpdateProgressBar'
+export type { UpdateProgressBarParams, UpdateProgressBarData } from './UpdateProgressBar'
```
AleX83Xpert, line 13: imports are mixed with exports. Not good for auto-sorting.

apps/condo/domains/miniapp/components/AppFrameWrapper.tsx
SavelevMatthew: Width is always 100%.
Maybe you can specify small width by yourself, but in most cases it's redundant

apps/condo/domains/common/components/PostMessageProvider/globalHandlers.ts
SavelevMatthew: It's a like platform-dependent const, and change like this will include global product decision.

So, we'll need to change Bridge API -> which will cause changing this place to change.

For now const is enough here

apps/condo/domains/miniapp/components/GlobalApps/GlobalAppsContainer.tsx
SavelevMatthew: Like completely reload?
When we should probably add some mark in schema, and use it here

apps/condo/domains/common/components/PostMessageProvider/globalHandlers.ts
SavelevMatthew: Please see handle logic...

We use ajv to validate params....so there (inside handler) will always be correct params (in terms of Bridge API) ['info', 'success', 'error', 'warning']. Even if not and this part failed somehow it will throw error. 

All event handlers are wrapped in `try...catch` block, so it will catch error and send error event with type `HANDLER_ERROR`

AleX83Xpert (approved):

nomerdvadcatpyat (approved):

apps/miniapp/domains/common/utils/oidcAuth.tsx
```diff
@@ -25,36 +25,37 @@ const OidcAuthProvider = ({ children }) => {
     const NoAccessToStorageMessage = intl.formatMessage({ id: 'NoAccessToStorage' })
     const AskForAccessButtonMessage = intl.formatMessage({ id: 'AskForAccessButton' })
 
-    const { organization: condoOrganization, isLoading: isOrganizationLoading } = useOrganization()
-    const { result: condoUser, isLoading: isCondoUserLoading } = condoBridge.useUser()
+    // const { organization: condoOrganization, isLoading: isOrganizationLoading } = useOrganization()
+    // const { result: condoUser, isLoading: isCondoUserLoading } = condoBridge.useUser()
```
sitozzz, line 29:
```suggestion
```

apps/miniapp/domains/common/utils/oidcAuth.tsx
```diff
@@ -25,36 +25,37 @@ const OidcAuthProvider = ({ children }) => {
     const NoAccessToStorageMessage = intl.formatMessage({ id: 'NoAccessToStorage' })
     const AskForAccessButtonMessage = intl.formatMessage({ id: 'AskForAccessButton' })
 
-    const { organization: condoOrganization, isLoading: isOrganizationLoading } = useOrganization()
-    const { result: condoUser, isLoading: isCondoUserLoading } = condoBridge.useUser()
+    // const { organization: condoOrganization, isLoading: isOrganizationLoading } = useOrganization()
```
sitozzz, line 28:
```suggestion
```

PR #2795, opened 2023-01-23
feat(global): DOMA-5185 Turborepo for dependency-management

SavelevMatthew:
## Stage 1. Local cache

https://user-images.githubusercontent.com/50069872/213964083-988853d1-6abd-45ea-8236-3c0a2abd90fa.mov


AleX83Xpert (approved):

.github/workflows/nodejs.address-service.test.yml
```diff
@@ -52,6 +52,7 @@ jobs:
       - name: install packages
         working-directory: ./apps/address-service
         run: |
+          npm i -g turbo
```
sitozzz, line 55: Why not `yarn global add turbo`?

sitozzz (approved):

.github/workflows/nodejs.address-service.test.yml
SavelevMatthew: yarn 2 does not support global :(

SavelevMatthew: Only dlx (similar to npx)

PR #2805, opened 2023-01-23
refactor(condo): DOMA-5149 refactoring syncOrganization in SBBOL integration

VKislov:
Refactoring that removes the ability to work with deleted entities in order to avoid possible errors and update the code.
If you delete a user who had an sbbol account, and then log into the system through sbbol authorization, you will get an infinite page 500 renderer, because the further code does not see these entities. So that such errors would not be repeated, I refactored using the same utilities as in the further part of the system.

### DIFF
--- a/apps/condo/domains/organization/integrations/sbbol/sync/syncOrganization.js
+++ b/apps/condo/domains/organization/integrations/sbbol/sync/syncOrganization.js
@@ -1,12 +1,13 @@
-const { updateItem, getItems } = require('@keystonejs/server-side-graphql-client')
-const { uniqBy } = require('lodash')
+
+const get = require('lodash/get')
 
 const conf = require('@open-condo/config')
 const { getSchemaCtx } = require('@open-condo/keystone/schema')
 
 const { CUSTOMER_IMPORTANT_NOTE_TYPE } = require('@condo/domains/notification/constants/constants')
 const { sendMessage } = require('@condo/domains/notification/utils/serverSchema')
-const { REGISTER_NEW_ORGANIZATION_MUTATION } = require('@condo/domains/organization/gql.js')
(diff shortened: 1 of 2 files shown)


VKislov:
^^ I don't understand why it helped, but it works well. The test passed locally, but it failed in CI if you use the admin account

apps/condo/domains/organization/integrations/sbbol/sync/syncOrganization.js
```diff
@@ -96,56 +72,57 @@ const createOrganization = async ({ context, user, organizationInfo }) => {
...
-                    organization: {
-                        id: importedOrganization.id,
-                    },
-                    name: 'employee.role.Administrator.name',
+            const allRoles = await OrganizationEmployeeRole.getAll(adminContext, {
+                organization: {
+                    id: importedOrganization.id,
                 },
-                returnFields: returnFields,
+                name_in: 'employee.role.Administrator.name',
```
dkoviazin, line 123: name_in is for array of possible variants 
you can use simple
```
name: 'employee.role.Administrator.name',
```

apps/condo/domains/organization/integrations/sbbol/sync/syncOrganization.js
```diff
@@ -66,18 +50,10 @@ const createOrganization = async ({ context, user, organizationInfo }) => {
...
-        item: {
-            id: organization.id,
-            data: {
-                ...importInfo,
-                ...dvSenderFields,
-            },
-        },
-        returnFields: 'id',
-        ...context,
+    const { obj: createdOrganization } = data
```
dkoviazin, line 53:
```
const { data: { obj: createdOrganization } } = await userContext.executeGraphQL({
...
```

dkoviazin (approved):

PR #2822, opened 2023-01-27
DOMA-5209 fix isExpired method in SbbolSecretStorage

VKislov:
The `keyStorage.ttl` asynchronous function without `await` returns a `Promise` in the `pending` state. Therefore, the logic for `getAccessTokenForUser` did not work correctly and could not update the `accessToken`

### DIFF
--- a/apps/condo/domains/organization/integrations/sbbol/SbbolSecretStorage.js
+++ b/apps/condo/domains/organization/integrations/sbbol/SbbolSecretStorage.js
@@ -11,6 +11,7 @@ const SBBOL_REDIS_KEY_PREFIX = 'SBBOL'
 
 // Real TTL is 180 days, but it is updated every time the accessToken expires and we get a new pair
 const REFRESH_TOKEN_TTL_DAYS = 180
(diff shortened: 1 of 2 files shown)


sitozzz (approved):

apps/condo/domains/organization/integrations/sbbol/SbbolSecretStorage.js
```diff
@@ -130,7 +130,7 @@ class SbbolSecretStorage {
 
     async #isExpired (key) {
         // NOTE: `TTL` Redis command returns -2 if the key does not exist, -1 if the key exists but has no associated expire
-        return !this.keyStorage.ttl(this.#scopedKey(key)) > 0
+        return !((await this.keyStorage.ttl(this.#scopedKey(key))) > 0)
```
AleX83Xpert, line 133: Simpler a bit :)
```suggestion
        return (await this.keyStorage.ttl(this.#scopedKey(key))) <= 0
```

AleX83Xpert (approved):

PR #2826, opened 2023-01-27
DOMA-5149 change constraint in model and add deletedAt to gql for UserExternalIdentity

VKislov:
Changed the `UserExternalIdentity` `constraint` so that the check does not consider deleted entries. Added `deletedAt` to `gql` so that it would be possible to filter the search for entities. Added cleaning the entities associated with the user who is being reset via `ResetUserService`. 
Also fixed the problem of the infinite rerender with a 500 error when trying to go behind a deleted user.


apps/condo/domains/user/schema/ResetUserService.js
```diff
@@ -109,6 +110,16 @@ const ResetUserService = new GQLCustomSchema('ResetUserService', {
                     await OrganizationEmployee.softDelete(context, employee.id, { dv: 1, sender })
                 }
 
+                const accordingUserExternalIdentity = await UserExternalIdentity.getAll(context, {
```
pahaz, line 113: You need to add a test for that

apps/condo/domains/organization/integrations/sbbol/sync/syncUser.js
```diff
@@ -74,6 +74,7 @@ const syncUser = async ({ context: { context, keystone }, userInfo, identityId }
     const identityWhereStatement = {
         identityId,
         identityType: SBBOL_IDP_TYPE,
+        deletedAt: null,
```
pahaz, line 77: You don't need to use deleted at?

apps/condo/domains/common/components/UseDeskWidget.tsx
```diff
@@ -28,24 +28,29 @@ const UseDeskWidget: React.FC = () => {
...
-                            id: useDeskFieldsIdsMap.role, value: get(link, ['role', 'name'], null),
-                        },
-                    ],
-            })
+        if (UseDeskWidgetId && isFunction(userIdentify)) {
+            const name = get(link, 'name')
+            const email = get(user, 'email')
+            const phone = get(user, 'phone')
+            userIdentify(
+                {
```
pahaz, line 36:
```suggestion
            userIdentify({
```

apps/condo/domains/organization/integrations/sbbol/sync/syncUser.js
VKislov: Screenshot 2023-02-01 at 12 50 12

VKislov: without it I get a deleted instance

pahaz: Looks like a bug! You should not use it

sitozzz (approved):

PR #2834, opened 2023-02-01
fix(condo): DOMA-5218 registerResident mutation updated

AleX83Xpert:
Additionally, search existing resident by address key.


apps/condo/domains/resident/schema/RegisterResidentService.js
```diff
@@ -35,8 +41,20 @@ const RegisterResidentService = new GQLCustomSchema('RegisterResidentService', {
                     unitType,
                     user: { connect: { id: context.authedItem.id } },
                 }
+
+                // TODO(AleX83Xpert): Think about how to prevent copying this condition :-/
+                const client = conf.NODE_ENV === 'test' || get(conf, 'ADDRESS_SERVICE_CLIENT_MODE') === 'fake'
```
sitozzz, line 46: What if client can't be found in conf? Should this case been handled with an special error?

AleX83Xpert: Resident and Property models can't work without address service and without config field. If there is no config for address service, property and resident creation will fail.

sitozzz (approved):

apps/condo/domains/resident/schema/RegisterResidentService.js
dkoviazin, line 46: You can move this logic to createAddressServiceClientInstance function with all checks for NODE_ENV and ADDRESS_SERVICE_CLIENT_MODE
Reasons:
1. RegisterResidentService will have less code and checks (for address service environment)
2. Now you  need to copy-paste this logic to every place where you is using addressService

dkoviazin (approved):

apps/condo/domains/resident/schema/RegisterResidentService.js
AleX83Xpert: Unfortunately, functions have different signatures :(
The real client needs only the address service URL. The test client needs an object.

Alllex202 (approved):

apps/condo/domains/resident/schema/RegisterResidentService.js
pahaz: But you can change it

pahaz (approved):

PR #2850, opened 2023-02-06
feat(condo): DOMA-5252 added isAutoClassified field to Ticket schema

nomerdvadcatpyat:
In the future, we may need to distinguish auto-classified tickets from those classified by a human


apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -220,6 +220,12 @@ const Ticket = new GQLListSchema('Ticket', {
             knexOptions: { isNotNullable: false },
             kmigratorOptions: { null: true, on_delete: 'models.PROTECT' },
         },
+        isAutomaticallyClassified: {
```
akarjakin, line 223: maybe `isAutoClassified` ?

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -419,6 +425,7 @@ const Ticket = new GQLListSchema('Ticket', {
                 const classifierResult = await classifyTicket(context, resolvedData.details)
 
                 resolvedData.classifier = get(classifierResult, 'id')
+                resolvedData.isAutomaticallyClassified = true
```
sitozzz, line 428: Why this flag set's to true on every request? Shouldn't it depend on the result above?

apps/condo/domains/ticket/schema/Ticket.js
nomerdvadcatpyat: Yes, it will be shorter.
Renamed

Alllex202 (approved):

apps/condo/domains/ticket/schema/Ticket.js
nomerdvadcatpyat: Tickets created or updated from CRM are considered to be classified manually, because the operator has validated this classifier.
Fixed: now `isAutoClassified: true` is passed only if this ticket has `isAutoClassified: false`.

sitozzz (approved):
Now looks good

PR #2625, opened 2022-11-28
DOMA-4305 feat syncClientAccounts with SBBOL

VKislov:
When authorizing through SBBOL, after initializing the organization, fetch extended information about the organization of the SBBOL user. I use this information to instantiate the `BankAccount` entity. I save additional data in the json `meta` field. At the moment, synchronization involves only tracking new accounts in SberBusiness. If an invoice is received that is not stored in our system, it will be created, if a copy is found, nothing will be changed. Synchronization currently occurs only during authorization

PR #2855, opened 2023-02-06
feat(condo): DOMA-4305 not required BankAccount.property field

AntonAL:
In upcoming feature to sync accounts from SBBOL it will be technically unknown to what property the fetched account belongs. A user will be prompted to manually connect properties to bank accounts.

Needed for https://github.com/open-condo-software/condo/pull/2625


sitozzz (approved):

sitozzz (changes requested):
Please update migration number before merge

VKislov (approved):

sitozzz (approved):

PR #2856, opened 2023-02-06
feature(rb): DOMA-5253 added userId field to acquiring payment logs

dkoviazin:
1. added userId field to acquiring payment logs for analytics
2. rebuild migrations to match user domain from condo


sitozzz (approved):

sitozzz:
Why not `Uuid` type for that field?

PR #2860, opened 2023-02-07
DOMA-5263 fixed unhandled exception in SBBOL completeAuth

VKislov:
photo_2023-02-07 11 45 14

### DIFF
--- a/apps/condo/domains/organization/integrations/sbbol/routes.js
+++ b/apps/condo/domains/organization/integrations/sbbol/routes.js
@@ -39,7 +39,8 @@ class SbbolRoutes {
         const reqId = req.id
         const sbbolAuthApi = await initializeSbbolAuthApi()
         try {
-            if (!isObject(req.session[SBBOL_SESSION_KEY].checks)) {
+            const checks = get(req, ['session', SBBOL_SESSION_KEY, 'checks'])
+            if (!isObject(checks)) {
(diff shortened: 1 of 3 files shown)


apps/condo/domains/organization/integrations/sbbol/routes.js
```diff
@@ -39,7 +39,8 @@ class SbbolRoutes {
         const reqId = req.id
         const sbbolAuthApi = await initializeSbbolAuthApi()
         try {
-            if (!isObject(req.session[SBBOL_SESSION_KEY].checks)) {
+            const checks = get(req, ['session', SBBOL_SESSION_KEY, 'checks'])
+            if (!isObject(checks)) {
                 logger.info({ msg: 'SBBOL invalid nonce and state', reqId })
                 return res.status(400).send('ERROR: Invalid nonce and state')
```
AleX83Xpert, line 45: Invalid or empty?

AleX83Xpert (approved):

sitozzz (approved):

apps/condo/domains/organization/integrations/sbbol/routes.js
VKislov: Maybe, but I don't think it's that important. We generate them and put them in the session, that is, we are sure that they are definitely there. Deleting a session is not standard behavior for a user

PR #2861, opened 2023-02-07
fix(condo): DOMA-5259 open first contact address if no activeTab

nomerdvadcatpyat:
When opening a card on a call, a card is opened without the selected address.
Made it so that the first address is selected.


akarjakin (approved):

apps/condo/pages/phone/[number]/index.tsx
```diff
@@ -10,6 +10,7 @@ import { Col, Row, Typography } from 'antd'
 import { CarouselRef } from 'antd/es/carousel'
 import { Gutter } from 'antd/es/grid/row'
 import { EllipsisConfig } from 'antd/es/typography/Base'
+import { isEmpty } from 'lodash'
```
Comment (VKislov) [inline apps/condo/pages/phone/[number]/index.tsx:13]: Can you please import a specific module? We agreed to use the import of individual modules on the front end `import isEmpty from 'lodash/get'`

VKislov (approved):

PR #2866, opened 2023-02-08
DOMA-5264 add Select component to the UI-KIT

sitozzz:
This PR also fixes missing `source-map-loader` dependency that cause errors at production build stage

image

### DIFF
--- a/packages/ui/src/components/Select/select.tsx
+++ b/packages/ui/src/components/Select/select.tsx
@@ -22,7 +22,8 @@ export type OptionType = {
     textType?: TypographyTextProps['type']
 }
 
-export type SelectProps = Pick<DefaultSelectProps, 'disabled' | 'value' | 'onChange' | 'loading' | 'placeholder' | 'children'> & {
+export type SelectProps = Pick<DefaultSelectProps, 'disabled' | 'value' | 'onChange' | 'loading' | 'children'> & {
+    placeholder: string
     options?: Array<OptionType>
     displayMode?: 'fill-parent' | 'fit-content'
     type?: typeof SELECT_TYPES[number]
--- a/packages/ui/src/components/Select/select.tsx
+++ b/packages/ui/src/components/Select/select.tsx
@@ -41,8 +41,7 @@ const Select: ISelect = (props) => {
     })
 
     const handleChange = useCallback((value, option) => {
-        const title = option.title ? option.title : extractChildrenContent(option.children)
-
+        const title = extractChildrenContent(option.children)
         if (title) {
             sendAnalyticsSelectEvent('Select', { label: title, value })
         }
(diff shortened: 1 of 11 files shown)


packages/ui/src/components/Select/select.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+const Select: ISelect = (props) => {
+    const { options, children, displayMode = 'fill-parent', type, onChange, ...rest } = props
+    const className = classNames({
+        [`${SELECT_CLASS_PREFIX}-${displayMode}`]: displayMode,
+        [`${SELECT_CLASS_PREFIX}-${type}`]: type,
+    })
+
+    const handleChange = useCallback((value, option) => {
```
SavelevMatthew, line 43: Use strict types here, otherwise dev portal will not build (yeah, strange, I know) :)

But `@types/react@18` does not like implicit any :(

packages/ui/src/components/Select/style.less
```diff
@@ -0,0 +1,142 @@
...
+  &-disabled:not(.condo-select-customize-input) &-arrow {
+    color: @condo-global-color-gray-3;
+  }
+  &-open &-arrow {
+    transform: rotate(180deg);
+  }
+
+  &:not(.condo-select-customize-input) .condo-select-selector {
+    border-radius: @condo-global-border-radius-medium;
+    height: 46px;
```
SavelevMatthew, line 56: `46 % 4 != 0` how does it appears

packages/ui/src/components/Select/style.less
```diff
@@ -0,0 +1,142 @@
...
+  &-info &-selection-item > *,
+  &-secondary &-selection-item > *,
+  &-success &-selection-item > * {
+    color: @condo-global-color-white;
+  }
+
+  &-arrow {
+    position: absolute;
+    margin: 0;
+    top: 18px;
```
SavelevMatthew, line 43: Same about % 4, is it ok?

packages/ui/src/components/_utils/analytics.ts
```diff
@@ -23,12 +25,22 @@ type ComponentSpecificCheckEventProps = {
...
 
+type ComponentSpecificSelectEventProps = {
+    Select: { value: string, label: string }
+}
+
 
 
-type AnalyticsClickData<K extends keyof ComponentSpecificClickEventProps> = CommonAnalyticsProps<'click', K> & ComponentSpecificClickEventProps[K]
-type AnalyticsCheckData<K extends keyof ComponentSpecificCheckEventProps> = CommonAnalyticsProps<'check', K> & ComponentSpecificCheckEventProps[K]
+type AnalyticsClickData<K extends keyof ComponentSpecificClickEventProps> = CommonAnalyticsProps<'click', K>
```
SavelevMatthew, line 34: I guess we'll merge this :)

Since 1 component is 1 event usually....but for now it's ok

SavelevMatthew (approved):

packages/ui/src/components/Select/style.less
sitozzz: Thats because of border

packages/ui/src/components/Select/style.less
sitozzz: I guess this is related to icon container size

PR #2867, opened 2023-02-08
fix(condo): DOMA-5268 register multipayment for payments with different fees

dkoviazin:
In case we have different commissions for different categories inside one acquiringIntegrationContext 
For example, organization have implicitFee on housing category and explicit fee for overhaul category

### DIFF
--- a/apps/condo/domains/acquiring/schema/RegisterMultiPaymentService.test.js
+++ b/apps/condo/domains/acquiring/schema/RegisterMultiPaymentService.test.js
@@ -714,6 +714,49 @@ describe('RegisterMultiPaymentService', () => {
         })
     })
     describe('Real-life cases', () => {
+        test('Should correctly calculate commissions for different categories of payments', async () => {
+            const {
+                admin,
+                client,
+                acquiringIntegration,
+                acquiringContext,
+                billingReceipts,
+                serviceConsumer,
+            } = await makePayer(2)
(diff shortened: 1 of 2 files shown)


apps/condo/domains/acquiring/schema/RegisterMultiPaymentService.test.js
```diff
@@ -714,6 +714,49 @@ describe('RegisterMultiPaymentService', () => {
...
+            })
+            const payload = [
+                {
+                    serviceConsumer: { id: serviceConsumer.id },
+                    receipts: billingReceipts.map(({ id }) => ({ id })),
+                },
+            ]
+            const [{ multiPaymentId }] = await registerMultiPaymentByTestClient(client, payload)
+            const [multiPayment] = await MultiPayment.getAll(admin, { id: multiPaymentId })
+            expect(Big(multiPayment.explicitServiceCharge).toFixed(2)).toEqual('12.10')
```
SavelevMatthew, line 758: Const is not obvious to refactor later? How about `amount1 * percentage + amount2 * ...`

SavelevMatthew (approved):

apps/condo/domains/acquiring/schema/RegisterMultiPaymentService.test.js
dkoviazin: it has a complex formula from fee calculator. Don't know how to explain it here. Fee calculator has it's own tests for this case

PR #2881, opened 2023-02-10
Fix/condo/doma-5283/remove nojson logs

AleX83Xpert:
### Before
image

### After
image


sitozzz (approved):

toplenboren (approved):

packages/keystone/fields/Json/Implementation.js
```diff
@@ -58,6 +64,29 @@ class JsonImplementation extends Implementation {
         return this.extendGraphQLTypes
     }
 
+    gqlAuxFieldResolvers ({ schemaName }) {
```
pahaz, line 67: It looks like a hardcode! Package should know nothing about the apps.

AleX83Xpert: Refactored at https://github.com/open-condo-software/condo/pull/2936

PR #2885, opened 2023-02-11
use debug package instead of hardcoded env checks

pahaz:
Hide useless debug output:

```
Transform Organization GQLListSchema by preprocessors
✔ schemaDocPreprocessor
✔ escapeSearchPreprocessor
Transform OrganizationEmployeeHistoryRecord GQLListSchema by preprocessors
✔ schemaDocPreprocessor
✔ escapeSearchPreprocessor
Transform OrganizationEmployee GQLListSchema by preprocessors
✔ schemaDocPreprocessor
✔ escapeSearchPreprocessor
Transform OrganizationEmployeeRoleHistoryRecord GQLListSchema by preprocessors
✔ schemaDocPreprocessor
✔ escapeSearchPreprocessor
```

### DIFF
--- a/packages/keystone/package.json
+++ b/packages/keystone/package.json
@@ -39,6 +39,7 @@
     "crypto": "^1.0.1",
     "cuid": "^2.1.8",
     "dayjs": "^1.10.6",
+    "debug": "^4.3.4",
     "emittery": "^0.7.1",
     "encoding": "^0.1.0",
     "ensure-error": "^3.0.1",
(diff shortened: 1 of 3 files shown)


AntonAL (commented):
Why we don't rendering these log statements in production?
Why we need an extra package for this tiny place of code?

dkoviazin (approved):

pahaz:
> Why we don't rendering these log statements in production? Why we need an extra package for this tiny place of code?

We already have this package in our lock file. It already used inside knex, prisma, express and so on. It's not so small check the examples. It's better to use this package rather than just print this useless logs every run.

PR #2895, opened 2023-02-14
DOMA-5323 extend analytics with organization id

sitozzz:


apps/condo/domains/common/components/TrackingContext.tsx
```diff
@@ -229,6 +229,7 @@ const TrackingProvider: React.FC = ({ children }) => {
             if (link) {
                 trackingProviderValueRef.current.userProperties['role'] = get(link, 'role.name')
                 trackingProviderValueRef.current.userProperties['organization'] = get(link, 'organization.name')
```
AleX83Xpert, line 231: I know it's too late, but
```suggestion
                trackingProviderValueRef.current.userProperties['organizationName'] = get(link, 'organization.name')
```

AleX83Xpert (approved):

apps/condo/domains/common/components/TrackingContext.tsx
sitozzz: Yeah, originally was no plan to send id there. So change key right now will break all of the created reports

akarjakin (approved):

PR #2903, opened 2023-02-15
feat(dev-portal): DOMA-5232 Proof of concept: dev-portal (docs)

SavelevMatthew:


apps/dev-portal/domains/docs/hooks/useMenuItems.tsx
```diff
@@ -0,0 +1,81 @@
...
+    nav: Array<NavItem>,
+    locale: string | undefined,
+    currentRoute: string,
+    entryEndpoint: string,
+    currentTitleLevel: TitleLevel = 4,
+    maxTitleLevel: TitleLevel = 5,
+): Array<MenuItem> {
+    const nextTitleLevel = currentTitleLevel < maxTitleLevel
+        ? currentTitleLevel + 1 as TitleLevel
+        : maxTitleLevel
```
AleX83Xpert, line 24:
```suggestion
    const nextTitleLevel = Math.min(currentTitleLevel + 1, maxTitleLevel) as TitleLevel
```

packages/ui/.storybook/main.ts
```diff
@@ -23,9 +23,10 @@ const config: StorybookConfig = {
     },
     'staticDirs': [{ from: '../public', to: '/ui' }],
     'webpackFinal': async (config) => {
-        const configRules = config && config.module ? config.module.rules : []
+        const configRules = config && config.module && config.module.rules ? config.module.rules : []
```
AleX83Xpert, line 26:
```suggestion
import get from 'lodash/get'
...
        const configRules = get(config, ['module', 'rules'], [])
```

packages/ui/src/components/Select/select.tsx
```diff
@@ -40,7 +40,7 @@ const Select: ISelect = (props) => {
         [`${SELECT_CLASS_PREFIX}-${type}`]: type,
     })
 
-    const handleChange = useCallback((value, option) => {
+    const handleChange = useCallback((value: any, option: any) => {
```
AleX83Xpert, line 43: `string | number | null` instead of `any`, as described within `OptionType`
Or create a custom type for an option value `type OptionValueType = string | number | null`

AleX83Xpert (approved):

apps/dev-portal/domains/docs/hooks/useMenuItems.tsx
SavelevMatthew: fixed 👍

apps/dev-portal/next.config.js
```diff
@@ -1,16 +1,75 @@
...
+        const rules = get(config, ['module', 'rules'], [])
+
+        const newRules = rules.map(rule => {
+            if (!rule.oneOf) {
+                return rule
+            }
+
+            rule.oneOf = rule.oneOf.map(option => {
+                if (option && option.test && typeof option.test.test === 'function' &&
+                    option.test.test('my.module.css') && Array.isArray(option.use)) {
```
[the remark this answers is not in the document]
SavelevMatthew: Webpack rules have optional `test` field which is `RegExp`, filtering files on which specific chain of loaders will apply.
So `test.test(...)` is equal to `fileRegexp.test(...)`. And `my.module.css` is just a random file name, representing css module

apps/dev-portal/pages/docs/[...path].tsx
```diff
@@ -0,0 +1,237 @@
...
+
+    const openPaths = useMemo(() => {
+        const parts = currentRoute.split('/').filter(Boolean)
+        const result = []
+        for (let i = 0; i < parts.length - 2; ++i) {
+            const route = `/${parts.slice(0, parts.length - (i + 1)).join('/')}`
+            result.push(route)
+        }
+        return result
+    }, [currentRoute])
```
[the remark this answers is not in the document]
Comment (SavelevMatthew) [reply apps/dev-portal/pages/docs/[...path].tsx:97]: Since it's not received here. `Router.route` and `Router.pathname` returns `/docs/[...path]` which is not informative :)

packages/ui/src/components/Select/select.tsx
SavelevMatthew: Option has children here...And analytics receives only strings...So idk about this type implementation. Strict typing here needed for dev-portal to build (yeah, magic :) )

packages/ui/src/components/Select/select.tsx
```diff
@@ -40,10 +40,11 @@ const Select: ISelect = (props) => {
         [`${SELECT_CLASS_PREFIX}-${type}`]: type,
     })
 
-    const handleChange = useCallback((value: any, option: any) => {
+    // TODO(DOMA-5347): Handle this value and option properly
```
SavelevMatthew, line 43: @sitozzz Left a todo here....Would be awesome to fix this some day :)

AleX83Xpert (approved):

toplenboren (approved):

PR #2911, opened 2023-02-16
fix(condo) HOTFIX: Prohibit users to create two billing integration ctx

toplenboren:
This PR will forcefully prohibit users to create two billing contexts. You can no longer create one INACTIVE or PAUSED context, and one in CONNECTED status. 

Only the case with HIDDEN + CONNECTED will work.

### DIFF
--- a/apps/condo/domains/billing/schema/BillingIntegrationOrganizationContext.js
+++ b/apps/condo/domains/billing/schema/BillingIntegrationOrganizationContext.js
@@ -85,7 +85,7 @@ const BillingIntegrationOrganizationContext = new GQLListSchema('BillingIntegrat
     hooks: {
         validateInput: async ({ operation, resolvedData, addValidationError }) => {
             // should have only one explicit (hidden = false) context in organization!
-            if (operation === 'create' && resolvedData['status'] === CONTEXT_FINISHED_STATUS) {
+            if (operation === 'create') {
                 const integration = await getById('BillingIntegration', get(resolvedData, ['integration']))
                 if (get(integration, 'isHidden') === true) { return } // we can add as many b2c integrations as we like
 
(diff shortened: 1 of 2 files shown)


apps/condo/domains/billing/schema/BillingIntegrationOrganizationContext.js
```diff
@@ -85,7 +85,7 @@ const BillingIntegrationOrganizationContext = new GQLListSchema('BillingIntegrat
     hooks: {
         validateInput: async ({ operation, resolvedData, addValidationError }) => {
             // should have only one explicit (hidden = false) context in organization!
-            if (operation === 'create' && resolvedData['status'] === CONTEXT_FINISHED_STATUS) {
```
SavelevMatthew, line 88: Not the right way to do it!

Client must be allowed to request connecting another integration...

SavelevMatthew (changes requested):

apps/condo/domains/billing/schema/BillingIntegrationOrganizationContext.js
SavelevMatthew: Case is simple:
1. Client is using billing A (it has `Finished`) context
2. Client want to connect billing B instead (see https://confluence.in.doma.ai/pages/viewpage.action?pageId=1214678029)
3. We're creating second context with "InProgress" status, which allows marketing to get a new deal in AMO via meta base integration

SavelevMatthew: So, instead of restricting creating second context we should fix filtering in `useObject` by adding status filter:
https://github.com/open-condo-software/condo/blob/master/apps/condo/pages/billing/index.tsx#L29

toplenboren: Why? Right now we do not support two billing integrations. We only support stragne UPS case, but its a HIDDEN billing + CONNECTED billing

PR #2913, opened 2023-02-16
Feat/address service/doma 4705/values overriding

AleX83Xpert:
### Before
Let's look at the address model
image
When we ask for this model we receive the next json value
image

### After
Now let's override the `qc` field
image
Pay attention that the original `qc` value is still null
And ask the service again
image
Also, look at the end. There are original values kept here
image


apps/address-service/domains/address/schema/Address.js
```diff
@@ -34,6 +36,28 @@ const Address = new GQLListSchema('Address', {
...
+            type: Json,
+            isRequired: false,
+            access: {
+                create: access.canManageOverrides,
+                update: access.canManageOverrides,
+            },
+            hooks: {
+                validateInput: async (data) => {
+                    const { resolvedData, addFieldValidationError, existingItem, fieldPath } = data
+                    Object.entries(get(resolvedData, fieldPath, {}) || {}).forEach(([path, value]) => {
```
SavelevMatthew, line 51: What if I'll pass `[]` here? Guess object.enties will be broken

apps/address-service/domains/address/access/Address.js
```diff
@@ -20,11 +20,19 @@ async function canManageAddresses ({ authentication: { item: user }, originalInp
     return !!user.isAdmin
 }
 
+async function canManageOverrides ({ authentication: { item: user }, originalInput, operation, itemId }) {
+    if (!user) return throwAuthenticationError()
+    if (user.deletedAt) return false
+
+    return !!user.isAdmin || !!user.isSupport
```
SavelevMatthew, line 27: Optional, but I guess 1 babeled double reverse is better than 2 `!!(a || b)`

SavelevMatthew: )

apps/address-service/domains/address/schema/Address.js
```diff
@@ -34,6 +36,28 @@ const Address = new GQLListSchema('Address', {
...
+                        if (!has({ ...existingItem, ...resolvedData }, `${OVERRIDING_ROOT}.${path}`)) {
```
SavelevMatthew, line 52: I cannot add field to meta here?

Is it possible, that dadata returns `{a: 1, b: 2}` for 1 object, and `{a: 2}` for second? Or it will always be `b: null` in here?

SavelevMatthew (approved):

apps/address-service/domains/address/schema/Address.js
AleX83Xpert: Looks like it's no problem
https://user-images.githubusercontent.com/25384290/219364363-b5dbe5f7-fdf6-465d-ab76-b3c1c7819fd4.mov

apps/address-service/domains/address/schema/Address.js
AleX83Xpert: Yes, no new fields. Only overriding existing ones. If new fields will make sense I'll add the functionality.

Alllex202 (approved):

apps/address-service/domains/address/schema/Address.js
AleX83Xpert: Anyway, just improved validation and tests: https://github.com/open-condo-software/condo/pull/2913/commits/7021d91d9a495d818953eb9ce2c9c56bd966b6b2

apps/address-service/domains/address/schema/Address.js
AleX83Xpert: And this is already normalized data, so all fields will be there.

PR #2925, opened 2023-02-17
fix(condo): DOMA-5362 fixed timezone for ticket print to pdf

Alllex202:


pahaz (approved):

dkoviazin (approved):

apps/condo/domains/common/utils/timezone.js
```diff
@@ -13,6 +15,15 @@ const normalizeTimeZone = (timeZone) => {
     }
 }
 
+const dayjsWithTimezone = ({ date, timezone }) => {
```
SavelevMatthew, line 18: Why not just?
```
if (timeZone && normalizeTimeZone(timeZone)) {
    return dayjs(date).tz(timezone)
}

return dayjs(date)
```

apps/condo/domains/common/utils/timezone.js
```diff
@@ -13,6 +15,15 @@ const normalizeTimeZone = (timeZone) => {
...
+    try {
+        return dayjs(date).tz(timezone)
+    } catch (e) {
```
SavelevMatthew, line 21: And also are we need this fallback? Maybe just handle this on front-end part?
It can cause some unpredicted behaviours potentially.

1. Client A somehow has correct time zone in one tab and incorrect in another
2. Client will use this util at almost the same time, but he'll get few hours diff between 2 objects...

apps/condo/domains/common/utils/timezone.js
Alllex202: Removed this function

Alllex202: Now I use the normalized or default timezone for generate blank

apps/condo/domains/ticket/utils/serverSchema/exportTicketBlanksToPdf.js
```diff
@@ -140,24 +148,24 @@ const getAddressPartsByTicket = ({ ticket, locale }) => {
...
     }))
 }
 
-const convertTicketToTicketBlank = ({ ticket, comments, blankOptions, locale }) => {
+const convertTicketToTicketBlank = ({ ticket, comments, blankOptions, locale, timeZone }) => {
     return {
         number: ticket.number,
-        printDate: dayjs().format('DD.MM.YYYY'),
-        createdAt: dayjs(ticket.createdAt).format('DD.MM.YYYY'),
+        printDate: formatDate(undefined, timeZone),
```
SavelevMatthew, line 161: Not best practice to pass undefined values followed by defined vars....

If some arg is optional, it's better to pass strict `null` value, or use simple object as args `{ timeZone: string, date?: string }`

SavelevMatthew (approved):

apps/condo/domains/ticket/utils/serverSchema/exportTicketBlanksToPdf.js
Alllex202: Fixed it

PR #2931, opened 2023-02-19
feature(condo): DOMA-5395 added upload file support

dkoviazin:
Modify basic integration client to support file upload

### DIFF
--- a/packages/apollo-server-client/index.js
+++ b/packages/apollo-server-client/index.js
@@ -3,7 +3,7 @@ const { BatchHttpLink } = require('@apollo/client/link/batch-http')
 const { RetryLink } = require('@apollo/client/link/retry')
 const { onError }  = require('apollo-link-error')
 const { createUploadLink } = require('apollo-upload-client')
-const fetchPolyFill = require('cross-fetch/polyfill')
+const fetch = require('cross-fetch/polyfill').fetch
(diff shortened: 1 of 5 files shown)


packages/apollo-server-client/index.js
```diff
@@ -3,15 +3,23 @@ const { BatchHttpLink } = require('@apollo/client/link/batch-http')
 const { RetryLink } = require('@apollo/client/link/retry')
 const { onError }  = require('apollo-link-error')
 const { createUploadLink } = require('apollo-upload-client')
-const fetchPoltfill = require('cross-fetch/polyfill')
+const fetchPolyFill = require('cross-fetch/polyfill')
```
pahaz, line 6: We already have node-fetch. Should we add it to package.json ?

pahaz:
Could you please add some usage example? Is it for tests?

packages/apollo-server-client/index.js
dkoviazin: node-fetch is not working with Apollo client on server side (as it is not provided poly fill entry point)
so to use node-fetch we need to add
```
const fetch = require('node-fetch')

if (!globalThis.fetch) {
    globalThis.fetch = fetch
}
```
to have fetch available globally

PR #2932, opened 2023-02-19
feature(condo): DOMA-5272 extend meter schema to support integration

dkoviazin:
Added importId + importRemoteSystem to identify meter from integration
Added meta field to store additional settings from integration


pahaz:
Should we remove BillingMeter ?

pahaz (approved):

apps/condo/domains/meter/schema/Meter.js
```diff
@@ -163,6 +164,21 @@ const Meter = new GQLListSchema('Meter', {
             isRequired: true,
             defaultValue: false,
         },
+        meta: {
+            schemaDoc: 'Meter metadata. Can be used to store additional settings from integrations',
```
SavelevMatthew, line 168:
```suggestion
            schemaDoc: 'Meter metadata. Can be used to store additional settings from external sources, such as billing integrations or miniapps',
```

apps/condo/domains/meter/schema/Meter.js
```diff
@@ -163,6 +164,21 @@ const Meter = new GQLListSchema('Meter', {
...
+            type: Json,
+            isRequired: false,
+        },
+        importRemoteSystem: {
+            schemaDoc: 'External provider for meter',
+            type: Text,
+            kmigratorOptions: { null: true, unique: false },
+        },
+        importId: {
```
SavelevMatthew, line 177: Store it in meta, that's the purpose of this field

apps/condo/domains/meter/schema/Meter.js
```diff
@@ -163,6 +164,21 @@ const Meter = new GQLListSchema('Meter', {
             isRequired: true,
             defaultValue: false,
         },
+        meta: {
+            schemaDoc: 'Meter metadata. Can be used to store additional settings from integrations',
+            type: Json,
+            isRequired: false,
+        },
+        importRemoteSystem: {
```
SavelevMatthew, line 172: Why we need it here? We have `MeterReadingSource` schema to clarify, where readings are came from...

SavelevMatthew (changes requested):
Cannot understand what's importId is doing outside of meta, which is the place for storing external information for apps / integrations / data providers and so on (according to our conventions)

It's also not clear to me, how our contractors was able to sync all their data without this changes...And then billings are creating specific fields there for your own needs. IMO they should have unified flow...Since we treat billings integrations same as miniapps

SavelevMatthew (approved):

PR #2953, opened 2023-02-23
DOMA-5427 create records BankCategory and BankCostItem

VKislov:


apps/condo/migrations/20230223172911-0226_manual_create_BankCostItem_records.js
```diff
@@ -0,0 +1,135 @@
...
+
+        COMMIT;
+    `)
+}
+
+exports.down = async (knex) => {
+    await knex.raw(`
+        BEGIN;
+        
+DELETE FROM "BankCostItem" WHERE id = '8b589a43-4f0e-446f-b519-775f3d28dc1e';
```
sitozzz, line 73: Tabulation?)

sitozzz (approved):

apps/condo/lang/en/en.json
```diff
@@ -1577,5 +1577,65 @@
...
+  "banking.costItem.maintenance_and_operation.name": "Maintenance and operation",
+  "banking.costItem.overhaul.name": "Overhaul",
+  "banking.costItem.water_supply.name": "Water supply",
+  "banking.costItem.compensation_for_damage.name": "Compensation for damage",
+  "banking.costItem.garbage_collection_and_disposal.name": "Garbage collection and disposal",
+  "banking.costItem.gas.name": "Gas",
+  "banking.costItem.deratization_and_disinfestation.name": "Deratization and disinfestation",
+  "banking.costItem.materials_for_repair_and_maintenance.name": "Materials for repair and maintenance",
+  "banking.costItem.purchase_of_machinery_and_equipment.name": "Purchase of machinery and equipment",
+  "banking.costItem.employee_wages.name": "Employee wages",
```
AntonAL, line 1607: What exactly do you mean by "wages".
Why not "salary"?

A salary is when you pay employees the same amount each pay period regardless of how much they work. In contrast, a wage is an hourly rate you pay employees based on how much time they put in during a pay period.

apps/condo/lang/ru/ru.json
```diff
@@ -1577,5 +1577,65 @@
...
+  "banking.costItem.purchase_of_machinery_and_equipment.name": "Закупка техники и оборудования",
+  "banking.costItem.employee_wages.name": "Заработная плата сотрудников",
+  "banking.costItem.rental_fee.name": "Квартплата",
+  "banking.costItem.concierges.name": "Консьержи",
+  "banking.costItem.materials_for_maintenance_and_current_repair.name": "Материалы для содержания и текущего ремонта",
+  "banking.costItem.taxes.name": "Налоги",
+  "banking.costItem.equipment.name": "Оборудование",
+  "banking.costItem.maintenance_of_access_control_systems.name": "Обслуживание систем контроля управления доступом",
+  "banking.costItem.intercom_service.name": "Обслуживание домофонов",
+  "banking.costItem.maintenance_and_inspection_of_metering_devices.name": "Обслуживание и проверка приборов учета",
```
AntonAL, line 1615: It's not obligatory to include exact words in key, because in this way it does deviates from pattern "maintenance_of…".
Maybe inspection is included in maintenance of elevators, ventilation etc, but it does not explicitly specified in key.
Simpler to name, simpler to work with.

apps/condo/lang/en/en.json
VKislov: of course I'll rename it to salary. I used a translator

apps/condo/lang/en/en.json
```diff
@@ -1577,5 +1577,65 @@
...
+  "banking.costItem.materials_for_maintenance_and_current_repair.name": "Materials for maintenance and current repair",
+  "banking.costItem.taxes.name": "Taxes",
+  "banking.costItem.equipment.name": "Equipment",
+  "banking.costItem.maintenance_of_access_control_systems.name": "Maintenance of access control systems",
+  "banking.costItem.intercom_service.name": "Intercom service",
+  "banking.costItem.maintenance_and_inspection_of_metering_devices.name": "Maintenance and inspection of metering devices",
+  "banking.costItem.maintenance_of_elevator_equipment.name": "Maintenance of elevator equipment",
+  "banking.costItem.maintenance_of_fireplaces_and_hearths.name": "Maintenance of fireplaces and hearths",
+  "banking.costItem.maintenance_of_ventilation_systems.name": "Maintenance of ventilation systems",
+  "banking.costItem.maintenance_of_video_surveillance_systems.name": "Maintenance of video surveillance systems",
```
AntonAL, line 1619: Please consider following difference between CCTV and surveillance systems

Key difference: Basically, CCTV and Surveillance Camera are video cameras working on advanced technologies. CCTV known as Closed-circuit television transmits a signal to a specific monitor located at a distant place; while Surveillance Camera often works on IP networks which connect the security (surveillance) cameras from the remote location to the main location

Is it sure a system of cameras, transmitting signals over IP networks?

apps/condo/lang/ru/ru.json
VKislov: too many keys similar in name, I'm not sure that I can greatly simplify them all and not lose their uniqueness and meaning

apps/condo/lang/en/en.json
VKislov: I also can't be sure if security systems use a closed data loop, so I called it by a more general name

apps/condo/migrations/20230223172911-0226_manual_create_BankCostItem_records.js
```diff
@@ -0,0 +1,135 @@
+exports.up = async (knex) => {
+    await knex.raw(`
+        BEGIN;
+        --create BankCategory
```
AntonAL, line 4: This comment does not provides additional information.
We clearly see lines containing `INSERT INTO "BankCategory"` statement, and then lines containing `INSERT INTO "BankCostItem"`.
No visual differentiation is needed here IMO.

AntonAL:
I see cost all cost items have `isOutcome` fields with `true` value.
AFAIK, we should also have a pair of income-related records.

apps/condo/lang/en/en.json
AntonAL: ok

apps/condo/lang/ru/ru.json
pahaz: image

sitozzz:
> I see cost all cost items have `isOutcome` fields with `true` value. AFAIK, we should also have a pair of income-related records.

The main reason is that the classifier does not return these categories now. It's not necessary to add all of the categories here for me

apps/condo/lang/en/en.json
```diff
@@ -1577,5 +1577,65 @@
...
+  "banking.category.maintenance_and_current_repair_of_common_property_in_an_apartment_building.name": "Maintenance and current repair of common property in an apartment building",
+  "banking.category.apartment_building_management.name": "Apartment building management",
+  "banking.category.overhaul.name": "Overhaul",
+  "banking.category.drainage.name": "Drainage",
+  "banking.category.water_supply.name": "Water supply",
+  "banking.category.garbage_collection_and_disposal.name": "Garbage collection and disposal",
+  "banking.category.gas.name": "Gas",
+  "banking.category.hot_water_supply.name": "Hot water supply",
+  "banking.category.additional_services.name": "Additional services",
+  "banking.category.thermal_energy.name": "Thermal energy",
```
sitozzz, line 1592: If third token of intl key is a constant - you need to use upper snake case for that

sitozzz: For example, 
```suggestion
  "banking.category.THERMAL_ENERGY.name": "Thermal energy",
```

apps/condo/lang/en/en.json
```diff
@@ -1577,5 +1577,65 @@
   "billing.category.internet.name": "Internet",
   "billing.category.radio.name": "Radio",
   "billing.category.intercom.name": "Intercom",
-  "billing.category.telecom.name": "Telecom"
+  "billing.category.telecom.name": "Telecom",
+  "banking.integration.SBBOL.name": "SBBOL",
+  "banking.integration.1CClientBankExchange.name": "1C",
+  "banking.category.maintenance_and_current_repair_of_common_property_in_an_apartment_building.name": "Maintenance and current repair of common property in an apartment building",
```
sitozzz, line 1583: It would be nice, if most of these keys will consist of three words

sitozzz (approved):
Now looks pretty good

AntonAL (approved):

PR #2962, opened 2023-02-27
Remove useless `Dockerfile` args and simplify `docker-compose.yml`

pahaz:
So, to simplify CI and local development we need to build one docker container for all apps. 

It's after then build container every app and easier for local development.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,6 +13,3 @@
 **/.idea
 **/.kmigrator
 **/docs
-**/.config
-**/.yarn
-**/.cache
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,11 +10,9 @@ COPY --from=node:14-buster /opt/ /opt/
 # > At the moment, setting "LANG=C" on a Linux system *fundamentally breaks Python 3*, and that's not OK.
 ENV LANG C.UTF-8
 # Add app user/group! Clean packages and fix links! Check version! And install some extra packages!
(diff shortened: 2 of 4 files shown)


.dockerignore
```diff
@@ -13,6 +13,3 @@
 **/.idea
 **/.kmigrator
 **/docs
-**/.config
-**/.yarn
```
sitozzz, line 17: Why you don't want to ignore these cache dirs? Is that for speed up build time? May this change cause bugs related to cache?

Dockerfile
```diff
@@ -28,31 +26,33 @@ RUN set -ex \
...
-    ls -lah /home/app/
+WORKDIR /app
+
+RUN echo "# Build time .env config!" >> /app/.env && \
+	echo "COOKIE_SECRET=undefined" >> /app/.env && \
+	echo "DATABASE_URL=undefined" >> /app/.env && \
+	echo "REDIS_URL=undefined" >> /app/.env && \
+	echo "NODE_ENV=production" >> /app/.env
+
+ADD --chown=app:app . /app
```
sitozzz, line 44: IMO there is no reason to use `ADD` here. If you're working with local non compressed files and dirs -> use `COPY` as you did before and after that layer

.dockerignore
pahaz: We already have .yarn inside out repository folder. We can't ignore them

sitozzz (approved):

PR #2970, opened 2023-02-28
Doma-3904 paid field in billing receipt details

dkoviazin:
Added paid field to billing receipt details
in this field will be stored information from billings about already paid amount of the receipt
in GIS - it is paidCache field
in KV24 - it is a sum of payments 
This field will be displayed in receipts table and on mobile device

field formula is not a required field now (because, this field is not used)

### DIFF
--- a/apps/condo/domains/billing/gql.js
+++ b/apps/condo/domains/billing/gql.js
@@ -43,7 +43,7 @@ const BillingRecipient = generateGqlQueries('BillingRecipient', BILLING_RECIPIEN
 const BILLING_CATEGORY_FIELDS = `{ name nameNonLocalized ${COMMON_FIELDS} }`
 const BillingCategory = generateGqlQueries('BillingCategory', BILLING_CATEGORY_FIELDS)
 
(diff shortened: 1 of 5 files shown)


toplenboren (approved):

apps/condo/domains/billing/gql.js
```diff
@@ -43,7 +43,7 @@ const BillingRecipient = generateGqlQueries('BillingRecipient', BILLING_RECIPIEN
 const BILLING_CATEGORY_FIELDS = `{ name nameNonLocalized ${COMMON_FIELDS} }`
 const BillingCategory = generateGqlQueries('BillingCategory', BILLING_CATEGORY_FIELDS)
 
-const BILLING_RECEIPT_TO_PAY_DETAILS_FIELDS = 'toPayDetails { charge formula balance recalculation privilege penalty }'
```
pahaz, line 46: what about tests?

pahaz (approved):

apps/condo/domains/billing/gql.js
dkoviazin: we are using this field as an optional information. it's not used in calculations. so, I don't know what tests to add

PR #2993, opened 2023-03-03
fix(condo): DOMA-5480 fixed cypress tests

Alllex202:


apps/condo/cypress/objects/Ticket.ts
```diff
@@ -28,7 +34,7 @@ class TicketCreate {
 
     clickAndInputAddress (address: string): this {
         cy.get('[data-cy=ticket__property-address-search-input] input', {
-            timeout: 5000,
+            timeout: 10000,
```
sitozzz, line 37: Does it work with old timeout?

sitozzz (approved):

apps/condo/cypress/objects/Ticket.ts
Alllex202: There was a error with this content. I did not understand which of the two cases it refers to, so I increased the timeout in both cases. 
Снимок экрана 2023-03-07 в 13 14 47

akarjakin (approved):

PR #2999, opened 2023-03-06
fix(condo): DOMA-5500 fix changedBy role field logic

nomerdvadcatpyat:
If the user has a deleted employee in some organization and an existing employee in a related `from` organization, then "Deleted employee" was displayed.
Fixed, now employee role is displayed.
Also wrote tests for this field


apps/condo/domains/ticket/schema/TicketChange.js
```diff
@@ -77,32 +77,36 @@ const TicketChange = new GQLListSchema('TicketChange', {
...
                         organization: { id_in: relatedOrgIds },
                         user: { id: userId },
+                        deletedAt: null,
+                        isBlocked: false,
                     })
 
-                    const translationKey = relatedOrgEmployees.length ? 'ContactCenterEmployee' : 'DeletedEmployee'
-                    return getTranslation(translations, translationKey)
+                    if (relatedOrgEmployees.length) {
+                        changedByEmployeeMessage = getTranslation(translations, 'ContactCenterEmployee')
```
pahaz, line 105: it looks strange

Alllex202 (approved):

apps/condo/domains/ticket/schema/TicketChange.js
nomerdvadcatpyat: Removed 'ContactCenterEmployee' translation, used `role.name` instead

PR #3002, opened 2023-03-07
refactor(condo): DOMA-5201 added deleteName param to resetUser mutation

nomerdvadcatpyat:
When an employee is fired `resetUser` mutation is used and the user’s name is erased. 
It is important for other employees to know who made the changes to the ticket, so we will delete the username only if the user requests.

### DIFF
--- a/apps/condo/domains/user/schema/ResetUserService.js
+++ b/apps/condo/domains/user/schema/ResetUserService.js
@@ -58,7 +58,7 @@ const ResetUserService = new GQLCustomSchema('ResetUserService', {
     types: [
         {
             access: true,
-            type: 'input ResetUserInput { dv: Int! sender: SenderFieldInput! user: UserWhereUniqueInput! }',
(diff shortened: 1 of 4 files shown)


apps/condo/domains/user/schema/ResetUserService.js
```diff
@@ -98,7 +98,7 @@ const ResetUserService = new GQLCustomSchema('ResetUserService', {
                     phone: null,
                     email: null,
                     password: null,
-                    name: DELETED_USER_NAME,
+                    name: deleteName && DELETED_USER_NAME,
```
Alllex202, line 101: will it work if you pass `false` in `deleteName`?

nomerdvadcatpyat: update logic, now it works correctly

Alllex202 (approved):

apps/condo/domains/user/schema/ResetUserService.js
Alllex202: Maybe add a test for this case?

pahaz:
Should delete by default!

PR #3008, opened 2023-03-09
feat(condo): DOMA-5515 added incident opening in new tab when clicking on number

Alllex202:


AleX83Xpert (approved):

apps/condo/domains/ticket/utils/clientSchema/IncidentRenders.tsx
```diff
@@ -44,7 +44,8 @@ type GetRenderClassifiersType = (incidentClassifiers: IIncidentClassifierInciden
 const MAX_CELL_CONTENT_LENGTH = 150
 
 
-export const getRenderNumber: GetRenderNumberType = () => getTableCellRenderer()
+export const getRenderNumber: GetRenderNumberType = () => (number, incident) =>
+    getTableCellRenderer(null, false, null, null, null, null, `/incident/${incident.id}`)(number)
```
SavelevMatthew, line 48: `null, false, null, null` WTF?)

It's better to use object as params when you have a lot of optional params...

You can fix it here or add todo as a tech debt

SavelevMatthew (approved):

apps/condo/domains/ticket/utils/clientSchema/IncidentRenders.tsx
Alllex202: Yes, it looks bad... We have already discussed this. We'll fix it later :)

PR #3016, opened 2023-03-10
refactor(condo): DOMA-5522 refactored getTableCellRenderer

Alllex202:
Also fixed two errors from console with translates and keys


nomerdvadcatpyat (approved):

apps/condo/domains/common/components/Table/Renders.tsx
```diff
@@ -201,7 +252,7 @@ export const getUnitNameRender = <T extends Record<string, unknown>>(intl, text:
 
     const unitName = text && unitNamePrefix ? `${unitNamePrefix} ${text}` : text
 
-    return getTableCellRenderer(search, false, null, null, null, extraTitle)(unitName)
+    return getTableCellRenderer({ search, extraTitle })(unitName)
```
pahaz, line 255: It is better to avoid creating unnecessary dictionaries

apps/condo/domains/common/components/Table/Renders.tsx
```diff
@@ -96,7 +132,10 @@ export const getHighlightedContents: TGetHighlightedFN = (search, postfix, extra
     )
 }
 
-const renderLink = (content: JSX.Element | string, href: string) => {
+
+type renderLinkType = (props: { content: JSX.Element | string, href: string }) => React.ReactElement
+
+const renderLink: renderLinkType = ({ content, href }) => {
```
pahaz, line 138: it's better to reuse it in our left menu

apps/condo/domains/common/components/Table/Renders.tsx
Alllex202: In this case, it improves readability

apps/condo/domains/common/components/Table/Renders.tsx
Alllex202: fixed it

pahaz (approved):

PR #3028, opened 2023-03-14
feat(condo): DOMA-5394 added script for creating billing app

abshnko:


apps/condo/bin/create-billing-app.js
```diff
@@ -0,0 +1,70 @@
...
+    try {
+        return JSON.parse(data)
+    } catch (e) {
+        return undefined
+    }
+}
+
+async function main (args) {
+    const [email, name, options] = args
+    let optionsJson = getJson(options)
```
toplenboren, line 20: better to parse json in one place

const options = parseOptions(options)

or even better

const { contextDefaultStatus, category, dv, sender, billingIntegration, ... } = parseOptions(options)

SavelevMatthew (commented):
Why do you even need this script?

`bin/generate-miniapps.js` can generate billing and acquiring

PR #3031, opened 2023-03-15
feat(global): DOMA-5542 bridge modals

SavelevMatthew:
https://user-images.githubusercontent.com/50069872/225277910-dd7177b3-2839-4a12-9be7-b83e2bb9c076.mov

### DIFF
--- a/packages/ui/src/components/Modal/modal.tsx
+++ b/packages/ui/src/components/Modal/modal.tsx
@@ -68,7 +68,6 @@ const Modal: React.FC<ModalProps> = (props) => {
             centered
             closable
             className={classes}
-            // closeIcon={<Close size='small' color={colors.black}/>}
             closeIcon={<Cross/>}
             footer={footer || null}
             title={<Typography.Title level={3} ellipsis={{ rows: 2 }} children={title}/>}
--- a/packages/ui/src/components/Modal/modal.tsx
+++ b/packages/ui/src/components/Modal/modal.tsx
@@ -68,7 +68,8 @@ const Modal: React.FC<ModalProps> = (props) => {
             centered
             closable
             className={classes}
-            closeIcon={<Cross />}
+            // closeIcon={<Close size='small' color={colors.black}/>}
+            closeIcon={<Cross/>}
             footer={footer || null}
             title={<Typography.Title level={3} ellipsis={{ rows: 2 }} children={title}/>}
             width={CONDO_MODAL_WIDTH[width]}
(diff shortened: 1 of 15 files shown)


packages/ui/src/components/Modal/style.less
```diff
@@ -78,6 +78,7 @@
 
       .condo-modal-wrapper {
         width: fit-content;
+        min-width: 100%;
```
dkoviazin, line 81: maybe, width?

packages/ui/src/components/Modal/useModal.tsx
```diff
@@ -0,0 +1,141 @@
+import React, { useCallback, useEffect, useImperativeHandle, useRef, useState } from 'react'
+
+import { Modal } from './modal'
+
+import { usePatchElement } from '../_utils/hooks/usePatchElement'
+
+import type { ModalProps } from './modal'
+
+type AnyArgs = Array<any>
```
dkoviazin, line 9: do we really have no idea what args will be?

packages/ui/src/components/Modal/useModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+type ModalConfig = Omit<ModalProps, 'open'> & {
+    onCancel?: (...args: AnyArgs) => any
+}
+
+let uuid = 0
```
dkoviazin, line 15: UUID (Universal Unique Identifier) is a 128-bit value

dkoviazin (approved):

packages/ui/src/components/Modal/style.less
SavelevMatthew: nah :)

packages/ui/src/components/Modal/useModal.tsx
SavelevMatthew: Antd does not, that's part of their original code

SavelevMatthew: and I guess they're taking logic from `rc-dialog`, so it's nested type def, that's why it's any

packages/ui/src/components/Modal/modal.tsx
```diff
@@ -68,7 +68,8 @@ const Modal: React.FC<ModalProps> = (props) => {
             centered
             closable
             className={classes}
-            closeIcon={<Cross />}
+            // closeIcon={<Close size='small' color={colors.black}/>}
```
Alllex202, line 71: maybe remove it?

abshnko (approved):

Alllex202 (approved):

PR #3033, opened 2023-03-15
feat(webhooks): DOMA-5557 added a selection of operation for which can subscribe

Alllex202:
## Problem
Right now I'm getting objects every time they are updated.

But I only want to get the objects that have been added.

## Demo

https://user-images.githubusercontent.com/56914444/232767048-e642a1c7-b957-4e85-9808-36886667aa9d.mov

## Also
* was forbidden to change `webhook` field in `WebhookSubscription`
* added reset "failuresCount" and "syncedAmount" when pass other fields


packages/webhooks/tasks/sendWebhook.js
```diff
@@ -48,81 +49,110 @@ async function sendWebhook (subscriptionId) {
...
-            lastSyncTime = dayjs().toISOString()
-
-            // No more objects -> GOTO final update syncedAt
-            if (lastLoaded === 0) {
-                break
+        const innerAND = []
+        const innerOR = []
+        if (update)  {
+            where.updatedAt_gt = syncedAt
+        } else {
```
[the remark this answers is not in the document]
Alllex202: The `update` operation includes `create` and `delete` :)

apps/condo/jest.setupSpec.js
```diff
@@ -2,6 +2,8 @@ const falsey = require('falsey')
 
 const EXTRA_LOGGING = falsey(process.env.DISABLE_LOGGING)
 
+jest.setTimeout(15000)
```
SavelevMatthew, line 5: Should it be modified right here? May be it's possible to specify it only for a specific test case?

Alllex202: This value is set to run tests in ci. I don't think it will be redundant

Alllex202: This is also specified in `jest.setupTest.js`, but I think 60 seconds will be a lot

packages/webhooks/tasks/sendWebhook.spec.js
```diff
@@ -48,37 +85,150 @@ const SendWebhookTests = (appName, actorsInitializer, userCreator, userDestroyer
...
+                            ]))
+                            expect(STABLE_CALLS.flat()).toHaveLength(3)
+                        },
+                    ],
+                    [
+                        'update',
+                        { create: false, update: true, delete: false },
+                        () => {
+                            // NOTE: DELETED USER MUST BE IN SENT DATA
+                            expect(STABLE_CALLS[0]).toEqual(expect.arrayContaining([
```
SavelevMatthew, line 138: This call should not exist for update-only.
 
in beforeAll we create 3 users and delete 1 of them....So it `3 x create` + `1 x delete`. Why they're in update hook?

packages/webhooks/tasks/sendWebhook.spec.js
```diff
@@ -48,37 +85,150 @@ const SendWebhookTests = (appName, actorsInitializer, userCreator, userDestroyer
...
+                            ]))
+                            expect(STABLE_CALLS[1]).toEqual(expect.arrayContaining([
+                                expect.objectContaining({ id: lastUser.id }),
+                            ]))
+                            expect(STABLE_CALLS.flat()).toHaveLength(2)
+                        },
+                    ],
+                    [
+                        'create and delete',
+                        { create: true, update: false, delete: true },
```
SavelevMatthew, line 168: If you want to work with multiple operations at once, you should probably pass operation as a part of payload

packages/webhooks/tasks/sendWebhook.spec.js
```diff
@@ -48,37 +85,150 @@ const SendWebhookTests = (appName, actorsInitializer, userCreator, userDestroyer
...
+                            expect(STABLE_CALLS[0]).toEqual(expect.arrayContaining([
+                                expect.objectContaining({ id: firstUser.id }),
+                                expect.objectContaining({ id: deletedUser.id }),
+                                expect.objectContaining({ id: lastUser.id }),
+                            ]))
+                            expect(STABLE_CALLS[1]).toEqual(expect.arrayContaining([
+                                expect.objectContaining({ id: firstUser.id }),
+                            ]))
+                            expect(STABLE_CALLS[2]).toEqual(expect.arrayContaining([
+                                expect.objectContaining({ id: lastUser.id }),
```
SavelevMatthew, line 196: Why this user is here?
It was created in `beforeAll` and deleted in test body...create is in first call, and update != delete

apps/condo/jest.setupSpec.js
pahaz: It's better don't change it. We use jest options for that

pahaz (approved):

packages/webhooks/schema/models/fields/WebhookSubscriptionOperations.js
```diff
@@ -0,0 +1,77 @@
...
+        update: { type: 'boolean' },
+        delete: { type: 'boolean' },
+    },
+    required: ['create', 'update', 'delete'],
+    additionalProperties: false,
+}
+
+const webhookSubscriptionOperationsValidator = getValidator(ajv.compile(WebhookSubscriptionOperationsSchema))
+
+const WEBHOOK_SUBSCRIPTION_OPERATIONS_FIELD = {
```
pahaz, line 57: probably it's better to create some enums like so: `create`, `create+update`. Because we don't have any delete

packages/webhooks/schema/models/WebhookSubscription.test.js
```diff
@@ -253,49 +254,120 @@ const WebhookSubscriptionBasicTests = (appName, actorsInitializer) => {
...
+                        ['Boolean', true],
+                        ['Object', { update: true }],
+                        ['Array', ['update']],
+                    ]
+                    test.each(cases2)('%s', async (name, operation) => {
+                        await expectToThrowGraphQLRequestError(async () => {
+                            await createTestWebhookSubscription(actors.admin, hook, {
+                                operation,
+                            })
+                        }, 'got invalid value')
```
SavelevMatthew, line 330: Include field name here, since it's not 100% that error is thrown by `operation` field

packages/webhooks/schema/models/WebhookSubscription.js
```diff
@@ -63,6 +67,9 @@ function getWebhookSubscriptionModel (schemaPath) {
                 isRequired: true,
                 knexOptions: { isNotNullable: true },
                 kmigratorOptions: { null: false, on_delete: 'models.CASCADE' },
+                access: {
+                    update: false,
```
SavelevMatthew, line 71: Why do we need that?

packages/webhooks/tasks/sendWebhook.spec.js
```diff
@@ -48,37 +86,101 @@ const SendWebhookTests = (appName, actorsInitializer, userCreator, userDestroyer
...
+                                expect.objectContaining({ id: firstUser.id }),
+                            ]))
+                            expect(STABLE_CALLS.flat()).toHaveLength(1)
+                        },
+                    ],
+                    [
+                        'Delete',
+                        WEBHOOK_OPERATIONS.DELETE,
+                        () => {
+                            // NOTE: deletedUser MUST NOT BE IN SENT DATA because it was created and deleted between first and second syncs
```
SavelevMatthew, line 146: Why must not?
1. If I create a subscription with `syncAt` long time ago and `operation = create` I'm expecting to get all created since that time models. That's true in case above
2. If I create a subscription with `syncAt` long time ago and `operation = update` I'm expecting to get all updated models (I'm assuming create + update goes to create group). That's true in case above
3. So if I create same subscription with `operation = delete` I should get all deleted objects since last sync? So, why `deletedUser` is not in payload then?

Alllex202:
Freeze

PR #3045, opened 2023-03-18
Use one `rimraf` version everywhere

pahaz:


SavelevMatthew (commented):
Why? is there any security issues?

pahaz:
> Why? is there any security issues?

We don’t need a different versions of this package. I want to unify our packages versions. To better control them

SavelevMatthew:
> > Why? is there any security issues?
> 
> We don’t need a different versions of this package. I want to unify our packages versions. To better control them

Then it's not the way to do it. 

It's a dangerous fix, which can cause issues. If some of our apps uses `rimraf` directly, fix version in their package.json. If we have multiple package versions, which depends on multiple rimrafs, when we probably should fix parent package version.

This way you fix it for all nested levels, causing any lib using older / newer versions potentially fail. That's how `resolutions` field work. It's not designed for syncing dependencies versions. Its main goals are pathing any vulnerabilities, or downgrading / upgrading sub-dep of other package in case their PR is progressing slowly (or they missed it in deps-array)

PR #3051, opened 2023-03-20
Many = true rel lint-schema checks

toplenboren:


apps/condo/domains/billing/schema/BillingIntegration.js
```diff
@@ -98,6 +98,7 @@ const BillingIntegration = new GQLListSchema('BillingIntegration', {
             type: Relationship,
             ref: 'BillingIntegrationAccessRight.integration',
             many: true,
+            access: { create: false, update: false },
```
pahaz, line 101: Probably, we also need to lint that you define all possible operations: create, read, update (optional)

toplenboren: It seems that if you do not define any access, it auto sets to true. I added a check just in case ...

dkoviazin (approved):

abshnko (approved):

bin/lint-schema.js
```diff
@@ -32,6 +34,19 @@ function verifySchema (keystone) {
                     }
                 }
             }
+
+            if (field.isRelationship && field.many && !excludedManyRelationshipCheckLists.includes(list.key)) {
+                const access = field.access.public
+
+                if (typeof access.create !== 'boolean' || typeof access.update !== 'boolean') {
+                    report(`${list.key}->${field.path} For many=true fields access.update and access.create should be simple boolean!`)
```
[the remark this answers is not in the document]
toplenboren: Yes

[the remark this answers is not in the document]
SavelevMatthew: Change error message to accept only `false` instead of `bool`

apps/condo/domains/billing/schema/BillingIntegration.js
pahaz: explicit is better than implicit

PR #3060, opened 2023-03-21
fix(condo): DOMA-5608 function expectValuesOfCommonFields

AntonAL:
In each test module a created schema object is expected to contain common set of fields.

These checks spanning on multiple lines and gets repeated in almost every test module.

Because of this duplication, they are presented **not in every test module, but should**.

Proposed test function makes expectation checks easier and removes a lot of duplicated code.


packages/keystone/test.utils.js
```diff
@@ -659,6 +659,19 @@ const expectToThrowGraphQLRequestError = async (testFunc, message) => {
     })
 }
 
+const expectValuesOfCommonFields = ({ obj, attrs, client }) => {
```
pahaz, line 662: why dict?

VKislov (approved):

PR #3089, opened 2023-03-24
feature(address-service): DOMA-5656 added parsing of address from string

dkoviazin:
We need to support bulk address normalization. This is the first step for it
In billing we need to normalize thousands of addresses from billing receipts
As we have some limits on dadata or google normalize calls - we need first to extract unique houses addresses and normalize only them

Usage example
image


apps/address-service/domains/common/utils/parseAddressesFromString.spec.js
AleX83Xpert, line 1: It's good to add a test case containing the `%` symbol (`SPLIT_SYMBOL`)

AleX83Xpert (approved):

apps/address-service/domains/common/utils/parseAddressesFromString.js
```diff
@@ -0,0 +1,97 @@
+const KEYWORDS = {
+    parking: ['автоместо', 'парковка', 'паркинг', 'машиноместо', 'гараж', 'паокинг', 'м/м', 'мм', 'место', 'м/место', 'а/м', 'бокс', 'парк'],
```
sitozzz, line 2: Is word with typo must defined here as a keyword?

apps/address-service/domains/common/utils/parseAddressesFromString.spec.js
dkoviazin: Done

Alllex202 (approved):

sitozzz (approved):

PR #3092, opened 2023-03-27
feat(condo): DOMA-5677 use predictTransactionClassification in syncTransaction from SBBOL

VKislov:


sitozzz (approved):

apps/condo/domains/organization/integrations/sbbol/sync/requestTransactions.js
```diff
@@ -154,11 +154,13 @@ async function requestTransactionsForDate ({ userId, bankAccounts, context, stat
 
                 if (!foundTransaction) {
                     const bankIntegrationAccountContextId = get(bankAccount, 'integrationContext.id')
+                    const costItem = await predictTransactionClassification(context, { purpose: transaction.paymentPurpose })
```
sitozzz, line 157: I suppose you should check classification result. And if prediction service return something like `COST_ITEM_NOT_FOUND` you're able to save this transaction without any costItem connected to it

VKislov: Yes, you're right, thanks for noticing.

PR #3118, opened 2023-03-30
feat(condo): DOMA-5706 Request statement summary from SBBOL

VKislov:
Added a request for the statement of the bank account summary from the `SBBOL`. It will be displayed on the property page

### DIFF
--- a/apps/condo/domains/organization/integrations/sbbol/sync/requestTransactions.js
+++ b/apps/condo/domains/organization/integrations/sbbol/sync/requestTransactions.js
@@ -17,7 +17,7 @@ const { ERROR_PASSED_DATE_IN_THE_FUTURE } = require('../constants')
 
 
 const logger = getLogger('sbbol/SbbolSyncTransactions')
-const isTransactionsReceived = (response) => (get(response, 'error.cause', '') !== 'STATEMENT_RESPONSE_PROCESSING')
+const isResponceSuccessfull = (response) => (get(response, 'error.cause', '') !== 'STATEMENT_RESPONSE_PROCESSING')
 // ---------------------------------
 const FAILED_STATUS = 'failed'
 const IN_PROGRESS_STATUS = 'inProgress'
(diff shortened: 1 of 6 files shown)


apps/condo/domains/organization/tasks/syncSbbolBankAccounts.js
```diff
@@ -32,9 +32,11 @@ async function syncSbbolBankAccounts () {
     const integration = await BankIntegration.getOne(context, { id: BANK_INTEGRATION_IDS.SBBOL })
     if (!integration) throw new Error(`Cannot find SBBOL integration by id=" ${BANK_INTEGRATION_IDS.SBBOL}"`)
 
-    for (const identity of usersWithSBBOLExternalIdentity) {
+    for (let identity of usersWithSBBOLExternalIdentity) {
         const [employee] = await OrganizationEmployee.getAll(context, {
-            user: { id: identity.user.id },
+            user: {
+                id: identity.user.id,
+            },
```
AntonAL, line 39: It it possible to fetch users with one request, using `id_in` query field?

apps/condo/domains/organization/integrations/sbbol/sync/requestTransactions.js
```diff
@@ -52,22 +52,36 @@ async function requestTransactionsForDate ({ userId, bankAccounts, context, stat
...
+                                page
+                            )
+                        }
+
+                        if (!isResponceSuccessfull(summary)) {
+                            summary = await fintechApi.getStatementSummary(
+                                bankAccount.number,
+                                statementDate,
+                            )
+                        }
```
AntonAL, line 84: You have a call of this method before this "while" loop.
For what purpose you have this extra call?

VKislov: when I send a request to receive an account statement, I may get back an error saying that the statement is being generated. Therefore, I first request it and, if I don’t receive it, I go to request it in a loop with a limit of 5 reps and a timeout that doubles after each fail

apps/condo/domains/organization/tasks/syncSbbolBankAccounts.js
```diff
@@ -44,7 +46,9 @@ async function syncSbbolBankAccounts () {
             isBlocked: false,
         }, { first: 1 })
 
-        await syncBankAccounts(identity.user.id, employee.organization)
+        if (employee) {
```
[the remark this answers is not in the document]
VKislov: what should I do if more than a thousand instances of the `OrganizationEmployee` entity come with such a filter? Do I need to do chunking? It seems that in the current realities, I can only write such a task in the backlog as a debt

apps/condo/domains/organization/tasks/syncSbbolBankAccounts.js
VKislov: yes

apps/condo/domains/organization/tasks/syncSbbolBankAccounts.js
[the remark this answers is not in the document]
AntonAL: ok

apps/condo/domains/organization/integrations/sbbol/sync/requestTransactions.js
AntonAL: Could you please create a comment with all what you answered here, because it's not obvious from the fist look.

apps/condo/domains/organization/integrations/sbbol/sync/requestTransactions.js
```diff
@@ -17,7 +17,7 @@ const { ERROR_PASSED_DATE_IN_THE_FUTURE } = require('../constants')
 
 
 const logger = getLogger('sbbol/SbbolSyncTransactions')
-const isTransactionsReceived = (response) => (get(response, 'error.cause', '') !== 'STATEMENT_RESPONSE_PROCESSING')
+const isResponceSuccessfull = (response) => (get(response, 'error.cause', '') !== 'STATEMENT_RESPONSE_PROCESSING')
```
AntonAL, line 20: Since you are checking on processing status of the response, then maybe to use according name for the function?
```suggestion
const isResponceProcessing = (response) => (get(response, 'error.cause', '') !== 'STATEMENT_RESPONSE_PROCESSING')
```

sitozzz (approved):

PR #3119, opened 2023-03-30
DOMA-5716 UI-kit useBreakpoint hook

nomerdvadcatpyat:


packages/ui/src/hooks/_utils/responsiveObserve.ts
```diff
@@ -0,0 +1,74 @@
...
+    register () {
+        Object.keys(responsiveMap).forEach((screen) => {
+            const matchMediaQuery = responsiveMap[screen as Breakpoint]
+            const listener = ({ matches }: { matches: boolean }) => {
+                this.dispatch({
+                    ...screens,
+                    [screen]: matches,
+                })
+            }
+            const mql = window.matchMedia(matchMediaQuery)
```
SavelevMatthew, line 62: SSR-friendly?)

packages/ui/src/index.ts
```diff
@@ -57,3 +57,6 @@ export type { RadioProps, RadioGroupProps } from './components/Radio'
 
 export { Space } from './components/Space'
 export type { SpaceProps } from './components/Space'
+
+export { useBreakpoint } from './hooks/useBreakpoint'
```
SavelevMatthew, line 61: Should we pack it here along with components?

packages/ui/src/hooks/_utils/responsiveObserve.ts
```diff
@@ -0,0 +1,74 @@
+export type Breakpoint = 'mobileSmall' | 'mobileLarge' | 'tabletSmall' | 'tabletLarge' | 'desktopSmall' | 'desktopLarge'
+export type BreakpointMap = Record<Breakpoint, string>
+export type ScreenMap = Partial<Record<Breakpoint, boolean>>
+export type ScreenSizeMap = Partial<Record<Breakpoint, number>>
+
+export const responsiveArray: Breakpoint[] = ['mobileSmall', 'mobileLarge', 'tabletSmall', 'tabletLarge', 'desktopSmall', 'desktopLarge']
+
+export const responsiveMap: BreakpointMap = {
+    mobileSmall: '(max-width: 480px)',
```
SavelevMatthew, line 9: are you sure about naming conventions? Should we use upper case like condo const or camel case?

packages/ui/src/hooks/_utils/hooks/index.ts
```diff
@@ -0,0 +1 @@
+export { useForceUpdate } from './useForceUpdate'
```
SavelevMatthew, line 1: Why separate from `@open-condo/ui/src/components/_utils/hooks` ?

packages/ui/src/hooks/_utils/responsiveObserve.ts
SavelevMatthew, line 62: Can you test it on a simple NextJS server or something like that...not sure about SSR environment.

I'm sure, that calling `document` directly causes issues, but not sure about window

packages/ui/src/hooks/_utils/responsiveObserve.ts
```diff
@@ -0,0 +1,74 @@
+export type Breakpoint = 'mobileSmall' | 'mobileLarge' | 'tabletSmall' | 'tabletLarge' | 'desktopSmall' | 'desktopLarge'
+export type BreakpointMap = Record<Breakpoint, string>
+export type ScreenMap = Partial<Record<Breakpoint, boolean>>
+export type ScreenSizeMap = Partial<Record<Breakpoint, number>>
+
```
SavelevMatthew, line 5: It's probably better to make breakpoints as consts like:
`const MOBILE_SMALL = 'MOBILE_SMALL'`
then do smth like:
`breakpoints = [MOBILE_SMALL, ...] as const
and then
```typescript
type Breakpoint = typeof breakpoints[number]
// make sure every key is specified, record is a partial
type BreakpointMap =  {[Key in Breakpoint]: string}

const responsiveMap: BreakpointMap = {
   [MOBILE_SMALL]: '...'
}
```

This way you only need to specify string enum value once, and then just use a const

SavelevMatthew: And you also can export all consts, so you can use it like:
```
if (screen[breakpoints.MOBILE_SMALL]) {
// breakpoint specific logic
}
```

SavelevMatthew (commented):
You also should add a basic docs in package readme, which should include:
1. What can I get from hook
2. How to check if we've at least at some checkpoint width
3. How to check if we've at most at some checkpoint width
4. Notification, that hook is designed to be mobile-first, so checkpoints for smaller devices are also active on a larger ones

packages/ui/src/hooks/_utils/responsiveObserve.ts
nomerdvadcatpyat: `ResponsiveObserve` called only in `useEffect` in `useBreakpoint` hook.
This `useEffect` will not be called on ssr because it is independent of the data received on the server.

But just in case, add a window check in `useEffect`.

packages/ui/src/index.ts
nomerdvadcatpyat: I think it is better to store hooks that we export from the UI-kit to the `src/hooks` directory than `src/components/hooks`.
Hooks are not components

packages/ui/src/hooks/_utils/hooks/index.ts
nomerdvadcatpyat: Maybe put `_utils` folder on the level above? `src/_utils`?

packages/ui/src/index.ts
nomerdvadcatpyat: move it to `components/hooks` like in `ant`

packages/ui/src/hooks/_utils/responsiveObserve.ts
nomerdvadcatpyat: used upper case here

packages/ui/src/components/_utils/responsiveObserve.ts
```diff
@@ -0,0 +1,82 @@
...
+const MOBILE_LARGE = 'MOBILE_LARGE'
+const TABLET_SMALL = 'TABLET_SMALL'
+const TABLET_LARGE = 'TABLET_LARGE'
+const DESKTOP_SMALL = 'DESKTOP_SMALL'
+const DESKTOP_LARGE = 'DESKTOP_LARGE'
+
+export const breakpoints = [MOBILE_SMALL, MOBILE_LARGE, TABLET_SMALL, TABLET_LARGE, DESKTOP_SMALL, DESKTOP_LARGE] as const
+
+export type Breakpoint = typeof breakpoints[number]
+type BreakpointMap = { [Key in Breakpoint]: string }
```
Alllex202, line 11: You can use `Record` as `Record<Breakpoint, string>`

apps/condo/domains/common/components/Card/CardsContainer.tsx
```diff
@@ -12,9 +13,9 @@ interface CardsContainerProps {
     autosize?: boolean
 }
 
-const getColSpan = (breakpoints: Partial<Record<Breakpoint, boolean>>) => {
-    if (breakpoints.xxl) return 8
-    if (breakpoints.xl) return 12
+const getColSpan = (breakpoints: Partial<ScreenMap>) => {
```
Alllex202, line 16: ScreenMap is `Partial` already
Does it really need again?

`export type ScreenMap = Partial<Record<Breakpoint, boolean>>`

packages/ui/src/hooks/_utils/responsiveObserve.ts
Alllex202: This example show that "record" is not partial
Снимок экрана 2023-03-31 в 13 11 21

packages/ui/src/index.ts
SavelevMatthew: Antd does not have a folder like that as far as I remember...
Why not just in `_utils/hooks` with other hooks and pack some of them (like useBreakpoints + useContainerSize) for export (separate entry for webpack)

packages/ui/src/components/_utils/responsiveObserve.ts
SavelevMatthew: This way keys are partial. `[A in B]` notation is making sure that every key from `B` is presented

Alllex202: `Record` does the same

Снимок экрана 2023-03-31 в 13 07 34

packages/ui/README.md
```diff
@@ -56,3 +57,28 @@ import '@open-condo/ui/style-vars/css'
 ```less
 @import (reference) "@open-condo/ui/dist/style-vars/variables.less";
 ```
+
+## Hooks
+Hooks can be imported as follows:
+```js
+import { useBreakpoint, useContainerSize } from 'open-condo/ui';
```
SavelevMatthew, line 64:
```suggestion
import { useBreakpoint, useContainerSize } from '@open-condo/ui/hooks';
```

packages/ui/src/hooks.ts
```diff
@@ -0,0 +1,4 @@
+export { useBreakpoint } from './components/_utils/hooks/useBreakpoint'
+export { useContainerSize } from './components/_utils/hooks/useContainerSize'
+
+export type { Breakpoint, ScreenMap, breakpoints } from './components/_utils/responsiveObserve'
```
SavelevMatthew, line 4: export const as `CAPS_CASE`, since is can be misleading with `const breakpoints = useBreakpoints()`

SavelevMatthew (approved):

packages/ui/src/components/_utils/responsiveObserve.ts
```diff
@@ -0,0 +1,82 @@
+const MOBILE_SMALL = 'MOBILE_SMALL'
+const MOBILE_LARGE = 'MOBILE_LARGE'
+const TABLET_SMALL = 'TABLET_SMALL'
+const TABLET_LARGE = 'TABLET_LARGE'
+const DESKTOP_SMALL = 'DESKTOP_SMALL'
+const DESKTOP_LARGE = 'DESKTOP_LARGE'
+
+export const breakpoints = [MOBILE_SMALL, MOBILE_LARGE, TABLET_SMALL, TABLET_LARGE, DESKTOP_SMALL, DESKTOP_LARGE] as const
```
SavelevMatthew, line 8:
```suggestion
export const BREAKPOINTS = [MOBILE_SMALL, MOBILE_LARGE, TABLET_SMALL, TABLET_LARGE, DESKTOP_SMALL, DESKTOP_LARGE] as const
```

PR #3125, opened 2023-03-31
DOMA-5639 add margin for ant modal button footer

sitozzz:
before:

image

after:

image

### DIFF
--- a/apps/condo/domains/common/components/containers/GlobalStyle.tsx
+++ b/apps/condo/domains/common/components/containers/GlobalStyle.tsx
@@ -391,6 +391,9 @@ export default function GlobalStyle () {
                 line-height: 32px; 
                 font-weight: 700;
               }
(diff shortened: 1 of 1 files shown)


VKislov (approved):

apps/condo/domains/common/components/containers/GlobalStyle.tsx
```diff
@@ -391,6 +391,9 @@ export default function GlobalStyle () {
                 line-height: 32px; 
                 font-weight: 700;
               }
+              .ant-modal-footer button + .ant-btn:not(.ant-dropdown-trigger) {
```
pahaz, line 394: is it possible to move it to some component?

sitozzz: It's already a component, but that works with antd components only. So if we'll decide to rewrite modal form to a new uikit - we'll able to rid of GlobalStyle overrides

PR #3137, opened 2023-04-04
feat(condo): DOMA-5736 added income costItems

VKislov:


apps/condo/lang/en/en.json
```diff
@@ -1697,6 +1697,9 @@
   "banking.category.THERMAL_ENERGY.name": "Thermal energy",
   "banking.category.COLD_WATER_SUPPLY.name": "Cold water supply",
   "banking.category.ELECTRIC_ENERGY.name": "Electric energy",
+  "banking.category.INCOME.name": "Income payments",
+  "banking.costItem.INCOME_FROM_RESIDENT.name": "Income from resident",
+  "banking.costItem.INCOME_FROM_LEGAL.name": "Income from legal entities",
   "banking.costItem.EMERGENCY_SERVICE.name": "Emergency service",
   "banking.costItem.RENT.name": "Rent",
   "banking.costItem.BANKING_SERVICES.name": "Banking services",
```
AntonAL, line 1705:
```suggestion
  "banking.costItem.BANKING_SERVICES.name": "Banking service",
```

apps/condo/lang/en/en.json
```diff
@@ -1697,6 +1697,9 @@
   "banking.category.THERMAL_ENERGY.name": "Thermal energy",
   "banking.category.COLD_WATER_SUPPLY.name": "Cold water supply",
   "banking.category.ELECTRIC_ENERGY.name": "Electric energy",
+  "banking.category.INCOME.name": "Income payments",
+  "banking.costItem.INCOME_FROM_RESIDENT.name": "Income from resident",
```
AntonAL, line 1701:
```suggestion
  "banking.costItem.INCOME_FROM_RESIDENT.name": "Income from residents",
```

apps/condo/lang/en/en.json
```diff
@@ -1697,6 +1697,9 @@
...
+  "banking.costItem.INCOME_FROM_LEGAL.name": "Income from legal entities",
   "banking.costItem.EMERGENCY_SERVICE.name": "Emergency service",
```
AntonAL, line 1703:
```suggestion
  "banking.costItem.EMERGENCY_SERVICE.name": "Emergency services",
```

apps/condo/lang/en/en.json
AntonAL: Are we using plural or singular form for all cost items?
I think, plural fits better, because it will be displayed in a chart that accumulates set of transactions.

AntonAL (commented):
What you decided on usage of BankCategory?
We have discussed that there is no confusion in case when a BankCategory can be associated with income and outcome cost items, because a BankCategory is an entity, used for display purposes only

sitozzz (changes requested):
Should be rebased before merge due to outdated migration number

PR #3141, opened 2023-04-04
DOMA-5745 Responsive modal

nomerdvadcatpyat:
Added mobile version of `Modal`.
Used `Modal` from condo-ui instead modal from `common` domain in our components.

Mobile `Modal`: Paddings сhanged from 40px to 24px, modal sticks to the bottom of the window
изображение


packages/ui/src/components/Button/button.tsx
```diff
@@ -53,9 +53,13 @@ const Button: React.ForwardRefExoticComponent<ButtonProps & React.RefAttributes<
             ref={ref}
             onClick={handleClick}
         >
-            <span className={`${BUTTON_CLASS_PREFIX}-text`} data-before={children}>
-                {children}
-            </span>
+            {
+                children && (
```
SavelevMatthew, line 57: was it causing some troubles?)

nomerdvadcatpyat: space after icon
изображение

apps/condo/domains/common/components/Prompt.tsx
```diff
@@ -70,15 +68,14 @@ const Prompt: React.FC<IPromptProps> = ({ children, title, form, handleSave: for
...
     return (
         <Modal
-            visible={isModalVisible}
+            open={isModalVisible}
             onCancel={hideModal}
             title={title}
-            centered
             footer={[
-                <Button key='back' type='sberDanger' style={{ margin: '16px' }} onClick={handleCancel}>
+                <Button key='back' type='secondary' onClick={handleCancel}>
```
Alllex202, line 75: maybe should add property `danger`?
Снимок экрана 2023-04-05 в 11 06 04

apps/condo/domains/common/components/containers/FeatureFlag.tsx
Alllex202, line 1: You can also use new `Alert`

apps/condo/domains/ticket/hooks/useTicketExportToPdfTask.tsx
Alllex202, line 1: should drop StyledModal

apps/condo/domains/user/components/UserMenu/MobileUserMenu.tsx
```diff
@@ -1,18 +1,15 @@
 import { UserOutlined } from '@ant-design/icons'
 import styled from '@emotion/styled'
-import { Menu, Avatar } from 'antd'
+import { Menu, Avatar, Modal } from 'antd'
```
Alllex202, line 3: ???

Alllex202: why not from open-condo?

nomerdvadcatpyat: update

apps/condo/domains/ticket/components/ShareTicketModal.tsx
```diff
@@ -288,20 +283,17 @@ export const ShareTicketModal: React.FC<IShareTicketModalProps> = (props) => {
     return (
         <>
             <Button
-                type='sberDefaultGradient'
+                type='secondary'
```
Alllex202, line 286: is it not `primary`?

Alllex202: and other places

Alllex202 (commented):
You should also update icons in new buttons. 
And mark the old modal as deprecated

apps/condo/domains/ticket/components/ShareTicketModal.tsx
nomerdvadcatpyat: design solution

packages/ui/src/components/style/core/global.less
```diff
@@ -6,4 +6,11 @@
 :root {
   --condo-font-fallback: sans-serif;
   --condo-font-fallback-mono: consolas, monaco, "Andale Mono", "Ubuntu Mono", monospace;
-}
\ No newline at end of file
+}
+
+@condo-global-breakpoints-mobile-small: 0px;
```
SavelevMatthew, line 11: It's not usable like that....

You can use a vars like that
image


Or use a mixins:
image

packages/ui/src/components/Modal/style.less
```diff
@@ -87,9 +90,54 @@
...
+        margin: 0 @condo-modal-mobile-spacing;
+        padding: 0 0 @condo-modal-mobile-spacing;
+      }
+
+      .condo-modal-footer {
+        padding: @condo-modal-mobile-spacing;
+
+        flex-direction: column;
+
+        button {
```
SavelevMatthew, line 137: use `.condo-` notation, you should not control elements from other design systems...

SavelevMatthew: and button with a href is `<a/>` btw

packages/ui/src/components/Modal/style.less
```diff
@@ -87,9 +90,54 @@
       padding: @condo-modal-half-padding-without-border @condo-modal-padding @condo-modal-half-padding;
       border-top: @condo-global-border-width-default solid @condo-modal-header-scrolling-border-bottom-color;
 
+      display: flex;
```
SavelevMatthew, line 93: Are you sure, that it's our responsibility to control footer elements?

packages/ui/src/components/Modal/modal.tsx
```diff
@@ -65,7 +66,7 @@ const Modal: React.FC<ModalProps> = (props) => {
         <DefaultModal
             {...rest}
             prefixCls={MODAL_CLASS_PREFIX}
-            centered
+            centered={centered}
```
SavelevMatthew, line 69: why? all our modal are in center by default according to UI styleguides...

Don't add it to support some Frankenstein workarounds

apps/condo/pages/ticket/index.tsx
```diff
@@ -504,7 +502,7 @@ const FiltersButton = ({ appliedFiltersCount, setIsMultipleFiltersModalVisible }
     }, [setIsMultipleFiltersModalVisible])
 
     return (
-        <Button
+        <CommonButton
```
SavelevMatthew, line 505: ?

apps/condo/pages/ticket/[id]/index.tsx
```diff
@@ -453,9 +452,8 @@ export const TicketPageContent = ({ ticket, refetchTicket, loading, organization
                             <Button
                                 disabled={disabledEditButton}
                                 color='green'
```
Comment (SavelevMatthew) [inline apps/condo/pages/ticket/[id]/index.tsx:454]: wtf is this prop about?

apps/condo/domains/user/components/UserMenu/MobileUserMenu.tsx
```diff
@@ -28,27 +26,27 @@ export const StyledMenu = styled(Menu)`
...
-    position: 'absolute',
-    left: 0,
-    right: 0,
-    bottom: 0,
-    pointerEvents: 'initial',
-    top: 'initial',
-    transformOrigin: 'initial',
-    maxWidth: '100%',
-}
+const StyledModal = styled(Modal)`
```
SavelevMatthew, line 29: I'm not sure, that it should be done with styled modal...

Btw, aren't we switching from emotion to css/less.modules?

packages/ui/src/components/Modal/style.less
SavelevMatthew: Since it's not clear:
1. Not obvious that React.Node accept list
2. Not obvious, how should it render it (like first element should be on a right or left side), since all of them are floating to right.

That's why I prefer to use Space there....Then you're 100% sure what's going on

apps/condo/domains/ticket/hooks/useTicketCancelModal.tsx
```diff
@@ -36,12 +31,12 @@ export const useTicketCancelModal: useTicketCancelModalType = (updateTicket) =>
 
     const cancelTicketModal = useMemo(() => (
         <Modal
-            visible={isModalVisible}
```
SavelevMatthew, line 39: You can use `Modal.useModal`, and then just spawn it dinamically. That's recommended option if source of opening = user interaction. (you do not need to destroy it manually)

apps/condo/domains/subscription/hooks/useSubscriptionPaymentModal.tsx
```diff
@@ -14,31 +12,33 @@ interface IApplySubscriptionModal {
 
 export const useSubscriptionPaymentModal = (): IApplySubscriptionModal => {
     const intl = useIntl()
+    const ModalTitle = intl.formatMessage({ id: 'subscription.modal.complete.title' })
```
SavelevMatthew, line 15: Same here....seems like all `useBlahBlahModal` are redundant, if you're spawning it from user, if not - than it' viable option

apps/condo/domains/common/hooks/useSearchByPhoneModal.tsx
```diff
@@ -224,18 +217,21 @@ export const useSearchByPhoneModal = (searchByPhoneFn, canManageContacts) => {
...
         <StyledModal
-            visible={isSearchByPhoneModalVisible}
+            isSmall={isSmall}
+            open={isSearchByPhoneModalVisible}
             title={SearchByPhoneMessage}
             onCancel={handleCloseModal}
             footer={null}
-            width={1150}
             destroyOnClose
+            centered={false}
```
SavelevMatthew, line 234: You should not destroy our component just because you need a custom in on place. All our modals are centered by default then used as a real modals....

apps/condo/domains/common/hooks/useSearchByPhoneModal.tsx
```diff
@@ -8,39 +8,34 @@ import React, { useCallback, useMemo, useState } from 'react'
...
-import { colors } from '@condo/domains/common/constants/style'
 import {
     ClientType,
     mapSearchItemToOption,
     redirectToForm,
 } from '@condo/domains/contact/utils/clientCard'
 
 
-const StyledModal = styled(Modal)`
+const StyledModal = styled(Modal)<{ isSmall }>`
```
SavelevMatthew, line 25: You can override centered behaviour here for you custom usage (and even better to do it in css/less module.

apps/condo/domains/common/components/containers/FeatureFlag.tsx
```diff
@@ -16,12 +16,11 @@
 
 import { Switch, Alert } from 'antd'
 import getConfig from 'next/config'
-import React, { ReactNode, useState } from 'react'
+import React, { ReactNode, useMemo, useState } from 'react'
```
SavelevMatthew, line 19: Seems like we don't use them anymore and use Growthbook instead, should we delete it?

apps/condo/domains/common/components/Modal.tsx
```diff
@@ -1,6 +1,7 @@
 import styled from '@emotion/styled'
 import { Modal as DefaultModal } from 'antd'
 
+/** @deprecated use Modal from @open-condo/ui */
 export const Modal = styled(DefaultModal)`
```
SavelevMatthew, line 5: Aren't you removed all usage in this PR?)

apps/condo/domains/common/components/LayoutContext.tsx
```diff
@@ -39,7 +39,7 @@ export const LayoutContextProvider: React.FC = (props) => {
         setIsCollapsed(!isCollapsed)
     }
 
-    const isSmall = !breakpoints.DESKTOP_SMALL
+    const isSmall = !breakpoints.TABLET_LARGE
```
SavelevMatthew, line 42: It's a bit strange to continue support this API if we use breakpoints :)))
At least add a todo to rework / kill `isSmall` as a part of responsive screens task

SavelevMatthew (changes requested):
1. `centered prop` should not be added, since it's all our modals are centered according to styleguides and we do not support a custom like this in UI-kit (more strictly we can do = better)
2. Static consts as a breakpoints seems not that usable. There's a mixin and string-var option. I liked mixins more, but its a code duplication in case:
```
.class {
    .breakpoint-mobile-small({
        // styles
    })
    .breakpoint-mobile-large({
        // same styles
    })
}
```
Media is a bit better since you can do:
```
@media @mobile-small,
@media @mobile-large {
   /// code
}
```
However it will produce a lot of css as well, even if don't need to write it twice.
So, there's my third variant.
Let's add some aliases:
```
@from-mobile-small = only screen and (min-width: <breakpoint>)
@to-desktop-small = and (max-width: <breakpoint>)
```
So we can call them like
```
@media @from-mobile-small {
  // rules for [mobile-smal; +inf)
} 
@media @from-mobile-small @to-desktop-small {
  // rules for [mobile-smal; desktop-small]
}
```
Which will produce only 1 rule set...
And we can have some aliases like:
```
@mobile-small = @from-mobile-small @to-mobile-small
``` 
to call them like:
``` 
@media @mobile-small {
  // styles specific for mobile-small breakpoint
}
```

3. What is `color='green'` on a button is about?

Also did not understand `color='green'` on button.

apps/condo/pages/ticket/index.tsx
nomerdvadcatpyat: We have not yet Text Button in UI-kit

apps/condo/domains/common/components/containers/FeatureFlag.tsx
nomerdvadcatpyat: Not part of this pr

apps/condo/domains/common/components/Modal.tsx
nomerdvadcatpyat: removed it

apps/condo/domains/banking/hooks/useCategoryModal.tsx
```diff
@@ -145,12 +145,7 @@ export const useCategoryModal: IUseCategoryModal = ({
                 open={open}
                 onCancel={closeModal}
                 width='big'
-                footer={<Button
-                    type='primary'
-                    disabled={isNull(selectedCostItem) || loading}
-                    onClick={handleSave}>
-                    {SaveTitle}
-                </Button>}
+                footer={<Space size={40}>123</Space>}
```
Alllex202, line 148: what it is?

Alllex202 (changes requested):

apps/condo/domains/banking/hooks/useCategoryModal.tsx
nomerdvadcatpyat: oops
fixed

packages/ui/src/components/style/core/global.less
```diff
@@ -6,4 +6,16 @@
 :root {
   --condo-font-fallback: sans-serif;
   --condo-font-fallback-mono: consolas, monaco, "Andale Mono", "Ubuntu Mono", monospace;
-}
\ No newline at end of file
+}
+
+@from-mobile-small: ~"screen and (min-width: 0px)";
+@from-mobile-large: ~"screen and (min-width: 360px)";
```
SavelevMatthew, line 12: 360px is intersecting, which can cause some bugs. Use semi-intervals:

[a; b) .... [b; c) .... [c ; d) .... [ d; +inf)

So, in particular case:
[0; 359] ... [360; 479] and so on

nomerdvadcatpyat: done

Alllex202 (approved):

apps/condo/domains/billing/components/BillingPageContent/ReceiptsTable.tsx
```diff
@@ -147,7 +147,6 @@ export const ReceiptsTable: React.FC<IContextProps> = ({ context }) => {
             <ServicesModal
                 receipt={detailedReceipt}
                 visible={modalIsVisible}
-                onOk={hideServiceModal}
```
SavelevMatthew, line 150: Did it change some logic there? Or there was no ok button before?

apps/condo/domains/common/components/Modal.tsx
```diff
@@ -1,8 +0,0 @@
-import styled from '@emotion/styled'
```
SavelevMatthew, line 1: 🥇

apps/condo/domains/common/components/Text.tsx
```diff
@@ -1,14 +0,0 @@
-import styled from '@emotion/styled'
```
SavelevMatthew, line 1: 🥇

apps/condo/domains/billing/components/BillingPageContent/ReceiptsTable.tsx
nomerdvadcatpyat: There was the same logic as with `onCancel`.
But this modal window already had `footer: null`, so the OK button was not there

packages/icons/src/wrappers.tsx
```diff
@@ -34,7 +34,7 @@ const getIconSize = (size: IconSize) => {
 
 export const IconWrapper: React.FC<IconWrapperProps> = ({
     icon,
-    size = 'large',
+    size = 'medium',
```
SavelevMatthew, line 37: That'a breaking change. Why would you need it?

packages/ui/src/components/Modal/style.less
```diff
@@ -1,5 +1,6 @@
 @import "antd/lib/modal/style/index.less";
 @import (reference) "@open-condo/ui/src/tokens/variables.less";
+@import "@open-condo/ui/src/components/style/core/global.less";
```
SavelevMatthew, line 3: inside `src/index.tsx` global styles are already imported, so you can just use it here

packages/ui/src/components/Modal/style.less
```diff
@@ -87,9 +90,58 @@
...
 
-      .condo-btn + .condo-btn:not(.condo-dropdown-trigger, :first-child) {
+      > .condo-btn + .condo-btn:not(.condo-dropdown-trigger) {
         margin-left: 0;
       }
     }
+
+    @media @to-tablet-large {
+      position: absolute;
+      left: 0;
```
SavelevMatthew, line 100: Can you attach a screen, how it looks like...because it's seems a bit strange right now

packages/ui/src/components/style/core/global.less
```diff
@@ -6,4 +6,16 @@
 :root {
   --condo-font-fallback: sans-serif;
   --condo-font-fallback-mono: consolas, monaco, "Andale Mono", "Ubuntu Mono", monospace;
-}
\ No newline at end of file
+}
+
+@from-mobile-small: ~"screen and (min-width: 0px)";
```
SavelevMatthew, line 11: Does it mean that syntax is:
```
@media @from-mobile-small and @to-tablet-large
```
and not like?
```
@media @from-mobile-small @to-tablet-large
@media @from-mobile-small
@media @mobile-small
```
as proposed originally?
what was the concerns?

SavelevMatthew:
> Added mobile version of `Modal`. Used `Modal` from condo-ui instead modal from `common` domain in our components.
> 
> Mobile `Modal`: Paddings сhanged from 40px to 24px, modal sticks to the bottom of the window изображение

Why stick to bottom on mobile? What if modal is extremely short? It's not following our styleguides (checked it in Figma, they're still the same)

packages/ui/src/components/style/core/global.less
nomerdvadcatpyat: If we do the second option, we will not be able to write a media query like this: `@media @to-tablet-large` because `@to-tablet-large` starts with `and`.

nomerdvadcatpyat:
> 

изображение

packages/ui/src/components/Modal/style.less
nomerdvadcatpyat: screen attached in the pr description

SavelevMatthew:
> изображение

That can be particular case, since I cannot see anything in web-kit styleguides

nomerdvadcatpyat:
> > изображение
> 
> That can be particular case, since I cannot see anything in web-kit styleguides


No, we need this behavior in all our modals.
Asked the designer to update the web-kit layouts

packages/ui/src/components/Modal/style.less
nomerdvadcatpyat: ok

PR #3146, opened 2023-04-05
fix(condo): DOMA-5614 fix meterReading contact connection

nomerdvadcatpyat:
There was an error creating a contact if there was already contact with the same `phone`, `property`, `unitName`, but other `unitType`

### DIFF
--- a/apps/condo/domains/meter/schema/MeterReading.js
+++ b/apps/condo/domains/meter/schema/MeterReading.js
@@ -81,7 +81,7 @@ const MeterReading = new GQLListSchema('MeterReading', {
             }
 
             if (operation === 'create' && user.type === RESIDENT) {
-                await addClientInfoToResidentMeterReading(context, resolvedData)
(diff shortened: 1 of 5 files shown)


sitozzz (approved):

apps/condo/domains/meter/schema/MeterReading.js
```diff
@@ -81,7 +81,7 @@ const MeterReading = new GQLListSchema('MeterReading', {
             }
 
             if (operation === 'create' && user.type === RESIDENT) {
-                await addClientInfoToResidentMeterReading(context, resolvedData)
+                addClientInfoToResidentMeterReading(context, resolvedData)
```
Alllex202, line 84: without `await`?

nomerdvadcatpyat: is no longer an async function

Alllex202 (approved):

PR #3154, opened 2023-04-06
hotfix(condo): fixed migration 246 - comment about not reversible migration

Alllex202:
problem: can't reapply migration


pahaz (approved):

AntonAL:
Wrong commit message
Not «can be reapplied without error», but «cannot…»

Simpler:
hotfix(condo): fixed migration 246 - comment about not reversible migration

PR #3156, opened 2023-04-06
fix(condo): DOMA-5720 added access for resident user to read BankAccount

AntonAL:
Resident user will post from mobile client following filtering conditions in GraphQL query to read BankAccountReport:

```
{
   bankAccount: {
       property: {
           id: residentProperty,
       }
   }
}
```

To make this possible, it should be able to read BankAccount records, connected to its Property.


sitozzz (approved):

apps/condo/domains/banking/schema/BankAccount.test.js
```diff
@@ -252,6 +254,47 @@ describe('BankAccount', () => {
                 expect(readObjects).toHaveLength(0)
             })
 
+            test('resident user can read for its property', async () => {
```
akarjakin, line 257: `resident user can read for its property`

I guess there's a typo, something is missing

apps/condo/domains/banking/access/BankAccount.js
```diff
@@ -22,41 +22,58 @@ async function canReadBankAccounts ({ authentication: { item: user } }) {
...
-            {
-                integrationContext_is_null: true,
-            }],
+        })
+        conditions.OR.push({
+            integrationContext_is_null: true,
+        })
+    }
 
+    if (user.type === RESIDENT) {
```
akarjakin, line 56: looks like heavy calculations that will be executed for every corresponding request... :(

akarjakin (approved):

apps/condo/domains/banking/access/BankAccount.js
AntonAL: It will be executed for every read request, for example, one `getAll…` request.
Exactly as for `create/update` as in all modules.

PR #3164, opened 2023-04-07
DOMA-5140 bank account report

sitozzz:
Web part of bank account report
image

### DIFF
--- a/apps/condo/domains/banking/components/BankAccountReport.tsx
+++ b/apps/condo/domains/banking/components/BankAccountReport.tsx
@@ -3,10 +3,10 @@ import styled from '@emotion/styled'
 import { Row, Col } from 'antd'
 import dayjs from 'dayjs'
 import ReactECharts from 'echarts-for-react'
+import find from 'lodash/find'
 import get from 'lodash/get'
-import React, { useState, useCallback, useRef, useMemo } from 'react'
+import React, { useState, useCallback, useRef, useMemo, useEffect } from 'react'
 
-import { Wallet } from '@open-condo/icons'
 import { useIntl } from '@open-condo/next/intl'
 import { Tabs, Card, Typography, Select, Option, Space } from '@open-condo/ui'
 import type { TypographyTextProps } from '@open-condo/ui'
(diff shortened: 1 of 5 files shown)


apps/condo/domains/banking/components/BankAccountReport.tsx
```diff
@@ -0,0 +1,401 @@
...
+    onClick?: CardProps['onClick']
+    currencyCode?: string
+    valueType?: TypographyTextProps['type']
+}
+interface IInfoCard { (props: InfoCardProps): React.ReactElement }
+const InfoCard: IInfoCard = ({ value, currencyCode = 'RUB',  ...props }) => {
+    const intl = useIntl()
+    const CurrencyValue = intl.formatNumber(parseFloat(value), { style: 'currency', currency: currencyCode })
+
+    const { isSmall, label, icon, valueType } = props
```
AleX83Xpert, line 96: `valueType` is not required, but has no default value. Is it ok?

AleX83Xpert (approved):

apps/condo/domains/banking/components/BankAccountReport.tsx
```diff
@@ -0,0 +1,401 @@
...
+import { useLayoutContext } from '@condo/domains/common/components/LayoutContext'
+import { Loader } from '@condo/domains/common/components/Loader'
+
+import type { BankAccount as BankAccountType } from '@app/condo/schema'
+import type { RowProps } from 'antd'
+import type { EChartsOption, EChartsReactProps } from 'echarts-for-react'
+
+const BANK_ACCOUNT_REPORT_ROW_GUTTER: RowProps['gutter'] = [40, 40]
+const LABEL_TRUNCATE_LENGTH = 17
+const CHART_COLOR_SET = ['#E45241', '#9036AA', '#4154AE', '#4595EC', '#51B7D1', '#419488', '#96C15B', '#D1DB58', '#FBE960', '#F5C244', '#F29B38']
```
nomerdvadcatpyat, line 28: What are these colors? Why not from our UI-kit?

apps/condo/domains/banking/components/BankAccountReport.tsx
```diff
@@ -0,0 +1,401 @@
...
+    radius: ['80%', '88%'],
+    avoidLabelOverlap: true,
+    top: 'center',
+    left: 'center',
+    width: '100%',
+    height: '100%',
+    labelLine: { show: false },
+}
+
+function truncate (str: string, n: number): string {
```
nomerdvadcatpyat, line 54: it can be common util

sitozzz: It can. But I don't think, that we need it as a common util. Because most of the text display components already has ellipsis logic under the hood. The reason I've added this function is lack of API from echarts to done this right way

apps/condo/domains/banking/components/BankAccountReport.tsx
sitozzz: Maybe it could be a colors from ui-kit. I'll check this out

sitozzz: So I double checked this question & these colors didn't match any string from UI-kit

pahaz (approved):

PR #3167, opened 2023-04-10
refactor(ui): DOMA-5842 fix useBreakpoints linter warnings

nomerdvadcatpyat:


AleX83Xpert (approved):

SavelevMatthew (approved):

sitozzz (approved):

packages/ui/src/components/_utils/hooks/useBreakpoints.tsx
```diff
@@ -25,7 +25,7 @@ export function useBreakpoints (refreshOnChange = true): ScreenMap {
 
             return () => ResponsiveObserve.unsubscribe(token)
         }
-    }, [])
+    }, [forceUpdate, refreshOnChange])
```
akarjakin, line 28: Very strange logic inside forceUpdate, how does it work using the counter, why reducer instead of state ?

packages/ui/src/components/_utils/responsiveObserve.ts
```diff
@@ -67,7 +69,7 @@ const responsiveObserve = {
                 })
             }
             const mql = window.matchMedia(matchMediaQuery)
-            mql.addListener(listener)
+            mql.addEventListener('change', listener)
```
akarjakin, line 72: You will be unable to remove this listener. If it is used inside effects, it should be removed on unmount. To remove it you have to store handlers reference somewhere.

nomerdvadcatpyat: it's store in `matchHandlers` below and removed in `unregister()`

pahaz (approved):

PR #3171, opened 2023-04-10
DOMA-5846 remove reportVisible field from BankAccount & change ui logic to the publishedAt field

sitozzz:


Alllex202 (approved):
you should update migrations

sitozzz:
> you should update migrations

Yup, I'll do it

PR #3176, opened 2023-04-11
fix(condo): DOMA-5826 RecurrentPaymentContext.js - validate omit fiel…

ekabardinsky:
…ds properly

### DIFF
--- a/apps/condo/domains/acquiring/schema/RecurrentPaymentContext.js
+++ b/apps/condo/domains/acquiring/schema/RecurrentPaymentContext.js
@@ -108,7 +108,7 @@ const RecurrentPaymentContext = new GQLListSchema('RecurrentPaymentContext', {
                 throw new GQLError(GQL_ERRORS.RECURRENT_PAYMENT_CONTEXT_NO_TRIGGER_SET_UP_ERROR)
             }
         },
-    }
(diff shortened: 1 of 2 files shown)


apps/condo/domains/acquiring/schema/RecurrentPaymentContext.js
```diff
@@ -104,6 +99,14 @@ const RecurrentPaymentContext = new GQLListSchema('RecurrentPaymentContext', {
         delete: false,
         auth: true,
     },
+    hooks: {
+        validateInput: async ({ resolvedData }) => {
+            //check at least one trigger are set
+            if (!resolvedData['autoPayReceipts'] && isNil(resolvedData['paymentDay'])) {
```
Alllex202, line 105: Perhaps you should also check the presence of these fields in an already created item (when update operation)

Like  `const newItem = { ...existingItem, ...resolvedData }`
and use `newItem` to check `if (!newItem.someField) {...}`

ekabardinsky: 10x

dkoviazin (approved):

pahaz (approved):

PR #3178, opened 2023-04-12
feat(condo): DOMA-5615 Make billing standalone again :)

SavelevMatthew:


apps/condo/domains/billing/components/BillingPageContent/PaymentsTab.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+
+export const PaymentsTab: React.FC = () => {
+    const intl = useIntl()
+    const NoPaymentsTitle = intl.formatMessage({ id:'accrualsAndPayments.payments.noPaymentsYet.title' })
+    const NoPaymentsMessage = intl.formatMessage({ id:'accrualsAndPayments.payments.noPaymentsYet.message' })
+
+    const { organization } = useOrganization()
+    const orgId = get(organization, 'id', null)
+    const { objs: anyPayments, loading: anyPaymentsLoading, error: anyPaymentsError } = Payment.useObjects({
```
sitozzz, line 26: Why not `useObject`?

SavelevMatthew: I think it's a matter of taste, since they're all using basic `useQuery` under the hood. For me `useObjects` is more break-safe for future refactoring.

If you have any points for `useObject` usage over `useObjects` I'll change that with no problem :)

apps/condo/bin/generate-miniapps.js
```diff
@@ -8,6 +8,17 @@ const { AcquiringIntegration } = require('@condo/domains/acquiring/utils/serverS
...
 const { B2BApp } = require('@condo/domains/miniapp/utils/serverSchema')
 
+const bannerVariants = [
+    { bannerColor: '#9b9dfa', bannerTextColor: 'WHITE' },
+    { bannerColor: 'linear-gradient(90deg, #4cd174 0%, #6db8f2 100%)', bannerTextColor: 'BLACK' },
+    { bannerColor: '#d3e3ff', bannerTextColor: 'BLACK' },
+]
+
+function randomChoice (choices) {
+    const index = Math.floor(Math.random() * choices.length)
```
sitozzz, line 18: You can use `random` from `lodash` for this

apps/condo/domains/billing/components/BillingPageContent/PaymentsTab.tsx
sitozzz: It's all about readability. If you just need to receive one object IMO you should use `useObject`. Otherwise you can use `useObjects`. There are no any other points, so you're correct

SavelevMatthew: I need to receive not 1 object, but "at least" one object. First is just a limitation to probably cache-hit better

apps/condo/pages/miniapps/[id]/about.tsx
Comment (AleX83Xpert) [inline apps/condo/pages/miniapps/[id]/about.tsx:1]: This file and `apps/condo/pages/miniapps/[id]/index.tsx` looks similar. Is it as planned?

apps/condo/domains/billing/components/BillingPageContent/ContextProvider.ts
```diff
@@ -0,0 +1,18 @@
+import { createContext, useContext } from 'react'
+
+import type { BillingIntegrationOrganizationContext, AcquiringIntegrationContext } from '@app/condo/schema'
+
+type IBillingAndAcquiringContexts = {
+    billingContext: BillingIntegrationOrganizationContext
+    acquiringContext: AcquiringIntegrationContext
```
AleX83Xpert, line 7: Didn't find usages of acquiring context. Are both contexts really needed to be placed in the same react context?
image

apps/condo/domains/billing/components/OnBoarding/SetupAcquiring.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+    const { obj: billingCtx, loading: billingCtxLoading, error: billingCtxError } = BillingContext.useObject({
+        where: {
+            status: CONTEXT_FINISHED_STATUS,
+            organization: { id: orgId },
+        },
+    })
+
+    // NOTE: On practice there's only 1 acquiring and there's no plans to change it soon
+    const { objs: acquiring, loading: acquiringLoading, error: acquiringError } = AcquiringIntegration.useObjects({
```
AleX83Xpert, line 41: Maybe it's better to use `.useObject(...)` to be sure there is only one object?

apps/condo/domains/billing/schema/BillingIntegration.js
```diff
@@ -107,16 +201,24 @@ const BillingIntegration = new GQLListSchema('BillingIntegration', {
...
-        afterChange: logoMetaAfterChange,
+        afterChange: async (args) => {
+            await logoMetaAfterChange(args)
+            await bannerPromoMetaAfterChange(args)
+        },
+        validateInput: ({ resolvedData, addValidationError, existingItem }) => {
+            const newItem = { ...existingItem, ...resolvedData }
+            if ((newItem['instruction'] && newItem['setupUrl']) ||
+                (!newItem['instruction'] && !newItem['setupUrl'])) {
+                addValidationError(BILLING_INTEGRATION_SINGLE_CONNECT_WAY_ERROR)
```
AleX83Xpert, line 215: We are about to use `throw new GQLError()` instead. Are there reasons to use `addValidationError()`?

apps/condo/domains/billing/components/OnBoarding/SetupAcquiring.tsx
SavelevMatthew: Nah, we need to take first one if multiple found...It's safer than to throw a error here

apps/condo/domains/billing/components/BillingPageContent/ContextProvider.ts
SavelevMatthew: Yep, added them in case of need in payments / accruals tab. However, payments somehow does not use that. But left them both in case they'll be needed

apps/condo/domains/billing/schema/BillingIntegration.js
SavelevMatthew: It's not working in fields error, has awful DX, and nobody except tech. support read it. Should I continue?

AleX83Xpert: 😿 
What about keeping "old style" errors within field hooks?

AleX83Xpert (approved):
LGTM
Only concerned about GQLError within model hooks. But looks like, GQLError does not support field hooks.

PR #3180, opened 2023-04-12
fix(condo): DOMA-5717 do not create a new BankAccountReport instance if the transaction pool has not changed. Add tests

VKislov:


sitozzz (approved):

apps/condo/migrations/20230412150028-0257_auto_20230412_1000.js
```diff
@@ -0,0 +1,32 @@
...
+
+exports.up = async (knex) => {
+    await knex.raw(`
+        BEGIN;
+            --
+            -- [CUSTOM] Set Statement Timeout to some large amount - 25 min (25 * 60 => 1500 sec)
+            --
+            SET statement_timeout = '1500s';  
+            UPDATE "OrganizationEmployeeRole"
+            SET "canManageBankAccountReports" = true
```
sitozzz, line 12: It's an incorrect syntax of SET command. You should use only one `SET` word for each update operation

sitozzz (changes requested):
Please use correct syntax of migration file

sitozzz (approved):

PR #3188, opened 2023-04-13
DOMA-5837 UI-kit ActionBar

nomerdvadcatpyat:
\>= `TABLET_LARGE`: 
изображение

< `TABLET_LARGE`:
изображение

Also made DOMA-5851 task (refactoring mobile layouts)


packages/ui/src/components/ActionBar/actionBar.tsx
```diff
@@ -0,0 +1,26 @@
...
+const ACTION_BAR_CLASS_PREFIX = 'condo-affix'
+const AFFIX_CONTENT_WRAPPER_CLASS = `${ACTION_BAR_CLASS_PREFIX}-content-wrapper`
+
+export type ActionBarProps = {
+    message?: string
+    actions: Array<ReactElement>
+}
+
+export const ActionBar: React.FC<ActionBarProps> = (props) => {
+    const { actions, message } = props
```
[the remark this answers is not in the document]
nomerdvadcatpyat: Yeah, it shows an empty action bar.
I don’t think we need to make logic with hide component if passed empty actions. It’s not obvious.

изображение

apps/condo/domains/common/components/ButtonWithDisabledTooltip.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { Button, ButtonProps, Tooltip } from '@open-condo/ui'
+
+
+type ButtonWithDisabledTooltipProps = ButtonProps & { title: string }
+
+const DisabledButtonWrapper = styled.div<{ disabled }>`
+    ${props => props.disabled ? 'cursor: not-allowed;' : 'cursor: auto;'}
+    .condo-btn {
+      width: 100%;
```
sitozzz, line 12: Look's little hackky) Why it's not a component behaviour by itself?

apps/condo/domains/common/components/ExportToExcelActionBar.tsx
```diff
@@ -27,12 +30,14 @@ export const ExportToExcelActionBar: React.FC<IExportToExcelActionBarProps> = (p
         useTimeZone,
     })
 
-    return (
+    return !hidden &&  (
```
sitozzz, line 33: There is no default hidden prop for ActionBar?

packages/ui/src/components/ActionBar/actionBar.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { Space, SpaceProps } from '../Space'
+import { Typography } from '../Typography'
+
+const ACTION_BAR_CLASS_PREFIX = 'condo-affix'
+const AFFIX_CONTENT_WRAPPER_CLASS = `${ACTION_BAR_CLASS_PREFIX}-content-wrapper`
+
+const SPACE_SIZE: SpaceProps['size'] = [24, 24]
+
+export type ActionBarProps = {
```
sitozzz, line 12: Don't you think that it'll be great to have `visible` prop here to manage component visibility with custom condition?

packages/ui/src/components/ActionBar/actionBar.tsx
```diff
@@ -0,0 +1,28 @@
...
+    message?: string
+    actions: Array<ReactElement>
+}
+
+export const ActionBar: React.FC<ActionBarProps> = (props) => {
+    const { actions, message } = props
```
sitozzz, line 18: You can declare a minimum length of this prop with that construction: `actions: [ReactElement, ...ReactElement[]]`

apps/condo/domains/common/components/ExportToExcelActionBar.tsx
nomerdvadcatpyat: Why? It’s the same as hiding it from the outside.
Also in many places this component is displayed inside the `<Сol><Col/>` and when it is hidden we also hide `<Сol><Col/>`

packages/ui/src/components/ActionBar/actionBar.tsx
nomerdvadcatpyat: Why? It’s the same as hiding it from the outside.
Also in many places this component is displayed inside the <Сol><Col/> and when it is hidden we also hide <Сol><Col/>

apps/condo/domains/common/components/ButtonWithDisabledTooltip.tsx
nomerdvadcatpyat: Ant has a bug that tooltip doesn’t work on the disabled button.
Because it is a hack, I thought it is not worth putting it in the UI-kit

sitozzz (approved):

Alllex202 (approved):

PR #3189, opened 2023-04-13
feat(condo): DOMA-5674 Added the required keys for the 1C transactions parser

VKislov:
Added the required keys for the 1c parser. And the error is now thrown only if there are no required keys.
I didn't add check for required node. If the required node is not found, the transaction simply will not be saved.


sitozzz (commented):
I would better prefer to see special switch that can help you fast change behaviour of parser from strict error throwing to ignoring some of them.
In this case you don't need to modify existing test. You just need to add new one with ignore critical errors cases

sitozzz (approved):

PR #3195, opened 2023-04-14
DOMA-5899 restrict resident access to read BankAccount schema

sitozzz:


nomerdvadcatpyat (approved):

VKislov (approved):

AntonAL (approved):

apps/condo/domains/banking/access/BankAccount.js
```diff
@@ -53,18 +51,6 @@ async function canReadBankAccounts ({ authentication: { item: user } }) {
...
-            const propertyIds = uniq(map(residents, 'property'))
-            conditions.OR.push({
-                property: {
-                    id_in: propertyIds,
-                },
-            })
-        }
-    }
-
     return conditions
```
akarjakin, line 54: does this mean that any resident will get empty conditions here, like support/admin ?

SavelevMatthew: Yep, that should resolve to empty filter, which will cause reading everything...

PR #3198, opened 2023-04-17
DOMA-5906 extend postmessage analytics event name validator list

sitozzz:


apps/condo/domains/common/components/PostMessageProvider/validators.ts
```diff
@@ -16,7 +16,7 @@ const NoParamsValidator = ajv.compile(NoParamsSchema)
 const CondoWebSendAnalyticsEventParamsSchema = {
     type: 'object',
     properties: {
-        event: { type: 'string', enum: ['click', 'check'] },
+        event: { type: 'string', enum: ['click', 'check', 'change', 'select'] },
```
SavelevMatthew, line 19: How about removing select to change btw if you're touching analytics btw?)

sitozzz: Done that

packages/ui/src/components/Select/select.tsx
```diff
@@ -66,8 +66,7 @@ const Select = <ValueType extends SelectValueTypeBase>(props: SelectProps<ValueT
         }
 
         if (title) {
-            // TODO(DOMA-5597): Move it to change event; Also key should be used as well
```
SavelevMatthew, line 69: left a todo about key and multiple select options please, it's still not fixed :)

SavelevMatthew (approved):

packages/ui/src/components/Select/select.tsx
sitozzz: Done

PR #3204, opened 2023-04-18
fix(condo): DOMA-3258 hide details field from residents if ticket has canReadByResident false

nomerdvadcatpyat:
Now we completely hide the ticket if it has `canReadByResident: false` and in the case when there was first `canReadByResident: true`, then became `canReadByResident: false`, in the resident’s mobile app this ticket will continue to be displayed.

We agreed that we will make a filter on `canReadByResident: false` on the client in the mobile app, and from the server will give tickets with `canReadByResident: false`, but hide details

### DIFF
--- a/apps/condo/domains/ticket/schema/Ticket.js
+++ b/apps/condo/domains/ticket/schema/Ticket.js
@@ -300,6 +300,15 @@ const Ticket = new GQLListSchema('Ticket', {
                     return normalizeText(resolvedData['details'])
                 },
             },
+            access: {
+                read: ({ authentication: { item: user }, existingItem }) => {
+                    if (user.type === RESIDENT && !existingItem.canReadByResident) {
+                        return false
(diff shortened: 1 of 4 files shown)


Alllex202 (approved):

akarjakin (approved):

nomerdvadcatpyat:
Waiting for improvements on the mobile app side

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -300,6 +300,15 @@ const Ticket = new GQLListSchema('Ticket', {
                     return normalizeText(resolvedData['details'])
                 },
             },
+            access: {
```
pahaz, line 303: extract it to access. Use the same pattern as `user/access/User.js`

pahaz (approved):

PR #3214, opened 2023-04-20
feat(condo): DOMA-5955 added active state for card

abshnko:
Screenshot 2023-04-20 at 13 37 01

### DIFF
--- a/packages/ui/src/components/Card/style.less
+++ b/packages/ui/src/components/Card/style.less
@@ -30,6 +30,16 @@
     overflow: visible;
   }
 
+  &-active {
+    .condo-card-body {
+      background:
+        linear-gradient(#fff, #fff) padding-box,
+        @condo-global-color-brand-gradient-5 border-box;
+      border-color: transparent;
+      box-shadow: @condo-global-box-shadow-active;
+    }
+  }
+
   &-hoverable {
     transition-duration: @condo-global-transition-duration-default;
     transition-property: box-shadow;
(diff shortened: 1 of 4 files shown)


packages/ui/src/components/Card/style.less
```diff
@@ -57,4 +57,24 @@
       }
     }
   }
+
+  &-active {
```
SavelevMatthew, line 61: Did you passed it through stylelint?

packages/ui/src/components/Card/style.less
```diff
@@ -57,4 +57,24 @@
...
+    .condo-card-body {
+      box-shadow: @condo-global-box-shadow-active;
+      background: linear-gradient(#fff, #fff) padding-box,
+      @condo-global-color-brand-gradient-5 border-box;
+      border: 1px solid transparent;
```
SavelevMatthew, line 66: use `default-border-width` token here

packages/ui/src/components/Card/style.less
```diff
@@ -57,4 +57,24 @@
       }
     }
   }
+
+  &-active {
+    .condo-card-body {
+      box-shadow: @condo-global-box-shadow-active;
```
SavelevMatthew, line 63: Why there's box shadow on card-body here, and on card itself below?

packages/ui/src/components/Card/style.less
```diff
@@ -43,18 +53,14 @@
       transition-duration: @condo-global-transition-duration-default;
       transition-property: border-color;
     }
+  }
 
-    &:hover {
-      box-shadow: @condo-global-box-shadow-default;
```
SavelevMatthew, line 48: Where's `hoverable:hover` state now?

packages/ui/src/components/Card/style.less
SavelevMatthew: Added CI check for that, not relevant

packages/ui/src/components/Card/style.less
```diff
@@ -43,18 +43,42 @@
...
+
+  &-active {
+    background:
+      linear-gradient(#fff, #fff) padding-box,
+      @condo-global-color-brand-gradient-5 border-box;
+    border: @condo-global-border-width-default solid transparent;
+    box-shadow: @condo-global-box-shadow-active;
+
+    .condo-card-body {
+      border: 0 solid transparent;
```
SavelevMatthew, line 56: why not just none?)

SavelevMatthew (approved):

packages/ui/src/components/Card/style.less
SavelevMatthew: got it

PR #3248, opened 2023-04-28
fix(condo): DOMA-6040 fix bankAccount creating in import from 1c

VKislov:


apps/condo/domains/banking/schema/BankSyncTask.test.js
```diff
@@ -525,6 +525,33 @@ describe('BankSyncTask', () => {
             })
         })
 
+        it('expect error when creating BankAccount in 1c import', async () => {
```
sitozzz, line 528: It's not a clear name to describe this test case

sitozzz, line 528: It might be better if you're done it something like that:
```suggestion
        it('throw an error if BankAccount already connected to requested property in file import', async () => {
```

apps/condo/domains/banking/tasks/importBankTransactionsFrom1CClientBankExchange.js
```diff
@@ -1,12 +1,13 @@
 const fetch = require('isomorphic-fetch')
 const { get } = require('lodash')
+const isEmpty = require('lodash/isEmpty')
```
sitozzz, line 3: It's already import from `lodash`. You can just append isEmpty to the previous line
```suggestion
const { get, isEmpty } = require('lodash')
```

apps/condo/domains/banking/tasks/importBankTransactionsFrom1CClientBankExchange.js
```diff
@@ -104,6 +105,24 @@ const importBankTransactionsFrom1CClientBankExchange = async (taskId) => {
...
+        })
+        if (!isEmpty(accountByProperty)) {
+            await BankSyncTask.update(context, task.id, {
+                status: BANK_SYNC_TASK_STATUS.ERROR,
+                meta: {
+                    errorMessage: `Already have an account with the same Property { id: ${property.id} }.`,
+                },
+                ...DV_SENDER,
+            })
+            return
```
sitozzz, line 123: Maybe throw an error and log some error to the logger instead of just return?

sitozzz (approved):

PR #3263, opened 2023-05-02
fix(condo): DOMA-5735 fixed details table

abshnko:
fixed borders for details table

Screenshot 2023-05-02 at 17 48 36


SavelevMatthew (changes requested):
We've decided not to use emotion. And this change will make it even harder to migrate to UI-kit.

Please change it globally or do not change at all

SavelevMatthew (approved):

PR #3268, opened 2023-05-03
feat(ui): DOMA-5597 updated Select, Radio 

Alllex202:


packages/ui/src/components/_utils/analytics.ts
```diff
@@ -129,6 +129,15 @@ export function sendAnalyticsChangeEvent<Component extends ComponentNames['chang
...
         }
+        console.group()
+        console.log('>>> sendMassage')
+        console.log({
+            handler: ANALYTICS_HANDLER_NAME,
+            params,
+            type: 'condo-ui',
+            version,
+        })
+        console.groupEnd()
```
Alllex202, line 140: should delete

packages/ui/src/components/Select/select.tsx
```diff
@@ -1,6 +1,8 @@
 import { Select as DefaultSelect } from 'antd'
 import classNames from 'classnames'
+import { isEmpty } from 'lodash'
```
Alllex202, line 3:
```suggestion
import isEmpty from 'lodash/isEmpty'
```

packages/ui/src/components/Select/hooks/useItems.tsx
```diff
@@ -0,0 +1,56 @@
...
+        .map((item,  index) => {
+            if (item === null) return null
+
+            const { key } = item
+            const mergedKey = key || index
+            const isOptionsGroup = Boolean(get(item, 'options', false))
+
+            if (isOptionsGroup) {
+                // Options group
+                const { label, options } = item as OptionsGroupType
```
SavelevMatthew, line 24: why typecasting here and below?

Alllex202: Without this, an error appears with the message `Property 'textType' does not exist on type 'OptionType | OptionsGroupType'`

Alllex202: fixed

packages/ui/src/components/Select/select.tsx
```diff
@@ -9,28 +10,54 @@ import type { TypographyTextProps } from '@open-condo/ui/src'
...
+
+import { Either } from '../_utils/types'
 
 import type { SelectProps as DefaultSelectProps } from 'antd'
 
-const SELECT_CLASS_PREFIX = 'condo-select'
-const Option: typeof DefaultSelect.Option = DefaultSelect.Option
-const OptGroup: typeof DefaultSelect.OptGroup = DefaultSelect.OptGroup
+export const SELECT_CLASS_PREFIX = 'condo-select'
+export const Option: typeof DefaultSelect.Option = DefaultSelect.Option
```
SavelevMatthew, line 20: Why you specifying type by yourself?
```
const a = 5 // typeof a = typeof 5
```

packages/ui/src/components/Select/select.tsx
```diff
@@ -9,28 +10,54 @@ import type { TypographyTextProps } from '@open-condo/ui/src'
...
+    value: OptionType['value'] | Array<OptionType['value']>,
+    option: React.PropsWithChildren<OptionType> | Array<React.PropsWithChildren<OptionType>>
+) => void
+
+/**
+ * NOTE: Ant have bad interface for Select
+ * It may work with multiple values, but it is not specified in the interface.
+ */
+type CustomSelectProps<ValueType = SelectValueTypeBase> = Either<{
+    value?: ValueType | null
```
SavelevMatthew, line 51: Why do you use `null`? Is there any logic to this?

Because `prop?: MyType` is equal to `prop: MyType | undefined`...
And yours are equal to `value: ValueType | null | undefined`. Seems like null is redundant in this scenario, isn't it?

packages/ui/src/components/Select/select.tsx
```diff
@@ -9,28 +10,54 @@ import type { TypographyTextProps } from '@open-condo/ui/src'
...
+}, {
+    mode: 'multiple'
+    value?: ValueType[] | null
```
SavelevMatthew, line 54: Af far as I remember we prefer to use `Array<Type>` instead of `Type[]`

packages/ui/src/components/Select/select.tsx
Alllex202: But we should be able to specify a value equal to `null`

For example, if we want to add an empty option

packages/ui/src/components/Select/select.tsx
Alllex202: It's already been written, I will remove the excess :)

packages/ui/src/components/Select/select.tsx
Alllex202: fixed

sitozzz (approved):

packages/ui/src/components/Select/select.tsx
SavelevMatthew: Got it 👍

packages/ui/src/components/Select/select.tsx
SavelevMatthew: Missed that part) Thank you for clarifying. 

But I guess in multiple select it's always is array, isn't it?
`Nothing selected = []`

SavelevMatthew (commented):
Fix multiple-case null value, everything else is LGTM 🔥

Also a question, why decided to use string `mode` instead of bool? Does Antd have other choices, we can implement there later?

In my personal opinion is simpler to use modifier similar to `disabled`:
```Typescript jsx
<Select>
// ...
</Select>
<Select multiple>
// ...
</Select>
```

Alllex202:
> Fix multiple-case null value, everything else is LGTM 🔥
> 
> Also a question, why decided to use string `mode` instead of bool? Does Antd have other choices, we can implement there later?
> 
> In my personal opinion is simpler to use modifier similar to `disabled`:
> 
> ```ts
> <Select>
> // ...
> </Select>
> <Select multiple>
> // ...
> </Select>
> ```

`Ant` `Select` also has a `tags` mode. It will probably need to be added later.

packages/ui/src/components/Select/select.tsx
Alllex202: Fixed it for multiple select

SavelevMatthew (approved):

PR #3280, opened 2023-05-04
DOMA-6067 Introduce ticket creation limits for Residents.

toplenboren:
## Task

Limit 10 requests per day per user of the Resident application (at user level)
No more than 2 requests per day with completely identical text

All limits apply to one ORG. That is, I as a resident can send 10 messages to the Managing Company "Romashka" and another 10 messages to the "Dvornik".

## Context

There are users of the "terrorists" who send several hundred applications to the control company to get an answer or a solution to their questions. Such cases were 2, at the request of the UK we have to block the user. Result: Negative from the Managing Company and possible negative from the user in storerooms

## PS

I still wait for texts from marketing department... So there are no explicit verbose messages yet..


apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -75,8 +86,46 @@ const ERRORS = {
...
+ * @param {string} phone
+ * @param {string} organizationId
+ * @param {string} details
+ * @param context
+ * @returns {Promise<void>}
+ */
+const checkDailyTicketLimit = async ({ phone, organizationId, details, context }) => {
+    const byPhoneAndOrgKey = phone + '_' + organizationId
+    const byPhoneAndOrgCounter = await redisGuard.incrementDayCounter(byPhoneAndOrgKey)
+    if (byPhoneAndOrgCounter > DAILY_TICKET_LIMIT && !phoneWhiteList.includes(phone)) {
```
nomerdvadcatpyat, line 117: Why we need a `phoneWhiteList`?

apps/condo/domains/ticket/schema/Ticket.test.js
```diff
@@ -1664,6 +1664,71 @@ describe('Ticket', () => {
...
+                })
+                for (let i = 0; i < DAILY_TICKET_LIMIT; ++i) {
+                    await createTestTicket(userClient, userClient.organization, userClient.property, { unitName })
+                }
+
+                await expectToThrowGQLError(async () => {
+                    await createTestTicket(userClient, userClient.organization, userClient.property, { unitName })
+                }, ERRORS.TICKET_FOR_PHONE_DAY_LIMIT_REACHED)
+
+                // User, banned from creating tickets to organization 1, can still create tickets to organization 2
```
VKislov, line 1716: It seems non-obvious to test two different cases in one "it". I have not seen anything like this in our code.

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -75,8 +86,46 @@ const ERRORS = {
...
+ * Whitelisted phones are excluded from this rule.
+ *
+ * @param {string} phone
+ * @param {string} organizationId
+ * @param {string} details
+ * @param context
+ * @returns {Promise<void>}
+ */
+const checkDailyTicketLimit = async ({ phone, organizationId, details, context }) => {
+    const byPhoneAndOrgKey = phone + '_' + organizationId
```
AntonAL, line 115: Could you please scope the key to the feature:
For example:
`dailyTicketLimit:organization:${organizationId}:phone:${phoneNumber}`

apps/condo/domains/ticket/schema/Ticket.js
toplenboren: I think, for debug purposes or for local deveopment. For example, QA or mobile app deveopers should be able to create more then 10 tickets if they are testing something.

apps/condo/domains/ticket/schema/Ticket.test.js
toplenboren: There are some code like this in billing domain.. It comes handy and more efficient in terms of test execution time. 

But I agree, that this is kinda against best practices.

apps/condo/domains/ticket/schema/Ticket.js
toplenboren: Done, thanks!

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -75,8 +86,50 @@ const ERRORS = {
...
+ * @param {string} details
+ * @param context
+ * @returns {Promise<void>}
+ */
+const checkDailyTicketLimit = async ({ phone, organizationId, details, context }) => {
+    if (phoneWhiteList.includes(phone)) {
+        return
+    }
+
+    const byPhoneAndOrgKey = `dailyTicketLimit:phone:${phone}:org:${organizationId}`
```
AntonAL, line 119:
```suggestion
    const byPhoneAndOrgKey = `dailyTicketLimit:phone:${phone}:organization:${organizationId}`
```

AntonAL (approved):

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -75,8 +86,50 @@ const ERRORS = {
         message: '"qualityControlValue" must be specified if there is "qualityControlAdditionalOptions" or "qualityControlComment"',
         messageForUser: 'api.ticket.QUALITY_CONTROL_VALUE_MUST_BE_SPECIFIED',
     },
+    TICKET_FOR_PHONE_DAY_LIMIT_REACHED: {
+        code: 'BAD_USER_INPUT',
+        type: 'TICKET_FOR_PHONE_DAY_LIMIT_REACHED',
+        message: 'TICKET_FOR_PHONE_DAY_LIMIT_REACHED',
```
Alllex202, line 92: maybe add a readable message? And `messageForUser`?

Alllex202 (approved):

.env.example
```diff
@@ -46,3 +46,6 @@ BANK_ACCOUNT_REQUEST_EMAIL_TARGET=help@doma.ai
 # APPLE_VOIP_PUSH_TOKEN_TEST=<TOKEN>
 
 MOBILE_APP_IDS = { "resident": { "ios": "ai.doma.clients", "android": "ai.doma.client" }, "master": { "ios": "ai.doma.master", "android": "ai.doma.master" }}
+
+DAILY_SAME_TICKET_LIMIT=2
```
SavelevMatthew, line 50: Seems like it's not necessary to build an app...

apps/condo/domains/ticket/schema/Ticket.js
SavelevMatthew, line 92: Why message is the same as type? And `messageForUser` must be here, since it resident-oriented error.

apps/condo/domains/ticket/schema/Ticket.test.js
```diff
@@ -1664,6 +1664,71 @@ describe('Ticket', () => {
...
+                for (let i = 0; i < DAILY_SAME_TICKET_LIMIT; ++i) {
+                    await createTestTicket(userClient, userClient.organization, userClient.property, { unitName, details })
+                }
+
+                await expectToThrowGQLError(async () => {
+                    await createTestTicket(userClient, userClient.organization, userClient.property, { unitName, details })
+                }, ERRORS.SAME_TICKET_FOR_PHONE_DAY_LIMIT_REACHED)
+
+                // User, banned from creating tickets to organization 1, can still create tickets to organization 2
+                const client2 = await makeClientWithProperty()
```
SavelevMatthew, line 1686: Can you rename it please to make a bit more clear? `userClient` and `client2` IMO should mean similar thing, but one of them is resident, and second one is Organization admin, which is kinda opposite.

apps/condo/domains/ticket/schema/Ticket.test.js
SavelevMatthew: IMO this scenario is ok. Since it will duplicate most of its logic (create N+1 ticket), receive a ban, then try another org.

SavelevMatthew (approved):
You should:
1. Clean .env.example, since we're trying to keep it MVP
2. Add user message
3. Squash :)

Otherwise, everything looks great 🥇

nomerdvadcatpyat (approved):

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -60,7 +63,15 @@ const {
     relatedManyToManyResolvers,
 } = require('@condo/domains/ticket/utils/serverSchema/TicketChange')
 const { RESIDENT } = require('@condo/domains/user/constants/common')
+const { RedisGuard } = require('@condo/domains/user/utils/serverSchema/guards')
+
+
+const phoneWhiteList = Object.keys(conf.SMS_WHITE_LIST ? JSON.parse(conf.SMS_WHITE_LIST) : {})
+
+const redisGuard = new RedisGuard()
```
pahaz, line 71: could you please create one global redisGuard instance and use it everywhere. Because every new RedisGuard generate a new redis connection.

.env.example
pahaz: We really can skip in here

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -580,6 +633,18 @@ const Ticket = new GQLListSchema('Ticket', {
             // Todo(zuch): add placeClassifier, categoryClassifier and classifierRule
             if (!hasDbFields(['organization', 'source', 'status', 'details'], resolvedData, existingItem, context, addValidationError)) return
 
+            const user = get(context, ['req', 'user'])
+            const userType = get(user, 'type')
+
+            if (userType === RESIDENT && operation === 'create') {
+                await checkDailyTicketLimit({
+                    phone: get(user, 'phone'),
```
pahaz, line 641: It's better to use a `user.id` here. Because the phone is not required field.

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -75,8 +86,50 @@ const ERRORS = {
...
+ *
+ * Whitelisted phones are excluded from this rule.
+ *
+ * @param {string} phone
+ * @param {string} organizationId
+ * @param {string} details
+ * @param context
+ * @returns {Promise<void>}
+ */
+const checkDailyTicketLimit = async ({ phone, organizationId, details, context }) => {
```
pahaz, line 114: You will get an unexpected behavior in case of `organizationId` is undefined / null please add a guards for that.

pahaz, line 114:
```suggestion
const checkDailyTicketLimit = async ({ phone, organizationId, details, context }) => {
    if (!organizationId || typeof organizationId !== 'string') throw new Error('unexpected organizationId value or type')
    if (!phone || typeof phone !== 'string') throw new Error('unexpected phone value or type')    
```

apps/condo/domains/ticket/schema/Ticket.test.js
pahaz: I aggree with @SavelevMatthew it's it looks like a real QA test case.

pahaz (approved):

apps/condo/domains/ticket/schema/Ticket.js
pahaz: Nice catch! please add some abstract message for user like: "You reached a daily ticket count limit"

toplenboren: As stated in description, the messages will be added once they are ready.

toplenboren: Yep, messageForUser will be added later, once the texts are ready. It is stated in descr

apps/condo/domains/ticket/schema/Ticket.js
toplenboren: Ok, nice catch

apps/condo/domains/ticket/schema/Ticket.js
toplenboren: Ok

sitozzz, line 114: What about to implement single redis connection & just use it here? It's better solution for a memory consumption

toplenboren: Yep, would be done in another PR

PR #3281, opened 2023-05-04
Hotfix/condo/remove old cron tasks for sbbol

AntonAL:


apps/condo/domains/organization/tasks/index.js
```diff
@@ -1,7 +1,13 @@
+const { removeCronTask } = require('@open-condo/keystone/tasks')
+
 const refreshSbbolClientSecret = require('./refreshSbbolClientSecret')
 const { syncSbbolBankAccountsCron: syncSbbolBankAccounts } = require('./syncSbbolBankAccounts')
 const { syncSbbolTransactionsCron: syncSbbolTransactions } = require('./syncSbbolTransactions')
 
+removeCronTask('syncSbbolSubscriptions', '20 9 * * *')
```
dkoviazin, line 7: Do we need such code in main repository?
You can add execute a script to clear old repeatable

https://github.com/open-condo-software/condorb/blob/master/bin/clearOldRepeatableTasks.js

pahaz (approved):

apps/condo/domains/organization/tasks/index.js
pahaz: o-O nice! probably we need to add this to https://github.com/open-condo-software/condo/blob/master/packages/keystone/tasks.js#L250

AntonAL: I've propozed to @pahaz an idea of a bin utility, that can be manually invoked to clear tasks
@pahaz said, that such manual operations should not be executed in server.
I was partially agree with it

PR #3284, opened 2023-05-04
Add metrics to .env.example

toplenboren:


.env.example
```diff
@@ -11,6 +11,11 @@ FAKE_ADDRESS_SUGGESTIONS=true
 GOOGLE_RECAPTCHA_CONFIG='{"SITE_KEY":"6LcPRvQaAAAAAJRyxsFIB4rP5VH036pFOkNH8lgh", "SERVER_KEY":"6LcPRvQaAAAAADn_h1440Es7fXIGD0E4lpXR_FyF"}'
 HELP_REQUISITES='{ "support_email": "help@doma.ai", "support_email_mobile": "helpmobile@doma.ai", "bot_email": "service@doma.ai", "support_phone": "8 800 700-36-62" }'
 
+# Statsd metrics settings
```
pahaz, line 14: Its should not be required. May be you need to add it as commented example here.

packages/keystone/metrics.js
```diff
@@ -18,6 +18,7 @@ const { getLogger } = require('./logging')
 const logger = getLogger('metrics')
 
 const STATSD_METRIC_PREFIX = conf['STATSD_METRIC_PREFIX'] || 'condo.'
+const STATSD_HOST = conf['STATSD_HOST'] || conf['DD_AGENT_HOST'] || '127.0.0.1'
```
pahaz, line 21: may be you don't need to init a StatsD if there is no ENV

PR #3296, opened 2023-05-10
fix(condo) HOTFIX: fix email translation function in templates.js

toplenboren:
if you put substituteTranslations(...) like this: 

1. if isObject(item) => return A
2. if isArray(item) => return B

then the 2 will never be executed, since typeof [] === 'object' in JS.


toplenboren:
https://www.githubstatus.com/ github is dead for the moment of creation of this pr, gonna restart the checks in 8-10 hours

akarjakin (approved):

nomerdvadcatpyat (approved):

Alllex202 (approved):

PR #3321, opened 2023-05-16
fix(condo): DOMA-5914 fixed multiple filter modal footer

Alllex202:
fix(ui): DOMA-5914 added minimum width for Button component

fix(condo): DOMA-5914 added label for ticket import button

### Before:

https://github.com/open-condo-software/condo/assets/56914444/91ab4c4b-3bb3-45e2-b7bc-b9390281d417

### After:

https://github.com/open-condo-software/condo/assets/56914444/20901620-5bce-4de8-bd16-a04bbba9c748

### DIFF
--- a/packages/ui/src/components/Button/style.less
+++ b/packages/ui/src/components/Button/style.less
@@ -18,6 +18,7 @@
   display: inline-flex;
   align-items: center;
   justify-content: center;
+  min-width: 64px;
   text-shadow: none;
   border-radius: @condo-button-border-radius;
   box-shadow: none;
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/Button/style.less
```diff
@@ -236,10 +237,10 @@
       background: @condo-global-color-brand-gradient-5 border-box;
       border: @condo-button-border-width solid transparent;
       opacity: 0;
+      transition: opacity @condo-global-transition-duration-default ease-out;
```
nomerdvadcatpyat, line 240: stylelint?

nomerdvadcatpyat (approved):

packages/ui/src/components/Button/style.less
Alllex202: yep

PR #3333, opened 2023-05-18
feat(ui): DOMA-6155 added ability to add extracontent to tabs

Kekmus:
Screenshot 2023-05-18 at 07 56 33

### DIFF
--- a/packages/ui/src/components/Tabs/style.less
+++ b/packages/ui/src/components/Tabs/style.less
@@ -10,6 +10,7 @@
 @tabs-hover-color: @condo-global-color-black;
 @tabs-horizontal-padding: @condo-global-spacing-4 0 @condo-global-spacing-12;
 @tabs-bar-margin: 0 0 @condo-global-spacing-40 0;
+@extra-content-margin: @condo-global-spacing-40;
 
 .condo-tabs {
   &-tab-btn {
@@ -55,5 +56,13 @@
 
   &-nav-operations {
     display: none !important;
+  }
+
+  &-nav {
+    .condo-tabs-extra-content {
+      &:nth-of-type(1) {
+        margin-right: @extra-content-margin;
+      }
+    }
   }
 }
--- a/packages/ui/src/components/Tabs/tabs.tsx
+++ b/packages/ui/src/components/Tabs/tabs.tsx
@@ -16,6 +16,11 @@ export type TabItem = {
     children?: React.ReactNode
 }
 
+type CondoTabsProps = {
+    leftExtraContent?: React.ReactNode
+    rightExtraContent?: React.ReactNode
+}
+
 export type TabsProps = Pick<DefaultTabsProps,
 'className' |
 'id' |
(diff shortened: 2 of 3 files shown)


dkoviazin (approved):

packages/ui/src/components/Tabs/style.less
```diff
@@ -56,4 +57,12 @@
   &-nav-operations {
     display: none !important;
   }
+
+  &-nav {
+    .condo-tabs-extra-content {
+      &:nth-of-type(1) {
```
SavelevMatthew, line 63: why do you select like this?
`a { b { c {  } } }` instead of simple `a b c {}`

SavelevMatthew: And why nth of type, if there's a single element?

SavelevMatthew (changes requested):
Why `leftExtraContent` is here? I didn't see it in any design layouts.
And what should it do when tabs are vertical?

Did you talk to the designer about this edit or did it yourself based on Dima's wishes?

packages/ui/src/components/Tabs/style.less
```diff
@@ -10,8 +10,13 @@
 @tabs-hover-color: @condo-global-color-black;
 @tabs-horizontal-padding: @condo-global-spacing-4 0 @condo-global-spacing-12;
 @tabs-bar-margin: 0 0 @condo-global-spacing-40 0;
+@extra-content-margin: @condo-global-spacing-40;
 
 .condo-tabs {
+  &-tab {
+    align-items: end;
```
SavelevMatthew, line 17: What's there purpose of that rule?

packages/ui/src/components/Tabs/style.less
```diff
@@ -56,4 +61,8 @@
   &-nav-operations {
     display: none !important;
   }
+
+  &-nav .condo-tabs-extra-content {
+    padding-left: @extra-content-margin;
```
SavelevMatthew, line 66: `padding: @margin`?

packages/ui/src/components/Tabs/tabs.tsx
```diff
@@ -16,6 +16,10 @@ export type TabItem = {
     children?: React.ReactNode
 }
 
+type CondoTabsProps = {
+    extraContent?: React.ReactNode
```
SavelevMatthew, line 20: Why not to keep name `tabBarExtraContent` from Antd?

packages/ui/src/components/Tabs/style.less
Kekmus: I discussed with the design and this task is a temporary solution until the button in the layout is moved and I was asked to make the text tabs and buttons on the same height 

with that rule:
Screenshot 2023-05-19 at 13 42 43

without  that rule:
Screenshot 2023-05-19 at 13 44 36

packages/ui/src/components/Tabs/style.less
```diff
@@ -56,4 +60,8 @@
   &-nav-operations {
     display: none !important;
   }
+
+  &-nav .condo-tabs-extra-content {
```
SavelevMatthew, line 64: It should be only applied to top/bottom tabs, shouldn't  it?

packages/ui/src/components/Tabs/tabs.tsx
```diff
@@ -22,6 +22,7 @@ export type TabsProps = Pick<DefaultTabsProps,
 'defaultActiveKey' |
 'activeKey' |
 'destroyInactiveTabPane' |
+'tabBarExtraContent' |
```
SavelevMatthew, line 25: Use `React.ReactElement` (or ReactNode if something like `[<Item/>, <Item/>, <Item/>]` is allowed), since Antd by default has config (or Node), with `left`, 'right`, `bottom`....props, which is not what we want

PR #3334, opened 2023-05-18
feat(condo): DOMA-6090 no receipts for proceeding recurrent payment

ekabardinsky:

PR #3335, opened 2023-05-18
feat(condo): DOMA-6121 do not send notification for removed recurrent payment context

ekabardinsky:


Alllex202:
It seems to be very similar to https://github.com/open-condo-software/condo/pull/3334.

Alllex202 (approved):

PR #3338, opened 2023-05-18
DOMA-6140 extending registry_ma config to support spp flow

dkoviazin:
We need to make SBBOL requests, so we have duplicated SBBOL config here
Now registry has 2 flows for SPP and ordinary users, so we need featureFlag config
Admin integration is used as a walk around for uploading files on server side and will be removed


Alllex202 (approved):

dkoviazin:
Already in master

PR #3346, opened 2023-05-19
fix(condo): DOMA-6045 read actual receipts in recurrent payment backg…

ekabardinsky:
…round process

### DIFF
--- a/apps/condo/domains/acquiring/utils/taskSchema/index.js
+++ b/apps/condo/domains/acquiring/utils/taskSchema/index.js
@@ -151,7 +151,8 @@ async function getReceiptsForServiceConsumer (context, date, serviceConsumer, bi
 
     const periodDate = dayjs(date)
     const period = periodDate.format('YYYY-MM-01')
-    const previousPeriod = periodDate.startOf('month').subtract(1, 'days').format('YYYY-MM-01')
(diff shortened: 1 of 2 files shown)


apps/condo/domains/acquiring/utils/taskSchema/index.js
```diff
@@ -151,7 +151,8 @@ async function getReceiptsForServiceConsumer (context, date, serviceConsumer, bi
 
     const periodDate = dayjs(date)
     const period = periodDate.format('YYYY-MM-01')
-    const previousPeriod = periodDate.startOf('month').subtract(1, 'days').format('YYYY-MM-01')
+    const monthBeforePeriod = periodDate.startOf('month').subtract(1, 'month').format('YYYY-MM-01')
+    const twoMonthBeforePeriod = periodDate.startOf('month').subtract(2, 'month').format('YYYY-MM-01')
```
sitozzz, line 155: What the difference between `YYYY-MM-01` and `YYYY-MM-DD` date formats?

ekabardinsky: Since I'm looking for receipts in certain periods, I have to follow billingReceipt.period field format. The format is: YYYY-MM-01, that is literally the first day of the month. I prefer to use exact the same format, like in billingReceipt.period

sitozzz (approved):

PR #3356, opened 2023-05-22
feat(condo): DOMA-6170 New fields for push messages

AleX83Xpert:


apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.js
```diff
@@ -92,6 +92,9 @@ async function sendNotifications (newsItem) {
                     userId: resident.user,
                     residentId: resident.id,
                     url: `${conf.SERVER_URL}/newsItem`,
+                    validBefore: get(newsItem, 'validBefore', null),
+                    // The first truthy value will be returned, or null if no values are found.
+                    dateCreated: ['sendAt', 'updatedAt', 'createdAt'].reduce((result, field) => (result || get(newsItem, field)), null),
```
sitozzz, line 97: I would rather use `find` here. If I'm understand the main idea correctly this line of code will do the job same way
```suggestion
                    dateCreated: ['sendAt', 'updatedAt', 'createdAt'].find(field => get(newsItem, field)) || null,
```

sitozzz (approved):

Alllex202 (approved):

PR #3362, opened 2023-05-23
fix(condo): DOMA-5148 fixed deleting a unit in map editor

Alllex202:
### Before: can't delete empty unit
Снимок экрана 2023-05-23 в 12 18 38

### After: can delete empty unit
(Also can delete unit with repeated name)
Снимок экрана 2023-05-23 в 12 18 08

### DIFF
--- a/apps/condo/domains/property/components/panels/Builder/forms/UnitForm.tsx
+++ b/apps/condo/domains/property/components/panels/Builder/forms/UnitForm.tsx
@@ -264,7 +264,7 @@ const UnitForm: React.FC<IPropertyMapModalForm> = ({ builder, refresh, setDuplic
                                         onClick={deleteUnit}
                                         type='sberDangerGhost'
                                         icon={<DeleteFilled />}
-                                        disabled={!(floor && section && label.trim() && !isValidationErrorVisible)}
+                                        disabled={!(floor && section)}
                                         data-cy='property-map__unit-form__delete-button'
                                     >{DeleteLabel}</Button>
                                 </Col>


apps/condo/domains/property/components/panels/Builder/forms/UnitForm.tsx
```diff
@@ -264,7 +264,7 @@ const UnitForm: React.FC<IPropertyMapModalForm> = ({ builder, refresh, setDuplic
                                         onClick={deleteUnit}
                                         type='sberDangerGhost'
                                         icon={<DeleteFilled />}
-                                        disabled={!(floor && section && label.trim() && !isValidationErrorVisible)}
```
AleX83Xpert, line 267: Is `isValidationErrorVisible ` not mandatory here?

AleX83Xpert (approved):

apps/condo/domains/property/components/panels/Builder/forms/UnitForm.tsx
Alllex202: I found only a name repetition error, but seem this is not relevant for deleting a unit

sitozzz (approved):

apps/condo/domains/property/components/panels/Builder/forms/UnitForm.tsx
Alllex202: This error should not block deletion
Снимок экрана 2023-05-23 в 15 33 17

PR #3365, opened 2023-05-23
fix(condo): DOMA-3289 added spaces trimming for "number" and "account Number"… 

Alllex202:
… fields in meter modal form

Problem: you can save field "number" and "accountNumber" for "Meter" only with spaces or without spaces trimming at start and end

Also added this validation for import and fixed translates


nomerdvadcatpyat (approved):

apps/condo/domains/meter/hooks/useUpdateMeterModal.tsx
```diff
@@ -45,10 +45,16 @@ export const useUpdateMeterModal = (refetch) => {
     }, [selectedMeter])
 
     const handleSubmit = useCallback(values => {
+        const number = String(get(values, 'number', '')).trim()
+        const accountNumber = String(get(values, 'accountNumber', '')).trim()
```
AntonAL, line 49: Why not to implement this trimming in `resolveInput` hook of `Meter` schema?

Alllex202: Yes, it happens in `resolveInput` hook

apps/condo/domains/meter/hooks/useCreateMeterModal.tsx
```diff
@@ -17,8 +18,13 @@ export function useCreateMeterModal (organizationId: string, propertyId: string,
 
     const handleMeterCreate = useCallback(values => {
         const numberOfTariffs = values.numberOfTariffs || 1
+        const number = String(get(values, 'number', '')).trim()
+        const accountNumber = String(get(values, 'accountNumber', '')).trim()
```
SavelevMatthew, line 22: Why explicit cast to string?

apps/condo/domains/meter/hooks/useUpdateMeterModal.tsx
Alllex202: removed it from submit handler

apps/condo/domains/meter/hooks/useCreateMeterModal.tsx
Alllex202: you're right, it's not needed here. Removed it

PR #3375, opened 2023-05-24
fix(condo): DOMA-6070 change of recurrent payment context trigger sho…

ekabardinsky:
…uld remove child recurrent payments too

### DIFF
--- a/apps/condo/domains/acquiring/schema/RecurrentPaymentContext.js
+++ b/apps/condo/domains/acquiring/schema/RecurrentPaymentContext.js
@@ -113,12 +113,14 @@ const RecurrentPaymentContext = new GQLListSchema('RecurrentPaymentContext', {
                 const recurrentPaymentContextId = get(existingItem, 'id', null)
                 const sender = get(originalInput, 'sender', null)
                 const dv = get(originalInput, 'dv', null)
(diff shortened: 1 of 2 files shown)


apps/condo/domains/acquiring/schema/RecurrentPaymentContext.js
```diff
@@ -113,12 +113,14 @@ const RecurrentPaymentContext = new GQLListSchema('RecurrentPaymentContext', {
...
                 const sender = get(originalInput, 'sender', null)
                 const dv = get(originalInput, 'dv', null)
+                const autoPayReceipts = get(existingItem, 'autoPayReceipts', null)
 
                 // data vars
                 const newServiceConsumerId = get(originalInput, 'serviceConsumer.connect.id', null)
                 const serviceConsumerId = get(existingItem, 'serviceConsumer', null)
                 const newPaymentDay = get(originalInput, 'paymentDay', null)
                 const paymentDay = get(existingItem, 'paymentDay', null)
+                const newAutoPayReceipts = get(originalInput, 'autoPayReceipts', null)
```
Alllex202, line 123: Is it correct that this value is taken from `originalInput` and not from `const newItem = { ...existingItem, ...resolvedData }`?

and other cases too?

PR #3387, opened 2023-05-25
fix(condo): DOMA-5070 added validation meter values when meters import

Alllex202:
### Before
You get error from meters import when imported file has invalid meter values

Снимок экрана 2023-05-25 в 21 45 22

### After
Errors are displayed in the table without completing the import with an error. 
Empty values are ignored. 

Снимок экрана 2023-05-25 в 21 47 51

### DIFF
--- a/apps/condo/domains/meter/hooks/useImporterFunctions.ts
+++ b/apps/condo/domains/meter/hooks/useImporterFunctions.ts
@@ -25,8 +25,6 @@ import {
     RowNormalizer,
     RowValidator,
 } from '@condo/domains/common/utils/importer'
-import { searchPropertyWithMap } from '@condo/domains/property/utils/clientSchema/search'
-
 import {
     COLD_WATER_METER_RESOURCE_ID,
     ELECTRICITY_METER_RESOURCE_ID,
@@ -34,9 +32,12 @@ import {
     HEAT_SUPPLY_METER_RESOURCE_ID,
     HOT_WATER_METER_RESOURCE_ID,
     IMPORT_CONDO_METER_READING_SOURCE_ID,
-} from '../constants/constants'
-import { Meter, MeterReading } from '../utils/clientSchema'
-import { searchMeter } from '../utils/clientSchema/search'
+} from '@condo/domains/meter/constants/constants'
+import { Meter, MeterReading } from '@condo/domains/meter/utils/clientSchema'
+import { searchMeter } from '@condo/domains/meter/utils/clientSchema/search'
+import { normalizeMeterValue, validateMeterValue } from '@condo/domains/meter/utils/helpers'
+import { searchPropertyWithMap } from '@condo/domains/property/utils/clientSchema/search'
+
 
 const MONTH_PARSING_FORMAT = 'YYYY-MM'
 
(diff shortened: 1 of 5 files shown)


nomerdvadcatpyat (approved):

apps/condo/domains/meter/hooks/useImporterFunction.ts
```diff
@@ -285,18 +297,18 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
...
-            meterType,
+            , // meterType
             meterNumber,
             numberOfTariffs,
-            value1,
-            value2,
-            value3,
-            value4,
-            readingSubmissionDate,
+            , // value1
```
sitozzz, line 307: Why this lines become commented? Could it be used in the future?

Alllex202: This is to avoid getting too many warnings about `Unused constant` from `ESLint`

Alllex202: you can use them if you need

apps/condo/domains/meter/hooks/useImporterFunction.ts
```diff
@@ -155,7 +157,7 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
     }), [ApartmentUnitTypeValue, CommercialUnitTypeValue, FlatUnitTypeValue, ParkingUnitTypeValue, WarehouseUnitTypeValue])
 
     const meterReadingNormalizer: RowNormalizer = async (row) => {
-        const addons = { address: null, unitType: null, propertyId: null, propertyMap: null, meterId: null, meterResourceId: null, readingSubmissionDate: null, invalidReadingSubmissionDate: null, valuesAmount: 0 }
+        const addons = { address: null, unitType: null, propertyId: null, propertyMap: null, meterId: null, meterResourceId: null, readingSubmissionDate: null, invalidReadingSubmissionDate: null, valuesAmount: 0, value1: undefined, value2: undefined, value3: undefined, value4: undefined }
```
SavelevMatthew, line 160: Why undefined if everything is null here?
And why not move it below ` ] = map(row, 'value')` line, so you can create initial values in place

Alllex202: Fixed

apps/condo/domains/meter/hooks/useImporterFunction.ts
```diff
@@ -237,6 +254,11 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
         if (!get(processedRow, ['addons', 'meterResourceId'])) errors.push(MeterResourceNotFoundMessage)
         if (!get(processedRow, ['addons', 'valuesAmount'])) errors.push(NoValuesErrorMessage)
 
+        if (!validateMeterValue(get(processedRow, ['addons', 'value1']))) errors.push(MeterValueInvalidMessage.replace('{columnName}', Value1ColumnMessage))
```
SavelevMatthew, line 257: You can use parametrized message like this:
image

apps/condo/domains/meter/hooks/useImporterFunction.ts
```diff
@@ -172,7 +173,23 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
...
+            address: null,
+            unitType: null,
+            propertyId: null,
+            propertyMap: null,
+            meterId: null,
+            meterResourceId: null,
+            readingSubmissionDate: null,
+            invalidReadingSubmissionDate: null,
+            valuesAmount: 0,
+            value1: normalizeMeterValue(value1),
```
SavelevMatthew, line 186: Should this helper return null as in columns above?

SavelevMatthew (approved):

apps/condo/domains/meter/hooks/useImporterFunction.ts
Alllex202: Yes, we use both. This one looks more compact.

apps/condo/domains/meter/hooks/useImporterFunction.ts
Alllex202: If `valueX` already exists and it will be return `null`, then `valueX` will be rewrite to `null` from some value. But in this case you should not change `valueX`. 
Therefore, `valueX` with `null` is invalid. And empty line is simply ignored.

apps/condo/domains/meter/hooks/useImporterFunction.ts
SavelevMatthew: But translation template provide more abstract syntax, since you can pass components there, use numbers, dynamic parameters based on locale. In other words Intl is translation tool we use, so it's best practice to pass parameters like this. (If we we switch this tool somehow this will probably can be uncaught), since it's using Intl format template without Intl itself

Alllex202: Ok, refactored it :)

PR #3405, opened 2023-05-30
chore(condo): DOMA-6244 allow admin&support to read scopes

AleX83Xpert:
## Before
image

## After
image

### DIFF
--- a/apps/condo/domains/news/schema/NewsItem.js
+++ b/apps/condo/domains/news/schema/NewsItem.js
@@ -124,7 +124,7 @@ const NewsItem = new GQLListSchema('NewsItem', {
             ref: 'NewsItemScope.newsItem',
             many: true,
             access: {
-                read: () => ({ authentication: { item: user } }) => (user.isAdmin || user.isSupport),
+                read: ({ authentication: { item: user } }) => (user.isAdmin || user.isSupport),
                 create: false,
                 update: false,
             },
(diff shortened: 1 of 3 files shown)


apps/condo/domains/news/schema/NewsItem.js
```diff
@@ -122,7 +123,11 @@ const NewsItem = new GQLListSchema('NewsItem', {
             type: 'Relationship',
             ref: 'NewsItemScope.newsItem',
             many: true,
-            access: { read: () => false, create: false, update: false },
+            access: {
+                read: ({ authentication: { item: user } }) => (user.isAdmin || user.isSupport),
```
SavelevMatthew, line 127: + test?

pahaz (approved):

apps/condo/domains/news/schema/NewsItem.js
AleX83Xpert: Done

packages/keystone/test.utils.js
```diff
@@ -710,6 +710,37 @@ const expectToThrowUniqueConstraintViolationError = async (testFunc, constraintN
...
 }
 
+/**
+ * @param testFunc
+ * @param {string} path
+ * @param {string} field
+ * @param {Number} [count]
+ * @returns {Promise<void>}
+ */
+const expectToThrowAccessDeniedToFieldError = async (testFunc, path, field, count = 1) => {
```
SavelevMatthew, line 720: basic `expect` + `path` won't work here?

SavelevMatthew (approved):

packages/keystone/test.utils.js
AleX83Xpert: I'd like to test the access to a particular field, so it's necessary to add the field name. I have had to add the field as a separate parameter due to the structure within the caught error object.
Need to check `caught -> errors -> n -> path`. The `path` is an array `[path, n, field]`.
At the same time need to check `caught -> data`. The `data` is an object `{path => array}`.
image
The current decision is fit to solve the problem, but of course, it may be changed in the future.

PR #3407, opened 2023-05-30
feat: DOMA-6240 OIDC bridge auth

SavelevMatthew:
### NOTE: Tested with local https + ngrok combo. 

Miniapp codebase is a mess btw, so don't copy from it :)

### DIFF
--- a/apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
+++ b/apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
@@ -183,7 +183,7 @@ export const PostMessageProvider: React.FC = ({ children }) => {
         setRegisteredHandlers((prev) => omit(prev, frameId))
     }, [])
 
-    const handleMessage = useCallback((event: MessageEvent) => {
+    const handleMessage = useCallback(async (event: MessageEvent) => {
         if (!event.isTrusted ||
             !event.source ||
             !('self' in event.source)) {
@@ -237,7 +237,7 @@ export const PostMessageProvider: React.FC = ({ children }) => {
             const validator = validators[method]
             if (validator(params)) {
                 try {
-                    const result = handler(params, origin)
+                    const result = await handler(params, origin)
                     return event.source.postMessage({
                         type: `${method}Result`,
                         data: {
--- a/apps/miniapp/domains/common/utils/oidcAuth.tsx
+++ b/apps/miniapp/domains/common/utils/oidcAuth.tsx
@@ -1,12 +1,18 @@
 import get from 'lodash/get'
 import isObject from 'lodash/isObject'
+import getConfig from 'next/config'
 import Router from 'next/router'
 import React, { useCallback, useContext, useEffect, useState } from 'react'
 
+import bridge from '@open-condo/bridge'
 import { useAuth } from '@open-condo/next/auth'
 import { useIntl } from '@open-condo/next/intl'
 
 import { useLaunchParams } from '@condo/domains/miniapp/hooks/useLaunchParams'
+
+const {
+    publicRuntimeConfig: { serverUrl },
+} = getConfig()
 
 interface IUseOidcAuthHookValue {
     user?: {
@@ -18,7 +24,7 @@ interface IUseOidcAuthHookValue {
     isLoading?: boolean,
 }
 
-const OIDC_AUTH_URL = '/oidc/auth'
+const OIDC_AUTH_URL = `${serverUrl}/oidc/auth`
 
 const OidcAuthContext = React.createContext<IUseOidcAuthHookValue>({})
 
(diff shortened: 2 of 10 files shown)


apps/condo/domains/common/components/PostMessageProvider/Provider.tsx
```diff
@@ -238,7 +238,7 @@ export const PostMessageProvider: React.FC = ({ children }) => {
             const validator = validators[method]
             if (validator(params)) {
                 try {
-                    const result = handler(params, origin)
+                    const result = await handler(params, origin)
```
AleX83Xpert, line 241: is the `handler` is an async function always?

apps/miniapp/domains/common/utils/oidcAuth.tsx
```diff
@@ -40,6 +46,8 @@ const OidcAuthProvider = ({ children }) => {
     const errorReason = get(error, 'errorReason')
     const errorData = isObject(error) ? JSON.stringify(error) : '<NonObject>'
     const { user, isLoading: isUserLoading, isAuthenticated } = useAuth()
+    const [authInProgress, setAuthInProgress] = useState(false)
+    const [, setAuthError] = useState(false)
```
AleX83Xpert, line 50: Is this state necessary?
Every call of `setAuthError` causes re-rendering, but the value from this state is never used.

apps/miniapp/index.js
```diff
@@ -158,7 +158,7 @@ class CondoOIDCMiddleware {
                 delete req.session[oidcSessionKey]
                 await req.session.save()
 
-                return res.redirect('/')
+                return res.json({ user: { id: user.id } })
```
AleX83Xpert, line 161: Callback without any redirect?

apps/miniapp/domains/common/utils/oidcAuth.tsx
```diff
@@ -18,7 +24,7 @@ interface IUseOidcAuthHookValue {
     isLoading?: boolean,
 }
 
-const OIDC_AUTH_URL = '/oidc/auth'
+const OIDC_AUTH_URL = `${serverUrl}/oidc/auth`
```
pahaz, line 27:
```suggestion
const OIDC_AUTH_URL = `${serverUrl || ''}/oidc/auth`
```

apps/miniapp/domains/common/utils/oidcAuth.tsx
SavelevMatthew: Please see the note above. It's just an example. We should redesign miniapp, since right now it's just a legacy mess. I'll do it outside this PR

apps/miniapp/index.js
SavelevMatthew: That's how mobile flow works right now. When you call auth from `cordova` plugin, something similar to webview is opened (but hidden), where v1 cookies are set. After this it follows auth flow (`miniapp/oidc/auth` -> `condo/oidc/auth/` ->  ... -> `miniapp/oidc/callback`) and returns response from latest server to original callback, so you can resolve it on client.

While processing, you can show loader or suggest alternative client route in case of no internet access.

We wanted to merge this flow with web via `bridge`, so you can write you auth uniformly for B2B and B2C apps

AleX83Xpert (approved):

apps/miniapp/domains/common/utils/oidcAuth.tsx
```diff
@@ -64,14 +72,37 @@ const OidcAuthProvider = ({ children }) => {
...
-                    pathname: OIDC_AUTH_URL,
-                    query: { organizationId: context.condoContextEntityId, condoUserId: context.condoUserId },
-                })
+                if (!hasNoRedirect) {
+                    setAuthInProgress(true)
+                    setAuthError(false)
+                    bridge.send('CondoWebAppRequestAuth', { url: OIDC_AUTH_URL }).then((data) => {
+                        if (data.response.status === 200) {
+                            // Process it in any way you like
+                            console.log(data.response)
```
pahaz, line 81: Debug

apps/miniapp/domains/common/utils/oidcAuth.tsx
```diff
@@ -64,14 +72,37 @@ const OidcAuthProvider = ({ children }) => {
...
+                            setAuthError(false)
+                        } else {
+                            setAuthError(true)
+                        }
+                    })
+                        .catch((err) => {
+                            console.debug(err)
```
pahaz, line 88: Debug

apps/miniapp/pages/index.tsx
```diff
@@ -1,89 +1,14 @@
...
-                {TasksUpdateMsg}
-            </Button>
-        </>
-    )
-}
-
-const IndexPage = () => {
+const IndexPage: React.FC = () => {
     const intl = useIntl()
-    const auth = useOidcAuth()
```
pahaz, line 81: Why you want to drop all oidc miniapp code?

SavelevMatthew: I'm not dropping it, just not using user info at this page. Btw, I would kill all miniapp code, since it's a legacy mess right now :)

Alllex202 (approved):

PR #3409, opened 2023-05-30
feat(condo): DOMA-6254 restrict unitName without unitType

AleX83Xpert:
Also, `queryFindResidentsByNewsItemAndScopes` renamed to `queryFindResidentsByOrganizationAndScopes`


AntonAL (commented):
Could you add following comment to `unitName` field of `NewsItemScope`:

"Because in case when a Property will have several units of different types, NewsItem can go to wrong recipients. In practice, when an organization needs to address specific residents it is usually assumed, that they are from units of some specific type, for example, a flat. Without this restriction, a NewsItem, targeted only to unitName 1 can potentially be received by residents of flat 1, parking 1 etc."

PR #3411, opened 2023-05-30
chore(condo) DOMA-6257 add side menu id

toplenboren:
image

### DIFF
--- a/apps/condo/domains/common/components/MenuItem.tsx
+++ b/apps/condo/domains/common/components/MenuItem.tsx
@@ -158,7 +158,7 @@ export const MenuItem: React.FC<IMenuItemProps> = (props) => {
         )
 
     const menuItem = (
-        <MenuItemWrapper onClick={handleClick} className={menuItemClassNames} isCollapsed={isCollapsed} {...restWrapperProps}>
+        <MenuItemWrapper id={`index-sidemenu-${path.replace('/', '')}`} onClick={handleClick} className={menuItemClassNames} isCollapsed={isCollapsed} {...restWrapperProps}>
             {(isCollapsed && !disabled) ? addToolTipForCollapsedMenu(linkContent, Message) : linkContent}
         </MenuItemWrapper>
     )
--- a/apps/condo/domains/common/components/MenuItem.tsx
+++ b/apps/condo/domains/common/components/MenuItem.tsx
@@ -70,6 +70,7 @@ const MenuItemWrapper = styled.div<IMenuItemWrapperProps>`
 `
 
 interface IMenuItemProps {
+    id: string
     path?: string
     icon: React.ElementType
     label: string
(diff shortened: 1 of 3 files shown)


apps/condo/domains/common/components/MenuItem.tsx
```diff
@@ -158,7 +158,7 @@ export const MenuItem: React.FC<IMenuItemProps> = (props) => {
         )
 
     const menuItem = (
-        <MenuItemWrapper onClick={handleClick} className={menuItemClassNames} isCollapsed={isCollapsed} {...restWrapperProps}>
+        <MenuItemWrapper id={`index-sidemenu-${path.replace('/', '')}`} onClick={handleClick} className={menuItemClassNames} isCollapsed={isCollapsed} {...restWrapperProps}>
```
SavelevMatthew, line 161: It's better not to generate ids based on path...Since it can be regex according to 128th line.

It's also not side, but top on mobile.

And what if we wanted to migrate route somehow (for example merge payments and billing sections)...

How about just static id prop?

toplenboren: ok

apps/condo/domains/common/components/MenuItem.tsx
```diff
@@ -70,6 +70,7 @@ const MenuItemWrapper = styled.div<IMenuItemWrapperProps>`
 `
 
 interface IMenuItemProps {
+    id: string
```
SavelevMatthew, line 73: Should it be optional prop?

SavelevMatthew (approved):

apps/condo/pages/_app.tsx
```diff
@@ -199,6 +212,7 @@ const MenuItems: React.FC = () => {
             <div>
                 {menuItemsData.map((menuItemData) => (
                     <MenuItem
+                        id={menuItemData.id}
                         key={`menu-item-${menuItemData.path}`}
```
Alllex202, line 216: now you can use `menuItemData.id`

Alllex202 (approved):

apps/condo/pages/_app.tsx
```diff
@@ -157,21 +165,25 @@ const MenuItems: React.FC = () => {
...
             label: 'global.section.miniapps',
             access: () => !isAssignedVisibilityType,
         }, {
+            id: 'menuitem-service-provider-profile',
             path: 'service-provider-profile',
             icon: Sber,
             label: 'global.section.SPP',
             access: () => sppBillingId && isSPPOrg,
         }, {
+            id: 'miniapp-settings',
```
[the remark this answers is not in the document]
toplenboren: Yep)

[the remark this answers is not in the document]
toplenboren: menu item :)

PR #3412, opened 2023-05-31
fix(condo): DOMA-6252 fixed blinking for organization select and orga…

Alllex202:
…nizations list in user page. Added sorting for organization list

* When use `useAllObjects` then should use `allDataLoaded` instead of `loading`, because `loading` updated on every request, but `allDataLoaded` updated only after loaded all item. 

* Also now organization list sorting by name

Снимок экрана 2023-05-31 в 13 18 02
Снимок экрана 2023-05-31 в 13 27 21

### DIFF
--- a/apps/condo/domains/user/components/UserOrganizationsList.tsx
+++ b/apps/condo/domains/user/components/UserOrganizationsList.tsx
@@ -1,7 +1,7 @@
 import { OrganizationEmployee as OrganizationEmployeeType } from '@app/condo/schema'
 import { Col, Row, Skeleton, Tag, Typography } from 'antd'
 import get from 'lodash/get'
-import React, { useCallback } from 'react'
+import React, { useCallback, useMemo } from 'react'
 
 import { useIntl } from '@open-condo/next/intl'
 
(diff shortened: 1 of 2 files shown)


apps/condo/domains/user/components/UserOrganizationsList.tsx
```diff
@@ -106,23 +106,29 @@ const OrganizationEmployeeItem: React.FC<IOrganizationEmployeeItem> = (props) =>
...
-    const list = userOrganizations.map((employee, index) => (
-        <OrganizationEmployeeItem
-            employee={employee}
-            key={index}
-            userOrganization={userOrganization}
-        />
-    ))
+    const list = useMemo(() => {
+        return userOrganizations
+            .sort((optionA, optionB) =>
```
SavelevMatthew, line 116: Can we just move sorting to query / db?

Alllex202: No, employees have id in the organization field, but need name

nomerdvadcatpyat (approved):

PR #3401, opened 2023-05-29
Feature/condo/DOMA-6235/add contact from resident for org mutation

Kanol:

PR #3415, opened 2023-05-31
Feat/condo/doma 6251/add delete meters mutation

abshnko:
added mutation to delete meters with specified property id(s)


apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
```diff
@@ -0,0 +1,72 @@
...
+                    console.info('[INFO] Could not found meters by specified property ids')
+                    return
+                }
+
+                console.info(`[INFO] Following meters will be deleted: [${meters.map(reading => `'${reading.id}'`).join(', ')}]`)
+
+                const deletedMeters = []
+                for (const meter of meters) {
+                    console.info(`Deleting Meter (id = "${meter.id}")`)
+                    const deletedMeter = await Meter.softDelete(context, meter.id, { dv, sender })
```
SavelevMatthew, line 56: how about bulk soft-delete here?

SavelevMatthew (changes requested):
What's the purpose of this mutation? Why current scripts does not solve problem? If it's appearing multiple time, them some feature probably is not working

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
abshnko: added bulk update

apps/condo/domains/meter/access/_internalDeleteMeterAndMeterReadingsService.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Generated by `createservice meter._internalDeleteMeterAndMeterReadingsService --type mutations`
+ */
+const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
+
+async function can_internalDeleteMeterAndMeterReadings ({ authentication: { item: user } }) {
```
dkoviazin, line 6: we do not use underscore in naming

apps/condo/domains/meter/gql.js
```diff
@@ -34,6 +34,14 @@ const METER_READING_FILTERS_FIELDS = '{ organization address accountNumber place
 const METER_READING_FILTER_TEMPLATE_FIELDS = `{ name employee { id } fields ${METER_READING_FILTERS_FIELDS} ${COMMON_FIELDS} }`
 const MeterReadingFilterTemplate = generateGqlQueries('MeterReadingFilterTemplate', METER_READING_FILTER_TEMPLATE_FIELDS)
 
+// TODO(codegen): write return type result!
+
+const DELETE_METER_AND_METER_READINGS_MUTATION = gql`
+    mutation _internalDeleteMeterAndMeterReadings ($data: _internalDeleteMeterAndMeterReadingsInput!) {
+        result: _internalDeleteMeterAndMeterReadings(data: $data) ${ METER_READING_FIELDS }
```
dkoviazin, line 41: no need to use _internal and mutation name needs to start with a verb
deleteMeterAndMeterReadings 
maybe, deleteMetersAndMeterReadingsForProperties

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+const _internalDeleteMeterAndMeterReadingsService = new GQLCustomSchema('_internalDeleteMeterAndMeterReadingsService', {
+    schemaDoc: 'Mutation to delete meters and meter readings for property',
+    types: [
+        {
+            access: true,
+            // TODO(codegen): write DeleteMeterAndMeterReadingsService input !
+            type: 'input _internalDeleteMeterAndMeterReadingsInput { dv: Int!, sender: JSON!, propertyIds: [String]! }',
+        },
```
dkoviazin, line 25: field sender has it's own type

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
```diff
@@ -0,0 +1,72 @@
...
+    
+    mutations: [
+        {
+            access: access.can_internalDeleteMeterAndMeterReadings,
+            schema: '_internalDeleteMeterAndMeterReadings(data: _internalDeleteMeterAndMeterReadingsInput!): [MeterReading]',
+            resolver: async (parent, args, context) => {
+                const { data } = args
+                const { dv, sender, propertyIds } = data
+
+                const findMeters = async () => {
```
dkoviazin, line 36: you do not use findMeters function more then once 
better to use direct find here 
```
const meters = await find('Meter', {
                        deletedAt: null,
                        property: {
                            id_in: propertyIds,
                        },
                    })
```

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
```diff
@@ -0,0 +1,72 @@
...
+                    })
+                }
+
+                const meters = await findMeters()
+                if (isEmpty(meters)) {
+                    console.info('[INFO] Could not found meters by specified property ids')
+                    return
+                }
+
+                console.info(`[INFO] Following meters will be deleted: [${meters.map(reading => `'${reading.id}'`).join(', ')}]`)
```
dkoviazin, line 51: debug
if you want to add some logs here - better to use appLogger

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
```diff
@@ -0,0 +1,72 @@
...
+
+                const deletedMeters = await Meter.updateMany(context, meters, { dv, sender, deletedAt: 'true' })
+                let foundedMeters = []
+
+                for (const meter of deletedMeters) {
```
dkoviazin, line 56: usually delete mutation do not returns deleted objects
maybe, better to change return type to something, like
{ deleted: 20 } or {status:'ok'}

apps/condo/domains/meter/access/_internalDeleteMeterAndMeterReadingsService.js
abshnko: this is only for internal mutations (_internal)

apps/condo/domains/meter/gql.js
abshnko: this is to differentiate that this mutation is only used internally (another example is mutation for synchronizing contacts -- https://github.com/open-condo-software/condo/pull/3401)

apps/condo/schema.graphql
```diff
@@ -59394,6 +59394,12 @@ type ExportMeterReadingsOutput {
   linkToFile: String!
 }
 
+input _internalDeleteMeterAndMeterReadingsInput {
+  dv: Int!
+  sender: JSON!
+  propertyIds: [String]!
+}
```
dkoviazin, line 59401: you need to use maketypes to change it to SenderField, (but in our schema.graphql we have lots of sender with type JSON and some sender with type SenderField)

dkoviazin (approved):

apps/condo/domains/meter/access/_internalDeleteMeterAndMeterReadingsService.js
SavelevMatthew: `can_internalSomething` seems strange. Even if mutation name contains underscore, not sure that access should contain it like this :)

apps/condo/domains/meter/gql.js
```diff
@@ -34,6 +34,14 @@ const METER_READING_FILTERS_FIELDS = '{ organization address accountNumber place
 const METER_READING_FILTER_TEMPLATE_FIELDS = `{ name employee { id } fields ${METER_READING_FILTERS_FIELDS} ${COMMON_FIELDS} }`
 const MeterReadingFilterTemplate = generateGqlQueries('MeterReadingFilterTemplate', METER_READING_FILTER_TEMPLATE_FIELDS)
 
+// TODO(codegen): write return type result!
```
SavelevMatthew, line 37: ?

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
```diff
@@ -0,0 +1,66 @@
...
+    ],
+    
+    mutations: [
+        {
+            access: access.can_internalDeleteMeterAndMeterReadings,
+            schema: '_internalDeleteMeterAndMeterReadings(data: _internalDeleteMeterAndMeterReadingsInput!): _internalDeleteMeterAndMeterReadingsOutput',
+            resolver: async (parent, args, context) => {
+                const { data } = args
+                const { dv, sender, propertyIds } = data
+                const logger = getLogger('DeleteMeters')
```
SavelevMatthew, line 38: Why it has different name here? It's not obvious to find messages by `DeleteMeters` keyword instead of `_internalDeleteMeterAndMeterReadings`. It may be fine if I can find mutation name directly in it.

SavelevMatthew: And why logger is created on every single call?

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
```diff
@@ -0,0 +1,66 @@
...
+                    },
+                })
+
+                if (isEmpty(meters)) {
+                    logger.warn({ msg: 'Could not find meters by specified property ids', data: { propertyIds } })
+                    return
+                }
+                logger.info({ msg: `Following meters will be deleted: [${meters.map(reading => `'${reading.id}'`).join(', ')}]` })
+
+                const deletedMeters = await Meter.updateMany(context, meters, { dv, sender, deletedAt: 'true' })
```
SavelevMatthew, line 53: Does it work fine with 100/1000 object (or should it contain batching?)

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
```diff
@@ -0,0 +1,66 @@
...
+                let deletedSuccessfully = meters.length === deletedMeters.length ? 'success' : 'error'
```
SavelevMatthew, line 54: `const`?

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
```diff
@@ -0,0 +1,66 @@
...
+const _internalDeleteMeterAndMeterReadingsService = new GQLCustomSchema('_internalDeleteMeterAndMeterReadingsService', {
+    schemaDoc: 'Mutation to delete meters and meter readings for property',
+    types: [
+        {
+            access: true,
+            type: 'input _internalDeleteMeterAndMeterReadingsInput { dv: Int!, sender: SenderFieldInput!, propertyIds: [String]! }',
+        },
+        {
+            access: true,
+            type: 'type _internalDeleteMeterAndMeterReadingsOutput { status: String! }',
```
SavelevMatthew, line 27: Specify `enum` if you have finite number of statuses

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
```diff
@@ -0,0 +1,66 @@
...
+
+                if (isEmpty(meters)) {
+                    logger.warn({ msg: 'Could not find meters by specified property ids', data: { propertyIds } })
+                    return
+                }
+                logger.info({ msg: `Following meters will be deleted: [${meters.map(reading => `'${reading.id}'`).join(', ')}]` })
+
+                const deletedMeters = await Meter.updateMany(context, meters, { dv, sender, deletedAt: 'true' })
+                let deletedSuccessfully = meters.length === deletedMeters.length ? 'success' : 'error'
+                logger.info({ msg: 'Deleted all Meter records with associated MeterReading' })
```
SavelevMatthew, line 55: associated MeterReading?

packages/codegen/generate.server.utils.js
```diff
@@ -233,6 +233,20 @@ function generateServerUtils (gql) {
...
     }
 
+    async function updateMany (context, data, extraAttrs = {}, options = {}) {
+        if (!context) throw new Error('no context')
+        if (!data) throw new Error('no data')
+        _checkOptions(options)
+
+        return await execGqlWithoutAccess(context, {
+            query: gql.UPDATE_OBJS_MUTATION,
+            variables: { data: data.map( data => ({ id: data.id, data: { ...data.data, ...extraAttrs } })) },
```
SavelevMatthew, line 243: `data.map(data => )` Seems like name duplication, how about to make it more clear to read?

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
abshnko: added batching

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
abshnko: added deleting associated meter reading with deleted meters

packages/codegen/generate.server.utils.js
```diff
@@ -233,6 +234,27 @@ function generateServerUtils (gql) {
...
 
+    async function updateMany (context, data, extraAttrs = {}, options = {}) {
+        if (!context) throw new Error('no context')
+        if (!data) throw new Error('no data')
+        _checkOptions(options)
+
+        const chunks = splitArray(data, MAX_CHUNK_SIZE)
+        let result = []
+
+        for (const chunk of chunks) {
```
SavelevMatthew, line 245: Are you sure, that chunks should go here and not to mutations itself?
And why number 50 is chosen? Did you test limits of it?

My comment was about the fact that we hardly ever used mass mutations before. That's why I asked, have you tested the limits? Can updateMany handle 100/1000/10000 objects per request?

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
SavelevMatthew: Where's enum?)

apps/condo/domains/meter/access/_internalDeleteMeterAndMeterReadingsService.js
SavelevMatthew: Just because someone else did it doesn't mean it's right. The underline in the middle of the camelCase name is odd.

toplenboren:
> What's the purpose of this mutation? Why current scripts does not solve problem? If it's appearing multiple time, them some feature probably is not working

It is bad that we have some root level unmanaged scripts that are run in production environment. In future we must remove them completely.

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
abshnko: removed it -- it deletes automatically when associated meter is deleted (`on_delete: 'models.CASCADE'`)

packages/codegen/generate.server.utils.js
abshnko: Changed max chunk size to 100 -- tested how many update operations updateMany can handle and it depends on the number of fields with type Relationship (for 5 fields it can do 200 operations, for 13 - 168 operations). So decided to make it smaller just in case.

apps/condo/domains/meter/schema/_internalDeleteMeterAndMeterReadingsService.js
```diff
@@ -0,0 +1,74 @@
...
+            access: true,
+            type: 'input _internalDeleteMeterAndMeterReadingsInput { dv: Int!, sender: SenderFieldInput!, propertyIds: [String]! }',
+        },
+        {
+            access: true,
+            type: 'enum Status { success, error }',
+        },
+        {
+            access: true,
+            type: 'type _internalDeleteMeterAndMeterReadingsOutput { status: Status }',
```
SavelevMatthew, line 34: Why it's optional?
```suggestion
            type: 'type _internalDeleteMeterAndMeterReadingsOutput { status: Status! }',
```

SavelevMatthew (approved):

SavelevMatthew (commented):
Don't forget to move chunk from hook to mutation :)

PR #3438, opened 2023-06-07
fix(condo): DOMA-6345 AutoIncrementInteger field was improved

AleX83Xpert:
There is an ability to calculate the next value within dynamic scopes. For example: to create number within organization use the following parameters:
```
fields: {
    ...
    number: {
        schemaDoc: 'The news item number',
        type: AutoIncrementInteger,
        isRequired: false,
        autoIncrementScopeFields: ['organization'],
    },
    ...
}
```


apps/condo/domains/news/schema/NewsItem.test.js
```diff
@@ -137,6 +138,21 @@ describe('NewsItems', () => {
                     await createTestNewsItem(anonymousClient, dummyO10n)
                 })
             })
+
+            test('The news item number autoincrement works fine', async () => {
+                const [o10n1] = await createTestOrganization(adminClient)
```
SavelevMatthew, line 143: o10n?)

packages/keystone/fields/AutoIncrementInteger/Implementation.js
```diff
@@ -1,14 +1,23 @@
...
-    
+
     async resolveInput ({ resolvedData, existingItem }) {
         if (!existingItem && !resolvedData[this.path]) {
-            const next = await this.adapter.nextValue()
+            const scopeFields = get(this, ['config', 'autoIncrementScopeFields'], [])
+            const scopeWhere = scopeFields.reduce((result, field) => {
+                const fieldValue = get(resolvedData, field)
+                if (fieldValue) {
+                    return { ...result, [field]: fieldValue }
```
SavelevMatthew, line 15: Copy inside reduce? How about assigning property to result and just return result itself?

packages/keystone/schema.js
```diff
@@ -142,20 +142,6 @@ async function find (schemaName, condition) {
     return await schemaList._keystone.lists[schemaName].adapter.find(condition)
 }
 
-/**
- * @param {string} schemaName
- * @param {Object} condition
- * @returns {Promise<number>}
- */
-async function count (schemaName, condition) {
```
SavelevMatthew, line 150: why did you remove count?

SavelevMatthew: Seems not relative to this PR

apps/condo/domains/news/schema/NewsItem.test.js
AleX83Xpert: o~~rganizatio~~n :)

packages/keystone/schema.js
AleX83Xpert: I was asked about deleting of this unused function by @pahaz and decided to do it within this PR.

packages/keystone/fields/AutoIncrementInteger/Implementation.js
AleX83Xpert: I'd like the reducer function to be pure, to make the code functional-ish. Yes, there are no disadvantages if I use mutating of the result object here. The `scopeFields` object always will be small, so there are no pros/cons for these two approaches for this task.

pahaz (approved):

PR #3443, opened 2023-06-09
fix(condo): DOMA-6368 fix empty screen on CreateMeterReadingsForm

VKislov:


apps/condo/domains/meter/components/CreateMeterReadingsForm.tsx
```diff
@@ -57,12 +57,11 @@ type MetersTableRecord = {
...
 }
 
-function getTableData (meters: MeterType[] | PropertyMeterType[], meterReadings: MeterReadingType[] | PropertyMeterReadingType[]): MetersTableRecord[] {
+function getTableData (meters: MeterType[] | PropertyMeterType[], meterReadings): MetersTableRecord[] {
     const dataSource: MetersTableRecord[] = []
 
-    // @ts-ignore
-    const lastMeterReadings = uniqWith(meterReadings,
-        (meterReading1, meterReading2) =>
+    const lastMeterReadings = uniqWith(meterReadings.filter((meterReading) => meterReading.meter !== null ),
```
Alllex202, line 63: `meterReadings` can be without `meter`?
`meter` is a required field

VKislov: yes, when the user deletes `meter`, a worker task is launched in the `afterChange` that deletes all its `MeterReadings`. Therefore, there is some delay when `readings` are still available, but the field `meter` already has a deleted `meter` and on the front it is not available on the front

apps/condo/domains/meter/components/CreateMeterReadingsForm.tsx
```diff
@@ -57,12 +57,11 @@ type MetersTableRecord = {
...
+        (meterReading1: any, meterReading2: any) =>
```
AleX83Xpert, line 64: `any`? Maybe `PropertyMeterReadingType` or something that?

apps/condo/domains/meter/hooks/useUpdateMeterModal.tsx
```diff
@@ -31,7 +31,7 @@ export const useUpdateMeterModal = (refetch, meterType: MeterTypes = METER_TYPES
     const DeleteMessage = intl.formatMessage({ id: 'pages.condo.meter.DeleteMeterAndReadings' })
     const DeletedMessage = intl.formatMessage({ id: 'Deleted' }).toUpperCase()
 
-    const [selectedMeter, setSelectedMeter] = useState<MeterType>()
```
AleX83Xpert, line 34: Why type was removed?

PR #3445, opened 2023-06-09
feat(ui): DOMA-6376 add new tabs type

Kekmus:
Screenshot 2023-06-09 at 12 11 14

### DIFF
--- a/packages/ui/src/components/Tabs/style.less
+++ b/packages/ui/src/components/Tabs/style.less
@@ -73,7 +73,7 @@
 
         &.condo-tabs-tab-active {
           background-color: @condo-global-color-gray-3;
-          transition: background-color 0.3s;
+          transition: background-color @condo-global-transition-duration-default ease-out;
         }
 
         & + .condo-tabs-tab {
--- a/packages/ui/src/components/Tabs/style.less
+++ b/packages/ui/src/components/Tabs/style.less
@@ -10,6 +10,9 @@
 @tabs-hover-color: @condo-global-color-black;
 @tabs-horizontal-padding: @condo-global-spacing-4 0 @condo-global-spacing-12;
 @tabs-bar-margin: 0 0 @condo-global-spacing-40 0;
+@tabs-rounded-line-height: @condo-global-typography-h4-line-height;
+@tabs-rounded-height: @tabs-rounded-line-height + (@condo-global-spacing-12 + @condo-global-border-width-default) * 2;
+@tabs-rounded-border-radius: (@tabs-rounded-height / 2);
 
 .condo-tabs {
   &-tab-btn {
@@ -66,7 +69,7 @@
       .condo-tabs-tab {
         padding: @condo-global-spacing-12;
         border: @condo-global-border-width-default solid @condo-global-color-gray-3;
-        border-radius: 1000px;
+        border-radius: @tabs-rounded-border-radius;
 
         &.condo-tabs-tab-active {
           background-color: @condo-global-color-gray-3;
--- a/packages/ui/src/components/Tabs/style.less
+++ b/packages/ui/src/components/Tabs/style.less
@@ -56,4 +56,35 @@
   &-nav-operations {
     display: none !important;
   }
+
+  &-rounded {
+    & > .condo-tabs-nav {
+      .condo-tabs-ink-bar {
+        visibility: hidden;
+      }
+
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/Tabs/style.less
```diff
@@ -56,4 +56,35 @@
   &-nav-operations {
     display: none !important;
   }
+
+  &-rounded {
+    & > .condo-tabs-nav {
+      .condo-tabs-ink-bar {
+        visibility: hidden;
```
SavelevMatthew, line 63: why not just `display: none` to remove it from DOM?

packages/ui/src/components/Tabs/style.less
```diff
@@ -56,4 +56,35 @@
...
+      }
+
+      .condo-tabs-tab {
+        padding: @condo-global-spacing-12;
+        border: @condo-global-border-width-default solid @condo-global-color-gray-3;
+        border-radius: 1000px;
```
SavelevMatthew, line 69: How about just specifying `height / 2` value here (you know height of tab, so you can calculate it), since it's took me a minute, while this magic constant works as expected :)

packages/ui/src/components/Tabs/tabs.tsx
```diff
@@ -16,6 +17,10 @@ export type TabItem = {
     children?: React.ReactNode
 }
 
+type CondoTabsTypeProp = {
+    type?: 'rounded'
```
SavelevMatthew, line 21: Why not just `rounded?: bool`. Type is fine, but where is default type then?

Kekmus: We didnt take type prop from default ant props before
For my task I didnt need it so decided that for now I would create my own type prop. And if we need to add default type in the future it will be easier to match default types with rounded type than to match how rounded property should behave with type prop

sitozzz, line 21: Maybe convert this prop to enum? It would be easier in the future to modify it

Kekmus: The point is that the default types are not enum and in the future if we want to add them all we have to do is the following

```
import { TabsType } from 'antd/lib/tabs'
...
type CondoTabsTypeProp = {
    type?: 'rounded' | TabsType
}
```

I dont really understand why it would be better to convert CondoTabsTypeProp to enum, can you explain a bit more please

SavelevMatthew: Didn't get it. There's 2 options to do it.
1. Modifier. This way you can pass `rounded` modifier to turn behaviour on or not pass. However, you can specify it directly like: `rounded={true/false}`
2. Select. This way you can specify option by `myAtrribute='value'` prop, or not specify to have default value. However you should be able to specify it directly as in modifier above: `myAttrubute='defaultOption'`.

In both cases you use explicit default value if you want to control appearance based on state (programmatically):
```typescript
const tabsType = condition ? 'rounded' : 'default'
``` 
Right now to do it with type approach, I need to write something like:
```typescript
const tabsType = condition ? 'rounded' : undefined
``` 
Which is way worse than having a default constant or using modifier

SavelevMatthew: Another reason in favor of the modifier is the consistency of the API. 

Let's imagine that tomorrow we want a rounded button (or input or something else). Would it be a type? I don't think so. Because any button or input can be rounded.

Then another question follows: Why do we have rounded as a modifier in one place and a type in another? There is an inconsistency in the API, making it more difficult to use the API.

sitozzz (approved):

packages/ui/src/components/Tabs/style.less
```diff
@@ -56,4 +59,31 @@
...
+        border: @condo-global-border-width-default solid @condo-global-color-gray-3;
+        border-radius: @tabs-rounded-border-radius;
+
+        &.condo-tabs-tab-active {
+          background-color: @condo-global-color-gray-3;
+          transition: all 0.3s;
+        }
+
+        & + .condo-tabs-tab {
+          margin: 0 0 0 @condo-global-spacing-8;
```
SavelevMatthew, line 80: Why not just `margin-left`?

SavelevMatthew (approved):

PR #3451, opened 2023-06-14
feat: DOMA-6389 organization types added

SavelevMatthew:
## What's inside?
- Separated all organisations into 3 types: `MANAGING_COMPANY`, `HOLDING` and `SERVICE_PROVIDER`.
- Added optional type prop to `registerNewOrganization` mutation
- Fixed resident connection and property map resolve bugs.
- Partly refactored tests in Organization / Property domains.
- Removed legacy `CheckIfPropertyAddressExistsService`

## About organization types:
- `MANAGING_COMPANY` is default organization type, which live in v1. Nothing new here.
- `HOLDING` is organization type for parent organizations in `OrganizationLink`. No extra logic will be added. Added to restrict organization-contractor flow, which will occur later (need to restrict holdings to invite contractors in parent organization because of architectural complexity)
- `SERVICE_PROVIDER` is type for contractors / settlement centers / resource providers. They will live on v1 as well, but will see slightly different interface.

## About residents.
- I've tried to change as little as possible to make everything work together, so I expect some of domains to not work properly with `SERVICE_PROVIDER `. (They'll be modified later "per domain" as `SERVICE_PROVIDER` part of v1 will grow_
- However it's not the case for resident domain, since we're having a mobile app, which must be working constantly.
- For now residents has no conflicts with orgazanition types, since CRM part of `SERVICE_PROVIDER` is empty.
- The only requirement is that **all residents must be connected / reconnected to `MANAGING_COMPANY` properties only**, since we're focusing on them right now.

## TODO

- [x] Add 3 organization types: `HOLDING`, `MANAGING_COMPANY` and `SERVICE_PROVIDER`
- [x] Migrate existing organizations manually (set `HOLDING` for each org existing with `OrganizationLink`)
- [x] Make sure existing tests are passing
- [x] Cover organization domain with new tests
- [x] Cover resident/property domain with tests
- [x] Make sure that API is backward compatible (`MANAGING_COMPANY` is created by default)
- [x] Make sure residents connects only to properties of `MANAGING_COMPANY`


dkoviazin (approved):

apps/condo/domains/organization/schema/OrganizationLink.js
```diff
@@ -50,6 +51,16 @@ const OrganizationLink = new GQLListSchema('OrganizationLink', {
         delete: false,
         auth: true,
     },
+    hooks: {
+        validateInput: async ({ resolvedData, addValidationError }) => {
+            if (resolvedData['from']) {
+                const org = await getById('Organization', resolvedData['from'])
+                if (!org || org.type !== HOLDING_TYPE) {
+                    addValidationError(`Parent organization ("from") must have "${HOLDING_TYPE}" type`)
```
Alllex202, line 59: you should use `GQLError`

apps/condo/domains/organization/schema/Organization.js
```diff
@@ -36,6 +37,20 @@ const Organization = new GQLListSchema('Organization', {
             isRequired: true,
             kmigratorOptions: { null: false },
         },
+        type: {
+            schemaDoc: 'Type of organization',
```
pahaz, line 41: It's better to write more info about this tricky field )

apps/condo/domains/organization/schema/OrganizationLink.js
pahaz: It's better to use our own GQL messages to help others developers.

apps/condo/domains/organization/schema/Organization.js
```diff
@@ -36,6 +37,20 @@ const Organization = new GQLListSchema('Organization', {
...
+            type: Select,
+            options: ORGANIZATION_TYPES,
+            dataType: 'string',
+            defaultValue: MANAGING_COMPANY_TYPE,
```
sitozzz, line 45: What about all existing organizations? Maybe provide special script that could manage it's type relying on `OrganizationLink` connection?

SavelevMatthew: See migration

apps/condo/domains/organization/schema/OrganizationLink.js
SavelevMatthew: I hate them, but for your sake I will add one...

pahaz (approved):

apps/condo/domains/organization/schema/Organization.js
sitozzz: Yup, that's good

sitozzz (approved):

Alllex202 (approved):

PR #3452, opened 2023-06-14
feat(condo): DOMA-6329 schema meter reporting period

VKislov:


apps/condo/domains/meter/schema/MeterReportingPeriod.js
```diff
@@ -0,0 +1,169 @@
...
+                const foundDefaultPeriod = await MeterReportingPeriodAPI.getAll(context, {
+                    organization_is_null: true,
+                    deletedAt: null,
+                })
+
+                if (!resolvedData.organization && foundDefaultPeriod.length > 0) {
+                    throw new GQLError(ERRORS.ORGANIZATION_IS_REQUIRED, context)
+                }
+            }
+        },
```
nomerdvadcatpyat, line 148: Can this be done with a constraint like with property in 115-122 lines?

apps/condo/domains/meter/schema/MeterReportingPeriod.js
```diff
@@ -0,0 +1,169 @@
...
+            if (operation === 'create') {
+                if (!resolvedData.organization && resolvedData.property) {
+                    const property = await Property.getOne(context, { id: resolvedData.property })
+
+                    resolvedData.organization = property.organization
+                }
+            }
+
+            return resolvedData
+        },
```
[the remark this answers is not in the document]
VKislov: property can be null if this period should be for the entire organization. At the same time, the plugin prohibits setting the organization in access when creating an entity. That is, it can only set the organization of the related entity. Therefore, the plugin will not cover all cases.

apps/condo/domains/meter/schema/MeterReportingPeriod.js
VKislov: I didn't manage to do it using constraint.
You need a record in the database that will be available to all users and use its settings as default. Therefore, a proposal was made to make a record not tied to an `organization`. To do this, I check that there is no such entry, and if there is, I think that the user wanted to create a period for his `organization` and for some reason it is not specified. If you use `constraint`, then you will not be able to create a record that is common to all. I did not find a way that would allow the `constraint` to fire only if there are no records with an empty `organization` field. It seems to me that there should be some kind of `count`, but I did not find it

apps/condo/domains/meter/access/MeterReportingPeriod.js
```diff
@@ -0,0 +1,84 @@
...
+    if (user.isSupport || user.isAdmin) return {}
+
+    if (user.type === RESIDENT) {
+        const availableMeterPeriods = await getAvailableResidentMeterReportPeriods(user.id)
+        const availableMeterPeriodIds = availableMeterPeriods.map(period => period.id)
+
+
+        return {
+            id_in: availableMeterPeriodIds,
+            deletedAt: null,
```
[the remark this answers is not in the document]
VKislov: No. I filter all records on the server. Even if the user specifies `deletedAt`, this filter will be applied on the server and all records with the `deletedAt` field filled in will not be returned to the user. Exactly as in the accesses of the `Meter` entity

apps/condo/domains/meter/schema/MeterReportingPeriod.test.js
```diff
@@ -0,0 +1,269 @@
...
+
+            test('anonymous can\'t', async () => {
+                await expectToThrowAuthenticationErrorToObjects(async () => {
+                    await MeterReportingPeriod.getAll(anonymous, {}, { sortBy: ['updatedAt_DESC'] })
+                })
+            })
+        })
+    })
+
+    describe('Validation tests', () => {
```
Alllex202, line 247: you should add tests for `INVALID_FINISH`, `INVALID_START`, `ORGANIZATION_IS_REQUIRED` errors

apps/condo/domains/meter/utils/serverSchema/index.js
```diff
@@ -67,6 +70,61 @@ const getAvailableResidentMeters = async (userId) => {
...
+        AND: [
+            selection,
+        ],
+    }))
+
+    selectionsByProperty.map(selection => orStatement.push({
+        AND: [
+            selection,
+        ],
+    }))
```
Alllex202, line 115: strange code `arr1.map(item => arr2.push(item))`

`map` is used for transformations like `const newArr = oldArr.map(item => someTransform(item))`

apps/condo/domains/meter/access/MeterReportingPeriod.js
```diff
@@ -0,0 +1,92 @@
...
+                id: propertyId,
+                deletedAt: null,
+            })
+
+            if (!property) return false
+
+            propertyOrganization = get(property, 'organization', null)
+        }
+
+        const organizationId = orgId ?? propertyOrganization
```
nomerdvadcatpyat, line 66: you need to check that `propertyOrganization` is equal `orgId`

nomerdvadcatpyat (approved):

apps/condo/domains/meter/access/MeterReportingPeriod.js
VKislov: `orgId` is what comes in `originalInput` with the `organization` field optional. There may be a case when only `property` are transferred. Therefore, it seems that such a check will fail in this case. Later, in the `resolveInput` hook, the `organization` is set if a `property` is passed. But these hooks are executed after the access check

apps/condo/migrations/20230606110411-0278_auto_20230606_0604.js
```diff
@@ -13,7 +13,8 @@ exports.up = async (knex) => {
             --
             -- Alter field type on meterreadingsource
             --
-            INSERT INTO "MeterReadingSource" (dv, sender, type, name, id, v, "createdAt", "updatedAt", "deletedAt", "newId", "createdBy", "updatedBy") VALUES (1, '{"dv": 1, "fingerprint": "initial"}', 'crm', 'meterReadingSource.crm.name', 'b09fec04-5bc2-4c58-a364-04b1ea505503', 1, '2023-06-05 00:00:00.000000', '2023-06-05 00:00:00.000000', null, null, null, null);
+            INSERT INTO "MeterReadingSource" (dv, sender, type, name, id, v, "createdAt", "updatedAt", "deletedAt", "newId", "createdBy", "updatedBy") VALUES (1, '{"dv": 1, "fingerprint": "initial"}', 'crm', 'meterReadingSource.crm.name', 'b09fec04-5bc2-4c58-a364-04b1ea505503', 1, '2023-06-05 00:00:00.000000', '2023-06-05 00:00:00.000000', null, null, null, null)
+            ON CONFLICT (id) DO NOTHING;;
```
sitozzz, line 17: Why are you modifying old migration at this pr?

apps/condo/domains/meter/utils/serverSchema/index.js
```diff
@@ -67,6 +70,61 @@ const getAvailableResidentMeters = async (userId) => {
...
+        property: { id: get(residentsByIds, [serviceConsumer.resident, 'property']) },
+    })))
+
+    const orStatement = selectionsByOrganization.map(selection => ({
+        AND: [
+            selection,
+        ],
+    }))
+
+    selectionsByProperty.map(selection => orStatement.push({
```
sitozzz, line 111: There are best way to solve this task. `forEach` would be better option If you're want to iterate through array items without mutation

apps/condo/domains/meter/schema/MeterReportingPeriod.test.js
```diff
@@ -0,0 +1,269 @@
...
+        describe('Constraints', () => {
```
sitozzz, line 248: Please add tests for all custom errors that you specify at schema

apps/condo/migrations/20230606110411-0278_auto_20230606_0604.js
VKislov: This is an irreversible manual migration. I was offered to add this so that conflicts would not fall during local development.

apps/condo/domains/meter/utils/serverSchema/index.js
VKislov: `map` doesn't mutate anything

PR #3484, opened 2023-06-21
DOMA-6450 add analytics events to ticket comment form send & file upload

sitozzz:


apps/condo/domains/common/components/Comments/CommentForm.tsx
```diff
@@ -152,7 +159,7 @@ const CommentForm: React.FC<ICommentFormProps> = ({
             <UploadComponent
                 initialFileList={editableCommentFiles}
                 UploadButton={
-                    <Button type='text'>
+                    <Button type='text' eventProperties={{ components: { id: 'comment-file-upload' } }}>
```
SavelevMatthew, line 162: This will not be supported in UI-kit Button

sitozzz: Yup, that's true. But does ui-kit button component support the same visual style as used on the page?

toplenboren (approved):

PR #3530, opened 2023-06-28
feat(condo) DOMA-6330 frontend of meter readings reporting period

VKislov:


apps/condo/domains/meter/schema/MeterReportingPeriod.js
```diff
@@ -66,46 +66,37 @@ const MeterReportingPeriod = new GQLListSchema('MeterReportingPeriod', {
             kmigratorOptions: { null: true, on_delete: 'models.CASCADE' },
         },
 
-        start: {
+        startAt: {
```
AntonAL, line 69: Naming format corresponds to timestamp, rather than integer.
Maybe `startDayOfMonth` and `finishDayMonth`?

apps/condo/domains/meter/schema/MeterReportingPeriod.js
```diff
@@ -66,46 +66,37 @@ const MeterReportingPeriod = new GQLListSchema('MeterReportingPeriod', {
...
+        finishAt: {
             schemaDoc: 'Finish day of the month for sending a push about the need to submit meter readings',
             type: Integer,
             defaultValue: 25,
             isRequired: true,
             hooks: {
                 validateInput: async ({ context, operation, resolvedData }) => {
                     if (operation === 'create' || operation === 'update') {
-                        if (resolvedData.finish > 31 || resolvedData.finish < 1) {
+                        if (resolvedData.finishAt > 31 || resolvedData.finishAt < 1) {
```
AntonAL, line 93: What if value will be negative?

apps/condo/domains/meter/schema/MeterReportingPeriod.js
VKislov: This name has already been approved at the architecture meeting

apps/condo/domains/meter/schema/MeterReportingPeriod.js
VKislov: -1 < 1 there will be an error. In addition, at the frontend, this is implemented by a select with numbers, and not by an input. Therefore, a negative number is only possible with a bug or creating a period in the mutation / admin panel

AntonAL: Could you please add check for negative values. Another weak point is an admin panel.
It's better to include data logic into schema itself, rather than UI.

AntonAL: Approvement diverges from our implicit standard. Up to you to decide what to do.

AntonAL (approved):

apps/condo/domains/meter/schema/MeterReportingPeriod.js
```diff
@@ -66,46 +66,37 @@ const MeterReportingPeriod = new GQLListSchema('MeterReportingPeriod', {
...
+                        if (resolvedData.startAt > 31 || resolvedData.startAt < 1) {
                             throw new GQLError(ERRORS.INVALID_START, context)
                         }
                     }
                 },
             },
         },
 
-        finish: {
+        finishAt: {
```
SavelevMatthew, line 85: We're using `At` postfix to specify, that field data is date time in ISO format. This field breaks this consistency, which is not good for third-party developers.

If you want to specify, that it is day of month, not specific date, include that in fieldName...Something like `finishDay` will be ok

apps/condo/domains/meter/schema/MeterReportingPeriod.js
AntonAL: Approvement diverges from our implicit standard. Up to you to decide what to do.

PR #3546, opened 2023-07-02
Doma-6526 pdf receipt for residents

dkoviazin:


apps/condo/domains/billing/schema/BillingReceiptFile.js
```diff
@@ -2,30 +2,77 @@
...
+            graphQLReturnType: 'File',
+            resolver: async (item, _, { authedItem }) => {
+                // We are changing link to publicData only for not verified residents. In other cases we return sensitive data files
+                let file = item.sensitiveDataFile
+                if (authedItem.type === RESIDENT) {
+                    // Resident already has an access to billing receipt, so we only need to check if he is approved
+                    const { account } = await getById('BillingReceipt', item.receipt)
+                    const isApproved = await isResidentVerified(authedItem, account)
+                    if (!isApproved) {
+                        file = item.publicDataFile
```
AleX83Xpert, line 48: It's better not to use sensitive data as default value
```suggestion
                let file = item.publicDataFile
                if (authedItem.type === RESIDENT) {
                    // Resident already has an access to billing receipt, so we only need to check if he is approved
                    const { account } = await getById('BillingReceipt', item.receipt)
                    const isApproved = await isResidentVerified(authedItem, account)
                    if (isApproved) {
                        file = item.sensitiveDataFile
```

apps/condo/domains/billing/schema/BillingReceiptFile.js
```diff
@@ -68,8 +115,19 @@ const BillingReceiptFile = new GQLListSchema('BillingReceiptFile', {
         auth: true,
     },
     hooks: {
-        // Set meta to file for checking permissions { listkey, id }
-        afterChange: fileMetaAfterChange,
+        afterChange: async ({ updatedItem, listKey }) => {
+            if (updatedItem && Adapter.acl && Adapter.acl.setMeta) {
+                const sensitiveFile = get(updatedItem, 'sensitiveDataFile.filename')
+                const publicFile = get(updatedItem, 'publicDataFile.filename')
```
AleX83Xpert, line 121: Lodash always casts a string path to an array. So we can decrease the number of calculations.
```suggestion
                const sensitiveFile = get(updatedItem, ['sensitiveDataFile', 'filename'])
                const publicFile = get(updatedItem, ['publicDataFile', 'filename'])
```

AleX83Xpert (approved):

apps/condo/domains/billing/schema/BillingReceiptFile.js
dkoviazin: it's hard to use such syntax on constructions like
```
const residentName = get(residents, '[0].name')
```
So, I prefer string variant

apps/condo/domains/billing/schema/BillingReceiptFile.js
dkoviazin: Done

PR #3560, opened 2023-07-04
feat(condo): DOMA-6537 Service provider billing section

SavelevMatthew:
https://github.com/open-condo-software/condo/assets/50069872/30c4eecc-0c6e-48fa-9341-fa10edc80939


apps/condo/domains/acquiring/access/AcquiringIntegrationContext.js
```diff
@@ -77,6 +78,29 @@ async function canManageAcquiringIntegrationContexts ({ authentication: { item:
...
+        const orgId = operation === 'create' ? get(originalInput, ['organization', 'connect', 'id']) : existingItem['organization']
+        if (!orgId) {
+            return false
+        }
+        const org = await getById('Organization', orgId)
+
+        return Boolean(org && org.type !== SERVICE_PROVIDER_TYPE)
+    }
+
+    return true
```
sitozzz, line 101: Is it correct to grant access by default?

apps/condo/domains/billing/components/OnBoarding/index.tsx
```diff
@@ -43,21 +47,36 @@ export const BillingOnboardingPage: React.FC<BillingOnboardingPageProps> = ({ on
     const router = useRouter()
     const [welcomeShown, setWelcomeShown] = useState(false)
 
-    const [currentStep] = useOnboardingProgress()
+    useEffect(() => {
+        setWelcomeShown(Boolean(localStorage.getItem(MODAL_STORAGE_KEY)))
```
sitozzz, line 51: Will it working with SSR?

apps/condo/pages/billing/index.tsx
```diff
@@ -4,14 +4,16 @@ import React, { useCallback } from 'react'
...
 
+import { CONTEXT_FINISHED_STATUS, CONTEXT_VERIFICATION_STATUS } from '@condo/domains/acquiring/constants/context'
 import { AcquiringIntegrationContext as AcquiringContext } from '@condo/domains/acquiring/utils/clientSchema'
 import { BillingPageContent } from '@condo/domains/billing/components/BillingPageContent'
 import { BillingAndAcquiringContext } from '@condo/domains/billing/components/BillingPageContent/ContextProvider'
 import { BillingOnboardingPage } from '@condo/domains/billing/components/OnBoarding'
 import { BillingIntegrationOrganizationContext as BillingContext } from '@condo/domains/billing/utils/clientSchema'
 import LoadingOrErrorPage from '@condo/domains/common/components/containers/LoadingOrErrorPage'
-import { CONTEXT_FINISHED_STATUS } from '@condo/domains/miniapp/constants'
+import { CONTEXT_FINISHED_STATUS as BILLING_FINISHED_STATUS } from '@condo/domains/miniapp/constants'
```
sitozzz, line 14: IMO it will be difficult to determine constant usages at the code below

apps/condo/domains/billing/components/OnBoarding/index.tsx
SavelevMatthew: useEffect is only triggered on a client as far as I remember, so it will.

Here's a reference in our code btw: https://github.com/open-condo-software/condo/blob/master/apps/condo/domains/common/components/LayoutContext.tsx#L44

apps/condo/domains/acquiring/access/AcquiringIntegrationContext.js
SavelevMatthew: Field-level access is triggered after model-level access. So we've already filtered all possible actors:
1. admin/support
2. employee
3. integration

All of them had all access to this field by default, since they're passing model-level access.

After that we want to add few restrictions on top for specific scenarios, but keep previous behaviour overall, that's why `true` is default here.

apps/condo/pages/billing/index.tsx
SavelevMatthew: This is due to the fact that prior to this, all integration lived as miniaps and had the same types of statuses. Now billing and acquiring have separated and acquiring required an additional status. So I moved them to separate constants. I did not move billing, because in general it still works in a similar way to mini-apps. 

If you know a more elegant solution to this problem, I'm willing to add it.

(I also didn't want to do something like `[...MINIAPP_STATUSES, NEW_STATUS]` because of some extra logic on acquiring statuses)

apps/condo/domains/acquiring/access/AcquiringIntegrationContext.js
sitozzz: Oh, okay. I didn't catch that it is a field access resolver

apps/condo/domains/billing/components/OnBoarding/index.tsx
sitozzz: First call of useEffect will proceed at server render. There is reference to similar [problem](https://stackoverflow.com/questions/73853069/solve-referenceerror-localstorage-is-not-defined-in-next-js)

apps/condo/pages/billing/index.tsx
sitozzz: It seems worth considering moving these constants to the current domain. Anyway it's an optional, so you can leave it as is

apps/condo/domains/billing/components/OnBoarding/index.tsx
```diff
@@ -43,21 +47,36 @@ export const BillingOnboardingPage: React.FC<BillingOnboardingPageProps> = ({ on
...
+
+    const [currentStep] = useOnboardingProgress(withVerification)
 
     const stepItems: Array<StepItem> = useMemo(() => {
-        return [
+        const steps: Array<StepItem> = [
             { title: BillingSelectTitle },
             { title: SetupBillingTitle, breakPoint: true },
-            { title: SetupAcquiringTitle },
+            { title: SetupAcquiringTitle, breakPoint: true },
```
sitozzz, line 60: Seems like `breakPoint` its an redundant property. It's last element of the array, so you have no options to go further. What is the meaning of its definition?

sitozzz (approved):

apps/condo/domains/billing/components/OnBoarding/index.tsx
SavelevMatthew: There can be a verification step after, so it's not redundant.

AleX83Xpert (approved):

apps/condo/domains/acquiring/access/AcquiringIntegrationContext.js
```diff
@@ -77,6 +78,29 @@ async function canManageAcquiringIntegrationContexts ({ authentication: { item:
...
+    // Admin / support can freely change status field
+    if (user.isAdmin || user.isSupport) return true
+
+    // Nobody else can update status field if it's on verification currently
+    if (existingItem && existingItem.status === CONTEXT_VERIFICATION_STATUS) {
+        return false
+    }
+
+    // Service provider org employee should not be able to finish connecting by itself
+    if ('status' in originalInput && originalInput.status === 'Finished') {
```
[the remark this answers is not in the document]
SavelevMatthew: left in so explicitly specify that I expect property to be in input, changed hard-coded string to const

PR #3568, opened 2023-07-04
feat(condo): DOMA-6449 fixed display RecipientCounter for many newsIt…

Kekmus:
…emScopes

### DIFF
--- a/apps/condo/pages/news/[id]/index.tsx
+++ b/apps/condo/pages/news/[id]/index.tsx
@@ -106,7 +106,7 @@ const NewsItemCard: React.FC = () => {
         objs: newsItemScopes,
         loading: newsItemScopesLoading,
         error: newsItemScopesError,
-    } = NewsItemScope.useObjects({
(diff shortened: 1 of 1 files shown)


apps/condo/pages/news/[id]/index.tsx
```diff
@@ -106,7 +106,7 @@ const NewsItemCard: React.FC = () => {
         objs: newsItemScopes,
         loading: newsItemScopesLoading,
         error: newsItemScopesError,
-    } = NewsItemScope.useObjects({
+    } = NewsItemScope.useAllObjects({
```
Comment (pahaz) [inline apps/condo/pages/news/[id]/index.tsx:109]: need to test it for big buildings

Comment (Kekmus) [reply apps/condo/pages/news/[id]/index.tsx:109]: what do you mean why would that be a problem?

Comment (AleX83Xpert) [reply apps/condo/pages/news/[id]/index.tsx:109]: What about scopes created for two sections, each with 120 flats?

AleX83Xpert (approved):

PR #3574, opened 2023-07-05
feat(condo): DOMA-6172 added 2 billing categories 

dkoviazin:
and extend model to contain keywords
image

image


apps/condo/domains/billing/schema/BillingCategory.js
```diff
@@ -1,24 +1,39 @@
...
             schemaDoc: 'Localized name of billing category: Hot water, Cold water, Housing Services',
             type: LocalizedText,
             isRequired: true,
             template: 'billing.category.*.name',
         },
+        serviceNames: {
+            schemaDoc: 'Possible names of services to auto-detect receipt category',
+            type: Json,
+            isRequired: false,
+            hooks: {
```
sitozzz, line 28: Is it possible to provide gql return type?

sitozzz (approved):

AleX83Xpert (approved):

apps/condo/domains/billing/schema/BillingCategory.js
dkoviazin: This field will be used only in unified mutation. It only contains an array of strings and I didn't include this field to gql queries

PR #3575, opened 2023-07-05
feat(condo): DOMA-6556 call discoverServiceConsumers when creating Re…

abshnko:
…sident and BillingAccount


apps/condo/domains/billing/schema/BillingAccount.js
```diff
@@ -101,6 +103,16 @@ const BillingAccount = new GQLListSchema('BillingAccount', {
                 return addValidationError(`${UNEQUAL_CONTEXT_ERROR}:property:context] Context is not equal to property.context`)
             }
         },
+        afterChange: async ({ context, updatedItem, operation }) => {
+            if (operation === 'create'){
```
SavelevMatthew, line 107: Why there's no update? According to accesses we can freely change billing accounts.

SavelevMatthew: Should old service consumers be deleted if unitName changes for example, or should hook be called and so on?

SavelevMatthew: Also what should be done if BillingPorperty address changed?

SavelevMatthew:
How should we migrate existing data?

apps/condo/domains/resident/gql.js
```diff
@@ -17,7 +17,7 @@ const RESIDENT_ORGANIZATION_FIELDS = 'id name country tin'
 const RESIDENT_PROPERTY_FIELDS = 'id name address'
 const ORGANIZATION_FEATURES_FIELDS = 'hasBillingData hasMeters'
 const PAYMENT_CATEGORIES_FIELDS = 'id categoryName billingName acquiringName'
-const RESIDENT_FIELDS = `{ user { id name locale } organization { id name tin country } residentOrganization { ${RESIDENT_ORGANIZATION_FIELDS} } property { id createdAt deletedAt } residentProperty { ${RESIDENT_PROPERTY_FIELDS} } address addressMeta { ${ADDRESS_META_SUBFIELDS_QUERY_LIST} } unitName unitType ${COMMON_FIELDS} organizationFeatures { ${ORGANIZATION_FEATURES_FIELDS} } paymentCategories { ${PAYMENT_CATEGORIES_FIELDS} } }`
+const RESIDENT_FIELDS = `{ user { id name locale phone } organization { id name tin country } residentOrganization { ${RESIDENT_ORGANIZATION_FIELDS} } property { id createdAt deletedAt } residentProperty { ${RESIDENT_PROPERTY_FIELDS} } address addressMeta { ${ADDRESS_META_SUBFIELDS_QUERY_LIST} } unitName unitType ${COMMON_FIELDS} organizationFeatures { ${ORGANIZATION_FEATURES_FIELDS} } paymentCategories { ${PAYMENT_CATEGORIES_FIELDS} } }`
```
dkoviazin, line 20: phone is a private data - better to use getById where you need phone and do not add this field to common queries

apps/condo/domains/billing/schema/BillingAccount.js
dkoviazin: There will be no cases when address for billingProperty will change
And discoverService consumer will not watch for serviceConsumers after it creates them

dkoviazin: we want discoverServiceConsumer to work only on new residents and new billing accounts
it's not an universal tool for managing service consumers for resident 
it's only a helper to get receipts for the first time

SavelevMatthew: Should we block update access on billing entities then?

SavelevMatthew (commented):
You should add more `real life` cases like:
```
1. Organization uploaded receipts
2. Resident registered
3. Resident can see receipts
```

```
1. Redisent registered
2. Receipts uploaded for a first time
3. Resident can see receipts
```

AleX83Xpert:
> How should we migrate existing data?

There will be a script for starting this mutation with different arguments.

PR #3581, opened 2023-07-06
feat(ui): DOMA-6580 added radio button only with icon

Alllex202:
Снимок экрана 2023-07-06 в 14 21 00

### DIFF
--- a/packages/ui/src/components/Radio/style.less
+++ b/packages/ui/src/components/Radio/style.less
@@ -87,8 +87,11 @@
 
 .condo-radio-icon {
   display: inline-flex;
-  margin-right: @condo-global-spacing-8;
   color: @condo-global-color-black;
+
+  & ~ .condo-typography {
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/Radio/style.less
```diff
@@ -87,8 +87,11 @@
 
 .condo-radio-icon {
   display: inline-flex;
-  margin-right: @condo-global-spacing-8;
   color: @condo-global-color-black;
+
+  & ~ .condo-typography {
+    margin-left: @condo-global-spacing-8 !important;
```
SavelevMatthew, line 93: why important here? Something from antd is blocking?

Alllex202: One of the rules overrides this and set margin to 0

SavelevMatthew (approved):

PR #3606, opened 2023-07-11
feat(condo): DOMA-6599 added type to news templates

Kekmus:


apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -375,6 +375,29 @@ export const BaseNewsForm: React.FC<BaseNewsFormProps> = ({
...
+    const commonTemplates = useMemo(() => {
+        const commonTemplates = {}
+
+        for (const id in templates) {
+            if (templates[id].type === NEWS_TYPE_COMMON || isNull(templates[id].type)) {
+                commonTemplates[id] = templates[id]
+            }
+        }
+
+        return commonTemplates
```
AleX83Xpert, line 387: Why not just filter the `templates` array?
```
const commonTemplates = useMemo(() => templates.filter(({type})=>type === NEWS_TYPE_COMMON), [...])
```

Why not fill both arrays in the same loop?

Or you may filter the templates list within the news type change handler

Kekmus: because its not array but an object of objects, I can change it to array but dont know why that would be better

its indeed possible to filter everything in one loop but I suppose that two loops for each type would be more readable

I also thought it wouldnt make sense to filter every time you change type since its potentially extra operation

AleX83Xpert (approved):

apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -375,6 +375,29 @@ export const BaseNewsForm: React.FC<BaseNewsFormProps> = ({
...
+    }, [templates])
+
+    const emergencyTemplates = useMemo(() => {
+        const commonTemplates = {}
+
+        for (const id in templates) {
+            if (templates[id].type === NEWS_TYPE_EMERGENCY || isNull(templates[id].type)) {
```
AntonAL, line 394: Is it ok, that templates without `type` will be presented in `commonTemplates` and in `emergencyTemplates`?

apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -375,6 +375,29 @@ export const BaseNewsForm: React.FC<BaseNewsFormProps> = ({
...
+        const commonTemplates = {}
+
+        for (const id in templates) {
+            if (templates[id].type === NEWS_TYPE_COMMON || isNull(templates[id].type)) {
+                commonTemplates[id] = templates[id]
+            }
+        }
+
+        return commonTemplates
+    }, [templates])
```
AntonAL, line 388: Could you avoid using local `commonTemplates` that matches the same named constant in upper level scope.

For example:

```suggestion
    const commonTemplates = useMemo(() => {
        _.transform(templates, (result, value, key) => {
	    if (value.type === NEWS_TYPE_COMMON || isNull(value.type)) {
		result[key] = value
	    }
        }, {})
    }, [templates])
```

PR #3613, opened 2023-07-12
Add an ability to get keystone schema with metadata in runtime

toplenboren:
We add a function `getSchema()` which outputs Map() with all available Lists and Custom schema.

--- 

Screenshot 2023-07-12 at 22 35 13

### DIFF
--- a/packages/keystone/plugins/utils/typing.js
+++ b/packages/keystone/plugins/utils/typing.js
@@ -14,6 +14,7 @@ const TEST_EXAMPLE = {
 }
 
 function assertPluginCall (fn) {
+    // NOTE: we want to check that the plugin is a pure function! It should not save or use any global state!
     fn(TEST_EXAMPLE, { schemaName: 'Test', addSchema: () => undefined })
 }
 
--- a/packages/keystone/schema.js
+++ b/packages/keystone/schema.js
@@ -172,6 +172,20 @@ async function getSchemaCtx (schemaObjOrName) {
     }
 }
 
+function getSchemaDependenciesGraph (schemaName) {
+    if (!SCHEMAS.has(schemaName)) throw new Error(`Schema ${schemaName} is not registered yet`)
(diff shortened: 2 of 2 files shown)


packages/keystone/schema.js
```diff
@@ -167,11 +168,25 @@ async function getSchemaCtx (schemaObjOrName) {
...
         name: schema.name,
+        list: get(schema, ['_keystone', 'lists', schema.name], null),
         keystone: schema._keystone,
     }
 }
 
+/**
+ * Outputs gql schema
+ */
+function getSchema () {
```
Alllex202, line 181:
```suggestion
function getSchemas () {
```

Alllex202 (approved):

packages/keystone/schema.js
AntonAL, line 181: According to function logic, it returns map of contexts
```suggestion
function getSchemaContexts () {
```

AntonAL (approved):

packages/keystone/plugins/utils/typing.js
```diff
@@ -14,6 +14,7 @@ const TEST_EXAMPLE = {
 }
 
 function assertPluginCall (fn) {
+    // NOTE: we want to check that the plugin is a pure function! It should not save or use any global state!
     fn(TEST_EXAMPLE, { schemaName: 'Test', addSchema: () => undefined })
```
AntonAL, line 18: Could you please explain in comments what is a TEST_EXAMPLE
I quess, this is a stub of an input data for a plugin, accepting that the function should not fail.

AntonAL, line 18: What about handling failed attempt of an assertion?
For example, throw error or input message to console.

packages/keystone/schema.js
toplenboren: Renamed to getSchemaContexts as @AntonAL suggested

PR #3615, opened 2023-07-13
Fix/condo/doma 6610/fixed all selections with only one property

Kekmus:


apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -785,6 +808,10 @@ export const BaseNewsForm: React.FC<BaseNewsFormProps> = ({
                     formValuesToMutationDataPreprocessor={(values) => {
                         values.property = selectedProperties[0]
 
+                        if (get(values, 'hasAllProperties', null) && get(values, 'property', null)) {
```
VKislov, line 811: if you do not specify the third parameter in get, then it will return undefined, I think the behavior in this case will be the same as with returning null, but shorter

apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -107,10 +108,19 @@ export const StyledTabs = styled(Tabs)`
...
     }
 `
-const COUNTER_COL_STYLE: React.CSSProperties = {
+const COUNTER_TITLE_STYLE: React.CSSProperties = {
+    top: '123px',
+    position: 'absolute',
+    right: 0,
+    margin: '12px',
+    padding: '2px 10px',
+    backgroundColor: 'black',
```
VKislov, line 117: it's better to import a collection of colors. It seems you can take here import { colors } from '@open-condo/ui/dist/colors'

VKislov (approved):

Alllex202 (approved):

apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -489,7 +522,17 @@ export const BaseNewsForm: React.FC<BaseNewsFormProps> = ({
     }, [Body, Title, templates])
     const propertyCheckboxChange = (form) => {
         return (value) => {
-            if (value) setSelectedPropertiesId([])
+            if (value) setSelectedPropertiesId(selectedPropertiesId => {
+                if (countPropertiesAvaliableToSelect.current === 1 && selectedPropertiesId.length === 1) 
+                    return selectedPropertiesId
```
AntonAL, line 527: Spelling

```suggestion
                    return selectedPropertyIds
```

PR #3630, opened 2023-07-17
feat(ui): DOMA-6671 added icon for "Tag" component

Alllex202:
With icon
Снимок экрана 2023-07-18 в 13 21 57
Снимок экрана 2023-07-18 в 13 21 52

Without icon
Снимок экрана 2023-07-18 в 13 21 42

### DIFF
--- a/packages/ui/src/components/Tag/tag.tsx
+++ b/packages/ui/src/components/Tag/tag.tsx
@@ -8,9 +8,11 @@ import { colors } from '@open-condo/ui/src/colors'
 const TAG_CLASS_PREFIX = 'condo-tag'
 
 export type TagProps = React.HTMLAttributes<HTMLSpanElement> & {
-    children: React.ReactNode
+    children: string
     textColor?: CSSProperties['color']
     bgColor?: CSSProperties['backgroundColor']
+    icon?: React.ReactNode
+    iconPosition?: 'start' | 'end'
 }
 
 const Tag = React.forwardRef<HTMLSpanElement, TagProps>((props, ref) => {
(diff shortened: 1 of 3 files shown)


sitozzz (approved):

sitozzz:
Why this change marked by you as `feat`?

sitozzz:
I'd better prefer to use `chore` instead

packages/ui/src/components/Tag/tag.tsx
```diff
@@ -8,7 +8,7 @@ import { colors } from '@open-condo/ui/src/colors'
 const TAG_CLASS_PREFIX = 'condo-tag'
 
 export type TagProps = React.HTMLAttributes<HTMLSpanElement> & {
-    children: string
+    children: React.ReactNode
```
SavelevMatthew, line 11: Why do you need Node here?

AntonAL (approved):

packages/ui/src/components/Tag/tag.tsx
Alllex202: I need to mark up text with an icon

Alllex202: Yes, the default component has an icon property, but that icon is on the left. But I need to place it on the right. And instead of having to hardcode the position of the icon in the component, it's better to allow any content to be added.

it seems we have already encountered this problem more than once, and changed interfaces

SavelevMatthew (approved):
remove 1px hack, rest of code is LGTM

PR #3634, opened 2023-07-17
feat(ui): DOMA-6670 added Button with dropdown items

Alllex202:
**New**
 - Added `Dropdown` component (with any trigger component)
 - Added `Dropdown.Button` component
 - Fixed animations for Dropdown and Select (animation `slide`)

Menu items in `Dropdown.Button` has 3 types:
1) Only label
2) Label + description
3) Icon + label

One menu can contain different types of items

https://github.com/open-condo-software/condo/assets/56914444/efc2d448-411e-4de8-8cde-3dc5a32289ff


packages/ui/src/components/Button/dropdownButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import { MoreHorizontal } from '@open-condo/icons'
+
+import { ButtonProps } from './button'
+import { useItems } from './hooks/useItems'
+
+import { Either } from '../_utils/types'
+
+
+const DROPDOWN_CLASS_PREFIX = 'condo-dropdown'
```
SavelevMatthew, line 14: Are you sure you want to use this prefix? What if we'll add a Dropdown as an component?

SavelevMatthew:
image

Seems like not what we expect to see in block mode, doesn't it?

SavelevMatthew:
It's also strange, why block dropdown takes more space than button itself?

image

https://632796770d6841d1ae37e18e-fqmoomqpui.chromatic.com/?path=/story/components-button--with-dropdown&args=block:true

packages/ui/src/components/Button/dropdownButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+const BUTTON_CLASS_PREFIX = 'condo-btn'
+
+
+export interface IDropdownButtonItem {
+    disabled?: boolean,
+    label: string
+    onClick?: MenuItemType['onClick']
```
SavelevMatthew, line 21: Why do you need `onClick` event if Button is not clickable?

packages/ui/src/components/Button/dropdownButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+    const menuItems = useItems(items, type)
+
+    return (
+        <Dropdown
+            disabled={disabled}
+            prefixCls={DROPDOWN_CLASS_PREFIX}
+            className={`${BUTTON_CLASS_PREFIX}-dropdown`}
+            menu={{ items: menuItems }}
+            dropdownRender={(menu) => (
+                <div className={dropdownWrapperClasses}>
```
SavelevMatthew, line 74: Why do you need a custom wrapper if you have `overlayClassname` prop?

packages/ui/src/components/Button/dropdownButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+                    {
+                        React.cloneElement(menu as React.ReactElement, {
```
SavelevMatthew, line 76: Why clone?

packages/ui/src/components/Button/dropdownButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+                id={id}
+                prefixCls={BUTTON_CLASS_PREFIX}
+                className={buttonClasses}
+                ref={ref}
+                block={block}
+                disabled={disabled}
+            >
+                {children}
+                <VerticalDividerSvg />
+                <span className={`${BUTTON_CLASS_PREFIX}-ellipsis`}>
```
SavelevMatthew, line 94: `@open-condo/icons` Icons has already `<span>` wrapper and className prop, aren't this enough?

packages/ui/src/components/Button/hooks/useItems.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    for (let index = 0; index < items.length; index++) {
+        const item = items[index]
+
+        if (!item) continue
+
+        const { key, description, disabled, onClick, label, icon, id } = item
+        const mergedKey = key || index
+
+        const isItemWithIcon = item.icon
```
SavelevMatthew, line 28: `Boolean(...)`?

packages/ui/src/components/Button/hooks/useItems.tsx
```diff
@@ -0,0 +1,91 @@
...
+        const withDividerAfterItem = index + 1 < items.length
+
+        const mergedItem: MenuItemType = {
+            key: mergedKey,
+            disabled,
+            onClick: (...args) => {
+                const stringContent = label
+
+                if (stringContent) {
+                    sendAnalyticsClickEvent('Button', { value: stringContent, type, id })
```
SavelevMatthew, line 40: Is it a button event? Sounds like dropdown one

SavelevMatthew: I also think that it should not include id of option, but id of button itself, because usually I want to receive a chosen option (value / label / key ?) and id of selector

packages/ui/src/components/Button/style.less
```diff
@@ -322,3 +323,111 @@ a.condo-btn {
...
+.condo-btn-dropdown {
+  cursor: default;
+
+  &.condo-btn > span {
+    margin-right: auto;
+  }
+
+  .condo-btn-vertical-divider {
+    z-index: 1;
+    margin-right: 16px;
```
SavelevMatthew, line 336: use tokens here :)

packages/ui/src/components/Button/style.less
```diff
@@ -322,3 +323,111 @@ a.condo-btn {
...
+        position: relative;
+        top: 1px;
+        display: inline-flex;
+      }
+
+      .condo-typography {
+        /* stylelint-disable-next-line value-no-vendor-prefix */
+        display: -webkit-box;
+        overflow: hidden;
+        -webkit-line-clamp: 2;
```
SavelevMatthew, line 395: Shouldn't it been added automatically by autoprefixer?

Alllex202: It doesn't work properly automatically

packages/ui/src/components/Button/dropdownButton.tsx
Alllex202: Made as in the example by antd https://codesandbox.io/s/tl4cf4?file=/demo.tsx

packages/ui/src/components/Button/dropdownButton.tsx
Alllex202: This is necessary in order to change the color of the icon.
Look at a regular button, the icon is arranged in a similar way.

packages/ui/src/components/Button/dropdownButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+                <div className={dropdownWrapperClasses}>
+                    {
+                        React.cloneElement(menu as React.ReactElement, {
+                            className: `${BUTTON_CLASS_PREFIX}-dropdown-menu`,
+                        })
+                    }
+                </div>
+            )}
+        >
+            <DefaultButton
```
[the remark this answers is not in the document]
Alllex202: Do you want to allow custom content for this in a simple button? :)

packages/ui/src/components/Button/dropdownButton.tsx
Alllex202: Changing the prefix is difficult for us
I added an extra class so they can be distinguished

Mb add dropdown another time?)

packages/ui/src/components/Button/dropdownButton.tsx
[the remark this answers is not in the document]
SavelevMatthew: Nope, but we can  ̶i̶g̶n̶o̶r̶e̶ ̶s̶o̶m̶e̶ ̶t̶h̶i̶n̶g̶s̶ sometimes :)

packages/ui/src/components/Button/dropdownButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+    const menuItems = useItems(items, triggerId)
+
+    return (
+        <div className={`${BUTTON_CLASS_PREFIX}-container`}>
+            <Dropdown
+                disabled={disabled}
+                prefixCls={DROPDOWN_CLASS_PREFIX}
+                className={`${BUTTON_CLASS_PREFIX}-dropdown`}
+                menu={{ items: menuItems }}
+                overlayClassName={dropdownWrapperClasses}
```
SavelevMatthew, line 75: How about `-overlay` prefix instead of `-wrapper`?

toplenboren (approved):

packages/ui/src/components/Button/extendedButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+
+const BUTTON_CLASS_PREFIX = 'condo-btn'
+
+export type ExtendedButtonProps = Omit<ButtonProps, 'icon' | 'href'>
+
+const VerticalDividerSvg: React.FC = () => {
+    return (
+        <svg className={`${BUTTON_CLASS_PREFIX}-vertical-divider`} xmlns='http://www.w3.org/2000/svg' width='1' height='48' viewBox='0 0 1 48' fill='none'>
+            <rect width='1' height='48' fill='currentColor' />
```
SavelevMatthew, line 17: svg?) Why can't we just spawn a div with height: 100%?)
(just curious)

Alllex202: That was easier for me)

SavelevMatthew: Hard-coded height can be a problem sometimes)
Otherwise, svg is fine

packages/ui/src/components/Button/extendedButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+import React from 'react'
+
+import { MoreHorizontal } from '@open-condo/icons'
+
+import { ButtonProps } from './button'
+
+
+const BUTTON_CLASS_PREFIX = 'condo-btn'
+
+export type ExtendedButtonProps = Omit<ButtonProps, 'icon' | 'href'>
```
SavelevMatthew, line 12: style and others still here :)

 (That's why I've suggested to use Condo Button with children type error suppression, so you can share common props, logic, types and modificators...)

SavelevMatthew: Oh, I've forgot, that we're not exporting it. So it's fine now...

(But placing it with styles closer to DropdownButton itself can help imho)

packages/ui/src/components/Dropdown/dropdown.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { isFunction } from 'lodash'
+import get from 'lodash/get'
+import React, { useCallback } from 'react'
+
+import { extractChildrenContent, sendAnalyticsClickEvent } from '../_utils/analytics'
+
+
+export const DROPDOWN_CLASS_PREFIX = 'condo-dropdown'
+
+export type DropdownProps = Omit<DefaultDropdownProps, 'visible' | 'prefixCls' | 'overlayStyle'>
```
SavelevMatthew, line 11: probably `style` too?

packages/ui/src/components/Dropdown/dropdown.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const Dropdown: React.FC<DropdownProps> = ({ menu, children,  ...props }) => {
+    const handleClickMenu = useCallback(async (info) => {
```
SavelevMatthew, line 14: should `info` be typed?

packages/ui/src/components/Dropdown/dropdown.tsx
```diff
@@ -0,0 +1,49 @@
+import { Dropdown as DefaultDropdown, DropdownProps as DefaultDropdownProps } from 'antd'
+import { isFunction } from 'lodash'
```
SavelevMatthew, line 2: CHANGE IMPORT! 
`import funcName from 'lodash/funcName'` is decreasing bundle size

packages/ui/src/components/Dropdown/style.less
```diff
@@ -0,0 +1,76 @@
...
+    .condo-dropdown-menu-item {
+      padding: @condo-global-spacing-20;
+
+      &:hover {
+        background-color: @condo-global-color-white;
+      }
+
+      .condo-dropdown-item-icon {
+        position: relative;
+        top: 1px;
```
SavelevMatthew, line 36: ?)

packages/ui/src/components/_utils/analytics.ts
```diff
@@ -16,6 +16,7 @@ type ComponentSpecificClickEventProps = {
     Banner: { title: string }
     Button: { value: string, type: string }
     'Typography.Link': { value: string, href?: string }
+    Dropdown: { optionValue: string, optionKey?: string, optionKeyPath?: Array<string>, triggerValue?: string }
```
SavelevMatthew, line 19: `keyPath` in used in case of nested menus?

packages/ui/src/components/Button/extendedButton.tsx
Alllex202: `ButtonProps` from `CondoButton`)

packages/ui/src/components/Dropdown/dropdown.tsx
Alllex202: No, this is the only property with a style

packages/ui/src/components/_utils/analytics.ts
Alllex202: Yes, it is for such cases. It seems we don't have them, but you never know

packages/ui/src/components/Dropdown/style.less
Alllex202: I forgot to remove it)

packages/ui/src/components/Button/extendedButton.tsx
Alllex202: I'll add a note that this button is for internal use.

packages/ui/src/components/Dropdown/style.less
Alllex202: we have different heights for the text and icon, so we need to move it down by 1 pixel so that it looks even
Снимок экрана 2023-07-31 в 13 26 11
Снимок экрана 2023-07-31 в 13 26 07

packages/ui/src/components/Dropdown/dropdown.tsx
Alllex202: fixed

packages/ui/src/components/Button/extendedButton.tsx
Alllex202: remade with `div`

packages/ui/src/components/Button/extendedButton.tsx
Alllex202: moved to `Dropdown/button/...`

packages/ui/src/components/Button/style.less
```diff
@@ -322,3 +322,63 @@ a.condo-btn {
     @condo-button-border-width);
   }
 }
+
+.condo-dropdown-trigger.condo-btn-extended {
```
SavelevMatthew, line 326: Should it be in `Dropdown` / `Dropdown.button` styles?

packages/ui/src/components/Button/style.less
```diff
@@ -322,3 +322,63 @@ a.condo-btn {
...
+  padding: 0;
+  cursor: default;
+
+  .condo-btn-ellipsis {
+    z-index: 1;
+    display: inline-flex;
+    padding: 12px 20px 12px 0;
```
SavelevMatthew, line 333: Why not tokens here?

packages/ui/src/components/Dropdown/button/extendedButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+        className,
+    )
+
+    return (
+        <CondoButton
+            {...rest}
+            prefixCls={BUTTON_CLASS_PREFIX}
+            className={buttonClasses}
+        >
+            {/*@ts-ignore*/}
```
SavelevMatthew, line 33: Pls add note here for future usage...
And should `eslint-ignore` be here?

packages/ui/src/components/Dropdown/button/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+    buttonProps?: Omit<ExtendedButtonProps, 'type'>,
+    dropdownProps?: Omit<DropdownProps, 'children' | 'disabled' | 'menu'>,
+}
+
+const DropdownButton: React.FC<DropdownButtonProps> = (props) => {
+    const { children, items = [], buttonProps, dropdownProps, id: triggerId, disabled, type } = props
+
+    const dropdownOverlayClasses = classNames(
+        {
+            [`${DROPDOWN_CLASS_PREFIX}-wrapper`]: true,
```
SavelevMatthew, line 45: Just string here? Since condition is always true...

SavelevMatthew (approved):

packages/ui/src/components/Button/style.less
Alllex202: moved to `Dropdown.button ` styles

PR #3655, opened 2023-07-24
feat(condo): DOMA-6693 added permissions a mini apps for requests on behalf of service user

Alllex202:
## Problem
From the miniapp on behalf of the service user it is impossible to receive data of organizations connected to this miniapp. 

---

#### What does it do?
 * In schemes that are somehow connected with the organization scheme, we add a custom access. What is it checking?
    - Request on behalf of service user
    - The organization is connected to miniapp `A` (have `B2BAppContext`)
    - Service user connected to miniapp `A` (have `B2BAppAccessRight`)
    - In the scheme `B2BAppAccessRight` for the miniapp `A` the necessary rights were issued to execute the request
 * Adds `canRead...`/`canManage...` fields to the `B2BAppAccessRightSet` schema, based on the name of the schemas for which you need to be able to grant access to service users who have `B2BAppAccessRight` (For example, `canReadContacts`, `canManageProperties`)

---

## Schema `B2BAppAccessRightSet`
Contains the `canRead...`/`canManage...` rights set and a link to `B2BApp`. You can read/create/delete. But you can't modify it!

Added constraint for the `app` field. You can only create one item for each miniapp.

---

## Schema `B2BAppAccessRight`
A new field `accessRightSet` is added with a reference to `B2BAppAccessRightSet`.

Added constraint for the `app` field. Тow you can only create one item for each miniapp.
 
---

## Schema `B2СAppAccessRight`

Added constraint for the `app` field. Тow you can only create one item for each miniapp.
 
---

## How does it work for the miniapp developer?
For example, we have created a miniapp `M`. And we want to be able to read and change contacts in organizations connected to your miniapp `M`.

1) (*) To do this, we need to create a user `S` with the type `service`.
2) (*) Then create `B2BAppAccessRight` with user `S` and miniapp `M`.
3) (*) After that, we can create `B2BAppAccessRightSet` with the necessary set of rights (in our case, these are `canReadOrganizations`, `canReadContacts` and `canManageContacts`).
4) (*) The next step is to attach our set of permissions to `B2BAppAccessRight` which refers to our miniapp `X`.

Congratulations! Now we can, on behalf of the service user `S`, read and change contacts for organizations that are connected to our miniapp `X`.

If we need to change the rights, then for this we just need to update the existing set of rights for the miniapp `X`.

_(*) - Some actions can only be done by privileged users (admin, support)_

---

 Now it only works for the schemes `Organization`, `Contact` and `Property`


apps/condo/domains/miniapp/constants.js
```diff
@@ -92,6 +92,28 @@ const PROMO_BLOCK_TEXT_VARIANTS_TO_PROPS = {
     [PROMO_BLOCK_LIGHT_TEXT_VARIANT]: { invertText: true },
 }
 
+const SCHEMAS_AVAILABLE_TO_B2B_APP = [
+    'Organization',
+    'Contact',
+    'Property',
+]
+
+const B2B_APP_ACCESS_CONFIG = {
```
SavelevMatthew, line 101: It will be cool to add a JSDoc here :)

apps/condo/domains/miniapp/gql.js
```diff
@@ -30,7 +33,11 @@ const B2BApp = generateGqlQueries('B2BApp', B2B_APP_FIELDS)
 const B2B_APP_CONTEXT_FIELDS = `{ app { id name appUrl icon menuCategory hasDynamicTitle } organization { id } status ${COMMON_FIELDS} }`
 const B2BAppContext = generateGqlQueries('B2BAppContext', B2B_APP_CONTEXT_FIELDS)
 
-const B2B_APP_ACCESS_RIGHT_FIELDS = `{ app { id } user { id } ${COMMON_FIELDS} }`
+const B2B_PERMISSION_FIELDS = SCHEMAS_AVAILABLE_TO_B2B_APP
+    .map(schemaName => `canManage${schemaName} canRead${schemaName}`)
```
SavelevMatthew, line 37: should it be pluralized?

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
```diff
@@ -186,5 +203,327 @@ describe('B2BAppAccessRight', () => {
                 })
             }, NON_SERVICE_USER_ERROR)
         })
+
+        test('Cannot be created 2 links for a app', async () => {
```
SavelevMatthew, line 207: Please move it to `Constraints` section of tests to keep it organized (Since it more db / migrator error), then keystone hooks

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
```diff
@@ -186,5 +203,327 @@ describe('B2BAppAccessRight', () => {
...
+
+            const [app] = await createTestB2BApp(admin)
+            await createTestB2BAppAccessRight(admin, serviceUser, app)
+
+            await expectToThrowUniqueConstraintViolationError(async () => {
+                await createTestB2BAppAccessRight(admin, serviceUser2, app)
+            }, 'b2b_app_access_right_unique_app')
+        })
+
+        test('One service user can be linked for different mini apps', async () => {
```
SavelevMatthew, line 220:
```suggestion
        test('One service user can be linked for different B2BApps', async () => {
```

apps/condo/domains/miniapp/schema/B2BAppAccessRight.js
```diff
@@ -24,6 +24,16 @@ const B2BAppAccessRight = new GQLListSchema('B2BAppAccessRight', {
             kmigratorOptions: { null: false, on_delete: 'models.PROTECT' },
         },
     },
+    kmigratorOptions: {
```
SavelevMatthew, line 27: Should we add similar constraint to B2CApp as well?

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
```diff
@@ -186,5 +203,327 @@ describe('B2BAppAccessRight', () => {
...
+            const [app] = await createTestB2BApp(admin)
+            const [app2] = await createTestB2BApp(admin)
+
+            await createTestB2BAppAccessRight(admin, serviceUser, app)
+            await createTestB2BAppAccessRight(admin, serviceUser, app2)
+        })
+    })
+})
+
+describe('Mini app permissions for service user', () => {
```
SavelevMatthew, line 233: Specify type of app to be clear in terminology

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
```diff
@@ -186,5 +203,327 @@ describe('B2BAppAccessRight', () => {
...
+        expect(countByUser).toBe(1)
+        const contactByUser = await Contact.getOne(user, { id: contact.id })
+        expect(contactByUser.id).toBe(contact.id)
+
+        // add permissions for B2BApp
+        await updateTestB2BAppAccessRight(admin, right.id, {
+            canReadContact: true,
+            canManageContact: true,
+            canReadOrganization: true, // required permission for manage contact
+            canReadProperty: true, // required permission for manage contact
```
SavelevMatthew, line 411: Did I guess it right, that each miniapp should resolve it manually?

SavelevMatthew: Do we want to build a relation-deps graph or we'll resolve it in UI-state?

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
```diff
@@ -186,5 +203,327 @@ describe('B2BAppAccessRight', () => {
...
+        await updateTestContact(serviceUserClient, contact.id, {})
+        await Contact.softDelete(serviceUserClient, contact.id)
+
+        const countByUser3 = await Contact.count(user, { organization: { id: organization.id } })
+        expect(countByUser3).toBe(1)
+    })
+
+    // NOTE: currently these schemes are not added to read/manage the service user.
+    //       It is used to test the preprocessor
+    test.skip('Ticket and TicketComment', async () => {
```
SavelevMatthew, line 434: Why add test and skip it? To run locally?

apps/condo/domains/miniapp/schema/B2BAppAccessRight.js
Alllex202: If we want to be consistent, then we should probably

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
Alllex202: Only the admin and the support can change permissions. A miniapp itself cannot change permissions.

There are no snippets yet. But it would be convenient

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
Alllex202: Yes, for local testing

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -0,0 +1,44 @@
...
+const access = require('@condo/domains/miniapp/access/B2BAppAccessRightSet')
+
+
+const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
+    schemaDoc: 'A set of access rights for a service user integrated with B2BApp.\n' +
+        ' The user can create a set of necessary rights for B2BApp to work.\n' +
+        ' Then it is:\n' +
+        '   - approved (for the corresponding B2BAppAccessRight the link in the "accessRightSet" field to the desired set is specified)\n' +
+        '   - or rejected.',
+    fields: {
```
AntonAL, line 17: How set records will be distinguished from each other from user point of view.
There is no "name" field or something like this…

Alllex202: For now, we have decided to prevent users from changing this model. Only supports and admins can do this. We also added a restriction so that only one set of rights can be created for one B2BApp. This is done for the reason that the functionality with the creation of several sets of rights on behalf of the service user is currently not a priority and has not been carefully developed. 
For the same reason, right now we don't need a "name" field or anything similar.

Alllex202: Description updated

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,617 @@
...
+                await expectToThrowAccessDeniedErrorToObj(async () => {
+                    await B2BAppAccessRightSet.softDelete(integratedToAnotherAppServiceUser, createdAccessRightSet.id)
+                })
+
+                await expectToThrowAccessDeniedErrorToObj(async () => {
+                    await B2BAppAccessRightSet.softDelete(integratedServiceUser, createdAccessRightSet.id)
+                })
+            })
+
+            test('Can not hard-delete', async () => {
```
sitozzz, line 211: IMO it's overkill to check hard delete for every user type

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
```diff
@@ -186,5 +191,64 @@ describe('B2BAppAccessRight', () => {
...
+            const [serviceUser] = await registerNewServiceUserByTestClient(admin)
+
+            const [app] = await createTestB2BApp(admin)
+            const [app2] = await createTestB2BApp(admin)
+
+            await createTestB2BAppAccessRight(admin, serviceUser, app)
+            await createTestB2BAppAccessRight(admin, serviceUser, app2)
+        })
+
+        test('Сan not create or update "B2BAppAccessRight" if "app" and "accessRightSet" refer to different "B2BApp"', async () => {
```
sitozzz, line 206: It would be looking better inside describe block with caption `Validations` isn't it?

sitozzz: Because of here you check code throws an error from `validateInput` hook of the schema

sitozzz (approved):

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
Alllex202: The test is inside this block))

apps/condo/domains/miniapp/access/B2BAppAccessRight.js
```diff
@@ -10,6 +10,7 @@ async function canReadB2BAppAccessRights ({ authentication: { item: user } }) {
 
     if (user.isSupport || user.isAdmin) return {}
 
+    // TODO(DOMA-6766): Add the ability to read the permissions of B2BApp for employees whose role has the flag "canManageIntegrations"
```
SavelevMatthew, line 13: Why do we need such a permission?

apps/condo/domains/miniapp/constants.js
```diff
@@ -95,6 +96,23 @@ const PROMO_BLOCK_TEXT_VARIANTS_TO_PROPS = {
...
+    'Property',
+]
+
+/**
+ * @type {B2BAppAccessConfigBySchemaName}
+ */
+const B2B_APP_ACCESS_CONFIG = {
+    Organization: {
+        pathToOrganizationId: ['id'],
+        canNotBeManage: true,
```
SavelevMatthew, line 111: a bit strange naming :)

Maybe `cannotBeManaged` or `disableManageOperations` 🤔 ?

apps/condo/domains/miniapp/schema/B2BAppAccessRight.js
```diff
@@ -2,27 +2,56 @@
...
             schemaDoc: 'Link to B2BApp.accessRights',
-            type: Relationship,
+            type: 'Relationship',
             ref: 'B2BApp.accessRights',
             isRequired: true,
             knexOptions: { isNotNullable: true }, // Required relationship only!
             kmigratorOptions: { null: false, on_delete: 'models.CASCADE' },
         },
+        accessRightSet: {
+            schemaDoc: 'Link to the set of access rights available to this application',
```
SavelevMatthew, line 38: Maybe available is not a good word right here...Since it's sound like an option to choose. Not like "this is the thing will be used to check your accesses" :)

apps/condo/domains/miniapp/schema/B2BAppAccessRight.js
```diff
@@ -32,8 +61,28 @@ const B2BAppAccessRight = new GQLListSchema('B2BAppAccessRight', {
...
+            if (accessRightSetId) {
+                const appId = get(newItem, 'app')
+                const accessRightSet = await getByCondition('B2BAppAccessRightSet', {
+                    app: { id: appId, deletedAt: null },
+                    deletedAt: null,
+                    id: accessRightSetId,
+                })
+
+                if (!accessRightSet) {
+                    throw new GQLError(ERRORS.ACCESS_RIGHT_SET_NOT_FOUND, context)
```
SavelevMatthew, line 78: It's not exactly "Not found" for user. It does not match a criteria (to have same app as accessRight)

apps/condo/domains/miniapp/schema/B2BAppAccessRight.js
```diff
@@ -32,8 +61,28 @@ const B2BAppAccessRight = new GQLListSchema('B2BAppAccessRight', {
         delete: false,
         auth: true,
     },
+    hooks: {
```
SavelevMatthew, line 64: Why not just field hook? Since it's will only be invoked when `accessRightSet` field is changed

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
```diff
@@ -186,5 +191,64 @@ describe('B2BAppAccessRight', () => {
...
+
+        test('One service user can be linked for different B2BApps', async () => {
+            const admin = await makeLoggedInAdminClient()
+            const [serviceUser] = await registerNewServiceUserByTestClient(admin)
+
+            const [app] = await createTestB2BApp(admin)
+            const [app2] = await createTestB2BApp(admin)
+
+            await createTestB2BAppAccessRight(admin, serviceUser, app)
+            await createTestB2BAppAccessRight(admin, serviceUser, app2)
```
SavelevMatthew, line 203: Maybe expect something? (at least that it is defined) :)

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
```diff
@@ -186,5 +191,64 @@ describe('B2BAppAccessRight', () => {
...
+            }, ERRORS.ACCESS_RIGHT_SET_NOT_FOUND)
+            await expectToThrowGQLError(async () => {
+                await updateTestB2BAppAccessRight(admin, b2BAppAccessRight.id, {
+                    accessRightSet: { connect: { id: accessRightSet2.id } },
+                })
+            }, ERRORS.ACCESS_RIGHT_SET_NOT_FOUND)
+        })
+    })
+    describe('Constraints', () => {
+        test('Cannot be created 2 links for a app', async () => {
```
SavelevMatthew, line 241:
```suggestion
        test('Cannot be created 2 active access rights for a single app', async () => {
```

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+const { historical, versioned, uuided, tracked, softDeleted, dvAndSender } = require('@open-condo/keystone/plugins')
+const { GQLListSchema } = require('@open-condo/keystone/schema')
+
+const access = require('@condo/domains/miniapp/access/B2BAppAccessRightSet')
+
+
+const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
+    schemaDoc: 'A set of access rights for a service user integrated with B2BApp.',
```
SavelevMatthew, line 12: Please provide more description to all miniapps models, since they're actively used by our partners, so it important for this descriptions to be as clear as possible

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -0,0 +1,50 @@
...
+            type: 'Relationship',
+            ref: 'B2BApp',
+            isRequired: true,
+            knexOptions: { isNotNullable: true }, // Required relationship only!
+            kmigratorOptions: { null: false, on_delete: 'models.CASCADE' },
+            access: {
+                update: false,
+            },
+        },
+
```
SavelevMatthew, line 26: Write a note, that a lot of fields will be added with preprocessor/plugin. Not obvious, where to find them :)

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,650 @@
...
+
+    let accessRightSet
+
+    beforeAll(async () => {
+        admin = await makeLoggedInAdminClient()
+        support = await makeClientWithSupportUser()
+        notServiceUser = await makeClientWithNewRegisteredAndLoggedInUser()
+        anonymous = await makeClient()
+
+        const [createdApp] = await createTestB2BApp(admin)
```
SavelevMatthew, line 60: Be true real-lief guy, use more support instead of admin :)

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,650 @@
...
+
+            test('Can read everything', async () => {
+                const [foundedAccessRightSet] = await B2BAppAccessRightSet.getAll(admin, {
+                    id: accessRightSet.id,
+                })
+                expect(foundedAccessRightSet).toHaveProperty('id', accessRightSet.id)
+            })
+
+            test('Can update', async () => {
+                const [updatedAccessRightSet] = await updateTestB2BAppAccessRightSet(admin, accessRightSet.id)
```
SavelevMatthew, line 103: Why empty payload? :(

Just to check its possible to update (not representative for real-life case imo)

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,650 @@
...
+                await expectToThrowAccessDeniedErrorToObj(async () => {
+                    await createTestB2BAppAccessRightSet(integratedServiceUser, app)
+                })
+
+                await expectToThrowAccessDeniedErrorToObj(async () => {
+                    await createTestB2BAppAccessRightSet(integratedToAnotherAppServiceUser, app)
+                })
+            })
+
+            test('Can only read if integrated with app', async () => {
```
SavelevMatthew, line 176: What do you mean by term "integrated". App must be connected or it should have access rights?

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,650 @@
...
+                const [foundedAccessRightSet] = await B2BAppAccessRightSet.getAll(integratedServiceUser, {
```
SavelevMatthew, line 177: just `getAll` with `id_in` and expect only 1 to come in array :)

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,650 @@
...
+    let admin
+    let user
+    let serviceUserClient
+    let serviceUser
+
+    beforeEach(async () => {
+        admin = await makeLoggedInAdminClient()
+        user = await makeClientWithNewRegisteredAndLoggedInUser()
+        const [newServiceUser] = await registerNewServiceUserByTestClient(admin)
+        serviceUser = newServiceUser
```
SavelevMatthew, line 328: How about to assign user to `.user` prop to be more consistent with all project tests?

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,650 @@
...
+})
+
+describe('B2BApp permissions for service user', () => {
+    let admin
+    let user
+    let serviceUserClient
+    let serviceUser
+
+    beforeEach(async () => {
+        admin = await makeLoggedInAdminClient()
```
SavelevMatthew, line 325: It's better to use support, since it's real-life test scenarios

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,650 @@
...
+        const [organization] = await registerNewOrganization(user)
+
+        const [app] = await createTestB2BApp(admin)
+        await createTestB2BAppContext(admin, app, organization, { status: 'Finished' })
+        const [right] = await createTestB2BAppAccessRight(admin, serviceUser, app)
+
+        // B2BApp without permissions
+        const countWithoutPermissions = await Organization.count(serviceUserClient, {})
+        expect(countWithoutPermissions).toBe(0)
+        const propertyWithoutPermissions = await Organization.getOne(serviceUserClient, { id: organization.id })
```
SavelevMatthew, line 356: property?)

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,650 @@
...
+            await createTestOrganization(serviceUserClient)
+        })
+        await expectToThrowAccessDeniedErrorToObj(async () => {
+            await updateTestOrganization(serviceUserClient, organization.id, {})
+        })
+        await expectToThrowAccessDeniedErrorToObj(async () => {
+            await Organization.softDelete(serviceUserClient, organization.id)
+        })
+
+        const countByUser = await Organization.count(user, {})
```
SavelevMatthew, line 369: If you want to check, that user can see specific organisation = pass `id` to filter. And expect matching id in response.

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,650 @@
...
+        expect(countByUser).toBe(1)
+
+        // you cannot create a field "canManageOrganizations: true". Only reading! Always false!
```
SavelevMatthew, line 372: Cannot get it :)

Note says I cannot set true value, but below is false and error is still happening, seems like it's an old description :)

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,650 @@
...
+            await updateTestB2BAppAccessRightSet(admin, accessRightSet.id, {
+                canManageOrganizations: false,
+            })
+        }, (e) => {
+            expect(e.errors[0].message).toContain('Field "canManageOrganizations" is not defined by type "B2BAppAccessRightUpdateInput"')
+        })
+
+        // B2BApp with permission "canReadOrganization"
+        const countWithPermissions = await Organization.count(serviceUserClient, {})
+        expect(countWithPermissions).toBe(1)
```
SavelevMatthew, line 403: It's cool to check that only one organization is available and no more, but add a check which exactly org is expected to be seen

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,650 @@
...
+
+        await expectToThrowAccessDeniedErrorToObj(async () => {
+            await createTestOrganization(serviceUserClient)
+        })
+        await expectToThrowAccessDeniedErrorToObj(async () => {
+            await updateTestOrganization(serviceUserClient, organization.id, {})
+        })
+        await expectToThrowAccessDeniedErrorToObj(async () => {
+            await Organization.softDelete(serviceUserClient, organization.id)
+        })
```
SavelevMatthew, line 415: I see 2 out of 3 cases here, it would be cool to add the third one

- [x] Have no accessRightSet
- [x] Have true permission
- [ ] Have false permissions

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,650 @@
...
+            await Property.softDelete(serviceUserClient, property.id)
+        })
+
+        const countByUser = await Property.count(user, {})
+        expect(countByUser).toBe(1)
+        const propertyByUser = await Property.getOne(user, { id: property.id })
+        expect(propertyByUser.id).toBe(property.id)
+
+        // add permissions for B2BApp
+        const [accessRightSet] = await createTestB2BAppAccessRightSet(admin, app, {
```
SavelevMatthew, line 449: How about separate read / manage checks?

For example:
I can update property area without having read permission to organization.
I can read property addresses without having access to org and so on.

SavelevMatthew: Also I've found an issue / question:

Do we allowing apps to change property address?

Alllex202: > Do we allowing apps to change property address?

In the plugin, at the moment, the accesses for fields do not change in any way. Only accesses to schemas change.

That is, default checks are used for fields, which are specified in schemas

If this is done now, it will complicate current plugin logic and its configuration. Therefore, I suggest leaving it for the next iteration)

apps/condo/domains/miniapp/schema/B2BAppAccessRight.js
Alllex202: I'll write exactly as you suggested (`this is the thing will be used to check your accesses`) :)

apps/condo/domains/miniapp/schema/B2BAppAccessRight.js
Alllex202: This is made in a common hook because the "app" field can be changed

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
Alllex202: This is enough to catch the error. 
But I'll add it to make it more obvious)

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
```diff
@@ -0,0 +1,470 @@
...
+const { isBoolean, isObject, isEmpty, isFunction, get, isArray, set, isString } = require('lodash')
+const pluralize = require('pluralize')
+
+const { execGqlWithoutAccess } = require('@open-condo/codegen/generate.server.utils')
+const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
+const { evaluateKeystoneAccessResult } = require('@open-condo/keystone/plugins/utils')
+const { GQL_LIST_SCHEMA_TYPE, find, getById } = require('@open-condo/keystone/schema')
+
+
+function fieldAccessWrapperIfNeeded (access, fnWrapper) {
```
SavelevMatthew, line 12: Please provide JSDocs to utils, since its not clear, what fnWrapper is or what it does

SavelevMatthew: So you need to build a puzzle yourself (which after thousands line of review is kinda hard :) )

SavelevMatthew: Also why prefix it with `ifNeeded`? if I need function, I call it :)

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
```diff
@@ -0,0 +1,470 @@
...
+ *
+ * @example
+ * Input: ['ticket', 'organization', 'id']
+ * Output: gql`
+ *         query getAllTickets ($where: TicketWhereInput, $first: Int = 1) {
+ *             objs: allTickets(where: $where, first: $first) { organization { id } }
+ *         }
+ *     `
+ */
+const generateGql = (pathToOrganizationId) => {
```
SavelevMatthew, line 72:
```suggestion
const generateGqlQuery = (pathToOrganizationId) => {
```

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
```diff
@@ -0,0 +1,470 @@
...
+    if (!pathToOrganizationId || !isArray(pathToOrganizationId) || isEmpty(pathToOrganizationId)) throw new Error('"pathToOrganizationId" should not be empty array')
+    if (pathToOrganizationId.length < 3) throw new Error(`To generate gql "pathToOrganizationId" must contain at least three elements! But was ${pathToOrganizationId}`)
```
SavelevMatthew, line 74: Where's this restriction comes from?

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
```diff
@@ -0,0 +1,470 @@
...
+    if (user.deletedAt) return false
+
+    if (!listKey) return false
+
+    const pathToOrganizationId = get(schemaConfig, 'pathToOrganizationId', ['organization', 'id'])
+
+    let organizationId
+
+    if (operation === 'create') {
+        if (pathToOrganizationId.length === 1) {
```
SavelevMatthew, line 105: Why 1, 2, 3, error cases? Aren't there only 2 of them:

1. Direct access: Not sure that it should take id from original input (since it generated), but ok. Anyway this case is not real, since we prevent organization from being created.
2. Direct relation access: get it from connect.id is fine.
3. Non-direct relation access. (any length > 2). Get itemId from connect, do request returning organization id. Access it.

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
```diff
@@ -0,0 +1,470 @@
...
+            })
+
+            if (!parentObject) return false
+
+            organizationId = get(parentObject, [pathToOrganizationId[1], pathToOrganizationId[2]])
+        } else {
+            throw new Error(`You should implement "canManageByServiceUser" when "pathToOrganizationId" have depth ${pathToOrganizationId.length}`)
+        }
+    } else if (operation === 'update') {
+        const item = await getById(listKey, itemId)
```
SavelevMatthew, line 129: This can also be simplified:
1. If length is one, take itemId directly.
2. If length is more, build a query and unpack it

SavelevMatthew: Please also note (probably add test for it?), that `itemId` is not always defined. In case of bulk-requests (`updateProperties` for example) keystone uses array named `itemIds` instead (for reference check out receipts in billing domain)

It'll be good to support bulk requests, since some miniapps update information by CRON, and we can save it from some request-spam

SavelevMatthew: It's not that hard to implement as well:
1. In case of update `itemId` becomes `[itemId]` (bulk of one). Then you can generate same query, but use `id_in` instead

SavelevMatthew: I also concerned about relations, since it's not checked there. Should we protect relations from being changed?

Since right now it's possible to change it from miniapp (change organization field on contact and other things like that)

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
Alllex202: This is left over from previous implementation)

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
Alllex202: in this case, we also check that there is no access to data from other organizations. And then we try to get data from specified organization

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
Alllex202: done

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
Alllex202: > How about separate read / manage checks?

Do you want to write this for all schemas? This is madness! :)

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
Alllex202: Otherwise you won't be able to generate gql query

[0]: all<name>s - name query
[1]: field parent item
[2]: field from first children
[3]: field from second children
[...]: field from ... children

There is an example for this function, look at it. 

But yes, you are right, this function can be run with two elements as well

I'll fix it

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
Alllex202: I don't know of cases where we need to check for  connection `A -> B -> C -> D (-> ...)`. (And actually at the moment it is not needed) Therefore, I preferred to leave the implementation of these cases for later, if necessary

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
Alllex202: Сhanged the names of the arguments. Seems like it should be a lot clearer

Alllex202: This function was taken from another preprocessor and has been slightly modified)

Alllex202: This is just a merging of accesses

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
Alllex202: > This can also be simplified:
> 
> 1. If length is one, take itemId directly.
> 2. If length is more, build a query and unpack it

done

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
Alllex202: refactored

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
Alllex202: > Please also note (probably add test for it?), that `itemId` is not always defined. In case of bulk-requests (`updateProperties` for example) keystone uses array named `itemIds` instead (for reference check out receipts in billing domain)
> 
> It'll be good to support bulk requests, since some miniapps update information by CRON, and we can save it from some request-spam

Please, not now)

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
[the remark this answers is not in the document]
Alllex202: done

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
SavelevMatthew: You've already wrote a helpful utils for this. So any depth can be done with an ease :)

Because right now it seems something like...
image

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
```diff
@@ -0,0 +1,505 @@
...
+        // NOTE: to prevent multiple wrapping the same function
+        if (defaultAccess.alreadyprocessedbyb2bappaccessplugin) return defaultAccess
+        else return customAccessFn
+    } else if (type === 'Object') {
+        const newAccess = {}
+        if (typeof defaultAccess.read !== 'undefined') newAccess.read = wrapInCustomAccess(defaultAccess.read, customAccessFn)
+        if (typeof defaultAccess.create !== 'undefined') newAccess.create = wrapInCustomAccess(defaultAccess.create, customAccessFn)
+        if (typeof defaultAccess.update !== 'undefined') newAccess.update = wrapInCustomAccess(defaultAccess.update, customAccessFn)
+        if (typeof defaultAccess.delete !== 'undefined') newAccess.delete = wrapInCustomAccess(defaultAccess.delete, customAccessFn)
+        if (typeof defaultAccess.auth !== 'undefined') newAccess.auth = wrapInCustomAccess(defaultAccess.auth, customAccessFn)
```
nomerdvadcatpyat, line 34: looks very similiar

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
```diff
@@ -0,0 +1,505 @@
...
+    if (pathToOrganizationId.length < 2) throw new Error(`To generate gql "pathToOrganizationId" must contain at least two elements! But was ${pathToOrganizationId}`)
+
+    const gqlKey = pathToOrganizationId.join(':')
+
+    // get already generated gql query to reuse them
+    if (ALL_GENERATED_GQL_QUERIES.has(gqlKey)) {
+        return ALL_GENERATED_GQL_QUERIES.get(gqlKey)
+    }
+
+    const listKey = pathToOrganizationId[0].charAt(0).toUpperCase() + pathToOrganizationId[0].slice(1)
```
nomerdvadcatpyat, line 108: field name may not match with listkey

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
```diff
@@ -0,0 +1,505 @@
...
+
+    if (!listKey) return false
+
+    const isBulkRequest = Array.isArray(originalInput)
+
+
+    if (isBulkRequest) return false
+
+    const pathToOrganizationId = get(schemaConfig, 'pathToOrganizationId', ['organization', 'id'])
+    if (!isArray(pathToOrganizationId) || isEmpty(pathToOrganizationId)) return false
```
nomerdvadcatpyat, line 130: `pathToOrganizationId` cannot be empty because you set default value in `get`

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
```diff
@@ -0,0 +1,505 @@
...
+
+    if (permissionFieldName.startsWith('canManage')) {
+        return 'Currently, this field is read-only. You cannot get manage access for the specified schema.'
+    }
+
+    throw new Error('Permission field name no starts with "canManage" or "canRead"! You should check the implementation!')
+}
+
+const addPermissionFieldsToSchema = (schema, schemaName, readOnlyPermissionFields) => {
+    if (ALL_PERMISSION_NAMES.size < 1) {
```
nomerdvadcatpyat, line 308: you use `ALL_PERMISSION_NAMES` before declaration

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
Alllex202: if you can set `pathToOrganizationId: []`

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
Alllex202: it seems that nothing can be done about it

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
Alllex202: Oh yeah, fixed

But this will not affect anything, because this function is called after const declaration

apps/condo/domains/miniapp/constants.js
```diff
@@ -95,6 +96,23 @@ const PROMO_BLOCK_TEXT_VARIANTS_TO_PROPS = {
...
+    'Property',
+]
+
+/**
+ * @type {B2BAppAccessConfigBySchemaName}
+ */
+const B2B_APP_ACCESS_CONFIG = {
+    Organization: {
+        pathToOrganizationId: ['id'],
+        cannotBeManage: true,
```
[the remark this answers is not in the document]
Alllex202: done

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
Alllex202: Now this is taken automatically from schema field

nomerdvadcatpyat (approved):

apps/condo/domains/miniapp/constants.js
```diff
@@ -95,6 +96,23 @@ const PROMO_BLOCK_TEXT_VARIANTS_TO_PROPS = {
...
+    'Property',
+]
+
+/**
+ * @type {B2BAppAccessConfigBySchemaName}
+ */
+const B2B_APP_ACCESS_CONFIG = {
+    Organization: {
+        pathToOrganizationId: ['id'],
+        canBeManage: false,
```
SavelevMatthew, line 111: `beManage`?)

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -0,0 +1,1017 @@
...
+        expect(foundRight).toHaveProperty('accessRightSet.canReadOrganizations', false)
+        expect(foundRight).toHaveProperty('accessRightSet.canManageOrganizations', false)
+        expect(foundRight).toHaveProperty('accessRightSet.canReadProperties', false)
+        expect(foundRight).toHaveProperty('accessRightSet.canManageProperties', false)
+        expect(foundRight).toHaveProperty('accessRightSet.canReadContacts', false)
+        expect(foundRight).toHaveProperty('accessRightSet.canManageContacts', false)
+    })
+
+    describe('Bulk-operations', () => {
+        test('Cannot create many Properties in one request', async () => {
```
SavelevMatthew, line 369: 😢

apps/condo/domains/miniapp/schema/preprocessors/B2BAppAccess.js
```diff
@@ -0,0 +1,521 @@
...
+    } else if (type === 'AsyncFunction') {
+        // NOTE: to prevent multiple wrapping the same function
+        if (defaultAccess.alreadyProcessedByB2BAppAccessPlugin) return defaultAccess
+        else return customAccessFn
+    } else if (type === 'Object') {
+        const newAccess = {}
+        if (typeof defaultAccess.read !== 'undefined') newAccess.read = wrapInCustomAccess(defaultAccess.read, customAccessFn)
+        if (typeof defaultAccess.create !== 'undefined') newAccess.create = wrapInCustomAccess(defaultAccess.create, customAccessFn)
+        if (typeof defaultAccess.update !== 'undefined') newAccess.update = wrapInCustomAccess(defaultAccess.update, customAccessFn)
+        if (typeof defaultAccess.delete !== 'undefined') newAccess.delete = wrapInCustomAccess(defaultAccess.delete, customAccessFn)
```
SavelevMatthew, line 52: Why should we wrap delete and auth operations? Shouldn't they be restricted?

Alllex202: They usually have a boolean value, so they have no effect

PR #3662, opened 2023-07-26
feat(condo): DOMA-6736 schema MobileFeatureConfig with tests

VKislov:


AleX83Xpert (commented):
1. Is this model will be used **only** for disabling the ability to send tickets?
2. Is there a new domain needed? What about using one of the existing domains (organization, user, resident, ticket)?
3. Also, please fix the code according to the linter rules.

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+        code: BAD_USER_INPUT,
+        variable: ['data', 'organization'],
+        type: 'ORGANIZATION_IS_REQUIRED',
+        message: 'The organization is required to create',
+        messageForUser: 'api.organization.MobileFeatureConfig.ORGANIZATION_IS_REQUIRED',
+    },
+}
+
+const MobileFeatureConfig = new GQLListSchema('MobileFeatureConfig', {
+    schemaDoc: 'The entity helps to manage the enabling/disabling of some features in the mobile application,' +
```
AntonAL, line 41:
```suggestion
    schemaDoc: 'Manages availability of some features in mobile application,' +
```

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+        ' as well as to store the data required in the disabled state of the feature.',
```
AntonAL, line 42:
```suggestion
        'stores data required in disabled state of a feature.',
```

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+            schemaDoc: 'Ref to the organization. Also needed for scope management',
+            type: Relationship,
+            ref: 'Organization',
+            isRequired: true,
+            knexOptions: { isNotNullable: true }, // Required relationship only!
+            kmigratorOptions: { null: false, on_delete: 'models.CASCADE' },
+        },
+
+        commonPhone: {
+            schemaDoc: 'The phone number where the organization wants to receive common calls',
```
AntonAL, line 55:
```suggestion
            schemaDoc: 'Phone number where the organization wants to receive common calls',
```

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+            type: Text,
+        },
+
+        ticketSubmittingIsDisabled: {
```
AntonAL, line 59: Simpler
In natural language: "ticket submission enabled/disabled".

```suggestion
        ticketSubmission: {
```

And don't forget to set default according to business logic.
I suppose, that they should can.

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+            schemaDoc: 'Disabling the ability to send tickets by the user of the resident\'s mobile application. ' +
```
AntonAL, line 60:
```suggestion
            schemaDoc: 'Ability to send tickets by user of resident\'s mobile application. ' +
```

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+                        }
+                        if (resolvedData['commonPhone'] && normalizePhone(resolvedData['commonPhone']) !== resolvedData['commonPhone']) {
+                            throw new GQLError(ERRORS.COMMON_PHONE_INVALID, context)
+                        }
+                    }
+                },
+            },
+        },
+
+        onlyProgressionMeterReadingsIsEnabled: {
```
AntonAL, line 77: "…IsEnabled" is a redundant suffix, because the field is a Checkbox.

```suggestion
        onlyProgressionMeterReadings: {
```

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+            schemaDoc: 'If enabled, the resident can only submit meter readings if they are greater than the previous ones. ' +
```
AntonAL, line 78:
```suggestion
            schemaDoc: 'Restricts residents to submit only meter readings, that are greater than previous. ' +
```

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: Naming the fields like that, I proceeded from a different logic. Now there are 5 fields needed to configure 2 features. When there are 30 fields, finding the right one will be a very difficult task. Therefore, I write the feature prefix and then the field itself. That is, the feature `ticketSubmitting` and then what the field stores. If you change `ticketSubmittingIsEnabled` to `ticketSubmission`, then searching for all fields related to the `ticketSubmitting` feature will be impossible and you will have to manually search for everything.

apps/condo/domains/settings/access/MobileFeatureConfig.js
```diff
@@ -0,0 +1,74 @@
...
+    if ( operation === 'update') {
+        if (!itemId) throw new Error('no itemId')
+
+        const foundConfig = await getById('MobileFeatureConfig', itemId)
+        if (!foundConfig) return false
+
+        organizationId = get(foundConfig, 'organization')
+    }
+
+    return await checkPermissionInUserOrganizationOrRelatedOrganization(user.id, organizationId, 'canManageOrganization')
```
pahaz, line 62: Add another permission

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+        code: BAD_USER_INPUT,
+        variable: ['data', 'organization'],
+        type: 'ORGANIZATION_IS_REQUIRED',
+        message: 'The organization is required to create',
+        messageForUser: 'api.organization.MobileFeatureConfig.ORGANIZATION_IS_REQUIRED',
+    },
+}
+
+const MobileFeatureConfig = new GQLListSchema('MobileFeatureConfig', {
+    schemaDoc: 'Manages availability of some features in mobile application,' +
```
pahaz, line 41: It’s unclear. It assumes they you have one model per one feature.

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+            knexOptions: { isNotNullable: true }, // Required relationship only!
+            kmigratorOptions: { null: false, on_delete: 'models.CASCADE' },
+        },
+
+        commonPhone: {
+            schemaDoc: 'Phone number where the organization wants to receive common calls',
+            type: Text,
+        },
+
+        ticketSubmittingIsDisabled: {
```
pahaz, line 59: Is ticket submission disabled
or is tickets disabled.

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+                        }
+                        if (resolvedData['commonPhone'] && normalizePhone(resolvedData['commonPhone']) !== resolvedData['commonPhone']) {
+                            throw new GQLError(ERRORS.COMMON_PHONE_INVALID, context)
+                        }
+                    }
+                },
+            },
+        },
+
+        onlyProgressionMeterReadingsIsEnabled: {
```
pahaz, line 77: How to be with meter counter overflow?

pahaz, line 77: Is allow only greater then previous meter reading submission enabled :)

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+            schemaDoc: 'Restricts residents to submit only meter readings, that are greater than previous. ' +
+                'Restrictions apply to mobile phones only.',
+            type: Checkbox,
+        },
+
+        meta: {
```
pahaz, line 83: Let it be. But, probably we want to restrict one miniapp data from another

Alllex202:
pls add description for your pr

apps/condo/domains/settings/components/ticketSubmitting/OnlyProgressionMeterReadingsForm.tsx
```diff
@@ -0,0 +1,126 @@
...
+        initialValues={initialValues}
+        action={action}
+        colon={false}
+        layout='horizontal'
+        formValuesToMutationDataPreprocessor={(values) => {
+            if (!mobileConfig) {
+                values.organization = { connect: { id: userOrganizationId } }
+            }
+
+            values.onlyProgressionMeterReadingsIsEnabled = onlyProgressionMeterReadingsIsEnabled
```
Alllex202, line 73: why don't you just use the value from the form?

You have a form with this field.  When you change it, you change the state above the form.  And here you put the value not from the form, but from the state. Why do that?

Alllex202: you can use `valuePropName` for checkbox

apps/condo/domains/settings/components/ticketSubmitting/OnlyProgressionMeterReadingsSettingCard.tsx
```diff
@@ -0,0 +1,39 @@
+import { MobileFeatureConfig as MobileFeatureConfigType } from '@app/condo/schema'
+import { Typography } from 'antd'
```
Alllex202, line 2: same^^

apps/condo/domains/settings/components/ticketSubmitting/TicketSubmittingSettingCard.tsx
```diff
@@ -0,0 +1,44 @@
+import { MobileFeatureConfig as MobileFeatureConfigType } from '@app/condo/schema'
+import { Typography } from 'antd'
```
Alllex202, line 2: same

apps/condo/domains/settings/components/ticketSubmitting/TicketSubmittingSettingsForm.tsx
```diff
@@ -0,0 +1,186 @@
+import { MobileFeatureConfig as MobileFeatureConfigType } from '@app/condo/schema'
+import { Col, Form, Row, Typography } from 'antd'
```
Alllex202, line 2: Same)

apps/condo/domains/settings/components/ticketSubmitting/TicketSubmittingSettingsForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+    const [ticketSubmittingIsDisabled, setTicketSubmittingIsDisabled] = useState<boolean>()
+
+    const initialValues = {
+        commonPhone: get(mobileConfig, 'commonPhone'),
+        ticketSubmittingIsDisabled: get(mobileConfig, 'ticketSubmittingIsDisabled'),
+    }
+
+    useEffect(() => {
+        if (mobileConfig) {
+            setCommonPhone(mobileConfig.commonPhone)
```
Alllex202, line 63: you can do like this `useState(defaultValue)`

apps/condo/domains/settings/components/ticketSubmitting/TicketSubmittingSettingsForm.tsx
```diff
@@ -0,0 +1,186 @@
...
+                                    <Form.Item
+                                        name='ticketSubmittingIsDisabled'
+                                        label={IsEnabledMessage}
+                                        labelAlign='left'
+                                        {...INPUT_LAYOUT_PROPS}
+                                    >
+                                        <Checkbox
+                                            checked={ticketSubmittingIsDisabled}
+                                            onChange={() => {
+                                                setTicketSubmittingIsDisabled(!ticketSubmittingIsDisabled)
```
Alllex202, line 126: can you use `valuePropName`?

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+const { historical, versioned, uuided, tracked, softDeleted, dvAndSender } = require('@open-condo/keystone/plugins')
+const { GQLListSchema } = require('@open-condo/keystone/schema')
+
+const { normalizePhone } = require('@condo/domains/common/utils/phone')
+const access = require('@condo/domains/settings/access/MobileFeatureConfig')
+
+
+
+const ERRORS = {
+    TICKET_SUBMITTING_PHONES_NOT_CONFIGURED: {
```
Alllex202, line 18: do you need to specify `variable` here?

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
```diff
@@ -0,0 +1,195 @@
...
+                test('can\'t read', async () => {
+                    const client = await makeClient()
+                    await expectToThrowAuthenticationErrorToObjects(async () => {
+                        await MobileFeatureConfig.getAll(client, {}, { sortBy: ['updatedAt_DESC'] })
+                    })
+                })
+            })
+
+
+            describe('user', () => {
```
Alllex202, line 102: Employee?

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
```diff
@@ -0,0 +1,195 @@
...
+
+                    expect(obj.id).toMatch(UUID_RE)
+                    expect(obj.dv).toEqual(1)
+                    expect(obj.sender).toEqual(attrs.sender)
+                    expect(obj.v).toEqual(2)
+                    expect(obj.updatedBy).toEqual(expect.objectContaining({ id: client.user.id }))
+                })
+
+                test('can\'t delete', async () => {
+                    const client = await makeClientWithNewRegisteredAndLoggedInUser()
```
Alllex202, line 141: in other tests you create employees

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
```diff
@@ -0,0 +1,195 @@
...
+
+                    expect(objs).toHaveLength(1)
+                    expect(objs[0]).toMatchObject({
+                        id: objCreated.id,
+                    })
+                })
+            })
+        })
+
+        describe('Validation tests', () => {
```
Alllex202, line 169: Add test for constraints

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
```diff
@@ -0,0 +1,195 @@
...
+describe('MobileFeatureConfig', () => {
+    let admin, commonOrganization
+    beforeAll(async () => {
+        admin = await makeLoggedInAdminClient();
+        [commonOrganization] = await createTestOrganization(admin)
+
+    })
+    describe('CRUD tests', () => {
+        describe('accesses', () => {
+            describe('admin', () => {
```
Alllex202, line 25: Tests for supports?)

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
```diff
@@ -0,0 +1,195 @@
...
+                    expect(objs[0]).toMatchObject({
+                        id: objCreated.id,
+                    })
+                })
+            })
+        })
+
+        describe('Validation tests', () => {
+            test('TICKET_SUBMITTING_PHONES_NOT_CONFIGURED', async () => {
+                await catchErrorFrom(async () => {
```
Alllex202, line 171: seems to have recently fixed a function to catch GQLError

apps/condo/pages/settings/mobileFeatureConfig/onlyProgressionMeterReadings/index.tsx
```diff
@@ -0,0 +1,60 @@
+import { Col, Row, Typography } from 'antd'
```
Alllex202, line 1: Same)

apps/condo/pages/settings/mobileFeatureConfig/onlyProgressionMeterReadings/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    
+    const userOrganization = useOrganization()
+    const userOrganizationId = get(userOrganization, ['organization', 'id'])
+
+    const { obj: mobileConfig, loading } = MobileFeatureConfig.useObject({
+        where: {
+            organization: { id: userOrganizationId },
+        },
+    })
+    return (loading ? <></> :
```
Alllex202, line 30: `!loading && ...`

Alllex202: Or you can add loader

apps/condo/pages/settings/mobileFeatureConfig/ticketSubmitting/index.tsx
```diff
@@ -0,0 +1,60 @@
+import { Col, Row, Typography } from 'antd'
```
Alllex202, line 1: ^^

apps/condo/pages/settings/mobileFeatureConfig/ticketSubmitting/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    
+    const userOrganization = useOrganization()
+    const userOrganizationId = get(userOrganization, ['organization', 'id'])
+
+    const { obj: mobileConfig, loading } = MobileFeatureConfig.useObject({
+        where: {
+            organization: { id: userOrganizationId },
+        },
+    })
+    return (loading ? <></> :
```
Alllex202, line 30: Same^^

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: `MobileFeaturesConfig`?

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: Naming the fields like that, I proceeded from a different logic. Now there are 5 fields needed to configure 2 features. When there are 30 fields, finding the right one will be a difficult task both on the client and on the server. Therefore, I write the feature prefix and then the field itself. That is, the feature `ticketSubmitting` and then what the field stores. If you change `ticketSubmittingIsEnabled` to `ticketSubmission`, then searching for all fields related to the `ticketSubmitting` feature will be impossible and you will have to manually search for everything.

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: Didn't understand what you mean. Haven't heard of counter overflow

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: but google translate says that MobileFeatureConfig is ok)

apps/condo/domains/settings/components/ticketSubmitting/OnlyProgressionMeterReadingsForm.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    const router = useRouter()
+    const [onlyProgressionMeterReadingsIsEnabled, setOnlyProgressionMeterReadingsIsEnabled] = useState<boolean>()
+
+    const initialValues = {
+        onlyProgressionMeterReadingsIsEnabled: get(mobileConfig, 'onlyProgressionMeterReadingsIsEnabled'),
+    }
+
+    useEffect(() => {
+        if (mobileConfig) {
```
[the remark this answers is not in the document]
VKislov: Yes, I think it's possible. This is the rest of the code after refactoring

apps/condo/domains/settings/components/ticketSubmitting/OnlyProgressionMeterReadingsForm.tsx
VKislov: This is a bug fix. For some reason, the submit form had the previous value. Using the value from the state here solves the problem

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
VKislov: I don't see the point in this

VKislov:
> pls add description for your pr

The title has a task. You need to open it task and there is all the context / links / pictures. I think writing the same thing in 100 places is a waste of resources

apps/condo/domains/settings/components/ticketSubmitting/OnlyProgressionMeterReadingsForm.tsx
```diff
@@ -0,0 +1,126 @@
+import { MobileFeatureConfig as MobileFeatureConfigType } from '@app/condo/schema'
+import { Col, Form, Row, Typography } from 'antd'
```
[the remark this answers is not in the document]
VKislov: jst `coz autoimport :)

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
Alllex202: Agree that it looks much better. (See example https://github.com/open-condo-software/condo/blob/e301d082f0999be098b61a1f1358b0d739864109/apps/condo/domains/miniapp/schema/B2BAppPermission.test.js#L364)
We also immediately make it clear that we expect `GQLError` and not some kind of error

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
Alllex202: How do you know if something is broken?)

apps/condo/domains/settings/components/ticketSubmitting/OnlyProgressionMeterReadingsForm.tsx
Alllex202: `valuePropName='checked'` doesn't help?

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
VKislov: ok changed it

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: No, I just wrote. Copy-pasted from another domain

apps/condo/domains/settings/components/ticketSubmitting/OnlyProgressionMeterReadingsForm.tsx
VKislov: unfortunately not working :(

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
VKislov: there is no special logic. The support, like the admin, just has full access to this entity. There are tests for the admin. So the support is doing well

apps/condo/domains/settings/schema/MobileFeatureConfig.js
AleX83Xpert: I think @pahaz means that you will not be able to customize other features using this model. See [my questions](https://github.com/open-condo-software/condo/pull/3662#pullrequestreview-1547300382) too.

apps/condo/domains/organization/gql.js
```diff
@@ -13,7 +13,7 @@ const COMMON_FIELDS = 'id dv sender { dv fingerprint } v deletedAt newId created
...
+const ORGANIZATION_EMPLOYEE_ROLE_FIELDS = '{ organization { id } name nameNonLocalized description descriptionNonLocalized statusTransitions canManageOrganization canManageCallRecords canDownloadCallRecords canManageEmployees canInviteNewOrganizationEmployees canManageRoles canManageTicketPropertyHints canManageIntegrations canReadBillingReceipts canReadPayments canManageProperties canManageTickets canManageContacts canManageContactRoles canManageTicketComments canManagePropertyScopes canShareTickets canBeAssignedAsResponsible canBeAssignedAsExecutor canManageMeters canManageMeterReadings ticketVisibilityType canManageBankAccounts canManageBankContractorAccounts canManageBankIntegrationAccountContexts canManageBankIntegrationOrganizationContexts canManageBankTransactions canManageBankAccountReports canManageBankAccountReportTasks canManageBankAccountReports canManageIncidents canManageNewsItems canManageNewsItemTemplates canManageMobileFeatureConfigs id dv sender { dv fingerprint } v createdBy { id name } updatedBy { id name } createdAt updatedAt }'
```
AntonAL, line 16: Are fields `id dv sender { dv fingerprint } v createdBy { id name } updatedBy { id name } createdAt updatedAt` not in COMMON_FIELDS?

VKislov: Oh, I didn't see it, it's an auto-generated code

apps/condo/domains/settings/access/MobileFeatureConfig.js
```diff
@@ -0,0 +1,73 @@
...
+
+    if (user.isAdmin || user.isSupport) return {}
+
+    if (user.type === RESIDENT) {
+        const residents = await find('Resident', { user: { id: user.id } })
+        const organizations = uniq(map(residents, 'organization'))
+
+        if (residents.length > 0) {
+            return {
+                organization: { id_in: organizations },
```
AntonAL, line 26: Organization can be soft-deleted
You can write a query builder `queryOrganization` in apps/condo/domains/organization/utils/accessSchema.js and incapsulate this logic.

apps/condo/domains/organization/gql.js
VKislov: oh wait, this is the OrganizationEmployeeRole model. I just added a field with permissions and didn't even look. Looks like it's some kind of legacy

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: In this case, the resident will call the organization and they will manually enter the testimony through CRM, after which the resident will be able to submit again

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+        type: 'ORGANIZATION_IS_REQUIRED',
+        message: 'The organization is required to create',
+        messageForUser: 'api.organization.MobileFeatureConfig.ORGANIZATION_IS_REQUIRED',
+    },
+}
+
+const MobileFeatureConfig = new GQLListSchema('MobileFeatureConfig', {
+    schemaDoc: 'Manages availability of some features in mobile application,' +
+        ' stores data required in disabled state of a feature.',
+    fields: {
```
AleX83Xpert, line 43: It's not understandable how auxiliary fields belong to feature (enabled/disabled) areas.

apps/condo/domains/settings/components/ticketSubmitting/TicketSubmittingSettingsForm.tsx
```diff
@@ -0,0 +1,177 @@
...
+    const RequiredCommonPhoneMessage = intl.formatMessage({ id: 'pages.auth.register.field.Phone' })
+    const InvalidPhoneMessage = intl.formatMessage({ id: 'global.input.error.wrongMobilePhone' })
+    const ErrorsContainerTitle = intl.formatMessage({ id: 'errorsContainer.requiredErrors' })
+    const CommonPhoneMessage = intl.formatMessage({ id: 'pages.condo.settings.mobileFeatureConfig.submittingPeriod.commonPhoneField' })
+    const IsEnabledMessage = intl.formatMessage({ id: 'pages.condo.settings.mobileFeatureConfig.submittingPeriod.isEnabled' })
+    const MessageAboutFeat = intl.formatMessage({ id: 'pages.condo.settings.mobileFeatureConfig.submittingPeriod.messageAboutFeat' })
+
+    const router = useRouter()
+    const [commonPhone, setCommonPhone] = useState<string>(get(mobileConfig, 'commonPhone'))
+    const [ticketSubmittingIsDisabled, setTicketSubmittingIsDisabled] = useState<boolean>(get(mobileConfig, 'ticketSubmittingIsDisabled'))
```
Alllex202, line 54: Do you really need this state?
It looks like you can remove it

apps/condo/domains/settings/components/ticketSubmitting/OnlyProgressionMeterReadingsForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+}
+
+export const OnlyProgressionMeterReadingsForm: React.FC<ITicketSubmittingSettingsForm> = ({ mobileConfig, userOrganizationId }) => {
+    const intl = useIntl()
+    const SaveMessage = intl.formatMessage({ id: 'Save' })
+    const MessageAboutFeat = intl.formatMessage({ id: 'pages.condo.settings.mobileFeatureConfig.OnlyProgressionMeterReadings.messageAboutFeat' })
+    const EnableMessage = intl.formatMessage({ id: 'pages.condo.settings.mobileFeatureConfig.OnlyProgressionMeterReadings.isEnabled' })
+
+    const router = useRouter()
+    const [onlyGreaterThanPreviousMeterReadingIsEnabled, setonlyGreaterThanPreviousMeterReadingIsEnabled] = useState<boolean>(get(mobileConfig, 'onlyGreaterThanPreviousMeterReadingIsEnabled'))
```
Alllex202, line 44: Same, it looks like you can remove it

apps/condo/domains/settings/schema/MobileFeatureConfig.js
AleX83Xpert: What about using the same prefix for auxiliary field naming?

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
AleX83Xpert: If we decide in the future to restrict access for support users, the tests will not show any errors. For now, we have a separate type of user and have no test for them.

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
```diff
@@ -0,0 +1,217 @@
...
+                        await MobileFeatureConfig.delete(client, 'id')
+                    })
+                })
+
+                test('can read', async () => {
+                    const client = await makeClientWithNewRegisteredAndLoggedInUser()
+                    const [organization] = await createTestOrganization(admin)
+
+                    const [role] = await createTestOrganizationEmployeeRole(admin, organization, {
+                        canManageMobileFeatureConfigs: true,
```
AleX83Xpert, line 152: `canManage` is used for controlling read access?

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+        organization: {
+            schemaDoc: 'Ref to the organization. Also needed for scope management',
+            type: Relationship,
+            ref: 'Organization',
+            isRequired: true,
+            knexOptions: { isNotNullable: true }, // Required relationship only!
+            kmigratorOptions: { null: false, on_delete: 'models.CASCADE' },
+        },
+
+        commonPhone: {
```
Alllex202, line 54: Add validation for this field and couple of tests

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
AleX83Xpert: For now, every employee can read models (according to access rules) despite the `can` rule.

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
```diff
@@ -0,0 +1,217 @@
...
+
+            test('organization uniq constraint', async () => {
+                const [organization] = await createTestOrganization(admin)
+
+                await createTestMobileFeatureConfig(admin, organization, {
+                    commonPhone: undefined,
+                    ticketSubmittingIsDisabled: false,
+                })
+
+                await catchErrorFrom(async () => {
```
AleX83Xpert, line 178: `expectToThrowGQLError`?

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+        messageForUser: 'api.organization.MobileFeatureConfig.ORGANIZATION_IS_REQUIRED',
+    },
+}
+
+const MobileFeatureConfig = new GQLListSchema('MobileFeatureConfig', {
+    schemaDoc: 'Manages availability of some features in mobile application,' +
+        ' stores data required in disabled state of a feature.',
+    fields: {
+
+        organization: {
```
Alllex202, line 45: You must not allow `organization` field to be updated!

Alllex202: You can use `ORGANIZATION_OWNED_FIELD` const

apps/condo/pages/settings/index.tsx
```diff
@@ -47,16 +50,21 @@ const SettingsPage: React.FC = () => {
     const ControlRoomTitle = intl.formatMessage({ id: 'ControlRoom' })
     const PropertyScopeTitle = intl.formatMessage({ id: 'pages.condo.settings.propertyScope.title' })
     const EmployeeRolesTitle = intl.formatMessage({ id: 'EmployeeRoles' })
+    const MobileFeatureConfigTitle = intl.formatMessage({ id: 'pages.condo.settings.barItem.MobileFeatureConfig' })
 
     const hasSubscriptionFeature = hasFeature('subscription')
-
+    const { useFlag } = useFeatureFlags()
+    const hasMobileFeatureConfigurationFeature = useFlag(MOBILE_FEATURE_CONFIGURATION)
```
AleX83Xpert, line 57: Looks like I still have the ability to create a model using graphql API directly.

apps/condo/domains/settings/schema/MobileFeatureConfig.js
Alllex202, line 1: Should this schema be in `organization` domain?

apps/condo/domains/common/components/settings/MobileFeatureConfigContent.tsx
```diff
@@ -0,0 +1,55 @@
...
+    const userOrganizationId = get(userOrganization, ['organization', 'id'])
+    const { obj: mobileConfig, loading } = MobileFeatureConfigAPI.useObject({
+        where: {
+            organization: { id: userOrganizationId },
+        },
+    })
+    const { useFlag } = useFeatureFlags()
+    const hasMobileFeatureConfigurationFeature = useFlag(MOBILE_FEATURE_CONFIGURATION)
+    const hasTicketSubmittingSettingFeature = useFlag(TICKET_SUBMITTING_SETTING)
+    const hasOnlyProgressionMeterReadingsSettingFeature = useFlag(ONLY_PROGRESSION_METER_READINGS_SETTING)
```
AleX83Xpert, line 31: Is every feature controlled by two feature flags combination?

apps/condo/domains/settings/constants.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Generated by `createschema settings.MobileFeatureConfig 'organization:Relationship:Organization:CASCADE; emergencyPhone:Text; commonPhone:Text; onlyGreaterThanPreviousMeterReadingIsEnabled:Checkbox; meta:Json; ticketSubmittingIsEnabled:Checkbox'`
+ */
+
+const MOBILE_FEATURE_CONFIGURATION = 'mobile-feature-configuration'
+const TICKET_SUBMITTING_SETTING = 'ticket-submitting-from-resident-mobile-app'
+const ONLY_PROGRESSION_METER_READINGS_SETTING = 'submit-only-progression-meter-readings'
```
AleX83Xpert, line 7: Maybe move these constants near to others?
apps/condo/domains/common/constants/featureflags.js

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+            schemaDoc: 'Phone number where the organization wants to receive common calls',
+            type: Text,
+        },
+
+        ticketSubmittingIsDisabled: {
+            schemaDoc: 'Disabling the ability to send tickets by the user of the resident\'s mobile application. ' +
+                'Instead, he will be shown a screen with phones where he can contact to send a request',
+            type: Checkbox,
+            hooks: {
+                validateInput: async ({ resolvedData, context }) => {
```
Alllex202, line 64: Your validation will not work if you follow these steps:
  1) Specify phone and this flag
  2) Delete phone

So move this validation in schema hooks

apps/condo/pages/settings/mobileFeatureConfig/onlyProgressionMeterReadings/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    return (!loading &&
+        <>
+            <PageHeader title={<Typography.Title>{PageTitle}</Typography.Title>} />
+            <Row gutter={ROW_GUTTER}>
+                <Col span={24}>
+                    <OnlyProgressionMeterReadingsForm mobileConfig={mobileConfig} userOrganizationId={userOrganizationId}/>
+                </Col>
+            </Row>
+        </>
+    )
```
AleX83Xpert, line 39:
```suggestion
    return !loading && (
        <>
            <PageHeader title={<Typography.Title>{PageTitle}</Typography.Title>} />
            <Row gutter={ROW_GUTTER}>
                <Col span={24}>
                    <OnlyProgressionMeterReadingsForm mobileConfig={mobileConfig} userOrganizationId={userOrganizationId}/>
                </Col>
            </Row>
        </>
    )
```

apps/condo/pages/settings/mobileFeatureConfig/ticketSubmitting/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    return (!loading &&
+        <>
+            <PageHeader title={<Typography.Title>{PageTitle}</Typography.Title>} />
+            <Row gutter={ROW_GUTTER}>
+                <Col span={24}>
+                    <TicketSubmittingSettingsForm mobileConfig={mobileConfig} userOrganizationId={userOrganizationId}/>
+                </Col>
+            </Row>
+        </>
+    )
```
AleX83Xpert, line 39:
```suggestion
    return !loading && (
        <>
            <PageHeader title={<Typography.Title>{PageTitle}</Typography.Title>} />
            <Row gutter={ROW_GUTTER}>
                <Col span={24}>
                    <TicketSubmittingSettingsForm mobileConfig={mobileConfig} userOrganizationId={userOrganizationId}/>
                </Col>
            </Row>
        </>
    )
```

apps/condo/pages/settings/mobileFeatureConfig/onlyProgressionMeterReadings/index.tsx
AleX83Xpert: What the user will see while `loading===true`?

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,111 @@
...
+            },
+        },
+
+        onlyGreaterThanPreviousMeterReadingIsEnabled: {
+            schemaDoc: 'Restricts residents to submit only meter readings, that are greater than previous. ' +
+                'Restrictions apply to mobile phones only.',
+            type: Checkbox,
+        },
+
+        meta: {
```
Alllex202, line 83: How is this field going to be used for miniapps and external sources? Shouldn't it be on their side?

Alllex202: By whom and how will this field be used if only employees and residents can read this scheme?

apps/condo/domains/settings/schema/MobileFeatureConfig.js
Alllex202: Also please add a test for this case

apps/condo/domains/settings/access/MobileFeatureConfig.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    return {
+        organization: {
+            OR: [
+                queryOrganizationEmployeeFor(user.id),
+                queryOrganizationEmployeeFromRelatedOrganizationFor(user.id),
+            ],
+        },
+    }
```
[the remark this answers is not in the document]
VKislov: Yes, and I think that's right. The `organization` settings page is available to everyone at the moment. Special rights are only used to check editing rights. I think it will be correct to repeat the behavior of the other sections of the `organization` settings

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: This data is necessary for the mobile application to determine what to display. In the case of the feature of disabling tickets - displaying the desired phone.
If you are describing the full changes that will occur in the mobile app when you disable/enable a feature, it will mean that this place must be maintained when changes are made in the mobile app, otherwise this data source will mislead the backend developer.

VKislov: >> What about using the same prefix for auxiliary field naming?

that's how it happens. All fields required for feature management are marked with a prefix. That is why the `ticketSubmittingIsDisabled` field is called so incorrectly (from the point of view of the English language). Pahaz asked for some fields, for example, the org phone, which will be reused with a high probability, not to attach to a specific feature, but to leave commonPhone

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
VKislov: Yes, you're right, most likely mechanically wrote.

apps/condo/pages/settings/index.tsx
VKislov: this is true

apps/condo/domains/common/components/settings/MobileFeatureConfigContent.tsx
VKislov: Yes, this is a business requirement. It is necessary that the org manages this itself. But at the same time, we do not want every user to see all the settings. Since we are developing them, not so that they would be turned off, but so that they would be used. This task is a rare case when it is necessary

apps/condo/domains/settings/constants.js
VKislov: Ok I will move these constants. I have never heard of such a rule, but it seems that many people follow it.

apps/condo/pages/settings/mobileFeatureConfig/onlyProgressionMeterReadings/index.tsx
VKislov: most likely a page with no content

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: No, we agreed to create a separate domain

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: yes, you are right. After refactoring, I did not think to change. This was originally done on purpose.

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: Yes, such validation is present only on the client side. I don't know if this validation is needed on the server. Might be worth adding

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: In the `meta`, they often write various auxiliary information. schemaDoc was just copied

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
VKislov: > If we decide in the future to restrict access for support users, the tests will not show any errors. For now, we have a separate type of user and have no test for them.

Hmm, the support now has the same read/modify rights as the admin. There is no other special logic for the support. All that can change is that rights can be taken away. But then nothing breaks. I don't think anyone does it by accident.

apps/condo/domains/settings/schema/MobileFeatureConfig.js
Alllex202: Yes, of course you need it :) You can change it without client

apps/condo/domains/settings/schema/MobileFeatureConfig.js
Alllex202: So you should update `schemaDoc`)

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
VKislov: It seems that this is not a `GQLError`, this is an error that the database throws. The `constraint` error is most likely the only way to catch

Alllex202: you can use `expectToThrowUniqueConstraintViolationError`

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: Yes, good point, thanks. Moved this to schema validation, not field validation)

apps/condo/domains/settings/schema/MobileFeatureConfig.test.js
VKislov: added tests for support

apps/condo/domains/common/components/settings/MobileFeatureConfigContent.tsx
```diff
@@ -0,0 +1,55 @@
...
+    OnlyProgressionMeterReadingsSettingCard,
+} from '@condo/domains/settings/components/ticketSubmitting/OnlyProgressionMeterReadingsSettingCard'
+import { TicketSubmittingSettingCard } from '@condo/domains/settings/components/ticketSubmitting/TicketSubmittingSettingCard'
+import { MobileFeatureConfig as MobileFeatureConfigAPI } from '@condo/domains/settings/utils/clientSchema'
+
+const CONTENT_GUTTER: Gutter | [Gutter, Gutter] = [0, 40]
+
+export const MobileFeatureConfigContent: React.FC = () => {
+    const userOrganization = useOrganization()
+    const userOrganizationId = get(userOrganization, ['organization', 'id'])
```
Alllex202, line 22:
```suggestion
    const userOrganizationId = get(userOrganization, ['organization', 'id'], null)
```
you should set it to null by default so that request returns nothing

apps/condo/domains/common/constants/featureflags.js
```diff
@@ -12,6 +12,9 @@ const PROPERTY_REPORT_DELETE_ENTITIES = 'property-report-delete-entities'
 const SERVICE_PROVIDER_PROFILE = 'service-provider-profile'
 const SHOW_ORGANIZATION_TYPES = 'show-organization-types'
 const ENABLE_DISCOVER_SERVICE_CONSUMERS = 'enable-discover-service-consumers'
+const MOBILE_FEATURE_CONFIGURATION = 'mobile-feature-configuration'
+const TICKET_SUBMITTING_SETTING = 'ticket-submitting-from-resident-mobile-app'
+const ONLY_PROGRESSION_METER_READINGS_SETTING = 'submit-only-progression-meter-readings'
```
Alllex202, line 17: Make name of constant same as its value.
This is done in other constants

apps/condo/domains/settings/access/MobileFeatureConfig.js
```diff
@@ -0,0 +1,74 @@
...
+const { RESIDENT } = require('@condo/domains/user/constants/common')
+
+async function canReadMobileFeatureConfigs ({ authentication: { item: user } }) {
+    if (!user) return throwAuthenticationError()
+    if (user.deletedAt) return false
+
+    if (user.isAdmin || user.isSupport) return {}
+
+    if (user.type === RESIDENT) {
+        const residents = await find('Resident', { user: { id: user.id } })
```
Alllex202, line 21: do we need to add `deletedAt: null` filter?

apps/condo/domains/settings/access/MobileFeatureConfig.js
```diff
@@ -0,0 +1,74 @@
...
+        const organizations = uniq(map(residents, 'organization'))
+
+        if (residents.length > 0) {
+            return {
+                organization: { id_in: organizations },
```
Alllex202, line 26: and for organization?

apps/condo/domains/settings/access/MobileFeatureConfig.js
```diff
@@ -0,0 +1,74 @@
...
+    if (user.deletedAt) return false
+    if (user.type === RESIDENT) return false
+    if (user.isAdmin || user.isSupport) return true
+
+    let organizationId
+    if (operation === 'create') {
+        organizationId = get(originalInput, 'organization.connect.id')
+    }
+    if ( operation === 'update') {
+        if (!itemId) throw new Error('no itemId')
```
Alllex202, line 55: why not just return false? `if (!itemId) return false`

apps/condo/domains/settings/components/ticketSubmitting/TicketSubmittingSettingsForm.tsx
```diff
@@ -0,0 +1,172 @@
...
+    const ExamplePhoneMessage = intl.formatMessage({ id: 'example.Phone' })
+    const RequiredCommonPhoneMessage = intl.formatMessage({ id: 'pages.auth.register.field.Phone' })
+    const InvalidPhoneMessage = intl.formatMessage({ id: 'global.input.error.wrongMobilePhone' })
+    const ErrorsContainerTitle = intl.formatMessage({ id: 'errorsContainer.requiredErrors' })
+    const CommonPhoneMessage = intl.formatMessage({ id: 'pages.condo.settings.mobileFeatureConfig.submittingPeriod.commonPhoneField' })
+    const IsEnabledMessage = intl.formatMessage({ id: 'pages.condo.settings.mobileFeatureConfig.submittingPeriod.isEnabled' })
+    const MessageAboutFeat = intl.formatMessage({ id: 'pages.condo.settings.mobileFeatureConfig.submittingPeriod.messageAboutFeat' })
+
+    const router = useRouter()
+    const [commonPhone, setCommonPhone] = useState<string>(get(mobileConfig, 'commonPhone'))
```
Alllex202, line 53: What is this state for? Isn't it stored in a form?

apps/condo/domains/settings/constants.js
Alllex202, line 1: Don't want to delete this empty file? :)

apps/condo/domains/settings/schema/MobileFeatureConfig.js
Alllex202: add `variable: ['data', 'commonPhone'],`

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -0,0 +1,97 @@
...
+    },
+}
+
+const MobileFeatureConfig = new GQLListSchema('MobileFeatureConfig', {
+    schemaDoc: 'Manages availability of some features in mobile application,' +
+        ' stores data required in disabled state of a feature.',
+    fields: {
+        organization: ORGANIZATION_OWNED_FIELD,
+
+        commonPhone: {
```
Alllex202, line 47: we have const `PHONE_FIELD`, use it

Alllex202, line 47: Also in `Organization` schema have `phone` field. 
Can we take it from there?

apps/condo/domains/settings/utils/MobileFeatureConfigSchemaValidations.js
```diff
@@ -0,0 +1,18 @@
+const { GQLError } = require('@open-condo/keystone/errors')
+
+const { normalizePhone } = require('@condo/domains/common/utils/phone')
+
+const ticketSubmittingValidations = async (resolvedData, context, ERRORS) => {
+    if (resolvedData.ticketSubmittingIsDisabled) {
```
Alllex202, line 6: You need to check existing value as well

apps/condo/pages/settings/mobileFeatureConfig/onlyProgressionMeterReadings/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { MobileFeatureConfig } from '@condo/domains/settings/utils/clientSchema'
+
+const ROW_GUTTER: [Gutter, Gutter] = [0, 60]
+
+const OnlyProgressionMeterReadingsContent: React.FC = () => {
+    const intl = useIntl()
+    const PageTitle = intl.formatMessage({ id: 'pages.condo.settings.barItem.MobileFeatureConfig.OnlyProgressionMeterReadings.title' })
+    
+    const userOrganization = useOrganization()
+    const userOrganizationId = get(userOrganization, ['organization', 'id'])
```
Alllex202, line 23:
```suggestion
    const userOrganizationId = get(userOrganization, ['organization', 'id'], null)
```

apps/condo/pages/settings/mobileFeatureConfig/onlyProgressionMeterReadings/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    const { obj: mobileConfig, loading } = MobileFeatureConfig.useObject({
+        where: {
+            organization: { id: userOrganizationId },
+        },
+    })
+    return !loading && (
```
Alllex202, line 30: Do we need to show loader?

apps/condo/pages/settings/mobileFeatureConfig/ticketSubmitting/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { MobileFeatureConfig } from '@condo/domains/settings/utils/clientSchema'
+
+const ROW_GUTTER: [Gutter, Gutter] = [0, 60]
+
+const TicketSubmittingContent: React.FC = () => {
+    const intl = useIntl()
+    const PageTitle = intl.formatMessage({ id: 'pages.condo.settings.mobileFeatureConfig.submittingPeriod.pageTitle' })
+    
+    const userOrganization = useOrganization()
+    const userOrganizationId = get(userOrganization, ['organization', 'id'])
```
Alllex202, line 23:
```suggestion
    const userOrganizationId = get(userOrganization, ['organization', 'id'], null)
```

apps/condo/pages/settings/mobileFeatureConfig/ticketSubmitting/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    const { obj: mobileConfig, loading } = MobileFeatureConfig.useObject({
+        where: {
+            organization: { id: userOrganizationId },
+        },
+    })
+    return !loading && (
```
Alllex202, line 30: Do we need to show loader?

apps/condo/domains/settings/access/MobileFeatureConfig.js
VKislov: Yes, you can return `false`. It just seemed that the error would make the behavior more predictable. Because the absence of `itemId` when trying to update an entity, returning `false` is more like a lack of access and not an error

apps/condo/domains/settings/constants.js
VKislov: For what. The next person to add a model to this domain will also create this file, you will need it later anyway. But I can remove it, now I don't need it

apps/condo/domains/settings/schema/MobileFeatureConfig.js
VKislov: in the `organization` the field is used by the `telephony` miniapp. Not sure if it can be reused

apps/condo/pages/settings/mobileFeatureConfig/onlyProgressionMeterReadings/index.tsx
VKislov: I think not, that's why I didn't add it)

apps/condo/pages/settings/mobileFeatureConfig/ticketSubmitting/index.tsx
VKislov: I think not, that's why I didn't add it)

apps/condo/domains/settings/schema/MobileFeatureConfig.js
```diff
@@ -45,6 +47,7 @@ const MobileFeatureConfig = new GQLListSchema('MobileFeatureConfig', {
         organization: ORGANIZATION_OWNED_FIELD,
 
         commonPhone: {
+            ...PHONE_FIELD,
             schemaDoc: 'Phone number where the organization wants to receive common calls',
             type: Text,
```
Alllex202, line 52: `type: Text`?  Isn't this in `PHONE_FIELD`? :)

apps/condo/domains/settings/utils/MobileFeatureConfigSchemaValidations.js
```diff
@@ -1,9 +1,11 @@
...
+
 const { GQLError } = require('@open-condo/keystone/errors')
 
 const { normalizePhone } = require('@condo/domains/common/utils/phone')
 
-const ticketSubmittingValidations = async (resolvedData, context, ERRORS) => {
-    if (resolvedData.ticketSubmittingIsDisabled) {
+const ticketSubmittingValidations = async (resolvedData, context, existingItem, ERRORS) => {
+    if (resolvedData.ticketSubmittingIsDisabled || get(existingItem, 'ticketSubmittingIsDisabled')) {
         if (!resolvedData['commonPhone']) {
```
Alllex202, line 9: In case, if you have already specified a phone without `ticketSubmittingIsDisabled` flag and then you specify the flag, then there will be an error.

Alllex202: you can write something like this `const newItem = { ...existingItem, ...resolvedData }` and use `newItem` in all conditions

apps/condo/pages/settings/mobileFeatureConfig/onlyProgressionMeterReadings/index.tsx
```diff
@@ -36,7 +37,7 @@ const OnlyProgressionMeterReadingsContent: React.FC = () => {
                 </Col>
             </Row>
         </>
-    )
+    ) || <Loader fill size='small'/>
```
Alllex202, line 40: some strange code) It's better to put that loader in a condition before the main return)

apps/condo/pages/settings/mobileFeatureConfig/ticketSubmitting/index.tsx
```diff
@@ -36,7 +37,7 @@ const TicketSubmittingContent: React.FC = () => {
                 </Col>
             </Row>
         </>
-    )
+    ) || <Loader fill size='small'/>
```
Alllex202, line 40: Same ^^

apps/condo/pages/settings/mobileFeatureConfig/onlyProgressionMeterReadings/index.tsx
VKislov: but it's just a logical expression. if `loading === true` => `Loader` if `loading === false` => content

Alllex202: Yes, it's same. But less readable)

Alllex202:
Why don't you add this config automatically for all organizations? It's a 1 to 1 relationship, why not?

Alllex202 (approved):

apps/condo/domains/settings/access/MobileFeatureConfig.js
```diff
@@ -0,0 +1,77 @@
...
+const { RESIDENT } = require('@condo/domains/user/constants/common')
+
+async function canReadMobileFeatureConfigs ({ authentication: { item: user } }) {
+    if (!user) return throwAuthenticationError()
+    if (user.deletedAt) return false
+
+    if (user.isAdmin || user.isSupport) return {}
+
+    if (user.type === RESIDENT) {
+        const residents = await find('Resident', { user: { id: user.id }, deletedAt: null })
```
pahaz, line 21: You need to chach others models. We need to get organizations from ServiceConsumer

apps/condo/domains/common/components/settings/MobileFeatureConfigContent.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { MOBILE_FEATURE_CONFIGURATION, TICKET_SUBMITTING_FORM_RESIDENT_MOBILE_APP, SUBMIT_ONLY_PROGRESSION_METER_READINGS } from '@condo/domains/common/constants/featureflags'
+import {
+    OnlyProgressionMeterReadingsSettingCard,
+} from '@condo/domains/settings/components/ticketSubmitting/OnlyProgressionMeterReadingsSettingCard'
+import { TicketSubmittingSettingCard } from '@condo/domains/settings/components/ticketSubmitting/TicketSubmittingSettingCard'
+import { MobileFeatureConfig as MobileFeatureConfigAPI } from '@condo/domains/settings/utils/clientSchema'
+
+const CONTENT_GUTTER: Gutter | [Gutter, Gutter] = [0, 40]
+
+export const MobileFeatureConfigContent: React.FC = () => {
```
pahaz, line 20: Confusing naming. It's look like some guard or some wrapper for feature flag checks.

It will be better to create some wraper or HOC for this task. For example:

```
const CardsContainer = ...

const CardsContainerWithAccessCheck = addFeatureFlagIsEnabled(CardsContainer, 'MOBILE_FEATURE')

```

PR #3666, opened 2023-07-27
DOMA-6733 one file in many call records

nomerdvadcatpyat:
When you save a conversation record for each organization that has the same `phone` as the conversation participant, the `CallRecord` object is created and a separate file is created in cloud for each `CallRecord`.
The real case: our callcenter has 14 organizations with one `phone` number. This means that for every conversation that is not attached to the ticket, 14 copies of the conversation files will be created. It can take up a lot of space.
The solution is to have the same file attached to such `CallRecord` objects.

1) Added `saveFileName` to `SberCloudAdapter` – if it's `true`, then original file name will be saved and if such file already exist in cloud, it will not be saved again
2) Pass `CallRecord` id in file meta `ids` field in `CallRecord.afterChange` hook when a new `CallRecord` is created; check access to file based on object ids from file meta `ids` field in `obsRouterHandler`

### DIFF
--- a/apps/condo/domains/common/utils/fileAdapter.js
+++ b/apps/condo/domains/common/utils/fileAdapter.js
@@ -79,10 +79,16 @@ class FileAdapter {
         if (!this.isConfigValid(conf, ['MEDIA_ROOT', 'MEDIA_URL', 'SERVER_URL'])) {
             return null
         }
-        return new LocalFileAdapter({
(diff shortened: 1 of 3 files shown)


toplenboren (approved):

Alllex202 (approved):
Should we do the same for `LocalFileApapter`?

PR #3669, opened 2023-07-28
DOMA-6615 Add ability to get list dependent relations

toplenboren:
A follow up to the previous task with the same key. 

We need a way to get all relations which depend on certain list.

Like:
BillingIntegrationOrgamnizationContext -> BillingProperty, BillingReceipt, BillingAccount

### DIFF
--- a/apps/condo/domains/organization/schema/Organization.test.js
+++ b/apps/condo/domains/organization/schema/Organization.test.js
@@ -64,8 +64,6 @@ describe('Organization', () => {
             test('Support can create organization directly', async () => {
                 const [obj, attrs] = await createTestOrganization(support)
 
-                const x = getListDependentRelations('BillingIntegrationOrganizationContext')
-
                 expect(obj.id).toMatch(UUID_RE)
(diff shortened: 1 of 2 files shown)


apps/condo/domains/organization/schema/Organization.test.js
```diff
@@ -4,6 +4,7 @@
 const { faker } = require('@faker-js/faker')
 const dayjs = require('dayjs')
 
+const { getListDependentRelations } = require('@open-condo/keystone/schema')
```
AleX83Xpert, line 7: Is this function used somewhere in this file? I didn't find it.

toplenboren: No, I will use this function when implementing restricting deletion of PROTECTED fields and cascade deletion

toplenboren: Its just a big PR, so I decided to split it to a lot of small increments

AleX83Xpert (approved):

sitozzz (approved):

PR #3671, opened 2023-07-31
feat(condo): DOMA-6784 B2BAppPermission name field added

SavelevMatthew:
image


apps/condo/domains/miniapp/schema/B2BAppPermission.js
```diff
@@ -58,7 +67,29 @@ const B2BAppPermission = new GQLListSchema('B2BAppPermission', {
...
+            hooks: {
+                resolveInput: ({ resolvedData, fieldPath }) => {
+                    if (fieldPath in resolvedData) {
+                        return resolvedData[fieldPath].trim()
+                    }
 
+                    return undefined
+                },
+                validateInput: ({ resolvedData, fieldPath, context }) => {
+                    if (resolvedData[fieldPath].length > MAX_PERMISSION_NAME_LENGTH) {
```
Alllex202, line 87: Is there a minimum length?

SavelevMatthew: Nope, there's not.

I've named this field `description` previously, but then I thought that it was too unclear, so I used field `shortDescription`. 

Then I thought, what is short? So I've added a limit.

And just before pushing when I was looking at designs again I thought that `name` option is the clearest one. But left old limitation :)

Alllex202: That is, I can make an empty name? What will the user see in the interface?

apps/condo/domains/miniapp/schema/B2BAppPermission.js
```diff
@@ -19,6 +23,11 @@ const ERRORS = {
         type: PERMISSION_KEY_WRONG_FORMAT_ERROR,
         message: 'Incorrect key format. The key must start with the prefix "can", have lowerCamelCase and answer the question: "what is allowed to the user with this key?". Example: canManagePasses, canReadConfig, etc.',
     },
+    TOO_LONG_SHORT_DESCRIPTION: {
```
Alllex202, line 26: Сan we name this error according to its type?)

apps/condo/domains/miniapp/schema/B2BAppPermission.js
```diff
@@ -19,6 +23,11 @@ const ERRORS = {
         type: PERMISSION_KEY_WRONG_FORMAT_ERROR,
         message: 'Incorrect key format. The key must start with the prefix "can", have lowerCamelCase and answer the question: "what is allowed to the user with this key?". Example: canManagePasses, canReadConfig, etc.',
```
AntonAL, line 24: There is a `canManagePasses` field for example values.
ADR: https://confluence.in.doma.ai/pages/viewpage.action?pageId=967573505

```suggestion
        message: 'Incorrect key format. The key must start with the prefix "can", have lowerCamelCase and answer the question: "what is allowed to the user with this key?"',
        correctExample: 'canManagePasses'
```

SavelevMatthew: `correctExample` is too strange. I have not seen it in any API I've worked with

apps/condo/domains/miniapp/schema/B2BAppPermission.js
SavelevMatthew: Refactored validation :)

Alllex202 (approved):

PR #3673, opened 2023-07-31
fix(condo): DOMA-6716 filter deleted users in query for ticket assigments

nomerdvadcatpyat:
No need to set employees with deleted users as executor or assignee in ticket

### DIFF
--- a/apps/condo/domains/ticket/components/BaseTicketForm/TicketAssignments.tsx
+++ b/apps/condo/domains/ticket/components/BaseTicketForm/TicketAssignments.tsx
@@ -69,7 +69,13 @@ const TicketAssignments = ({
     const { assignee: initialAssignee, executor: initialExecutor } = form.getFieldsValue(['assignee', 'executor'])
 
     const { objs: employees, allDataLoaded: allEmployeesLoaded } = OrganizationEmployee.useAllObjects({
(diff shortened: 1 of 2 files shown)


Alllex202 (approved):

apps/condo/domains/ticket/components/BaseTicketForm/TicketAssignments.tsx
```diff
@@ -69,7 +69,13 @@ const TicketAssignments = ({
     const { assignee: initialAssignee, executor: initialExecutor } = form.getFieldsValue(['assignee', 'executor'])
 
     const { objs: employees, allDataLoaded: allEmployeesLoaded } = OrganizationEmployee.useAllObjects({
-        where: { organization: { id: organizationId }, isBlocked: false, isRejected: false },
+        where: {
+            organization: { id: organizationId },
+            user: { deletedAt: null },
+            deletedAt: null,
```
sitozzz, line 75: Isn't `deletedAt` prop passed to the hook by default?

nomerdvadcatpyat: if we manually set deletedAt (in user), then it's will stop being passed automatically

PR #3681, opened 2023-08-02
fix(condo): DOMA-6735 change location of organization name message in phone page

nomerdvadcatpyat:
изображение


sitozzz (approved):

Alllex202 (approved):
Do you need to change in cc?

nomerdvadcatpyat:
> Do you need to change in cc?

cc reuse this component (also organization name message shows only in cc)

PR #3683, opened 2023-08-02
INFRA-92 : Proof of work: CLS for NodeJS, thanks node16 <3

toplenboren:
# CLS for NodeJS

Here we implement something like thread local context (check c#) for NodeJS. Turns out, node already has this tool and its called `continuation local storage` 

Continuation Local Storage is a tool to save and transfer data in async context.

Right now we use it to retrieve current web request ID in database adapter (inside cache module) 

-----

Here is the proof that it works (Taken from Proof of concept)

Screenshot 2023-08-03 at 04 07 46

Screenshot 2023-08-03 at 04 07 56

> The whole trick is that `adapterCache` or `adapter-knex.js` do not know anything about `express` context, but instead get request id from global async storage

----

Docs: https://nodejs.org/dist/latest-v16.x/docs/api/async_context.html#new-asynclocalstorage

Note: this is just a proof of concept, and pr is needed to start a conversation about the topic :-) so don't really review it. I will rework it :-)


packages/keystone/asyncLocalStorage.js
```diff
@@ -0,0 +1,27 @@
+const { AsyncLocalStorage } = require('node:async_hooks')
+
+const { getLogger } = require('./logging')
+
+const ASYNC_LOCAL_STORAGES = {}
```
sitozzz, line 5: Why do you need multiple copies of AsyncLocalStorage? You can't store all of your data inside single instance?

sitozzz (approved):

packages/keystone/asyncLocalStorage.js
toplenboren: You can technically store everything in single ALS, but I believe it would be better to split them to different objects

packages/keystone/asyncLocalStorage.js
```diff
@@ -0,0 +1,44 @@
...
+const logger = getLogger('asyncLocalStorage')
+
+/**
+ * Return current execution context
+ */
+function getExecutionContext () {
```
pahaz, line 11: why do we need more than one context? it's easy to get executionContext as an object.
```
const ctx = getExecutionContext ()
console.log(ctx.reqId, ctx.taskId, ctx.execId) // one should be not null!
```

pahaz:
```

if (ctx.reqId) {
  console.log(`we are in HTTP context! user =`, ctx.userId, `fingerprint =`, ctx.fingerprint, ctx.startedAt)
} else if (ctx.taskId) {
  console.log(`we are in worker task! scheduled by processing req =`, ctx.byReqId, ctx.startedAt)
} else {
  console.log(`we are in some script`, ctx.execArgv, ctx.startedAt)
}
```

packages/keystone/asyncLocalStorage.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * This function is intended to be used in internal modules only, like prepareKeystone.js
+ * Use this function if you need a specific asyncLocalStorage instance in project
+ *
+ * @param {string} name -- name of storage
+ * @returns {import('node:async_hooks').AsyncLocalStorage}
+ * @deprecated for any external usage!
+ */
+function internalGetAsyncLocalStorage (name = 'default') {
```
pahaz, line 30: add `_` to mark it as private

packages/keystone/adapterCache.js
```diff
@@ -189,15 +190,17 @@ class AdapterCache {
...
         if (!this.logging) return
 
         logger.info({
             type,
             functionName,
             listName,
             key,
+            reqId,
             result,
+            context: getExecutionContext(),
```
pahaz, line 203: ??

packages/keystone/KSv5v6/v5/prepareKeystone.js
```diff
@@ -38,6 +39,8 @@ function prepareKeystone ({ onConnect, extendExpressApp, schemas, schemasPreproc
     if (tasks && typeof tasks !== 'function') throw new Error('tasks should be a function like `() => [ Task ]`')
     if (apps && typeof apps !== 'function') throw new Error('apps should be a function like `() => [ App | Middleware ]`')
 
+    _internalGetExecutionContextAsyncLocalStorage().enterWith( { prepareKeystone: { id: v4(), argv: process.argv, execArgv: process.execArgv } })
```
pahaz, line 42: it's better to move it to [executionContext.js](https://github.com/open-condo-software/condo/pull/3683/files#diff-2b515060d90d3c4b6d869ad52f19005af586aa47dfbdf4973b349fd39a359854)

pahaz: It's better to use `exec` key for that

pahaz (approved):

packages/keystone/adapterCache.js
```diff
@@ -198,6 +199,7 @@ class AdapterCache {
             listName,
             key,
             result,
+            context: getExecutionContext(),
```
pahaz, line 202:
```suggestion
            ...getExecutionContext(),
```

PR #3684, opened 2023-08-03
feat(condo): DOMA-6776 add primaryFile field to BillingReceipt

ekabardinsky:


apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -176,6 +176,14 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
             },
         },
 
+        primaryFile: {
```
pahaz, line 179: Why it’s primary? We will have a secondary file?

ekabardinsky: probably in future we can have other files related to the billingReceipt, but one of them going to be "primary" one.

dkoviazin: There will be no other fields here. Better, to rename it

ekabardinsky: renamed to "file"

apps/condo/domains/billing/schema/AllResidentBillingReceiptsService.test.js
```diff
@@ -89,6 +89,86 @@ describe('AllResidentBillingReceiptsService', () => {
...
+            // set primary file
+            await updateTestBillingReceipt(integrationClient, receipt.id, {
+                file: { connect: { id: receiptFile.id } },
+            })
+
+            // getting the result
+            const residentBillingReceipts = await ResidentBillingReceipt.getAll(residentUser)
+
+            expect(residentBillingReceipts).toBeDefined()
+            expect(residentBillingReceipts).not.toHaveLength(0)
```
AleX83Xpert, line 133: Is the exact number not known?

AleX83Xpert (commented):
Didn't find a test for public/sensitive files.

apps/condo/domains/billing/schema/AllResidentBillingReceiptsService.test.js
ekabardinsky: Looks like not, since this is an exact copy of another test with a little changed set up

AleX83Xpert (approved):

PR #3686, opened 2023-08-03
DOMA-6786 fix number filter

nomerdvadcatpyat:
**Probem:** when user searching phone number in global search (on ticket/callRecord and other pages), then gql request fails because out of the 32bit integer fieilds error (like `Ticket.number` or `CallRecord.talkTime`)
изображение

**Solve:** Added validation for number filter. Now if search data out of the 32bit integer bound, then such fields not used in filter query:
изображение


apps/condo/domains/common/utils/tables.utils.ts
```diff
@@ -124,7 +124,7 @@ export const getFilter: (
         }
         switch (argData) {
             case 'number':
-                args = search.filter(value => Number(value) && String(value).indexOf('.') === -1).map(Number)
+                args = search.filter(value => Number(value) && is32BitInteger(value) && String(value).indexOf('.') === -1).map(Number)
```
toplenboren, line 127: I believe you should be able to set this up, because you may be using bigint or other types

nomerdvadcatpyat:
> This is really strange that user should care for INT32 size :-/

What user should care about it? User just typing in search input, if he pass in the search in global search something like '79995666620' – he is clearly not looking for the ticket number (and even if he is looking, we definitely do not have one, because `Ticket.number` is 32bit integer field).

> I think it would be better to also show an error, like, your number is too big, or use a string and _i to search (for phone number, for example)

`contains_i` is already used for the phone number and other `string` fields, for `Integer` fields we don't have `contains_i`.

toplenboren:
> > This is really strange that user should care for INT32 size :-/
> 
> What user should care about it? User just typing in search input, if he pass in global search something like '79995666620' – he is clearly not looking for the ticket number (and even if he is looking, we definitely do not have one, because `Ticket.number` is 32bit integer field).
> 
> > I think it would be better to also show an error, like, your number is too big, or use a string and _i to search (for phone number, for example)
> 
> `contains_i` is already used for the phone number and other `string` fields, for `Integer` fields we don't have `contains_i`.

Gotcha. Anyways, I still find implicitly removing these fields from the search a strange solution

PR #3694, opened 2023-08-07
feat(condo): DOMA-6796 make miniapps accessible to all 🤟😎

SavelevMatthew:


apps/condo/domains/miniapp/schema/AllMiniAppsService.js
```diff
@@ -3,108 +3,147 @@
...
+            type: 'input AllMiniAppsInput { dv: Int!, sender: SenderFieldInput!, organization: OrganizationWhereUniqueInput!, where: AllMiniAppsWhereInput, sortBy: SortAllMiniAppsBy }',
         },
         {
             access: true,
-            type: 'type MiniAppOutput { id: ID!, type: AppType!, connected: Boolean!, name: String!, shortDescription: String!, category: AppCategory!, logo: String, label: String }',
+            type: 'type MiniAppOutput { id: ID!, connected: Boolean!, accessible: Boolean!, name: String!, shortDescription: String!, category: AppCategory!, logo: String, label: String, icon: String, menuCategory: String }',
         },
     ],
     
     queries: [
```
Alllex202, line 80: Write at least in `doc` that only `B2BApp`s are returned here :)

apps/condo/domains/miniapp/schema/B2BAppContext.js
```diff
@@ -6,14 +6,59 @@ const { Relationship } = require('@keystonejs/fields')
...
 const access = require('@condo/domains/miniapp/access/B2BAppContext')
 const { NO_CONTEXT_STATUS_ERROR, CONTEXT_FINISHED_STATUS } = require('@condo/domains/miniapp/constants')
 const { STATUS_FIELD, getStatusResolver, getStatusDescription } = require('@condo/domains/miniapp/schema/fields/context')
-const { createDefaultB2BAppRoles, deleteB2BAppRoles } = require('@condo/domains/miniapp/tasks')
+const { deleteB2BAppRoles } = require('@condo/domains/miniapp/tasks')
+const { B2BAppRole } = require('@condo/domains/miniapp/utils/serverSchema')
+
+const logger = getLogger('miniapp/createDefaultB2BAppRoles')
+
+async function createDefaultB2BAppRoles (appId, organizationId, sender, context) {
```
Alllex202, line 22: Why did you remove this logic from task?

SavelevMatthew: I've realised, that we'll need to do it in sync. Since on a client we click a "Connect" button, which creates a context. And after receiving the response, we should be able to open an app (which is not possible before roles are created)

So we can do a polling to get notified when we can open an app (which is hard and can take a few minutes in case were queue is overwhelmed), or just create it "in sync manner" since response is not returned until `afterChange` is finished working, so we can guarantee on client, that everything is ready to use

apps/condo/domains/miniapp/schema/AllMiniAppsService.test.js
```diff
@@ -2,225 +2,382 @@
...
+                    id: app.id,
+                    connected: false,
+                    accessible: false,
+                    name: appAttrs.name,
+                    shortDescription: appAttrs.shortDescription,
+                    category: appAttrs.category,
+                    icon: appAttrs.icon,
+                    label: appAttrs.label,
+                    menuCategory: appAttrs.menuCategory,
+                    logo: expect.stringContaining(''),
```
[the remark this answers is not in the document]
SavelevMatthew: I'm expecting to contain any string (any because filename can be changed depending on file adapter). There's no `expect.anyString` in jest, but there's `stringContaining`. So I've used a trick, that any string contains empty string :)

PR #3700, opened 2023-08-09
feat(condo): DOMA-6613 changed indents to match layouts

Kekmus:
Screenshot 2023-08-09 at 13 24 10

### DIFF
--- a/apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
+++ b/apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
@@ -87,6 +87,7 @@ export type BaseNewsFormProps = {
     actionName: ActionNameProps,
 }
 
+//TODO(DOMA-6846) wrap form label with 0 margin and use default spacing (details in 6613 pr)
 const NO_RESIZE_STYLE: React.CSSProperties = { resize: 'none' }
 const FLEX_START_STYLE: React.CSSProperties = { alignItems: 'flex-start' }
 const BIG_MARGIN_BOTTOM_STYLE: React.CSSProperties = { marginBottom: '60px' }
--- a/apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
+++ b/apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
@@ -1046,7 +1046,7 @@ export const BaseNewsForm: React.FC<BaseNewsFormProps> = ({
                                 <Col span={24}>
                                     <Row gutter={BIG_HORIZONTAL_GUTTER}>
                                         <Col span={formFieldsColSpan}>
-                                            <Row gutter={[0, 0]}>
+                                            <Row>
                                                 <Col span={24} style={ { marginBottom: '24px' } }>
                                                     <Typography.Title level={2}>
                                                         {SelectSendPeriodLabel} ({tzInfo})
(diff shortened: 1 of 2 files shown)


apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -1043,8 +1046,8 @@ export const BaseNewsForm: React.FC<BaseNewsFormProps> = ({
                                 <Col span={24}>
                                     <Row gutter={BIG_HORIZONTAL_GUTTER}>
                                         <Col span={formFieldsColSpan}>
-                                            <Row gutter={SMALL_VERTICAL_GUTTER}>
-                                                <Col span={24}>
+                                            <Row gutter={[0, 0]}>
```
sitozzz, line 1049: I guess this property is zero by default
```suggestion
                                            <Row>
```

apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -813,13 +815,13 @@ export const BaseNewsForm: React.FC<BaseNewsFormProps> = ({
...
-                                <Col span={24}>
+                            <Row>
+                                <Col span={24} style={BIG_MARGIN_BOTTOM_STYLE}>
                                     <Row gutter={BIG_HORIZONTAL_GUTTER}>
                                         <Col span={formFieldsColSpan}>
-                                            <Row gutter={MEDIUM_VERTICAL_GUTTER}>
-                                                <Col span={24}>
-                                                    <Row>
+                                            <Row>
+                                                <Col span={24} style={ { marginBottom: '32px' } }>
```
sitozzz, line 823: You already declared some style constants above. Maybe move this object to the constants too?

sitozzz (approved):

apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -88,13 +88,20 @@ export type BaseNewsFormProps = {
...
 const NO_RESIZE_STYLE: React.CSSProperties = { resize: 'none' }
-const FORM_FILED_COL_PROPS = { style: { width: '100%', padding: 0 } }
+const FLEX_START_STYLE: React.CSSProperties = { alignItems: 'flex-start' }
+const BIG_MARGIN_BOTTOM_STYLE: React.CSSProperties = { marginBottom: '60px' }
+const MARGIN_BOTTOM_32_STYLE: React.CSSProperties = { marginBottom: '32px' }
+const MARGIN_BOTTOM_38_STYLE: React.CSSProperties = { marginBottom: '38px' }
+const MARGIN_BOTTOM_10_STYLE: React.CSSProperties = { marginBottom: '10px' }
+const MARGIN_BOTTOM_24_STYLE: React.CSSProperties = { marginBottom: '24px' }
+const MARGIN_TOP_8_STYLE: React.CSSProperties = { marginTop: '8px' }
+const MARGIN_TOP_44_STYLE: React.CSSProperties = { marginTop: '44px' }
```
nomerdvadcatpyat, line 98: Can you do this with Row gutter or our Space component from ui kit?

Kekmus: no because of the default styles of antd components, for example form.item labels I need to set different margins in different parts of the application to make the layout match the design layout. 
Previously it was done through Row gutter but the design team was not very satisfied so I had to manually rearrange the values to make the layout more pixel perfect)

nomerdvadcatpyat: But in other forms, we could handle the indentation without a bunch of custom margins.
I think it's better to wrap form label with 0 margin and use [default spacing](https://www.figma.com/file/Nbkug4IRM33sEyDuN0yO1C/UI-Kit-(web)?node-id=1240%3A5095&mode=dev)

Kekmus: Would it be a comfortable solution for both of us if I create new task  on code refactoring and add a TODO comment to the code with the number of this task?

At this moment I have tasks with higher priority and users are seeing a different interface than the design intended. I propose to upload these fixes unchanged with not hight quality code so that users can enjoy the beauty rn and rewrite the code later, when I have free slots for tasks

nomerdvadcatpyat (approved):

PR #3702, opened 2023-08-09
fix(condo): DOMA-6840 handle case when BillingReceiptFile is omit

ekabardinsky:


apps/condo/domains/billing/schema/AllResidentBillingReceiptsService.js
```diff
@@ -50,6 +50,12 @@ const getFile = (receipt, contacts) => {
     const file = propertyContacts.length > 0
```
dkoviazin, line 50: We need to check if Contact is verified

ekabardinsky: This is a part of query in main function:
const contacts = await Contact.getAll(context, {
                    phone: context.authedItem.phone,
                    isVerified: true,
                })

VKislov (approved):

PR #3723, opened 2023-08-11
fix(infra): DOMA-6887 fix server url for pass miniapp

Alllex202:


sitozzz (approved):

pahaz (approved):

pahaz:
its better to use domain name as the folder name

Alllex202:
> its better to use domain name as the folder name

It is important that word "miniapp" be contained.
This is necessary for performance of b2c applications

PR #3726, opened 2023-08-15
feat(condo): DOMA-6825 added sync task for pass

Alllex202:


pahaz (approved):

Alllex202:
> You do not need AND in conditions where only one element in array
> 
> ```
>  {
>                         AND: [
>                             {
>                                 organization: {
>                                     id_in: condoOrganizationIdsWithAllProperties,
>                                     deletedAt: null,
>                                 },
>                             },
>                         ],
>                     },
> ```

There was a problem with an empty array. If it was passed, then all elements in "OR" filter were returned. So just in case, I wrapped in in "AND" each of filters in "OR"

PR #3727, opened 2023-08-15
fix(pass): DOMA-6888 fixed bugs in pass miniapp

Alllex202:


Kekmus (approved):

Alllex202:
> You need to sync translates as line numbers are different

Oh.. Yes, a few translations were missed. Fixed it

---

> You can use dayjs format with locale support for this code

I will try to use this

PR #3731, opened 2023-08-16
fix(condo): DOMA-6927 fixed sonarlint suggestions

Kekmus:


Alllex202 (approved):

SavelevMatthew (approved):
LGTM, but I’m not a big fan of ‘.forEach’ syntax (there was an article telling that ‘.forEach’ was done as a hack, so it’s working a lot slower than a regular ‘for’)

PR #3753, opened 2023-08-22
DOMA-6980 fix actualUpdatedAt bug

toplenboren:
Hey!

We had a problem with field `actualUpdatedAt` filling with wrong values:

== before ==

image-2023-08-22-13-05-21-564

----

I fixed this by changing the implicit logic inside `createTicketChange` function:

This logic was first implemented in [this](https://jira.in.doma.ai/browse/DOMA-5956) task

> Now ONLY IF STATUS WAS CHANGED, the actualCreatedAt will fill with statusUpdatedAt value

== after ==

Screenshot 2023-08-22 at 19 59 11

### DIFF
--- a/apps/condo/domains/ticket/schema/TicketChange.test.js
+++ b/apps/condo/domains/ticket/schema/TicketChange.test.js
@@ -4,7 +4,7 @@
 
 const { faker } = require('@faker-js/faker')
 const dayjs = require('dayjs')
-
+const { sortBy } = require('lodash')
 
 const {
     expectToThrowAccessDeniedErrorToObj,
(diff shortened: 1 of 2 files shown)


apps/condo/domains/ticket/schema/TicketChange.test.js
```diff
@@ -517,6 +516,41 @@ describe('TicketChange', () => {
...
+                const statusUpdatedAt = dayjs().subtract('20', 'second').toISOString()
+
+                // Update 1 Update status explicitly. TicketChange actualUpdateDate should equal this number
+                await updateTestTicket(client, ticket.id, {
+                    status: { connect: { id: STATUS_IDS.IN_PROGRESS } },
+                    statusUpdatedAt,
+                })
+
+                // Explicitly sleep for some time
+                await new Promise(r => setTimeout(r, 500))
```
nomerdvadcatpyat, line 537: Can we use `waitFor` this?

nomerdvadcatpyat (approved):

apps/condo/domains/ticket/schema/TicketChange.test.js
sitozzz, line 537: waitFor would be even better here

toplenboren: No, since we need to sleep between requests, not wait for something

sitozzz (approved):

PR #3763, opened 2023-08-24
Update deploy_production.yaml

toplenboren:


.github/workflows/deploy_production.yaml
```diff
@@ -1,4 +1,4 @@
-name: "[production] Deploy"
+name: "⚠️💥💀⚡💢❗ [production] Deploy"
```
Alllex202, line 1: Why so few emojis? :)

Alllex202 (approved):

.github/workflows/deploy_production.yaml
SavelevMatthew, line 1: Why emojis looks dangerous? R u scary to deploy? You main branch is not prod ready?)

toplenboren: If you deploy review branches to prod a few times, you might find yourself fired, but that depends

PR #3766, opened 2023-08-24
refactor(condo): DOMA-6870 refactored B2BAppAccess preprocessor

Alllex202:


pahaz (approved):

apps/condo/domains/miniapp/schema/plugins/serviceUserAccessForB2BApp.js
```diff
@@ -397,125 +361,13 @@ const addCustomAccessToSchema = (schemaName, schema, schemaConfig) => {
     }
 
     schema.access = wrapInCustomAccess(access, customListAccess)
```
pahaz, line 363: It's better to call the helper function inside Mode.access rather than use an implicit wrapper here.

pahaz: It is better to explicitly call this function inside the access check

apps/condo/domains/contact/schema/Contact.js
```diff
@@ -169,7 +172,10 @@ const Contact = new GQLListSchema('Contact', {
             },
         ],
     },
-    plugins: [uuided(), versioned(), tracked(), softDeleted(), dvAndSender(), historical(), webHooked()],
+    plugins: [
+        uuided(), versioned(), tracked(), softDeleted(), dvAndSender(), historical(), webHooked(),
+        serviceUserAccessForB2BApp({ schemaConfig: SERVICE_USER_ACCESS_FOR_B2B_APP_CONFIG.Contact }),
```
SavelevMatthew, line 177: That's syntax is kinda strange.
1. Why do we need to pass config to every singe plugin call?
2. `lowerCamelCase: CAPS_CASE.UpperCamelCase` 🤣

Alllex202: 1) Config is optional. You can leave it empty, then default config values will be used
2) What's wrong with him?) `propKey: GLOBAL_VARIABLE.SchemaName`. I can move it in a variable with a different name and use it, but does that make sense?

PR #3769, opened 2023-08-25
fix(condo): DOMA-7016 fixed NewsItemRecipientsExportTask test

Kekmus:


apps/condo/domains/news/schema/NewsItemRecipientsExportTask.test.js
```diff
@@ -113,12 +113,14 @@ describe('NewsItemRecipientsExportTask', () => {
             test('admin can', async () => {
                 const [objCreated] = await createTestNewsItemRecipientsExportTask(adminClient, adminClient.user, dummyO10n)
 
-                const [obj, attrs] = await updateTestNewsItemRecipientsExportTask(adminClient, objCreated.id)
+                await waitFor(async () => {
```
SavelevMatthew, line 116: Why to you need `waitFor` you can update item and inspect it right after.

Or you're just starting updating it infinitely until `v` is `3`. Seems like a hack, not a fix

Kekmus: problem is that we can't guarantee if the worker will work in time. If it does, it will be v 3, if it doesn't, it will be v 2. I have looked at how it is implemented in other export tasks and it seems that it is common to do it via waitFor

if you know how you can ensure that check happens faster than the worker works and it is somehow the best solution just write about it

apps/condo/domains/news/schema/NewsItemRecipientsExportTask.test.js
```diff
@@ -113,12 +113,14 @@ describe('NewsItemRecipientsExportTask', () => {
...
+                    const [obj, attrs] = await updateTestNewsItemRecipientsExportTask(adminClient, objCreated.id)
 
-                expect(obj.dv).toEqual(1)
-                expect(obj.sender).toEqual(attrs.sender)
-                expect(obj.v).toEqual(2)
-                expect(obj.updatedBy).toEqual(expect.objectContaining({ id: adminClient.user.id }))
+                    expect(obj.dv).toEqual(1)
+                    expect(obj.sender).toEqual(attrs.sender)
+                    expect(obj.v).toEqual(3)
+                    expect(obj.updatedBy).toEqual(expect.objectContaining({ id: adminClient.user.id }))
```
AleX83Xpert, line 122: Are fields `sender` and `updatedBy same for `v=2` and `v=3` ?

Kekmus: done

apps/condo/domains/news/schema/NewsItemRecipientsExportTask.test.js
Kekmus: done

apps/condo/domains/news/schema/NewsItemRecipientsExportTask.test.js
```diff
@@ -113,12 +113,15 @@ describe('NewsItemRecipientsExportTask', () => {
...
 
-                const [obj, attrs] = await updateTestNewsItemRecipientsExportTask(adminClient, objCreated.id)
-
+                const [obj, attrs] = await updateTestNewsItemRecipientsExportTask(adminClient, objCreated.id, {
+                    status: COMPLETED,
+                })
+                
                 expect(obj.dv).toEqual(1)
                 expect(obj.sender).toEqual(attrs.sender)
                 expect(obj.v).toEqual(2)
```
AleX83Xpert, line 122: Looks like this expectation will fail in the case of the worker being done before the updating within the test.

SavelevMatthew (approved):

apps/condo/domains/news/schema/NewsItemRecipientsExportTask.test.js
```diff
@@ -113,12 +113,15 @@ describe('NewsItemRecipientsExportTask', () => {
             test('admin can', async () => {
                 const [objCreated] = await createTestNewsItemRecipientsExportTask(adminClient, adminClient.user, dummyO10n)
 
-                const [obj, attrs] = await updateTestNewsItemRecipientsExportTask(adminClient, objCreated.id)
-
+                const [obj, attrs] = await updateTestNewsItemRecipientsExportTask(adminClient, objCreated.id, {
```
AleX83Xpert, line 116: If the worker manages to process the task, this string will fail with an exception.
image

AleX83Xpert:
Trying to fix it with https://github.com/open-condo-software/condo/pull/3823/commits/c81d95ddb61fa78efcafd5a1030124929b426f13

PR #3803, opened 2023-09-01
fix(condo): DOMA-7025 added ellipsis to tabs

nomerdvadcatpyat:
Since the user may not guess to scroll tabs, added a default ellipsis from antd (the solution is consistent with the design, the UI-kit layouts should be updated soon)

изображение


sitozzz (approved):

packages/ui/src/components/Tabs/style.less
```diff
@@ -53,7 +53,20 @@
...
+      border-radius: 8px;
+
+      &-item-disabled {
+        opacity: @condo-global-opacity-disabled;
+      }
+    }
+
+    & .condo-tabs-tab-label {
+      color: @condo-global-color-gray-7;
+      font-weight: 600;
```
sitozzz, line 69: It would be better to use `@condo-global-font-weight-semibold` constant from `variables.less`

packages/ui/src/components/Tabs/style.less
```diff
@@ -53,7 +53,21 @@
     }
   }
 
-  &-nav-operations {
-    display: none !important;
+  &-dropdown {
+    font-family: "Open Sans", var(--condo-font-fallback);
```
SavelevMatthew, line 57: Please use typography token....Not hard-code

SavelevMatthew (approved):

PR #3811, opened 2023-09-04
DOMA-7068 check tag existence inside release job

sitozzz:


pahaz (approved):

AntonAL (approved):

SavelevMatthew (approved):

.github/workflows/deploy_production.yaml
```diff
@@ -16,6 +16,17 @@ jobs:
...
     steps:
+      - name: Check tag
+        id: 'check-tag'
+        run: |
+          if [[ ${{ github.event.ref }} =~ ^refs\/tags\/release-v[0-9]+\.[0-9]+\.[0-9]+$ ]]
+          then
+              echo "match=true" >> $GITHUB_OUTPUT
+              exit 0
+          else
+              echo "Wrong tag name was provided. Tag should be created with pattern 'release-vXX.XX.XX'"
```
AntonAL, line 27: Why we don't output result to ` >> $GITHUB_OUTPUT` here?

.github/workflows/deploy_production.yaml
```diff
@@ -16,6 +16,17 @@ jobs:
       PULL_USER: ${{ secrets.PULL_USER }}
       PULL_PERSONAL_ACCESS_KEY: ${{ secrets.PULL_PERSONAL_ACCESS_KEY }}
     steps:
+      - name: Check tag
+        id: 'check-tag'
+        run: |
+          if [[ ${{ github.event.ref }} =~ ^refs\/tags\/release-v[0-9]+\.[0-9]+\.[0-9]+$ ]]
```
AntonAL, line 22: Could you please write comment about reason of this check
Presence of a wildcard RegExp in `on` node above confuses in case when out of discussion context, because a related task contains no information.

sitozzz: There are two main reasons of this change. First one is restriction of action running with branch as an input source. And second reason is fix tag format with provided regular expression

PR #3814, opened 2023-09-04
feat(condo): DOMA-7047 user direct access (isSupport alternative) 😎😎😎😎

SavelevMatthew:


apps/condo/domains/miniapp/access/B2BAppPromoBlock.js
```diff
@@ -4,18 +4,32 @@
 
 const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
 
+const { canDirectlyManageSchemaObjects } = require('@condo/domains/user/utils/directAccess')
+
+/**
+ * B2BPromoBlocks are publicly available for any authed user including:
+ * 1. Admin / support
+ * 2. Users wit direct access
```
nomerdvadcatpyat, line 12:
```suggestion
 * 2. Users with direct access
```

apps/condo/domains/miniapp/access/B2CAppProperty.js
```diff
@@ -8,24 +8,46 @@ const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFo
 const { getById } = require('@open-condo/keystone/schema')
 
 const { RESIDENT, SERVICE } = require('@condo/domains/user/constants/common')
+const { canDirectlyReadSchemaObjects, canDirectlyManageSchemaObjects } = require('@condo/domains/user/utils/directAccess')
 
 const { checkB2CAppAccessRight } = require('../utils/accessSchema')
 
-async function canReadB2CAppProperties ({ authentication: { item: user } }) {
+/**
+ * B2CAppProperties can be read by:
```
nomerdvadcatpyat, line 16: and all residents

apps/condo/domains/user/schema/UserRightsSet.test.js
```diff
@@ -0,0 +1,509 @@
...
+                // Access removed from specific user
+                const [removedAccessUser] = await updateTestUser(support, executor.user.id, {
+                    rightsSet: { disconnectAll: true },
+                })
+                expect(removedAccessUser).toHaveProperty('rightsSet', null)
+                await expectToThrowAccessDeniedErrorToResult(async () => {
+                    await registerNewServiceUserByTestClient(executor)
+                })
+
+                // Access provided again
```
sitozzz, line 363: I don't understand what the reason to test case when user has lose his rights and after that you provide another one similar to the first part of the test

apps/condo/domains/miniapp/access/B2BApp.js
```diff
@@ -4,18 +4,32 @@
 
 const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
 
+const { canDirectlyManageSchemaObjects } = require('@condo/domains/user/utils/directAccess')
+
+/**
+ * B2BApps are publicly available to read for any authed user including:
```
nomerdvadcatpyat, line 10: any not deleted users

apps/condo/domains/miniapp/access/B2CApp.js
```diff
@@ -4,19 +4,32 @@
 
 const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
 
+const { canDirectlyManageSchemaObjects } = require('@condo/domains/user/utils/directAccess')
+
+/**
+ * B2CApps are publicly available for all authed users including:
```
nomerdvadcatpyat, line 10: any not deleted users

apps/condo/domains/user/utils/directAccess/config.js
```diff
@@ -0,0 +1,34 @@
...
+ * Note that adding a schema to this config will automatically add the required fields to the UserRightsSet schema,
+ * but accesses will not work automatically!
+ * To do this, you need to explicitly add one of the utilities from "@condo/domains/user/utils/access" to the model accesses.
+ *
+ * Also note that changing this config always generates a new migration!
+ * @typedef {{schemaName: string, readonly?: boolean}} DirectAccessList
+ * @typedef {{lists: Array<DirectAccessList | string>, services: Array<string>}} DirectAccessConfig
+ * @type {DirectAccessConfig}
+ */
+const DIRECT_ACCESS_AVAILABLE_SCHEMAS = {
```
sitozzz, line 13: Is it a task related list of available schemas? What if someone want to change this list? Did you thought about making that configurable not from a code source

apps/condo/domains/miniapp/schema/B2BAppPermission.test.js
```diff
@@ -189,8 +190,9 @@ describe('B2BAppPermission', () => {
                 })
             })
             test('No one in the organization can, including the administrator', async () => {
-                const allPermissions = await B2BAppPermission.getAll(manager, { id: permission.id })
-                expect(allPermissions).toHaveLength(0)
+                await expectToThrowAccessDeniedErrorToObjects(async () => {
+                    await B2BAppPermission.getAll(manager, { id: permission.id })
```
nomerdvadcatpyat, line 194: why error on `read` operation?

apps/condo/domains/miniapp/access/B2BAppPermission.js
```diff
@@ -9,33 +9,44 @@ const { getById } = require('@open-condo/keystone/schema')
 
 const { checkB2BAppAccessRight } = require('@condo/domains/miniapp/utils/accessSchema')
 const { SERVICE } = require('@condo/domains/user/constants/common')
+const { canDirectlyReadSchemaObjects, canDirectlyManageSchemaObjects } = require('@condo/domains/user/utils/directAccess')
 
 /**
  * B2B App permissions may only be read:
```
nomerdvadcatpyat, line 15: can all employees in organization with b2b context has this permission?

apps/condo/domains/user/utils/directAccess/config.js
```diff
@@ -0,0 +1,34 @@
...
+        'B2BAppAccessRightSet',
+        'B2BAppContext',
+        'B2BAppPermission',
+        'B2BAppPromoBlock',
+        'B2CApp',
+        'B2CAppAccessRight',
+        'B2CAppBuild',
+        'B2CAppProperty',
+        // Organization domain
+        { schemaName: 'Organization', readonly: true },
```
sitozzz, line 27: I'm a little confused by the different definition format

apps/condo/domains/miniapp/access/B2BAppContext.js
```diff
@@ -9,44 +9,54 @@ const { getById } = require('@open-condo/keystone/schema')
...
  */
-async function canReadB2BAppContexts ({ authentication: { item: user } }) {
+async function canReadB2BAppContexts ({ authentication: { item: user }, listKey }) {
     if (!user) return throwAuthenticationError()
     if (user.deletedAt) return false
 
     if (user.isAdmin || user.isSupport) return {}
 
+    const hasDirectAccess = await canDirectlyReadSchemaObjects(user, listKey)
+    if (hasDirectAccess) return true
```
pahaz, line 30: May be we can return not only true/false but also can use it for read restrictions?

pahaz, line 30:
```suggestion
    if (typeof hasDirectAccess !== "undefined") return hasDirectAccess
```

apps/condo/domains/miniapp/access/B2BAppPermission.js
```diff
@@ -9,33 +9,44 @@ const { getById } = require('@open-condo/keystone/schema')
...
 
+    const hasDirectAccess = await canDirectlyReadSchemaObjects(user, listKey)
+    if (hasDirectAccess) return true
+
+    if (user.type === SERVICE) {
+        return { app: { accessRights_some: { user: { id: user.id }, deletedAt: null } } }
+    }
 
-    return { app: { accessRights_some: { user: { id: user.id }, deletedAt: null } } }
+    return false
```
pahaz, line 33: Sure to rollback app access?

apps/condo/domains/user/schema/UserRightsSet.js
```diff
@@ -0,0 +1,75 @@
...
+                // (similar as find/getByConditions returns), so there's no softDeleted filter in it.
+                // NOTE 2: Yes, we're checking deletedAt in direct accesses, but the presence of a field "rightsSet" in user record
+                // is our sign to check it (do a sub-query).
+                // That's why we're clearing this fields now once to avoid extra sub-queries later.
+                const relatedUsers = await find('User', { rightsSet: { id: itemId } })
+                const updateData = relatedUsers.map(user => ({
+                    id: user.id,
+                    data: {
+                        dv: 1,
+                        sender: { dv: 1, fingerprint: USER_UPDATE_FINGERPRINT },
```
pahaz, line 52: Get real fingerprint from resolvedData

apps/condo/domains/user/utils/directAccess/config.js
pahaz: It’s better to use strong typing :) be friendly to type script

pahaz (approved):

apps/condo/domains/miniapp/access/B2BAppContext.js
SavelevMatthew: Why? Its returns only boolean, since you can read / manage anything or nothing

SavelevMatthew: Is there a real case when you need filtering? 

Since right now (webhooks / dev-portal user / support) will need access to entire table.

apps/condo/domains/miniapp/schema/B2BAppPermission.test.js
SavelevMatthew: That's default keystone behaviour then access is resolved to `false`. Previously we returned a filter, which was making no sense (and that's why returns empty results). So, this is more general "keystone" behaviour, not part of this PR

apps/condo/domains/user/utils/directAccess/config.js
SavelevMatthew: Didn't get it. All types are defined above. You use an object, when you need to pass extra config (like manage restriction). In 99% of cases there will be a strings, which are a lot easier to write / read.

apps/condo/domains/miniapp/access/B2BAppPermission.js
SavelevMatthew: I think they can get `read` access, but not in this PR I guess

apps/condo/domains/user/schema/UserRightsSet.js
```diff
@@ -0,0 +1,75 @@
...
+            isRequired: true,
+        },
+        ...generateRightSetFields(DIRECT_ACCESS_AVAILABLE_SCHEMAS),
+    },
+    hooks: {
+        beforeChange: async ({ resolvedData, existingItem, context }) => {
+            const isSoftDelete = Boolean(resolvedData['deletedAt'])
+            const itemId = get(existingItem, 'id', resolvedData['id'])
+
+            if (isSoftDelete) {
```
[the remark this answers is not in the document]
SavelevMatthew: Of course it will :)
(and covered with tests)

When you're soft-deleting items you're sending `deletedAt` at request, so it's presented in resolvedData. So we can perform side-effects before deletion

apps/condo/domains/user/schema/UserRightsSet.test.js
SavelevMatthew: Because it's 2 cases of loosing access. For example:
1. You create rights set called "Support level 1", and give it to user A and user B, so they both are available to do some stuff.
2. Later, you decided to drop user A from your group, so you're setting their rightsSet to `null` and expect them to loose direct access. And that's what is happening in first part of the test.
3. You can also delete entire rights set, so any users having it should have `null` in their relation (logic you've asked about below) (( P.S. by default keystone do not set null on soft-delete )). And you expect all users to loose relation (even with `flat` utils like `find`/`getByCondition`) and loose access. So that's what is being checked in the second part of the test.

apps/condo/domains/user/utils/directAccess/config.js
SavelevMatthew: I did not get this question. You need to do 2 steps to add new schemas / delete schemas / modify config:

1. Change the config, so it will change set of fields in `UserRightsSet` (false for new by default for adding).
2. Generate a migration, since list of fields will be changed
3. Use one of the provided utils in access (that's a design decision, since we want all accesses to specified explicitly)

SavelevMatthew: And all of that info is described in JSDoc

nomerdvadcatpyat (approved):

sitozzz (approved):

PR #3820, opened 2023-09-05
DOMA-7022 miniapp permissions settings table

nomerdvadcatpyat:
Added a role configuration table to the "Role Settings" page (for b2bAppRole only), closed with a feature flag.
[Figma layouts](https://www.figma.com/file/MZebKVygbECAtZiwQEEeLh/14-%D0%9D%D0%B0%D1%81%D1%82%D1%80%D0%BE%D0%B9%D0%BA%D0%B8?node-id=201%3A10443&mode=dev)

изображение


apps/condo/domains/organization/components/EmployeeRolesSettingsContent.tsx
```diff
@@ -1,62 +1,582 @@
...
+
+type TableCheckboxProps = {
+    employeeRoleId: string
+    b2bAppId?: string
+    permissionKey: string
+    permissionState: PermissionState
+    setPermissionState: Dispatch<SetStateAction<PermissionState>>
+}
+
+type SaveNotificationType = 'create' | 'delete'
```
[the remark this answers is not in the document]
nomerdvadcatpyat: removed it and replaced `getSaveNotificationType` function to `getSaveNotificationMessage`

apps/condo/domains/organization/components/EmployeeRolesSettingsContent.tsx
```diff
@@ -1,62 +1,582 @@
...
+                                    ...acc,
+                                    [b2bApp.id]: {
+                                        permissions: {
+                                            ...b2BAppPermissions
+                                                .filter(permission => permission.app.id === b2bApp.id)
+                                                .reduce((acc, permission) => ({ ...acc, [permission.key]: false }), {}),
+                                            [`canRead${b2bApp.id}`]: false,
+                                        },
+                                    },
+                                }
```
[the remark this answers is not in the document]
nomerdvadcatpyat: Simplified, wrote a docs for functions

sitozzz (approved):

apps/condo/domains/miniapp/access/B2BAppPermission.js
```diff
@@ -2,28 +2,43 @@
...
+    const b2bAppContexts =  await find('B2BAppContext', {
+        organization: { id_in: userOrganizations.map(organization => organization.id) },
+        deletedAt: null,
+    })
+    const b2bAppIds = uniq(compact(b2bAppContexts.map(context => get(context, 'app'))))
 
-    return { app: { accessRights_some: { user: { id: user.id }, deletedAt: null } } }
+    return {
+        OR: [
+            { app: { accessRights_some: { user: { id: user.id }, deletedAt: null } } },
```
SavelevMatthew, line 38: Split this `or` condition

User with type service have one query, staff another, residents is banned

apps/condo/domains/miniapp/schema/B2BAppPermission.test.js
```diff
@@ -188,9 +188,9 @@ describe('B2BAppPermission', () => {
                     expect(allPermissions).toHaveLength(0)
                 })
             })
-            test('No one in the organization can, including the administrator', async () => {
-                const allPermissions = await B2BAppPermission.getAll(manager, { id: permission.id })
-                expect(allPermissions).toHaveLength(0)
+            test('Organization employee can', async () => {
```
SavelevMatthew, line 191: Are you really need to give access to all employees?
Seems like ones with `canManageRoles` should be able to see this

apps/condo/domains/miniapp/access/B2BAppPermission.js
```diff
@@ -30,6 +34,26 @@ async function canReadB2BAppPermissions ({ authentication: { item: user }, listK
         return { app: { accessRights_some: { user: { id: user.id }, deletedAt: null } } }
     }
 
+    if (user.type === STAFF) {
+        const userEmployeesWithCanManageRoles = await find('OrganizationEmployee',
+            {
```
SavelevMatthew, line 39: Why not just using filter from queryEmployee? (to change it later in single place)

apps/condo/domains/organization/components/EmployeeRolesSettingsContent.tsx
```diff
@@ -1,62 +1,629 @@
...
+import {
+    B2BApp,
+    B2BAppRoleCreateInput,
+    B2BAppRoleUpdateInput,
+    SortOrganizationEmployeeRolesBy,
+    OrganizationEmployeeRole as OrganizationEmployeeRoleType,
+    B2BAppPermission as B2BAppPermissionType,
+    B2BAppRole as B2BAppRoleType,
+} from '@app/condo/schema'
+import { Col, notification, Row, RowProps, Typography } from 'antd'
```
SavelevMatthew, line 10: why old typography here?

apps/condo/domains/organization/components/EmployeeRolesSettingsContent.tsx
```diff
@@ -1,62 +1,629 @@
...
+    employeeRoleId: string
+    b2bAppId?: string
+    permissionKey: string
+    permissionsState: PermissionsState
+    setPermissionsState: Dispatch<SetStateAction<PermissionsState>>
+}
+
+/**
+ * Changes value of the checkbox and related checkboxes.
+ * If user set "canManage" permission checkbox, then "canRead" checkbox sets to true automatically.
```
SavelevMatthew, line 104: Does this logic applied to miniapps?

Seems like it shouldn't. Because in condo we have strong naming agreement. Our contractors don't have one. So it can possibly be any key, not just `canRead` / `canManage`.

SavelevMatthew: Also aren't there any warnings? Seems like function has `Record<string, any>: any` type

apps/condo/domains/organization/components/EmployeeRolesSettingsContent.tsx
```diff
@@ -1,62 +1,629 @@
...
+ * If there is only a checkbox withdrawal (removal of permissions), then there will be a DeletePermissionsMessage message,
+ * in another case (adding permissions/adding and deleting permissions) will be a UpdatePermissionsMessage message.
+ */
+const getSaveNotificationMessage = ({ intl, initialPermissionsState, permissionsState }): string => {
+    const UpdatePermissionsMessage = intl.formatMessage({ id: 'pages.condo.settings.employeeRoles.notification.updatePermissions' })
+    const DeletePermissionsMessage = intl.formatMessage({ id: 'pages.condo.settings.employeeRoles.notification.deletePermissions' })
+
+    let notificationMessage = DeletePermissionsMessage
+
+    for (const roleId of Object.keys(initialPermissionsState)) {
```
SavelevMatthew, line 187: Triple `for` just for notification diff checking seems impressive :)

apps/condo/domains/organization/components/EmployeeRolesSettingsContent.tsx
```diff
@@ -1,62 +1,629 @@
...
+    createB2BAppRoleAction,
+    softDeleteB2BAppRoleAction,
+    updateB2BAppRoleAction,
+}) => {
+    for (const appId of Object.keys(newB2bRolePermissions)) {
+        const initialEmployeePermissionsInB2BApp = initialB2bRolePermissions[appId]
+        const newEmployeePermissionsInB2BApp = newB2bRolePermissions[appId]
+
+        if (!isEqual(initialEmployeePermissionsInB2BApp, newEmployeePermissionsInB2BApp)) {
+            const canReadAppKey = `canRead${appId}`
```
SavelevMatthew, line 226: What if I have a permission B2BAppPermission with this key? Seems like naming collision

apps/condo/domains/organization/components/EmployeeRolesSettingsContent.tsx
```diff
@@ -1,62 +1,629 @@
...
+    const CancelSelectionLabel = intl.formatMessage({ id: 'global.cancelSelection' })
+
+    const tableColumns = useEmployeeRolesTableColumns(employeeRoles)
+    const tableData: PermissionsGroup[] = useMemo(() => connectedB2BApps.map((b2bApp): PermissionsGroup => ({
+        id: b2bApp.id,
+        groupName: b2bApp.name,
+        permissions: [
+            // Client side mark that employee role has b2bRole (Due to our logic this employee role can read b2bApp).
+            {
+                id: `canRead${b2bApp.id}`,
```
SavelevMatthew, line 391: Same about name collision

apps/condo/domains/organization/components/EmployeeRolesSettingsContent.tsx
```diff
@@ -1,62 +1,629 @@
...
+                .filter(permission => permission.app.id === b2bApp.id)
+                .map((permission): PermissionRow => ({
+                    id: permission.id,
+                    key: permission.key,
+                    name: permission.name,
+                })),
+        ],
+    })), [CanReadServiceTitle, b2BAppPermissions, connectedB2BApps])
+
+    console.log('tableData', b2BAppPermissions, tableData)
```
SavelevMatthew, line 405: ?

apps/condo/domains/organization/components/EmployeeRolesSettingsContent.tsx
SavelevMatthew, line 1: It's probably worth it to split up this 600-liner into multiple parts. It will be difficult to read once we've added `OrganizationEmployeeRole` here.

apps/condo/domains/organization/components/EmployeeRolesSettingsContent.tsx
nomerdvadcatpyat: We check `permission` in the `group of permissions` (appId for b2bApps) in the `employee role`.
I think it's ok

apps/condo/domains/miniapp/access/B2BAppPermission.js
nomerdvadcatpyat: `queryOrganizationEmployeeFor` not includes `isAccepted` filter (and refactor `queryOrganizationEmployeeFor` not part of this pr). But ok, use `queryOrganizationEmployeeFor` here and add additional filter

apps/condo/domains/organization/components/EmployeeRolesSettingsContent.tsx
nomerdvadcatpyat: > Does this logic applied to miniapps?
> 
> Seems like it shouldn't. Because in condo we have strong naming agreement. Our contractors don't have one. So it can possibly be any key, not just `canRead` / `canManage`.


`canManage` here it's just an example: if we have not checked `canRead` checkbox (for b2bApps it's only on client-side)  and set any other checkbox in the group for employee role, then `canRead` checkbox becomes true.

About name collisions: in the current logic I find it difficult to represent a case when we have `b2bAppRole` (= can read app) and it has a separate right `canRead{AppId}` = false. Contractors have created such right and it means something other than the ability to view app? So it’s bad naming of this right. And if, logically, this right should mean the same as here, then it should not exist at all, as we are responsible for this having `B2bAppRole` (And even if there is nothing will break).

> Also aren't there any warnings? Seems like function has `Record<string, any>: any` type

Added function return type `PermissionsType`.

apps/condo/domains/organization/components/EmployeeRolesSettingsContent.tsx
nomerdvadcatpyat: So far, I don’t see the point of putting components/functions/types into separate files. They are used only here.

VKislov (approved):

PR #3827, opened 2023-09-06
fix(condo): DOMA-6513 moved CHECK_TLS_CLIENT_CERT_CONFIG variable into condo microservice

AntonAL:
Because Sber URLS are not belonging to origin of Doma.ai, there is no possibility to make requests.
The only way was using "no-cors" mode of `fetch`.
Due to using "no-cors" mode, responses of `fetch` does not containing any useful information.
Previously fetch has thrown an Error, and I've relied on it, but now it does not.

It seems to be impossible to do anything now.

So, a manual disabling of feature has been implemented.

### DIFF
--- a/apps/condo/domains/common/hooks/useCheckTLSClientCert.tsx
+++ b/apps/condo/domains/common/hooks/useCheckTLSClientCert.tsx
@@ -1,6 +1,8 @@
 import getConfig from 'next/config'
 import { useState } from 'react'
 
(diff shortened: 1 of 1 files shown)


AntonAL:
> it's not a infra master! use [57bf39c](https://github.com/open-condo-software/infrastructure-condo/commit/57bf39c3a96c34e2d5b8c65c903fc05f905c4bf4)

It is an infra master at exactly the same SHA, you have specified
image

PR #3829, opened 2023-09-07
Feat/condo/doma 6963/discover service consumers script

AleX83Xpert:
The new script's aim is to create `ServiceConsumer` models for a given organization.

Usage:
```bash
yarn workspace @app/condo node bin/billing/discoverServiceConsumers -h
```

### DIFF
--- a/apps/condo/domains/common/utils/serverSchema/index.js
+++ b/apps/condo/domains/common/utils/serverSchema/index.js
@@ -1,4 +1,5 @@
 const { getItems } = require('@keystonejs/server-side-graphql-client')
+const { isFunction } = require('lodash')
 
 const { getSchemaCtx } = require('@open-condo/keystone/schema')
 const GLOBAL_QUERY_LIMIT = 1000
@@ -114,7 +115,7 @@ class GqlWithKnexLoadList {
  * @param {String[]} sortBy
  * @param {Number} chunkSize
  * @param {Number} limit
- * @param {function(Array): Array} chunkProcessor A place to use or/and modify just loaded chunk
+ * @param {function(Array): Array | Promise<Array>} chunkProcessor A place to use or/and modify just loaded chunk
  * @returns {Promise<*[]>}
  */
 const loadListByChunks = async ({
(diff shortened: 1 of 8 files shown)


dkoviazin (commented):
new package only for parsing arguments seems like over-kill
why you need  chunkProcessor ? you can get all models and then use chunkProcessor to them

AleX83Xpert:
> new package only for parsing arguments seems like over-kill

We have dozens of scripts with custom parsing of arguments. It's time to get a chance to improve it. As a bonus - autogenerated help and long&short arguments :)

> why you need chunkProcessor ? you can get all models and then use chunkProcessor to them

I'd like to reduce memory usage. `chunkProcessor` returns an empty array, so there are not all models listed within memory.

apps/condo/domains/common/utils/serverSchema/index.js
```diff
@@ -114,7 +115,7 @@ class GqlWithKnexLoadList {
  * @param {String[]} sortBy
  * @param {Number} chunkSize
  * @param {Number} limit
- * @param {function(Array): Array} chunkProcessor A place to use or/and modify just loaded chunk
+ * @param {function(Array): Array | Promise<Array>} chunkProcessor A place to use or/and modify just loaded chunk
  * @returns {Promise<*[]>}
  */
 const loadListByChunks = async ({
```
dkoviazin, line 121: I don't like an idea to use this function for processing data
The name of the function is starting with **load**
It's like to add another argument and it will start saving models by chunks


Maybe, better is to make a separated function for this

dkoviazin (approved):

apps/condo/domains/common/utils/serverSchema/index.js
AleX83Xpert: The `chunkProcessor` allows modifying loaded data on the fly. One of the options is to filter chunks or return an empty array.
Load is not equal to return. Load from the database and optionally pass through chunkProcessor.

The separated function will have absolutely the same functionality. Btw, current functionality is used in many places. I just have to add the ability to pass the async function.

PR #3837, opened 2023-09-11
DOMA-7118 migrate ANALYTICS_V3 feature flag from organization to growthbook

sitozzz:


apps/condo/domains/analytics/schema/GetOverviewDashboardService.test.js
```diff
@@ -69,9 +70,7 @@ describe('GetOverviewDashboardService', () => {
         organizationAdminUser = await makeClientWithNewRegisteredAndLoggedInUser()
         organizationAdminUser.organization = organization
 
-        await updateTestOrganization(admin, organization.id, {
-            features: [ANALYTICS_V3],
-        })
+        await updateTestOrganization(admin, organization.id)
```
SavelevMatthew, line 73: Why do you need an empty update?

apps/condo/domains/analytics/schema/GetOverviewDashboardService.test.js
```diff
@@ -109,11 +108,13 @@ describe('GetOverviewDashboardService', () => {
         ticket = createdTicket
     })
 
+    beforeEach(() => {
```
SavelevMatthew, line 111: It's a bit strange to see `beforeEach` at the end of `describe` block :)

SavelevMatthew: Also should feature flag be toggled down in `afterEach`?

apps/condo/domains/organization/constants/features.js
```diff
@@ -1,16 +1,13 @@
...
 const ALL_FEATURES = [
     SERVICE_PROVIDER_PROFILE_FEATURE,
-    ANALYTICS_V3,
 ]
 
 const ORGANIZATION_FEATURE_ENUM_NAME = 'OrganizationFeature'
 
 module.exports = {
     SERVICE_PROVIDER_PROFILE_FEATURE,
-    ANALYTICS_V3,
```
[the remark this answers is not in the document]
sitozzz: Yes it should. I would write a migration script for that

apps/condo/domains/analytics/schema/GetOverviewDashboardService.test.js
sitozzz: In most of the test cases feature flag should be enabled. But in a single test we should check if value of the `ANALYTICS_V3` flag is `false`, error was provided to the output.

From the other side, there is a case when developer want to run single test (for example with error about disabled feature). So the easiest way to make this work is enable feature before each test and disable feature if necessary

apps/condo/domains/analytics/schema/GetOverviewDashboardService.test.js
sitozzz: That's my fault. Fixed it

SavelevMatthew (approved):

nomerdvadcatpyat (approved):

PR #3838, opened 2023-09-12
DOMA-7049 telephony postmessage api

nomerdvadcatpyat:
1. Moved logic with connect tickets to call to condo local storage
2. Added 2 postMessage handlers – `CondoWebAppSetActiveCall` (call start) and `CondoWebAppSaveCallRecord` (call end, save call record)
3. Added `globalUrl` to `B2BApp`, removed `isGlobal`, made global b2bApps dependent on B2bAppContext


apps/condo/domains/miniapp/components/GlobalApps/GlobalAppsContainer.tsx
```diff
@@ -126,8 +130,8 @@ export const GlobalAppsContainer: React.FC = () => {
         }
     }, [user, loading])
 
-    // // Global miniapps allowed only for authenticated users
-    if (!user) {
+    // // Global miniapps allowed only for authenticated (not support or admin) users
+    if (!user || user.isSupport || user.isAdmin) {
```
sitozzz, line 134: Why admin or support can't use global miniapps?

apps/condo/migrations/20230912113650-0313_auto_20230912_0636.js
```diff
@@ -0,0 +1,50 @@
...
+
+exports.up = async (knex) => {
+    await knex.raw(`
+    BEGIN;
+--
+-- Remove field isGlobal from b2bapp
+--
+ALTER TABLE "B2BApp" DROP COLUMN "isGlobal" CASCADE;
+--
+-- Remove field isGlobal from b2bapphistoryrecord
```
sitozzz, line 12: By what logic is the fact that this mini-application is now determined to be global?

sitozzz: If it done with `globalUrl` prop, I suppose you should take into account already existing records in the database

apps/condo/domains/miniapp/components/GlobalApps/GlobalAppsContainer.tsx
nomerdvadcatpyat: Because you can steal an admin token or support token from a third-party application.
For not global miniapps we have such check (B2BAppPage).

apps/condo/migrations/20230912113650-0313_auto_20230912_0636.js
nomerdvadcatpyat: If it has `globalUrl`. Miniapp can have `appUrl` to connect and configure miniapp and `globalUrl` for background work in service

nomerdvadcatpyat: > If it done with `globalUrl` prop, I suppose you should take into account already existing records in the database

updated in migration

sitozzz (approved):

apps/condo/migrations/20230912113650-0313_auto_20230912_0636.js
sitozzz: > > If it done with `globalUrl` prop, I suppose you should take into account already existing records in the database
> 
> updated in migration

That's what I'm talking about. Looks good now

pahaz (approved):

PR #3845, opened 2023-09-13
feat(condo): DOMA-7104 make category field optional for RegisterBilli…

ekabardinsky:
…ngReceiptsService.js


apps/condo/domains/billing/schema/RegisterBillingReceiptsService.test.js
```diff
@@ -398,35 +401,84 @@ describe('RegisterBillingReceiptsService', () => {
...
             test('BillingReceipts are created if they have different categories', async () => {
-
                 const [organization] = await createTestOrganization(admin)
                 const [integration] = await createTestBillingIntegration(admin)
                 const [billingContext] = await createTestBillingIntegrationOrganizationContext(admin, organization, integration)
 
+                // create categories
+                const services = generateServicesData()
+                const defaultCategory = await BillingCategory.getOne(admin, {
+                    name: 'billing.category.housing.name',
```
dkoviazin, line 411: You can use constants, as all BillingCategories has the same id (UUID) on all servers

dkoviazin (approved):

apps/condo/domains/billing/schema/RegisterBillingReceiptsService.test.js
ekabardinsky: Good point, 10x

apps/condo/domains/billing/schema/RegisterBillingReceiptsService.js
```diff
@@ -334,13 +334,16 @@ const RegisterBillingReceiptsService = new GQLCustomSchema('RegisterBillingRecei
 
                 // Step 1:
                 // Parse properties, accounts and receipts from input
+                const categoryResolver = new CategoryResolver()
```
toplenboren, line 337: why use class here ?

apps/condo/domains/billing/schema/helpers/categoryResolver.js
```diff
@@ -0,0 +1,74 @@
...
+const { generateServerUtils } = require('@open-condo/codegen/generate.server.utils')
+
+const { DEFAULT_BILLING_CATEGORY_ID } = require('@condo/domains/billing/constants/constants')
+const { loadListByChunks } = require('@condo/domains/common/utils/serverSchema')
+
+const BILLING_CATEGORY_FIELDS = '{ id name serviceNames }'
+const BillingCategoryGQL = generateGqlQueries('BillingCategory', BILLING_CATEGORY_FIELDS)
+const BillingCategoryApi = generateServerUtils(BillingCategoryGQL)
+
+class CategoryResolver {
```
toplenboren, line 13: Why do we need a class for this? 

The only thing that is stored in `this` is categories, which can be fetched from DB / JSON anyways, why not make a module with static functions? It would be easier to read and use

I would get it if CategoryResolver were a singletone, or if Resolver pattern is common in project, but I'm not sur on this.

It also would be really cool if you added docstrings

apps/condo/domains/billing/schema/RegisterBillingReceiptsService.js
dkoviazin: Because, there will be a single class for every instance the unified mutation will work: Category, Account, Recipient, Property, Receipt. They all will have base class for logging, error handling and caching mechanics. Work with receipt will be like using the chain of classes. And yes, they will be singleton . To separate logic to different classes is better then to have one super-function 5 pages in length. It will be also easy to write tests for them as they have predictable input and needs to produce predictable output.

apps/condo/domains/billing/schema/helpers/categoryResolver.js
dkoviazin: Because, there will be a single class for every instance the unified mutation will work: Category, Account, Recipient, Property, Receipt. They all will have base class for logging, error handling and caching mechanics. Work with receipt will be like using the chain of classes. And yes, they will be singleton . To separate logic to different classes is better then to have one super-function 5 pages in length. It will be also easy to write tests for them as they have predictable input and needs to produce predictable output.

ekabardinsky: I prefer to use that way to handle category resolve logic, since we have state in code and we have methods that manage/read state

apps/condo/domains/billing/schema/helpers/categoryResolver.js
```diff
@@ -0,0 +1,74 @@
+const { isEmpty, get } = require('lodash')
```
pahaz, line 1: why you don't want to use `utils` folder for such helpers

PR #3852, opened 2023-09-14
fix(condo): DOMA-7172 fixed case when processed chunk in loadListByChunks is empty

nomerdvadcatpyat:
if we use `chunkProcessor` for data loading we can return an empty list in it (It is necessary that `all` variable in `loadListByChunks` does not take up much memory).
Therefore, the length of the new chunk need to be counted before converting the received data in `chunkProcessor`

### DIFF
--- a/apps/condo/domains/common/utils/serverSchema/index.js
+++ b/apps/condo/domains/common/utils/serverSchema/index.js
@@ -133,18 +133,24 @@ const loadListByChunks = async ({
     let maxIterationsCount = Math.ceil(limit / chunkSize)
     let newChunk = []
     let all = []
+    let newChunkLength
+
     do {
         newChunk = await list.getAll(context, where, { sortBy, first: chunkSize, skip: skip })
-        if (newChunk.length > 0) {
+        newChunkLength = newChunk.length
+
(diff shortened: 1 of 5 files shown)


apps/condo/domains/news/schema/GetNewsItemsRecipientsCountersService.js
```diff
@@ -61,7 +61,7 @@ const GetNewsItemsRecipientsCountersService = new GQLCustomSchema('GetNewsItemsR
                                 orConditions.push(...queryConditionsByUnits(property))
                             }
 
-                            return []
+                            return chunk
```
AleX83Xpert, line 64: The result of `loadListByChunks` is not using. Chunk processor is used here to reduce memory consumption.
Looks like a bug within of `loadListBuChunks`. We need to increase the `skip` variable before start chunkProcessor.

nomerdvadcatpyat: Refactored

AleX83Xpert (approved):

dkoviazin (approved):

PR #3853, opened 2023-09-14
INFRA-126 - fix an error that kills pods [Proof of concept]

toplenboren:


sitozzz (approved):

apps/condo/domains/common/utils/sberCloudFileAdapter.js
```diff
@@ -229,25 +231,30 @@ const obsRouterHandler = ({ keystone }) => {
...
-            // check property access case
-            if (hasAccessToReadFile && isString(propertyQuery) && !isNil(propertyValue)) {
-                const propertyPath = propertyQuery
-                    .replaceAll('}', '') // remove close brackets of sub props querying
-                    .split('{') // work with each path parts separately
-                    .map(path => path.trim()) // since gql allow to have spaces in querying - let's remove them
-                    .join('.') // join by . for lodash get utility
-                hasAccessToReadFile = get(item, propertyPath) == propertyValue
+            try {
+                const item = await getItem({
```
pahaz, line 237: it looks like you don't need to check access here. Use `getByID` util here

toplenboren: No, I think this code was supposed to check access by design

toplenboren: Gonna look into it :/

pahaz: We also have execute gql by user. It looks like we need to use it here?!

PR #3859, opened 2023-09-15
docs(condo): DOMA-1932 updated data-access.md

AntonAL:


AntonAL:
@pahaz, could you please write negative side-effects of using

```
const { getItem, getItems } = require('@keystonejs/server-side-graphql-client')
```

pahaz:
> @pahaz, could you please write negative side-effects of using
> 
> ```
> const { getItem, getItems } = require('@keystonejs/server-side-graphql-client')
> ```

The primary issue is the incompatibility with Keystone 6. Additionally, we are attempting to address a problem that can readily be resolved using existing utilities, such as the `execGqlAsUser` function.

pahaz (approved):

PR #3863, opened 2023-09-16
fix(condo): DOMA-7186 used status() instead sendStatus() in obsRouterHandler

nomerdvadcatpyat:
`sendStatus` it's a terminate operation and in case with `shallow-redirect` it throws error


toplenboren (approved):
Best to drop a comment in the code on why it is done so

pahaz (approved):

PR #3875, opened 2023-09-19
feat(dev-api): DOMA-7048 dev portal API introduced 🧑‍💻 🧑‍💻 

SavelevMatthew:


apps/dev-api/domains/common/utils/sms.js
```diff
@@ -0,0 +1,53 @@
...
+    #internalAdapter = null
+
+    constructor (type = 'fake') {
+        switch (type) {
+            case 'condo-dev':
+                this.#internalAdapter = new CondoSMSAdapter(developmentClient)
+                break
+            case 'condo-prod':
+                this.#internalAdapter = new CondoSMSAdapter(productionClient)
+                break
```
sitozzz, line 33: Why not just `dev` and `prod`? IMO `condo-` prefix is redundant here

apps/dev-api/domains/user/utils/testSchema/index.js
```diff
@@ -0,0 +1,214 @@
...
+    throwIfError(data, errors)
+    return [data.result, attrs]
+}
+/* AUTOGENERATE MARKER <FACTORY> */
+
+async function startConfirmPhoneActionByTestClient(attrs = {}, client) {
+    client ??= await makeClient()
+    attrs.dv ??= 1
+    attrs.sender ??= { dv: 1, fingerprint: faker.random.alphaNumeric(8) }
+    attrs.phone ??= createTestPhone()
```
sitozzz, line 116: What is the reason for using conditional assignment in this case?

.github/workflows/waitForLocalhostApiReady.sh
```diff
@@ -11,19 +11,3 @@ while [[ "${API_RESULT}" != *"appVersion"* && "${API_RESULT}" != "" && ((${API_W
   API_RESULT=$(curl -s 'http://localhost:3000/admin/api' --data-raw '{"operationName":null,"variables":{},"query":"{appVersion}"}' -H 'content-type: application/json' | tr -d ' \n')
   ((API_WAIT_TIMEOUT-=${API_SLEEP_INTERVAL}))
 done
-
-AUTH_USERNAME=admin@example.com
```
pahaz, line 15: We need to wait until the `dev` command completes.

packages/cli/index.js
```diff
@@ -183,21 +183,24 @@ async function prepareMinimalAppEnv (appName, dbName, redisName, port, sport, se
     await updateAppEnvFile(appName, 'COOKIE_SECRET', `${appName}-secret`)
 }
 
-async function prepareAppEnvLocalAdminUsers (appName) {
+async function prepareAppEnvLocalAdminUsers (appName, identity = 'email') {
     // TODO(pahaz): check create-user.js logic!
-    const justUserEmail = await getAppEnvValue(appName, 'DEFAULT_TEST_USER_IDENTITY') || 'user@example.com'
+    const justUserDefaultIdentity = identity === 'email' ? 'user@example.com' : '+79067777777'
```
pahaz, line 188: It's better to avoid any hardcoded credentials

packages/cli/index.js
```diff
@@ -183,21 +183,24 @@ async function prepareMinimalAppEnv (appName, dbName, redisName, port, sport, se
...
+    const justUserIdentity = await getAppEnvValue(appName, 'DEFAULT_TEST_USER_IDENTITY') || justUserDefaultIdentity
     const justUserPassword = await getAppEnvValue(appName, 'DEFAULT_TEST_USER_SECRET') || getRandomString()
     const justUserOpts = JSON.stringify({ password: justUserPassword, isAdmin: false, name: 'JustUser' })
-    await safeExec(`yarn workspace @app/${appName} node ./bin/create-user.js ${JSON.stringify(justUserEmail)} ${JSON.stringify(justUserOpts)}`)
-    const adminUserEmail = await getAppEnvValue(appName, 'DEFAULT_TEST_ADMIN_IDENTITY') || 'admin@example.com'
+    await safeExec(`yarn workspace @app/${appName} node ./bin/create-user.js ${JSON.stringify(justUserIdentity)} ${JSON.stringify(justUserOpts)}`)
+
+    const adminDefaultIdentity = identity === 'email' ? 'admin@example.com' : '+79068888888'
```
pahaz, line 194: it's better to raise an error if we can't find any required data

packages/cli/index.js
SavelevMatthew: It's the same behaviour as was before.
Before: 
1. took email from `.env` or use default
Now:
1. If auth is email use this default, or this default otherwise
2. Use identity from `.env` or use default

packages/cli/index.js
SavelevMatthew: It's not required. In case, where you're launching `prepare.js` for a first time, there'll be no identities in `.env`.
So the app will use default values (from initialData in old good times) and then update `.env`

apps/dev-api/domains/user/utils/testSchema/index.js
SavelevMatthew: It's just a bit different syntax from old one. (Instead of):
```
function createTestSomething(extraAttrs = {}) {
    // generate a lot of data
    const attr = {
        // a lot of duplicated data
        ...extraAttrs
    }
}
```

apps/dev-api/domains/common/utils/sms.js
SavelevMatthew: The reason for that is dev portal API not works like usual app (dev stand uses dev condo, prod uses prod condo).

In reality dev-api uses both stands at the same time. Prefix condo needed to specify that `condo` is a provider of SMS.

(Maybe we'll want to create a separate SMSC for this, so it will became SMSC-dev and SMSC-prod or smth like this)

apps/dev-api/bin/create-user.js
```diff
@@ -0,0 +1,50 @@
...
+
+const BASIC_PHONE_REGEX = /^\+\d+$/
+
+function getJSON (opts) {
+    try {
+        return JSON.parse(opts)
+    } catch {
+        return undefined
+    }
+}
```
sitozzz, line 17: IMO it is little excessively to define method for single usage. Here it's okay to put the content of this function inside `main`

apps/dev-api/domains/common/utils/serverClients.js
```diff
@@ -0,0 +1,60 @@
...
+
+function _getClientRequisites (config) {
+    return {
+        endpoint: get(config, 'apiUrl', 'http://localhost:4004'),
+        authRequisites: {
+            email: get(config, 'email', 'dev-api-bot@example.com'),
+            password: get(config, 'password', 'your-secret-pass'),
+        },
+    }
+}
```
sitozzz, line 20: Why not disable the use of a function without environment variables?

sitozzz: Instead I would prefer to throw an error

apps/dev-api/domains/common/utils/serverClients.js
```diff
@@ -0,0 +1,60 @@
...
+
+    async sendMessage (to, message) {
+        await this.executeAuthorizedMutation({
+            mutation: SEND_MESSAGE_MUTATION,
+            variables: {
+                data: {
+                    ...this.dvSender(),
+                    to,
+                    type: 'DEV_PORTAL_MESSAGE',
+                    lang: this.locale,
```
sitozzz, line 36: Does `this.locale` had a fallback to default value?

SavelevMatthew: In this case it's not important. If defines from which template to pick a message in condo.
But we've already formed a localised message inside dev-api. So template in condo in all languages  looks like {{meta.body}}, so it's not important.

apps/dev-api/bin/create-user.js
SavelevMatthew: Just `Ctrl-C+V`'ed that from condo, but yeah...

sitozzz (approved):

apps/dev-api/domains/common/utils/serverClients.js
SavelevMatthew: I guess we should think about `.env` validation outside of this PR

packages/codegen/index.js
```diff
@@ -223,7 +223,7 @@ function createapp (argv) {
         .coerce('name', opt => {
             let name = opt.toLowerCase()
             if (name.length < 3) throw new Error('<name> is too short!')
-            if (!/^[a-z_][a-z0-9_]+$/.test(name)) throw new Error('<name> should be [a-z0-9_]+ string')
+            if (!/^[a-z_][a-z0-9-]+$/.test(name)) throw new Error('<name> should be [a-z0-9-]+ string')
```
nomerdvadcatpyat, line 226:
```suggestion
            if (!/^[a-z][a-z0-9-]+$/.test(name)) throw new Error('<name> should be [a-z0-9-]+ string')
```
изображение

nomerdvadcatpyat (approved):

packages/codegen/index.js
nomerdvadcatpyat: or is this expected behavior?

apps/dev-api/domains/user/utils/testSchema/index.js
```diff
@@ -0,0 +1,214 @@
...
+        const { code } = await ConfirmPhoneAction.getOne(adminClient, {id: actionId})
+
+        await completeConfirmPhoneActionByTestClient(actionId, {
+            sender,
+            code,
+        }, client)
+    }
+
+
+    const { data, errors } = await client.mutate(REGISTER_NEW_USER_MUTATION, {
```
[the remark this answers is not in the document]
SavelevMatthew: There're few reasons for that:
1. I didn't want to make `ConfirmPhoneAction` visible for anyone initially. Some tests requires me to view and update it as admin, so I've added `read` and `update specific fields` option to it. Creating it by bypassing `startConfirmPhoneAction` mutation should be prohibited.
2. I want tests to be as close to real-life scenarios as possible, that's why I've decided to recreate registration in-depth (instead of recreating complex step from admin user)

packages/codegen/index.js
SavelevMatthew: fixed

PR #3879, opened 2023-09-20
DOMA-7216 added ticket domain permission settings

nomerdvadcatpyat:
Added tickets and incidents permissions to permissions (read/manage) table in settings


sitozzz (approved):

apps/condo/domains/ticket/access/TicketComment.js
```diff
@@ -48,8 +48,8 @@ async function canReadTicketComments ({ authentication: { item: user } }) {
         ticket: {
             organization: {
                 OR: [
-                    queryOrganizationEmployeeFor(user.id),
```
Alllex202, line 51: Should there be a `canReadTicketComments` permission here?
In function `checkManageCommentAccess` we use `canManageTicketComments`

apps/condo/domains/ticket/access/TicketCommentFile.js
```diff
@@ -42,8 +42,8 @@ async function canReadTicketCommentFiles ({ authentication: { item: user } }) {
             {
                 organization: {
                     OR: [
-                        queryOrganizationEmployeeFor(user.id),
```
Alllex202, line 45: same

apps/condo/domains/ticket/access/TicketFilterTemplate.js
```diff
@@ -13,7 +13,7 @@ async function canReadTicketFilterTemplates ({ authentication: { item: user } })
     if (user.isAdmin) return {}
 
     return {
-        employee: { organization: { ...queryOrganizationEmployeeFor(user.id) } },
```
Alllex202, line 16: 1) Is permission check necessary here?
2) If necessary, is it necessary to check this right when creating/updating?

apps/condo/domains/ticket/access/UserFavoriteTicket.js
```diff
@@ -21,8 +21,8 @@ async function canReadUserFavoriteTickets ({ authentication: { item: user } }) {
         user: { id: user.id },
         organization: {
             OR: [
-                queryOrganizationEmployeeFor(user.id),
```
Alllex202, line 24: same
1) Is permission check necessary here?
2) If necessary, is it necessary to check this right when creating/updating?

apps/condo/domains/ticket/access/TicketComment.js
nomerdvadcatpyat: I think it's redundant to add `canReadTicketComments` because we can not be able to manage it from ui.
If we need it, we’ll add

Alllex202 (commented):
How about hiding or blocking controls that should only be accessible with manage rights on pages with read rights? (Update status, edit button, etc...)

apps/condo/domains/ticket/access/TicketFilterTemplate.js
nomerdvadcatpyat: added checks, updated tests

Alllex202 (commented):
There are also buttons for creating when a table is empty.

And you need to check that everything works well with importing tickets

Alllex202 (approved):

AntonAL (approved):

PR #3880, opened 2023-09-20
feat(condo): DOMA-7214 added approved flag for organizations

Alllex202:
1) Added `canManageOrganizations` flag for `UserRightsSet`

2) Added `Organization.isApproved` field. 
It can be read any user who can get organization.
But update this field can only admin, support or user with special rights


toplenboren (approved):

apps/condo/domains/organization/access/Organization.js
```diff
@@ -76,10 +76,15 @@ async function canReadOrganizations ({ authentication: { item: user }, listKey }
...
-async function canManageOrganizations ({ authentication: { item: user }, operation }) {
+async function canManageOrganizations ({ authentication: { item: user }, operation, listKey, originalInput }) {
     if (!user) return throwAuthenticationError()
     if (user.deletedAt) return false
     if (user.isSupport || user.isAdmin) return true
+    const hasDirectAccess = await canDirectlyManageSchemaObjects(user, listKey)
+    if (hasDirectAccess) return true
+
+    // NOTE: The "isApproved" field can only be managed by the admin, support, users with special rights.
+    if ('isApproved' in originalInput) return false
```
SavelevMatthew, line 87: Should we add a common fields-access function? Seems like we need to protect many fields (like importId, importRemoteSystem and so on)

apps/condo/domains/organization/schema/Organization.js
```diff
@@ -156,6 +156,20 @@ const Organization = new GQLListSchema('Organization', {
             kmigratorOptions: { null: true, unique: false },
         },
         features: ORGANIZATION_FEATURES_FIELD,
+
+        isApproved: {
+            // NOTE: This field is used for analytics, you should not implement logic using it
+            // By default, all organizations are set to true.
+            // If the organization is test or fraudulent, then you need to set value to false.
+            schemaDoc: '@deprecated' +
```
SavelevMatthew, line 164: Born deprecated? 😆

apps/condo/domains/organization/schema/Organization.test.js
```diff
@@ -364,6 +366,66 @@ describe('Organization', () => {
...
+                expect(createdO10n.isApproved).toBeFalsy()
+
+                const o10n = await Organization.getOne(support, { id: createdO10n.id })
+                expect(o10n.isApproved).toBeFalsy()
+
+                const [updatedO18n] = await updateTestOrganization(support, createdO10n.id, { isApproved: true })
+                expect(updatedO18n.isApproved).toBeTruthy()
+            })
+
+            test('User with custom access: can register organization, read and update field "isApproved"', async () => {
```
SavelevMatthew, line 393: What? User with custom access cannot create organizations

apps/condo/domains/organization/access/Organization.js
SavelevMatthew: We should block organization creating here. Since the only source of organization should be `registerNewOrganization` mutation

apps/condo/domains/organization/schema/Organization.js
Alllex202: in order to scare away particularly persistent developers who want to use this field :)

apps/condo/domains/organization/access/Organization.js
Alllex202: > Should we add a common fields-access function? Seems like we need to protect many fields (like importId, importRemoteSystem and so on)

It seems our fields are protected. But not everything is possible

apps/condo/domains/organization/access/Organization.js
```diff
@@ -76,18 +76,23 @@ async function canReadOrganizations ({ authentication: { item: user }, listKey }
...
 
-    if (operation === 'create') {
-        return false
-    } else if (operation === 'update') {
-        // user is inside employee list and is not blocked
-        return {
-            employees_some: { user: { id: user.id }, role: { canManageOrganization: true }, isBlocked: false, deletedAt: null },
-        }
+    // You should use "registerNewOrganization"
+    if (operation === 'create') return false
```
SavelevMatthew, line 85: I guess it's should be true for admin and supports as well...

Alllex202: A complete access update involves some refactoring, let's create a separate task for this?

Alllex202: I don’t want a small task about adding a new field to turn into a huge task with refactoring :)

Alllex202: Now there is a task DOMA-7240 for this

apps/condo/domains/organization/access/Organization.js
```diff
@@ -76,18 +76,23 @@ async function canReadOrganizations ({ authentication: { item: user }, listKey }
...
+
+    const hasDirectAccess = await canDirectlyManageSchemaObjects(user, listKey)
+    if (hasDirectAccess) return true
+
+    // NOTE: The "isApproved" field can only be managed by the admin, support, users with special rights.
+    if ('isApproved' in originalInput) return false
+
+    // user is inside employee list and is not blocked
+    return {
+        employees_some: { user: { id: user.id }, role: { canManageOrganization: true }, isBlocked: false, deletedAt: null },
```
nomerdvadcatpyat, line 95: you can use `queryOrganizationEmployeeFor` for it

nomerdvadcatpyat (approved):

apps/condo/domains/organization/access/Organization.js
Alllex202: This function does not filter by role

PR #3882, opened 2023-09-22
feature(condo): DOMA-7107 getBankInfo and getOrganizationInfo

dkoviazin:
added dadata wrapper to get information about bank and organization


packages/clients/finance-info-client/FinanceInfoClient.js
```diff
@@ -0,0 +1,177 @@
+const { get } = require('lodash')
+const fetch = require('node-fetch')
+
+const conf = require('@open-condo/config')
+const { getRedisClient } = require('@open-condo/keystone/redis')
+
+const { apiToken: API_KEY } = conf['ADDRESS_SUGGESTIONS_CONFIG'] ? JSON.parse(conf['ADDRESS_SUGGESTIONS_CONFIG']) : {}
```
AleX83Xpert, line 7: Direct queries to dadata? Why do not use the address service?

AleX83Xpert (approved):

packages/clients/finance-info-client/FinanceInfoClient.js
dkoviazin: Do address service supports such queries as 
```
/findById/party
/findById/bank
```

PR #3884, opened 2023-09-22
fix(condo): DOMA-7243 fix calc isPayable for receipts in AllResidentBillingReceiptsService.js

VKislov:


dkoviazin (commented):
You need to add tests for this case

dkoviazin (commented):
It's hard to understand what case you solved with this fix

VKislov:
> It's hard to understand what case you solved with this fix


203: `const existingRecipientPeriod = dayjs(get(receiptsByAccountAndRecipient[key], 'period'), 'YYYY-MM-DD')`

Here you can see that they are trying to get the `period` from the `receiptsByAccountAndRecipient` object. Moreover, in this object the `key` is a string of the format `const key = accountNumber + '-' + recipientId + '-' + categoryId`
and the value is `receipt.id` . If you try to get a non-existent period value, `dayjs` will get `undefined` and it will return `invalidDate` , not the correct date. Naturally, the subsequent comparison of invalidDate and valid date will be incorrect

apps/condo/domains/billing/schema/AllResidentBillingReceiptsService.js
```diff
@@ -195,18 +195,18 @@ const AllResidentBillingReceiptsService = new GQLCustomSchema('AllResidentBillin
...
                     // If we have a receipt with later period -- we take it
                     const existingRecipientPeriod = dayjs(get(receiptsByAccountAndRecipient[key], 'period'), 'YYYY-MM-DD')
                     if (existingRecipientPeriod < period) {
-                        receiptsByAccountAndRecipient[key] = receipt.id
+                        receiptsByAccountAndRecipient[key] = { id: receipt.id, period }
                     }
                 }
 
-                const payableReceipts = Object.values(receiptsByAccountAndRecipient)
+                const payableReceipts = (Object.values(receiptsByAccountAndRecipient)).map(r => r.id)
```
sitozzz, line 209:
```suggestion
                const payableReceipts = Object.values(receiptsByAccountAndRecipient).map(r => r.id)
```

sitozzz (approved):

PR #3889, opened 2023-09-25
DOMA-7244 add addressService plugin into B2CAppProperty

sitozzz:


apps/condo/domains/miniapp/schema/B2CAppProperty.js
```diff
@@ -2,13 +2,13 @@
  * Generated by `createschema miniapp.B2CAppProperty 'app:Relationship:B2CApp:PROTECT; address:Text;' --force`
  */
 
-const { Text, Relationship } = require('@keystonejs/fields')
+const { Relationship } = require('@keystonejs/fields')
```
SavelevMatthew, line 5: Can you please replace field with string representation to be KS6 compatible?
`type: Relationship` -> `type: 'Relationship'`

apps/condo/domains/miniapp/schema/B2CAppProperty.js
```diff
@@ -37,37 +37,43 @@ const B2CAppProperty = new GQLListSchema('B2CAppProperty', {
...
+                            return addFieldValidationError(INCORRECT_ADDRESS_ERROR)
+                        }
+
+                        const suggestionAddress = get(resolvedData, ['addressMeta', 'value'], '')
+                        const suggestionHouseType = get(resolvedData, ['addressMeta', 'data', 'house_type_full'])
+
+                        if (!VALID_HOUSE_TYPES.includes(suggestionHouseType)) {
+                            return addFieldValidationError(`${INCORRECT_HOUSE_TYPE_ERROR}. Valid values are: [${VALID_HOUSE_TYPES.join(', ')}]`)
+                        }
+                        if (operation === 'create' && suggestionAddress.toLowerCase() !== inputAddress.toLowerCase()) {
```
SavelevMatthew, line 64: Why did this check become create-only?

apps/condo/domains/miniapp/schema/B2CAppProperty.js
```diff
@@ -37,37 +37,43 @@ const B2CAppProperty = new GQLListSchema('B2CAppProperty', {
...
+                        const inputAddress = resolvedData[fieldPath]
+                        if (!inputAddress) {
+                            return addFieldValidationError(INCORRECT_ADDRESS_ERROR)
+                        }
+
+                        const suggestionAddress = get(resolvedData, ['addressMeta', 'value'], '')
+                        const suggestionHouseType = get(resolvedData, ['addressMeta', 'data', 'house_type_full'])
+
+                        if (!VALID_HOUSE_TYPES.includes(suggestionHouseType)) {
+                            return addFieldValidationError(`${INCORRECT_HOUSE_TYPE_ERROR}. Valid values are: [${VALID_HOUSE_TYPES.join(', ')}]`)
```
SavelevMatthew, line 62: Should it be replaced with GQLError?)

apps/condo/domains/miniapp/utils/testSchema/index.js
```diff
@@ -496,13 +498,24 @@ async function updateTestB2BAppAccessRightSet (client, id, extraAttrs = {}) {
...
+
+    let invalidAddress
+
+    if (!validAddress) {
+        invalidAddress = address.replace(/д\s\d+\s/gm, '')
+    }
+
+    if (!validHouse) {
+        invalidAddress = address.replace('д', 'б')
+        addressMeta.data.house_type_full = 'б'
```
AleX83Xpert, line 515: `'б'` looks like `house_type`

apps/condo/domains/miniapp/schema/B2CAppProperty.js
AleX83Xpert: Looks like this code was moved as is :)

apps/condo/domains/miniapp/schema/B2CAppProperty.js
AleX83Xpert: Agree with the question asked.
What if we change the address?

AleX83Xpert (approved):

SavelevMatthew (approved):

apps/condo/domains/miniapp/schema/B2CAppProperty.js
```diff
@@ -37,37 +52,47 @@ const B2CAppProperty = new GQLListSchema('B2CAppProperty', {
...
+                // NOTE: address string in the lower case is required for address + app unique constraint
+                return {
+                    address: address.toLowerCase(),
+                    ...restAddressFields,
+                }
+            },
+            fieldsHooks: {
+                address: {
+                    validateInput: ({ resolvedData, fieldPath, context }) => {
+                        const inputAddress = resolvedData[fieldPath]
```
Alllex202, line 72: `resolvedData['address']` already contains a value from the address service.

Below you check whether the input data matches the data from the address service:
`if (suggestionAddress.toLowerCase() !== inputAddress.toLowerCase())`

But since the value in `resolvedData` was updated in the plugin, it will always be true. And the input values will not be checked

PR #3894, opened 2023-09-25
fix(condo): DOMA-7225 fix file names in utf8

abshnko:


apps/condo/domains/common/utils/sberCloudFileAdapter.js
```diff
@@ -87,7 +87,7 @@ class SberCloudFileAdapter {
     save ({ stream, filename, id, mimetype, encoding, meta = {} }) {
         const fileData = {
             id,
-            originalFilename: filename,
+            originalFilename: Buffer.from(filename, 'latin1').toString('utf8'),
```
SavelevMatthew, line 90: How do you know, that incoming filename had `latin1` encoding?

apps/condo/domains/ticket/schema/TicketFile.js
```diff
@@ -43,6 +43,12 @@ const TicketFile = new GQLListSchema('TicketFile', {
                 await Adapter.delete(existingItem.file)
             }
         },
+        resolveInput: async ({ resolvedData }) => {
+            if (resolvedData['file']) {
+                resolvedData['file'].originalFilename = Buffer.from(resolvedData['file'].originalFilename, 'latin1').toString('utf-8')
```
sitozzz, line 48: First of all you should check the original encoding

apps/condo/domains/common/utils/sberCloudFileAdapter.js
dkoviazin: You can use chardet as we already have it in project

apps/condo/domains/ticket/schema/TicketFile.js
abshnko: i checked the incoming endocing using chardet and it can be different (ascii, utf8 etc), but when i tried converting it to buffer and specifying the correct encoding the result was incomprehensible.
then i tried converting different encodings specifying 'latin1' and it worked every time. i checked it on ascii, unicode, latin1, 7bit

pahaz: it looks suspicious

apps/condo/domains/common/utils/fixFileNameEncoding.js
```diff
@@ -0,0 +1,9 @@
+const iconv = require('iconv-lite')
+
+const getFixedFileNameEncoding = (filename) => {
+    return iconv.decode(Buffer.from(filename, 'binary'), 'utf-8')
+}
+
+module.exports = {
+    getFixedFileNameEncoding,
```
dkoviazin, line 8: name means "get encoding" 
maybe: fixFileNameEncoding or convertFileNameToUTF8

dkoviazin (approved):

Alllex202 (approved):

PR #3915, opened 2023-09-29
feat(condo): DOMA-7297 added ability to directly manage fields

Alllex202:
## Basic logic

<table>
  <tr>
    <th>canManage&lt;Schema&gt;</th>
    <th>canManage&lt;Schema&gt;&lt;Fields&gt;Field</th>
    <th>User capabilities</th>
  </tr>
  <tr>
    <td>✅</td>
    <td>✅</td>
    <td>Can direct create and update objects</td>
  </tr>
  <tr>
    <td>✅</td>
    <td>❌</td>
    <td>Can direct create and update objects but without using &lt;Fields&gt; field in payload</td>
  </tr>
  <tr>
    <td>❌</td>
    <td>✅</td>
    <td>Cannot direct create objects. <br/>But can only direct update &lt;Fields&gt; field in objects</td>
  </tr>
  <tr>
    <td>❌</td>
    <td>❌</td>
    <td>Cannot direct create or update objects</td>
  </tr>
</table>


apps/condo/domains/organization/access/Organization.js
```diff
@@ -87,6 +88,9 @@ async function canManageOrganizations ({ authentication: { item: user }, operati
     const hasDirectAccess = await canDirectlyManageSchemaObjects(user, listKey)
     if (hasDirectAccess) return true
 
+    const hasDirectAccessToFieldUpdating = await canDirectlyManageSchemaFields(user, listKey, originalInput, operation)
```
SavelevMatthew, line 91: Maybe we should merge this with `canDirectlyManageSchemaObjects`?

apps/condo/domains/user/utils/directAccess/common.utils.js
```diff
@@ -26,6 +44,10 @@ function generateManageSchemaFieldName (schemaName) {
     return `canManage${pluralize.plural(schemaName)}`
 }
 
+function generateFieldNameToManageField (schemaName, fieldName) {
+    return `canManage${schemaName}${fieldName.slice(0, 1).toUpperCase() + fieldName.slice(1)}Field`
```
SavelevMatthew, line 48: There's `_capitalize` function in the same file :)

apps/condo/domains/user/utils/directAccess/server.utils.js
```diff
@@ -90,7 +111,26 @@ async function canDirectlyReadSchemaObjects (user, schemaName) {
...
+    return await _hasSpecificRights(user, [generateManageSchemaFieldName(schemaName)])
+}
+
+const FIELD_NAMES_TO_SKIP_ACCESS = ['dv', 'sender']
+
+async function canDirectlyManageSchemaFields (user, schemaName, originalInput, operation) {
+    const isUpdateOperation = operation === 'update'
+    const fieldNamesWithAccess = getFieldNamesBySchemaName(schemaName)
+
+    if (isUpdateOperation && isArray(fieldNamesWithAccess) && !isEmpty(fieldNamesWithAccess)) {
```
SavelevMatthew, line 123: Why `isUpdate...` filter here?

Alllex202: You can only update a field, you cannot create or delete it (as a schema object). 

Even if there is only one field in schema, then you have access to control only this one field, and not entire schema object

apps/condo/domains/organization/access/Organization.js
Alllex202: Will it be so obvious that function `canDirectlyManageSchemaObjects` checks not only direct access to entire schema, but also direct access only to some of its fields?

apps/condo/domains/user/utils/directAccess/server.utils.js
SavelevMatthew: Seems like I don't get it:
1. If fields-access function resolves to `false` in `create` operation, then it means that I cannot pass this fields in `create` mutation.
2. So, what if I have schema `A`, and `canManageA: true` and `canManageAField: true`? Seems like I should be able to create entities of `A` with `Field`, shouldn't I?

Alllex202: You should check accesses in the following sequence:
1) `canDirectlyManageSchemaObjects` - Direct access to object management (creating/deleting an object and updating its fields)
2) `canDirectlyManageSchemaFields` - Direct access to updating object fields
3) Default accesses

It looks like it's really worth combining these functions together so that the order doesn't change

- If you have `canManageA: true`, then it doesn’t matter what value you have specified for `canManageAField`. You can create/update/delete objects.

- If you have `canManageA: false` and `canManageAField: true`, then you can only update the specified fields.

- If you have `canManageA: false` and `canManageAField: false`, then you will not have direct access to any actions that bypass standard checks

apps/condo/domains/user/schema/UserRightsSet.test.js
```diff
@@ -387,152 +409,532 @@ describe('UserRightsSet', () => {
...
+                                await updateTestOrganization(executor, org.id, { name: faker.company.name(), isApproved: false })
+                            })
+                            await expectToThrowAccessDeniedErrorToObj(async () => {
+                                await updateTestOrganization(executor, org.id)
+                            })
+                        })
+                    })
+                })
+
+                test('Employee with custom access: can read and update field "isApproved". Cannot update "isApproved" at the same time as other fields', async () => {
```
SavelevMatthew, line 898: It should not be employee

SavelevMatthew (approved):

nomerdvadcatpyat (approved):

PR #3956, opened 2023-10-11
feat(condo): DOMA-7211 fix contractor permissions

nomerdvadcatpyat:
1. Remove redirects from any page to /ticket for Contractor role
2. Added new permissions: `canReadMeters`, `canReadSettings`, `canReadExternalReports`, `canReadServices`, `canReadCallRecords` – `true` for all roles except Contractor.
3. Allow create contacts from tickets if role has `canManageContacts` but has not `canReadContacts` (Contractor case).

Also fix settings table layout


apps/condo/pages/settings/contactRole/[id]/update.tsx
```diff
@@ -43,6 +43,6 @@ const UpdateContactRolePage = () => {
     )
 }
 
-UpdateContactRolePage.requiredAccess = OrganizationRequired
+UpdateContactRolePage.requiredAccess = SettingsReadPermissionRequired
```
Comment (Alllex202) [inline apps/condo/pages/settings/contactRole/[id]/update.tsx:46]: Should there be a `canManageContactRoles` check here?

apps/condo/pages/settings/hint/create.tsx
```diff
@@ -37,6 +37,6 @@ const CreateTicketPropertyHintPage = () => {
     )
 }
 
-CreateTicketPropertyHintPage.requiredAccess = OrganizationRequired
+CreateTicketPropertyHintPage.requiredAccess = SettingsReadPermissionRequired
```
Alllex202, line 40: Should there be a `canManageTicketPropertyHints` check here?

And should there be similar checks on other settings pages if they are needed to manage objects?

apps/condo/pages/settings/contactRole/create.tsx
```diff
@@ -37,6 +37,6 @@ const CreateContactRolePage = () => {
     )
 }
 
-CreateContactRolePage.requiredAccess = OrganizationRequired
+CreateContactRolePage.requiredAccess = SettingsReadPermissionRequired
```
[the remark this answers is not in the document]
nomerdvadcatpyat: When we add (and if add) these permissions to the settings table, then we update the access to the pages.

apps/condo/pages/settings/hint/create.tsx
nomerdvadcatpyat: When we add (and if add) these permissions to the settings table, then we update the access to the pages.

apps/condo/pages/reports/index.tsx
```diff
@@ -41,13 +42,18 @@ const IndexPage = () => {
...
         if (isDashboardEnabled && canReadAnalytics) {
             return <Dashboard organizationId={organization.id} />
         }
 
         const isEmptyReports = isEmpty(externalReports)
+        const canReadExternalReports = get(link, ['role', 'canReadExternalReports'], false)
+        
+        if (!canReadExternalReports) {
+            return <AccessDeniedPage />
+        }
```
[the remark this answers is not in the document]
nomerdvadcatpyat: Access to this page provides 2 permissions: `canReadAnalytics` for new analytics and `canReadExternalReports` for metabase reports (if no `canReadAnalytics` or featureflag).
Soon, the metabase reports should go away and leave only a new analytics (and permission `canReadExternalReports` will be removed), so it doesn’t matter

pahaz (approved):

PR #3958, opened 2023-10-11
refactor(condo): DOMA-7396 refactored importer

Alllex202:
Before: any unexpected error terminates the file import without report back.
After: an unexpected error will not abort the file import. The row with the error is added to report with all errors

fix(condo): DOMA-7386 fixed a import report. Fixed a validation for dates 
Before: If the import file contains cells with a date format, then in the report they were filled in in this format: `Fri Oct 20 2023 13:17:02 GMT+0500 (GMT+05:00)`
After: Now such cells are returned in the format: `20.10.2023`

fix(condo): DOMA-7394 fixed empty errors in a import report

fix(condo): DOMA-7395 fixed a meter import validation


nomerdvadcatpyat (approved):

apps/condo/domains/common/utils/importer.ts
```diff
@@ -172,98 +185,102 @@ export class Importer implements IImporter {
...
+            } catch (e) {
+                const mutationErrors = get(e, 'graphQLErrors', []) || []
+
+                if (!isArray(mutationErrors) || isEmpty(mutationErrors)) {
+                    console.error('Unexpected error in "objectCreator"!')
+                    console.error(e)
+                    processedRow.errors.push(this.errors.creation)
+                } else {
+                    for (const mutationError of mutationErrors) {
+                        const mutationErrorMessages = get(mutationError, ['data', 'messages'], []) || []
```
VKislov, line 257: it seems `get(mutationError, ['data', 'messages'], [])` is sufficient

apps/condo/domains/meter/hooks/useImporterFunctions.ts
```diff
@@ -207,83 +221,65 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
...
-            console.error('meterReadingNormalizer error')
-            console.error(error)
-            console.error(row)
+        // Used tell whether suggestion API has found specified address at all
+        const properties = await searchPropertyWithMap(client, {
+            organization: { id: userOrganizationIdRef.current },
+            address_i: suggestion.value,
+        }, undefined)
+
+        const propertyId = !isEmpty(properties) ? get(properties[0], 'id') : null
```
VKislov, line 251: it seems like this will work too
```suggestion
        const propertyId = get(properties[0], 'id',  null)
```

apps/condo/domains/meter/hooks/useImporterFunctions.ts
```diff
@@ -207,83 +221,65 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
...
+        const propertyMap = !isEmpty(properties) ? get(properties[0], 'map') : null
```
VKislov, line 252: same as above

apps/condo/domains/meter/hooks/useImporterFunctions.ts
```diff
@@ -207,83 +221,65 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
...
 
+        const meterOptions = await searchMeter(client, searchMeterWhereConditions, SortMetersBy.CreatedAtDesc)
+        addons.meterId = meterOptions.length > 0 ? meterOptions[0].value : null
+
         return { row, addons }
     }
 
-    const meterReadingValidator: RowValidator = (processedRow) => {
-        if (!processedRow) return Promise.resolve(false)
+    const meterReadingValidator: RowValidator = async (processedRow) => {
```
VKislov, line 274: what is `async` for? I didn't see `await` or `promise` inside

VKislov (approved):

apps/condo/domains/common/utils/importer.ts
Alllex202: `get` replace only `undefined` value. If this value is `null`, then the replacement will not occur.
It seems that you can do without `|| []`, but who knows what the answer might come)

apps/condo/domains/meter/hooks/useImporterFunctions.ts
Alllex202: I just moved the existing code)

VKislov: I’ve already approved it, correct it if you want, leave it if you don’t) I don’t block pull requests

apps/condo/domains/meter/hooks/useImporterFunctions.ts
Alllex202: `RowValidator` have `Promise` type
Снимок экрана 2023-10-20 в 16 47 26

PR #4021, opened 2023-10-26
fix(condo): DOMA-7560 fix notification about meter readings and receipts

VKislov:
The first commit fixes the problem with connected but deleted/missing `billingContext` in `ServiceConsumer`. Since the deprecated field no one fills it with current data and it interferes with the correct search for SC instances.

the second commit fixes the case when at 3 addresses with different beginnings of the period, but the same end of the period, also for the reverse case


Alllex202 (commented):
What about tests? Looks like they don't cover all cases

VKislov:
> What about tests? Looks like they don't cover all cases

tests already cover this case.

dkoviazin (approved):

PR #4033, opened 2023-10-31
fix(condorb): DOMA-7574 add scope to accept schema in rb

VKislov:


dkoviazin (approved):

Alllex202 (commented):
Accesses will not work automatically! To do this, you need to explicitly add one of the utilities from `@condo/domains/user/utils/directAccess` to the model accesses

apps/condo/domains/marketplace/access/InvoiceContext.js
```diff
@@ -12,13 +12,19 @@ const {
...
-async function canReadInvoiceContexts ({ authentication: { item: user } }) {
+
+async function canReadInvoiceContexts ({ authentication: { item: user }, listKey }) {
     if (!user) return throwAuthenticationError()
     if (user.deletedAt) return false
     if (user.isAdmin || user.isSupport) return {}
 
+    if (user.type === SERVICE) {
+        return await canDirectlyReadSchemaObjects(user, listKey)
+    }
```
Alllex202, line 26: I don’t think it’s worth adding restrictions by user type for this check

These are accesses like support, only with flexible settings

```suggestion
    const hasDirectAccess = await canDirectlyReadSchemaObjects(user, listKey)
    if (hasDirectAccess) return true
```

apps/condo/domains/marketplace/access/InvoiceContext.js
```diff
@@ -30,12 +36,16 @@ async function canReadInvoiceContexts ({ authentication: { item: user } }) {
...
-async function canManageInvoiceContexts ({ authentication: { item: user }, originalInput, operation, itemId }) {
+async function canManageInvoiceContexts ({ authentication: { item: user }, originalInput, listKey, operation, itemId }) {
     if (!user) return throwAuthenticationError()
     if (user.deletedAt) return false
     if (user.isAdmin || user.isSupport) return true
     if (user.type === RESIDENT) return false
 
+    if (user.type === SERVICE) {
+        return await canDirectlyManageSchemaObjects(user, listKey, originalInput, operation)
+    }
```
Alllex202, line 47: ^^ same

Alllex202 (approved):

PR #4041, opened 2023-11-01
fix(condo): DOMA-7604 fix `TicketPropertyHintProperty` duplicates

nomerdvadcatpyat:
1) In `TicketPropertyHintProperty` there was a constraint on the uniqueness of `property` and `ticketPropertyHint`, which was incorrect, since we cannot create two `propertyHints` on one `property`. Fixed it, change constraint to only `property` uniqueness and remove duplicates in migration.
2) In create ticket hint form there is a properties select, which should not show addresses to which already created `TicketHint`. Since `useObjects` was used there, only the first 100 `PropertyHintProperties` were searched. Changed it to `useAllObjects`


apps/condo/domains/ticket/schema/TicketPropertyHintProperty.test.js
Alllex202, line 1: rename test

apps/condo/migrations/20231101165212-0339_auto_20231101_1152.js
```diff
@@ -0,0 +1,56 @@
...
+
+UPDATE "TicketPropertyHintProperty" AS t1
+SET "deletedAt" = NOW()
+WHERE EXISTS (
+  SELECT 1
+  FROM "TicketPropertyHintProperty" AS t2
+  WHERE t1.property = t2.property
+  AND t1."deletedAt" IS NULL
+  AND t2."deletedAt" IS NULL
+  AND t1.id > t2.id
```
Alllex202, line 20: What is the condition `t1.id > t2.id`?

Alllex202 (approved):

PR #4052, opened 2023-11-07
feat(condo): DOMA-7555 meter resource owner UI changes

sitozzz:


apps/condo/domains/meter/schema/Meter.js
```diff
@@ -212,33 +220,49 @@ const Meter = new GQLListSchema('Meter', {
...
+                    throw new GQLError(ERRORS.METER_RESOURCE_OWNED_BY_ANOTHER_ORGANIZATION, context)
+                }
+
+            }
         },
         afterChange: async ({ context, operation, originalInput, updatedItem }) => {
-            if (operation === 'create') {
-                const property = await Property.getOne(context, {
-                    id: updatedItem.property,
+            let hasMeterResourceOwnership = false
```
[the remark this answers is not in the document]
sitozzz: That's an artifact. Now logic has changed a little, so I can make it const like you said

apps/condo/domains/meter/schema/Meter.test.js
```diff
@@ -1103,5 +1122,45 @@ describe('Meter', () => {
...
+                    addressKey: client.property.addressKey,
+                })
+
+                expect(meterResourceOwners).toHaveLength(2)
+                expect(meterResourceOwners).toEqual(expect.arrayContaining([
+                    expect.objectContaining({
+                        organization: { id: client.organization.id },
+                        resource: { id: COLD_WATER_METER_RESOURCE_ID },
+                        addressKey: client.property.addressKey,
+                    }),
```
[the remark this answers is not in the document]
sitozzz: I don't think so. It's has a lot of special cases that might be worth to consider.
This is also related to business logic. You can only lose your right to manage resources through support. And the one who first creates `Meter` can acquire this right

apps/condo/domains/meter/schema/Meter.js
```diff
@@ -212,33 +220,50 @@ const Meter = new GQLListSchema('Meter', {
...
-                    if (!hasMeterResourceOwnership) {
-                        await MeterResourceOwner.create(context, {
-                            dv: 1,
-                            sender: originalInput.sender,
-                            address: property.address,
-                            organization: { connect: { id: updatedItem.organization } },
-                            resource: { connect: { id: updatedItem.resource } },
-                        })
+                    if (meterResourceOwner && meterResourceOwner.organization.id !== newItem.organization) {
+                        throw new GQLError(ERRORS.METER_RESOURCE_OWNED_BY_ANOTHER_ORGANIZATION, context)
```
Alllex202, line 237: You should do this check when updating `resource` or `property` too

sitozzz: Great suggestion, fixed that and add test for this case

apps/condo/domains/meter/schema/Meter.test.js
```diff
@@ -1103,5 +1148,45 @@ describe('Meter', () => {
                 expect(meterResourceOwner).toHaveProperty(['sender', 'fingerprint'], meter.sender.fingerprint)
             })
         })
+
+        test('should create MeterResourceOwner after meter resource field update', async () => {
```
Alllex202, line 1152: What about when updating other fields? I didn’t see any restrictions, maybe change test name or update logic? :)

apps/condo/domains/meter/schema/Meter.js
```diff
@@ -212,33 +220,53 @@ const Meter = new GQLListSchema('Meter', {
...
+            }
+        },
+        afterChange: async ({ context, operation, originalInput, updatedItem }) => {
+            const property = await Property.getOne(context, {
+                id: updatedItem.property,
+                deletedAt: null,
+            })
+
+            if (property) {
+                const hasMeterResourceOwnership = await MeterResourceOwner.getOne(context, {
```
Alllex202, line 253:
```suggestion
                const hasMeterResourceOwnership = await MeterResourceOwner.count(context, {
```

Alllex202 (approved):

nomerdvadcatpyat (approved):

PR #4074, opened 2023-11-10
refactor(condo): DOMA-7654 moved server user accesses for b2b apps from plugin to schemas accesses

Alllex202:


apps/condo/domains/contact/access/Contact.js
```diff
@@ -7,32 +7,49 @@ const get = require('lodash/get')
 const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
 const { getById } = require('@open-condo/keystone/schema')
 
+const {
+    serviceUserCanReadSchemaObjectsIfOrganizationConnectedToLinkedB2BApp,
+    serviceUserCanManageSchemaObjectsIfOrganizationConnectedToLinkedB2BApp,
```
SavelevMatthew, line 12: I think util names are too short, may be we should create longer names?)

apps/condo/domains/miniapp/utils/b2bAppServiceUserAccess/utils/server.utils.js
```diff
@@ -0,0 +1,216 @@
...
+}
+
+/**
+ * Merge two access filters or return false
+ *
+ * @param a {any}
+ * @param b {any}
+ * @return {Record<string, any>|{OR: [{AND: Record<string, any>[]},{AND: Record<string, any>[]}]}||false}
+ */
+const mergeAccessFilters = (a, b) => {
```
SavelevMatthew, line 197: Why we need some OR's here?

SavelevMatthew: OR will double db-table scan. 

Can we just figure out once, that it's a service user. and if so, return his access.
Else return staff / resident access

Alllex202: Currently, the plugin combines two queries - default filter + filter by miniapp.

If accesses are not explicitly set for a service account, then it uses as for employee. And if you change the filters everywhere for the service user, then I’m not sure that nothing will break in miniapps/for partners/contractors/integrations

apps/condo/domains/contact/access/Contact.js
Alllex202: I tried to make them shorter, but then it becomes more difficult to understand what is happening in them))

apps/condo/domains/miniapp/utils/b2bAppServiceUserAccess/utils/server.utils.js
SavelevMatthew: We can assume, that all users, linked to apps are of `SERVICE` type

apps/condo/domains/meter/access/Meter.js
```diff
@@ -29,21 +39,29 @@ async function canReadMeters ({ authentication: { item: user } }) {
         }
     }
 
-    return {
+    const accessFilterForB2BAppServiceUser = await b2bAppServiceUserCanReadObjects(args)
+    return mergeAccessFilters(accessFilterForB2BAppServiceUser, {
```
pahaz, line 43: It’s not a good idea to merge. It’s override.

apps/condo/domains/miniapp/utils/b2bAppServiceUserAccess/server.utils.js
```diff
@@ -0,0 +1,217 @@
...
+}
+
+/**
+ * Merge (if it needed) two access filters or return false
+ *
+ * @param a {any}
+ * @param b {any}
+ * @return {Record<string, any>|{OR: [{AND: Record<string, any>[]},{AND: Record<string, any>[]}]}|boolean}
+ */
+const mergeAccessFilters = (a, b) => {
```
pahaz, line 197: It’s not a good idea to use some math to calculate accesses. It should be as match simple as possible. No any transformations and code logic

apps/condo/domains/meter/access/Meter.js
Alllex202: removed this

apps/condo/domains/miniapp/utils/b2bAppServiceUserAccess/server.utils.js
Alllex202: removed this function

apps/condo/domains/miniapp/utils/b2bAppServiceUserAccess/utils/server.utils.js
Alllex202: removed this function

apps/condo/domains/contact/access/Contact.js
```diff
@@ -7,16 +7,28 @@ const get = require('lodash/get')
...
+    const { authentication: { item: user } } = args
 
-async function canReadContacts ({ authentication: { item: user } }) {
     if (!user) return throwAuthenticationError()
     if (user.deletedAt) return false
     
     if (user.isAdmin) return {}
 
+    if (user.type === SERVICE) {
+        return await b2bAppServiceUserCanReadObjects(args)
```
SavelevMatthew, line 29: How about something like this to stay consistent with naming?
```suggestion
        return await canReadObjectsAsB2BAppServiceUser(args)
```

apps/condo/domains/miniapp/utils/b2bAppServiceUserAccess/config.js
```diff
@@ -0,0 +1,52 @@
...
+ * // 5) You should apply new migrations
+ * // run 'yarn migrate'
+ *
+ * // 6) You can also update the "B2BApp permissions for service user > Get all permissions for app" test so that
+ * in "@condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js"
+ *
+ * @type {B2bAppServiceUserAccessConfig}
+ */
+const B2B_APP_SERVICE_USER_ACCESS_AVAILABLE_SCHEMAS = {
+    Contact: {},
```
SavelevMatthew, line 39: Should we decide, hot to sort them in case of growing in the future? Place it by domain / alphabetically, like UserRightsSet for example?

apps/condo/domains/miniapp/utils/b2bAppServiceUserAccess/server.utils.js
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * @param args
+ * @return {Promise<boolean>}
+ */
+const b2bAppServiceUserCanManageObjects = async (args) => {
+    const { listKey } = args
+    if (!isServiceUser(args)) return false
+    const schemaConfig = get(B2B_APP_SERVICE_USER_ACCESS_AVAILABLE_SCHEMAS, listKey)
+    if (!isObject(schemaConfig)) return false
+    const canBeManage = get(schemaConfig, 'canBeManage', true)
```
SavelevMatthew, line 184: I guess this PR is the right place to fix old typo, isn't it?)
```suggestion
    const canBeManaged = get(schemaConfig, 'canBeManaged', true)
```

apps/condo/domains/organization/access/Organization.js
```diff
@@ -65,15 +73,16 @@ async function canReadOrganizations ({ authentication: { item: user }, listKey }
             },
             deletedAt: null,
         })
+
         const serviceOrganizationIds = uniq(billingContexts
             .map(({ organization }) => organization )
             .concat(acquiringContexts.map(({ organization }) => organization )))
             .concat(bankIntegrationOrganizationContext.map(({ organization }) => organization))
+            .concat(get(accessFilterForB2BAppServiceUser, 'id_in', []) || [])
```
SavelevMatthew, line 81: What will happen, if something different will be returned here?

packages/keystone/test.utils.js
```diff
@@ -521,11 +521,11 @@ const expectToThrowAccessDeniedError = async (testFunc, path) => {
     await catchErrorFrom(testFunc, (caught) => {
         expect(caught).toMatchObject({
             name: 'TestClientResponseError',
-            data: { [path]: null },
+            data: set({}, path, null),
```
SavelevMatthew, line 524: Field accesses or what?)

apps/condo/domains/organization/access/Organization.js
Alllex202: try to get `id_in` from filter, otherwise you will get an empty array (in other words, nothing will be added in result array).
Filter (B2BAppServiceUser) can return only something like `{ id_in: [1,2,3] }` or `false`

packages/keystone/test.utils.js
Alllex202: `path` can be nested, as in the case of counting objects. 
So we had to add a little more variability :)

`obj` | `objs` | `meta.count`

SavelevMatthew (approved):
LGTM

apps/condo/domains/miniapp/utils/b2bAppServiceUserAccess/helpers.utils.js
```diff
@@ -0,0 +1,108 @@
...
+ *         organization: {
+ *             id_in: ['cdcbf4f8-a5b2-4831-bfb0-354a0abe0aa3'],
+ *             deletedAt, null,
+ *         },
+ *         deletedAt, null,
+ *     },
+ *     deletedAt, null,
+ * }
+ */
+const getFilter = (pathToOrganizationId, organizationIds) => {
```
pahaz, line 86: this name is too common

apps/condo/domains/miniapp/utils/b2bAppServiceUserAccess/helpers.utils.js
```diff
@@ -0,0 +1,108 @@
...
+
+    return {
+        [pathToOrganizationId[0]]: getFilter(pathToOrganizationId.slice(1), organizationIds),
+        deletedAt: null,
+    }
+}
+
+module.exports = {
+    getFilter,
+    generateGqlQuery,
```
pahaz, line 105: its better to add something organization related inside all of this names

pahaz (approved):

PR #4077, opened 2023-11-13
hotfix(global): cypress EOF

SavelevMatthew:
String without `\n` at the end is not valid line:
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206

And .env files are just bash....


pahaz (approved):

.env.example
```diff
@@ -48,4 +48,4 @@ CC_DOMAIN=http://cc_domain.app
 # APPLE_PUSH_TOKEN_TEST=<TOKEN>
 # APPLE_VOIP_PUSH_TOKEN_TEST=<TOKEN>
 
-MOBILE_APP_IDS='{ "resident": { "ios": "ai.doma.clients", "android": "ai.doma.client" }, "master": { "ios": "ai.doma.master", "android": "ai.doma.master" }}'
\ No newline at end of file
+MOBILE_APP_IDS='{ "resident": { "ios": "ai.doma.clients", "android": "ai.doma.client" }, "master": { "ios": "ai.doma.master", "android": "ai.doma.master" }}'
```
pahaz, line 51: It looks like this variable should be optional.

PR #4079, opened 2023-11-13
feat(pass): DOMA-7653 added access to manage pass ticket status field

Alllex202:
You can now allow employees to update pass ticket statuses, but prohibit pass ticket creation

This is necessary so that security guards cannot create passes unless they are explicitly allowed to do so.

----
Before release, need to add a new B2BAppPermission and run the script to fill new field

### DIFF
--- a/apps/condo/bin/update-b2b-app-roles-for-passapp.js
+++ b/apps/condo/bin/update-b2b-app-roles-for-passapp.js
@@ -1,0 +1,89 @@
+const path = require('path')
+
+const { GraphQLApp } = require('@keystonejs/app-graphql')
+const chunk = require('lodash/chunk')
+const get = require('lodash/get')
+const isEmpty = require('lodash/isEmpty')
+const isString = require('lodash/isString')
+
+const { find } = require('@open-condo/keystone/schema')
+
+const { B2BAppRole } = require('@condo/domains/miniapp/utils/serverSchema')
+
+class UpdateB2BAppRoles{
+    constructor (b2bAppId) {
+        this.b2bAppId = b2bAppId
+        this.context = null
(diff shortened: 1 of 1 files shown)


apps/condo/bin/update-b2b-app-roles-for-passapp.js
```diff
@@ -0,0 +1,89 @@
...
+const chunk = require('lodash/chunk')
+const get = require('lodash/get')
+const isEmpty = require('lodash/isEmpty')
+const isString = require('lodash/isString')
+
+const { find } = require('@open-condo/keystone/schema')
+
+const { B2BAppRole } = require('@condo/domains/miniapp/utils/serverSchema')
+
+class UpdateB2BAppRoles{
```
nomerdvadcatpyat, line 13:
```suggestion
class UpdateB2BAppRoles {
```

apps/condo/bin/update-b2b-app-roles-for-passapp.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    async findRoles () {
+        return await find('B2BAppRole', {
+            deletedAt: null,
+            app: { id: this.b2bAppId },
+        })
+    }
+
+    async updateRoles () {
```
nomerdvadcatpyat, line 36: Сan this be done by migration like org employee role?

nomerdvadcatpyat (approved):

apps/condo/bin/update-b2b-app-roles-for-passapp.js
Alllex202: At the time of migration, we do not know miniapp id in order to update only the necessary roles

PR #4099, opened 2023-11-17
fix(condo): DOMA-7749 get cookie_keys for oidc provider from env

sitozzz:
This PR fix security issue with OIDC provider

- [x] Secret values updated at helm submodule


.env.example
```diff
@@ -3,6 +3,7 @@ REDIS_URL=redis://127.0.0.1:6379
 NODE_ENV=development
 DISABLE_LOGGING=true
 COOKIE_SECRET=random
+COOKIE_KEYS='["OEK5zjaAMPc3L6iK7PyUjCOziUH3rsrMKB9u8H07La1SkfwtuBoDnHaaPCkG5Brg", "MNKeIebviQnCPo38ufHcSfw3FFv8EtnAe1xE02xkN1wkCV1B2z126U44yk2BQVK7"]'
```
SavelevMatthew, line 6: Maybe add OIDC_ prefix?

apps/condo/domains/user/oidc/configuration.js
```diff
@@ -8,6 +8,11 @@ const HTTPS_REGEXP = /^https:/
 
 module.exports = function createConfiguration (context, conf) {
     const jwksStr = get(conf, 'JWKS')
+    const cookieKeys = JSON.parse(get(conf, 'COOKIE_KEYS', '[]'))
+
+    if (isEmpty(cookieKeys)) {
```
Alllex202, line 13: Do we need to check that it is an array?

.env.example
sitozzz: Fixed

apps/condo/domains/user/oidc/configuration.js
sitozzz: Done

SavelevMatthew (approved):

PR #4104, opened 2023-11-20
fix(condo): DOMA-7763 don't show organization create modal if active invites exists

sitozzz:


apps/condo/domains/organization/components/OrganizationSelect.tsx
```diff
@@ -32,28 +32,38 @@ function compareEmployees (lhs: OrganizationEmployeeType, rhs: OrganizationEmplo
...
+    }, { skip: isNull(userId) })
+
+    const { count: hasInvites, loading: isInvitesLoading } = OrganizationEmployee.useCount({
+        where: {
+            user: { id: userId },
+            isAccepted: false,
+            ...ORGANIZATION_EMPLOYEE_WHERE_QUERY,
         },
-    })
+    }, { skip: isNull(userId) })
```
SavelevMatthew, line 66: Isn't skip is `Int`?

SavelevMatthew: Why not just `!userId` :)

sitozzz: If I'm not mistaken, I used an apollo graphql client hook parameter. According to the documentation it has a boolean type
image

sitozzz: > Why not just `!userId`? :)

Yup, it is possible

Alllex202 (commented):
Will modal be shown again if user declines all invitations?

Alllex202 (approved):

sitozzz:
> Will modal be shown again if user declines all invitations?

No it didn't show again. Only if user declines all of his invitations and after that reload the page

PR #4110, opened 2023-11-22
fix(condo): DOMA-7776 remove empty return if no property in getAddressDetails

nomerdvadcatpyat:


sitozzz (approved):

Alllex202 (approved):

apps/condo/domains/common/utils/helpers.ts
```diff
@@ -47,8 +47,6 @@ type ObjectWithAddressInfo = Pick<Property, 'address' | 'addressMeta'>
  * @param property
  */
 export const getAddressDetails = (property: ObjectWithAddressInfo) => {
-    if (isEmpty(property)) return
```
AleX83Xpert, line 50: Is it OK that some address parts may be "undefined"?
For example, `const streetWithType = get(addressMeta, 'street_with_type')` is used in concatenated strings.
image

PR #4115, opened 2023-11-24
feat(condo): DOMA-7799 Prepare B2CApps for dev-api

SavelevMatthew:


apps/condo/migrations/20231124091332-0343_remove_b2capp_shortdescription_and_more.js
```diff
@@ -0,0 +1,34 @@
...
+    await knex.raw(`
+    BEGIN;
+--
+-- Remove field shortDescription from b2capphistoryrecord
+--
+ALTER TABLE "B2CAppHistoryRecord" ADD COLUMN "shortDescription" text NULL;
+--
+-- Remove field shortDescription from b2capp
+--
+ALTER TABLE "B2CApp" ADD COLUMN "shortDescription" text NOT NULL;
```
AleX83Xpert, line 30: It looks like this down-migration will not work in the case of `B2CApp` table having some records, because of "NOT NULL". Is it ok?

Alllex202 (commented):
Who will create `B2CApp` and `B2CAppBuild` from the `dev-portal`?
Currently only admins/supports and users with direct access have access

AleX83Xpert (approved):

SavelevMatthew:
> Who will create `B2CApp` and `B2CAppBuild` from the `dev-portal`?
> Currently only admins/supports and users with direct access have access

Service integration account for dev-portal. There're few tests for it :)

Alllex202 (approved):

PR #4134, opened 2023-11-30
fix(condo): DOMA-7713 add email from acquiring offer

AleX83Xpert:


apps/condo/domains/marketplace/components/MarketplaceOnboarding/OfferSetup.tsx
```diff
@@ -65,48 +69,52 @@ export const OfferSetup: React.FC<{ launchContext: LaunchContextType }> = ({ lau
...
         loading: organizationIsLoading,
-    } = Organization.useObject(
-        { where: { id: organizationId } }
-    )
+    } = Organization.useObject({ where: { id: organizationId } })
+
+    const {
+        obj: acquiringOffer,
+        loading: isAcquiringOfferLoading,
+    } = Accept.useObject({ where: { scope: SCOPE_TYPES.acquiring, organizationId }, sortBy: [SortAcceptsBy.CreatedAtDesc], first: 1 })
```
sitozzz, line 77: IMO useObject already returns single value, so you don't need to filter with first:1 at the options prop

AleX83Xpert: `useObject` will throw an error in the case of 2+records. I just need to load the last created object.

sitozzz (approved):

VKislov (approved):

PR #4136, opened 2023-11-30
refactor(condo): HOTFIX: Make JSON in sender help text valid

toplenboren:
I just made that help text near `sender` helping text is actually a valid JSON, so it can be copy-pasted to the field below

Here is how it looks now:

telegram-cloud-photo-size-2-5424677925917153232-y

And a proof its a valid JSON now

telegram-cloud-photo-size-2-5424677925917153233-y

### DIFF
--- a/apps/address-service/schema.ts
+++ b/apps/address-service/schema.ts
@@ -47,7 +47,7 @@ export type Address = {
   newId?: Maybe<Scalars['String']>;
   /**  Data structure Version  */
   dv?: Maybe<Scalars['Int']>;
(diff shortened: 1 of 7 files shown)


Alllex202 (commented):
Seems like you should run `yarn maketypes` for all apps which use it :)

toplenboren:
ffs =/ I have two unregistered apps, need to find someone who is going to run make types for me =)

sitozzz (approved):

VKislov (approved):

Alllex202 (approved):

PR #4144, opened 2023-12-04
fix(condo): DOMA-7890 tax percent default value in `null`

AleX83Xpert:


apps/condo/domains/marketplace/components/MarketplaceOnboarding/RequisitesSetup.tsx
```diff
@@ -115,10 +113,8 @@ export const RequisitesSetup: React.FC = () => {
     const taxPercentOptions = useMemo<SelectProps['options']>(() => selectedTaxType === TAX_REGIME_GENEGAL ? [noTaxOption, ...TAX_PERCENT_OPTIONS] : [noTaxOption, ...TAX_PERCENT_OPTIONS.slice(1)], [noTaxOption, selectedTaxType])
 
     useEffect(() => {
-        if (isTaxTypeTouched) {
-            form.resetFields(['taxPercent'])
-        }
-    }, [form, isTaxTypeTouched, selectedTaxType])
+        form.setFieldValue('taxPercent', undefined)
```
sitozzz, line 116: Inside initial state of the form you set taxPercent to `null`. But here you set it to `undefined`. 
It will look better, if you'll use identical initial type

PR #4170, opened 2023-12-11
fix(condo): DOMA-7890 Make 'no vat' as default option

AleX83Xpert:
Also, husky pre-commit script was fixed

### DIFF
--- a/apps/condo/domains/marketplace/components/MarketplaceOnboarding/RequisitesSetup.tsx
+++ b/apps/condo/domains/marketplace/components/MarketplaceOnboarding/RequisitesSetup.tsx
@@ -102,7 +102,7 @@ export const RequisitesSetup: React.FC = () => {
     })
     const updateAction = InvoiceContextApi.useUpdate({ status: INVOICE_CONTEXT_STATUS_INPROGRESS })
 
-    const [noTaxOption]: SelectProps['options'] = [{ label: NoTax, key: NoTax, value: null }]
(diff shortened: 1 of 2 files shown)


apps/condo/domains/marketplace/components/MarketplaceOnboarding/RequisitesSetup.tsx
```diff
@@ -102,7 +102,7 @@ export const RequisitesSetup: React.FC = () => {
     })
     const updateAction = InvoiceContextApi.useUpdate({ status: INVOICE_CONTEXT_STATUS_INPROGRESS })
 
-    const [noTaxOption]: SelectProps['options'] = [{ label: NoTax, key: NoTax, value: null }]
+    const [noTaxOption]: SelectProps['options'] = [{ label: NoTax, key: NoTax, value: '' }]
```
sitozzz, line 105: You can do it little bit simple with typescript trick
```suggestion
    const noTaxOption: SelectProps['options'][number] = { label: NoTax, key: NoTax, value: '' }
```

sitozzz: With `number` shorthand you can get access to array item type, so you don't need to create array and then unpack it to a single value

sitozzz (approved):

apps/condo/domains/marketplace/components/MarketplaceOnboarding/RequisitesSetup.tsx
AleX83Xpert: Refactored

PR #4172, opened 2023-12-11
feat(address-service): INFRA-141 add balance & daily limits healthchecks

sitozzz:
- [x] Update env's for address-service at helm submodule
- [x] Update all other submodules

### DIFF
--- a/apps/address-service/domains/common/utils/healthchecks.js
+++ b/apps/address-service/domains/common/utils/healthchecks.js
@@ -1,31 +1,37 @@
-const { v4: uuid } = require('uuid')
+const { get } = require('lodash')
+
+const conf = require('@open-condo/config')
 
 const { DADATA_PROVIDER } = require('@address-service/domains/common/constants/providers')
-const { getSuggestionsProvider } = require('@address-service/domains/common/utils/services/providerDetectors')
+const { DadataHealthChecker } = require('@address-service/domains/common/utils/services/DadataHealthChecker')
 
 const prepareAddressServiceHealthcheck = () => {
-    const provider = getSuggestionsProvider({ req: { id: uuid() } })
+    const provider = get(conf, 'PROVIDER')
 
-    return provider.getProviderName() === DADATA_PROVIDER ? provider : null
+    if (provider === DADATA_PROVIDER) {
+        this.healthChecker = new DadataHealthChecker()
+    } else {
+        this.healthChecker = null
+    }
 }
 
 const getAddressProviderBalanceHealthCheck = () => ({
     name: 'address-provider-balance',
+    prepare: prepareAddressServiceHealthcheck,
     run: async () => {
-        const provider = prepareAddressServiceHealthcheck()
-        if (!provider) return false
+        if (!this.healthChecker) return true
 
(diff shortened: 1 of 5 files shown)


sitozzz:
This needs to be merged for address-service healthchecks proper op
https://github.com/open-condo-software/infrastructure-condo/pull/310

apps/address-service/domains/common/utils/healthchecks.js
```diff
@@ -0,0 +1,50 @@
...
+const { DadataSuggestionProvider } = require('@address-service/domains/common/utils/services/suggest/providers/DadataSuggestionProvider')
+
+const getAddressProviderBalanceHealthCheck = () => {
+    return {
+        name: 'address-provider-balance',
+        prepare: () => {
+            const provider = get(conf, 'PROVIDER')
+
+            if (provider === DADATA_PROVIDER) {
+                this.addressProvider = new DadataSuggestionProvider()
```
AleX83Xpert, line 15: The provider uses `req.id` for logging. You need to pass args to DadataSuggestionProvider:
```js
DadataSuggestionProvider({ req: { id: 'some id' } })
```
Please, see the constructor.

apps/address-service/domains/common/utils/healthchecks.js
```diff
@@ -0,0 +1,50 @@
...
+            } else {
+                this.addressProvider = null
+            }
```
AleX83Xpert, line 18: Maybe it's better to use the [providerDetector](https://github.com/open-condo-software/condo/blob/master/apps/address-service/domains/common/utils/services/providerDetectors.js)?

apps/address-service/domains/common/utils/healthchecks.js
sitozzz: I should create req.id by myself? It should be something like uuid? @AleX83Xpert

AleX83Xpert: Yes, UUID will fit.

apps/address-service/domains/common/utils/services/suggest/providers/DadataSuggestionProvider.js
```diff
@@ -415,6 +420,71 @@ class DadataSuggestionProvider extends AbstractSuggestionProvider {
...
         ))
     }
+
+    /**
+     * Checks token balance.
+     * In case of returned value less than DADATA_PROFILE_BALANCE_WARNING or any unhandled error -> returns false
+     * Otherwise returns true
+     * @return {Promise<boolean>}
+     */
+    async profileBalance () {
```
AleX83Xpert, line 430: Moving these new functions outside is better because they do not belong to suggestions and do not use encapsulated methods.
For example, you may add `utils/services/DadataHealtsChecker.js`. In this case, you even won't need to detect the provider.

sitozzz: Done, check plz

AleX83Xpert (approved):

toplenboren (approved):

PR #4175, opened 2023-12-11
fix(condo): DOMA-7969 fixed a imported date

Alllex202:
**Problem**: V8 (JavaScript engine) has issue with Date.getTimeZoneOffset. Because of this, dates from files are not imported correctly and have some deviation from the specified date of a few seconds

**Example**: 
1) Client input in cell: 2023-12-01 00:00:00
2) We get in code and show to client: 2023-11-30 23:59:27
3) Client: WTF!?

**Links**:
https://bugs.chromium.org/p/v8/issues/detail?id=7863 - V8 (JavaScript engine) has issue with Date.getTimezoneOffset
https://github.com/SheetJS/sheetjs/issues/1565#issuecomment-548491331 - Description and solution to the problem

### DIFF
--- a/apps/condo/domains/common/components/DataImporter.tsx
+++ b/apps/condo/domains/common/components/DataImporter.tsx
@@ -15,7 +15,7 @@ type UploadCols = {
 }
 
 type UploadDataItem = {
-    value: number | string
+    value: number | string | Date
 }
 
 type OnUpload = (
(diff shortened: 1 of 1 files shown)


toplenboren (approved):

nomerdvadcatpyat (approved):

Alllex202:
The solution does not work as expected in all cases. Let's come back to this problem later

PR #4179, opened 2023-12-12
fix(condo): DOMA-6655 fix ReadingSubmissionDate when meter importing

Alllex202:
**Problem**: 
`ReadingSubmissionDate` is set the wrong date if the value is in the `DD.MM.YYYY` format

**Example**: 
1) Input: 10.12.2023
2) Output: 01.12.2023

**Solution**: 
If a full date is specified (`DD.MM.YYYY` or `YYYY-MM-DD`), then the full date is used, and if only a month is indicated (`YYYY-MM`), then the first day of the month is used

**Also**:
* Fixed error output if it is not our `graphQLError` format (for example `unique constraint` error has a different format)
* Import error details were not displayed in the console, this has now been fixed

### DIFF
--- a/apps/condo/domains/meter/hooks/useImporterFunctions.ts
+++ b/apps/condo/domains/meter/hooks/useImporterFunctions.ts
@@ -19,8 +19,8 @@ import {
 import { useAddressApi } from '@condo/domains/common/components/AddressApi'
 import {
     Columns,
-    DATE_PARSING_FORMAT,
-    DATE_PARSING_FORMAT_2,
+    ISO_DATE_FORMAT,
+    EUROPEAN_DATE_FORMAT,
     ObjectCreator,
     ProcessedRow,
     RowNormalizer,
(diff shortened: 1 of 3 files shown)


apps/condo/domains/meter/hooks/useImporterFunctions.ts
```diff
@@ -20,6 +20,7 @@ import { useAddressApi } from '@condo/domains/common/components/AddressApi'
 import {
     Columns,
     DATE_PARSING_FORMAT,
+    DATE_PARSING_FORMAT_2,
```
nomerdvadcatpyat, line 23: mb rename it to
```
export const ISO_DATE_FORMAT = 'YYYY-MM-DD';
export const EUROPEAN_DATE_FORMAT = 'DD.MM.YYYY';
```

nomerdvadcatpyat (approved):

apps/condo/domains/meter/hooks/useImporterFunctions.ts
Alllex202: Fixed

PR #4182, opened 2023-12-12
fix(condo) DOMA-7581: Fix circular import warning during `yarn dev`

toplenboren:
Just a fix, that was meant to be in original PR >__<


toplenboren:
I do self-accept, because this is 

1) 1 loc change
2) not relevant
3) tests are ok

PR #4186, opened 2023-12-13
feat(dev-api): INFRA-191 add mutation to import apps

SavelevMatthew:


packages/keystone/upload.js
```diff
@@ -0,0 +1,27 @@
...
+ * @typedef {{
+ *     stream: ReadableStream,
+ *     filename?: string
+ *     mimetype?: string
+ *     encoding?: string
+ * }} UploadFile
+ * @param {UploadFile} uploadFile - file to upload. Basically object with stream / filename / mimetype / encoding
+ * @return {Upload} - Upload from graphql-upload, which can be passed to server-side utils
+ */
+function wrapUploadFile (uploadFile) {
```
sitozzz, line 13: IMO it's not a wrapping op. It's creation of new object isn't it?
```suggestion
function getGQLUploadFile (uploadFile) {
```

sitozzz (approved):

toplenboren (approved):

packages/keystone/upload.js
SavelevMatthew: It's not an `UploadFile`. Since we can already pass UploadFile here.

SavelevMatthew: This util is wrapping object with an Upload class, so server utils can eat it.

```
const u = new Upload()
// { promise: Promise<pending>, resolve: Function, reject: Function }
u.resolve(something)
// { promise: Promise<something>, resolve: Function, reject: Function, file: something  }
```

So, `something` wrapped with Upload can be passed as file, no wrapped = Invalid upload error

PR #4234, opened 2023-12-21
Fix/condo/doma 8032/fix marketplace setup steps

AleX83Xpert:


apps/condo/domains/marketplace/components/MarketplaceOnboarding/RequisitesSetup.tsx
```diff
@@ -139,7 +147,7 @@ export const RequisitesSetup: React.FC = () => {
                 bic: get(acquiringContext, ['invoiceRecipient', 'bic'], ''),
                 account: get(acquiringContext, ['invoiceRecipient', 'bankAccount'], ''),
                 taxType: get(acquiringContext, 'invoiceTaxRegime'),
-                taxPercent: get(acquiringContext, 'invoiceVatPercent'),
+                taxPercent: Number(get(acquiringContext, 'invoiceVatPercent', '')).toString(),
```
VKislov, line 150: It seems that a type error will occur when creating “without VAT”. Empty string invalid value for `Decimal` type
image

AleX83Xpert: This casting type is needed to convert '10.0000' to '10' to set the previously selected value within the option group. The database stores the `Decimal` type converted to string at the front-end ('10.0000'). We have options for VAT as numbers without zeros at the end ('10').

VKislov (approved):

apps/condo/domains/marketplace/components/MarketplaceOnboarding/RequisitesSetup.tsx
AleX83Xpert: Fixed. Now pass `null` instead of `''`

PR #4253, opened 2024-01-09
feat(condo) HOTFIX: add the ability to extend keystone config

toplenboren:
Hi, I want to add a small feature to `prepareKeystone()` function. 

My usecase: I need to override `cookie` param, so a `miniapp` can access cookies on other domains.

I can achieve such functionality using built in `extendExpressApp` but this way code is cleaner, and we do not override cookie secret and other variables. 

Example usecase: 

```(js)
const extendKeystoneConfig =(defaultConfig) => {
  cookie: {
    ...defaultConfig.cookie,

    // Enable cross-site usage
    sameSite: 'none',
    secure: true,
  }
}

module.exports = prepareKeystone({
    lastApp, schemas, apps, extendKeystoneConfig
})
```

### DIFF
--- a/packages/keystone/KSv5v6/v5/prepareKeystone.js
+++ b/packages/keystone/KSv5v6/v5/prepareKeystone.js
@@ -65,7 +65,7 @@ const sendAppMetrics = () => {
     }
 }
 
-function prepareKeystone ({ onConnect, extendKeystoneConfig = {}, extendExpressApp, schemas, schemasPreprocessors, tasks, apps, lastApp, graphql, ui }) {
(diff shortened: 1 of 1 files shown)


Alllex202 (approved):

Alllex202:
Maybe it would be more convenient to do something like this, so as not to import the default config from somewhere?

```
module.exports = prepareKeystone({
    lastApp, schemas, apps, 
    extendKeystoneConfig: (defaultConfig) => ({ 
      ...defaultConfig, 
      cookie: { ...defaultConfig.cookie, /* custom config */ } ,
    },
})
```

PR #4256, opened 2024-01-10
fix(condo): DOMA-8186 fixed a date range for analytic reports

Alllex202:


apps/condo/domains/analytics/hooks/useDashboardFilters.tsx
```diff
@@ -82,9 +82,7 @@ type UseDateRangeFilterType = () => {
 export const useDateRangeFilter: UseDateRangeFilterType = () => {
     const [dateRange, setDateRange] = useState<[Dayjs, Dayjs]>([dayjs().subtract(6, 'days'), dayjs()])
 
-    const disabledDate = useCallback((currentDate) => {
-        return currentDate && currentDate < dayjs().startOf('year')
-    }, [])
+    const disabledDate = useCallback(() => false, [])
```
nomerdvadcatpyat, line 85: I think the restriction was there for a reason.
Perhaps we should limit the last year (currentDate < now - year)

sitozzz: Yeah, thats good point. It would be nice to limit aggregation range to some period. For example it might be 1 year

sitozzz, line 85: Disabling aggregation limit is not a good idea

Alllex202: Now all years are disabled off except the current and previous one

Alllex202: Now all years are disabled off except the current and previous one

nomerdvadcatpyat (approved):

sitozzz (approved):

PR #4258, opened 2024-01-11
fix(condo): DOMA-8196 exclude ticket with invoice from throttling by ticket details

VKislov:


apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -852,10 +855,12 @@ const Ticket = new GQLListSchema('Ticket', {
             const userType = get(user, 'type')
 
             if (userType === RESIDENT && operation === 'create') {
+                const locale = extractReqLocale(context.req) || conf.DEFAULT_LOCALE
                 await checkDailyTicketLimit({
                     userId: get(user, 'id'),
                     organizationId: get(resolvedData, 'organization'),
                     details: get(resolvedData, 'details'),
+                    isInvoiceTicket: get(resolvedData, 'details') === i18n('marketplace.invoice.newTicket.details', { locale }),
```
Alllex202, line 863: Should the ticket have the "isPayable" flag selected? Can we add this condition here?

apps/condo/domains/marketplace/schema/Invoice.test.js
```diff
@@ -1603,5 +1603,51 @@ describe('Invoice', () => {
                 messageForUser: 'api.marketplace.invoice.error.PublishingWithoutDefinedPricesForbidden',
             })
         })
+
+        test('can create more then 2 invoice with different ticket', async () => {
+            const createInvoiceWithTicket = async () => {
```
Alllex202, line 1608:
```suggestion
            const createInvoiceWithTicket = async (client) => {
```
Please pass `client` here. I tried for a long time to understand where this variable comes from - but it turns out it is declared below)

Alllex202: Maybe move this function outside the test? It may be needed in other tests too

Alllex202 (approved):

apps/condo/domains/marketplace/schema/Invoice.test.js
VKislov: This is a local simplification; in other tests, the process of creating entities for an invoice uses only test utilities like 'createTest...'. I simplified it here just to avoid a long duplication

apps/condo/domains/ticket/schema/Ticket.js
VKislov: yes, you're right, you can additionally secure this check

nomerdvadcatpyat (approved):

PR #4262, opened 2024-01-11
feat(ui): DOMA-8189 added tip component

nomerdvadcatpyat:
изображение
изображение


packages/ui/src/components/Tip/style.less
```diff
@@ -0,0 +1,53 @@
+@import "antd/lib/tooltip/style/index.less";
+@import (reference) "@open-condo/ui/src/tokens/variables.less";
+
+.condo-tooltip.condo-tip {
+  max-width: 350px;
+  padding: 4.3px;
```
Alllex202, line 6: 4.3? :)

packages/ui/src/components/Tip/style.less
```diff
@@ -0,0 +1,53 @@
...
+      font-family: @condo-global-font-family-default;
+      line-height: @condo-global-typography-text-s-line-height;
+      background-color: @condo-global-color-black;
+      border-radius: @condo-global-border-radius-large;
+      box-shadow: @condo-global-box-shadow-default;
+
+      & .condo-tip-icon-wrapper {
+        position: absolute;
+        top: 8px;
+        right: 8px;
```
Alllex202, line 23: How about this?
```suggestion
        top: @condo-global-spacing-8;
        right: @condo-global-spacing-8;
```

packages/ui/src/components/Tip/tip.tsx
```diff
@@ -0,0 +1,74 @@
...
+    useEffect(() => {
+        setOpen(initialOpen)
+    }, [initialOpen])
+
+    const handleClose = useCallback(() => {
+        if (onClose) {
+            onClose()
+        }
+
+        setOpen(prev => !prev)
```
Alllex202, line 39:
```suggestion
        setOpen(false)
```

packages/ui/src/components/Tip/style.less
sitozzz, line 23: It's better to use global spacing variables

packages/ui/src/components/Tip/style.less
nomerdvadcatpyat: changed to 4

sitozzz (approved):

Alllex202 (approved):

SavelevMatthew (changes requested):
What's the difference from current Tooltip component?
(except for the headline and visuals)

nomerdvadcatpyat:
> What's the difference from current Tooltip component? (except for the headline and visuals)

Different opening/closing behavior.
Tooltip is opened and closed by some trigger (in our case it is hover), this component closes either by clicking on the cross icon or by performing an external action (prop open). Combined with the different styles and the addition of tite/message (not just the title) there are a lot of changes to make it a Tooltip component new type.
In addition to this in UI-kit designs it is made as a separate component

toplenboren:
image

I think we should add a little padding here, so that tip triangle does not collide with button

toplenboren (approved):

packages/ui/src/components/Tour/tourContext.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { useContext, useState } from 'react'
+
+type TourContextValue = {
+    currentStep?: number
```
SavelevMatthew, line 4: Why everything is optional?

SavelevMatthew: You should provide proper default value instead

SavelevMatthew: Context if basically a 3 things (types + defaults + actual values)

SavelevMatthew: So you can solve it like this:

1. Types. You should say value is defined
```typescript
type TourContextType = {
    currentStep: number
    setCurrentStep: React.Dispatch<React.SetStateAction<number>>
}
```
2. Defaults. What to return if no Provider not found in parent components
```typescript
const TourContext = React.createContext<TourContextValue>({
    currentStep: 0
    setCurrentStep: () => ({})
})
```
So, now if you decided to use context without a provider, you will get this defaults

3. Actual values
```
const TourContextProvider = ...
```

SavelevMatthew (approved):
LGTM

PR #4266, opened 2024-01-12
feat(condo): DOMA-8011 add health check handlers for certificates and…

ekabardinsky:
… 3rd party integrations


apps/condo/index.js
```diff
@@ -100,6 +107,15 @@ const tasks = () => [
 const checks = [
     getRedisHealthCheck(),
     getPostgresHealthCheck(),
+    getCertificateHealthCheck({
+        certificateName: 'sber_id_client',
+        certificate: SBER_ID_CONFIG.cert,
+    }),
```
sitozzz, line 113: It would be more consistent if you move config parsing code inside your healthcheck function body

dkoviazin (approved):

PR #4278, opened 2024-01-16
feat(condo): DOMA-8209 hide marketplace for organization with service_provider type (RSO)

VKislov:


apps/condo/domains/marketplace/schema/Invoice.js
```diff
@@ -85,7 +85,7 @@ const sendPush = async ({ originalInput, userId, propertyId, unitName, unitType,
                             userId,
                             residentId: resident.id,
                             invoiceId: updatedItem.id,
-                            url: `${conf.SERVER_URL}/payments`,
+                            url: `${conf.SERVER_URL}/payments/`,
```
nomerdvadcatpyat, line 88: why `/` in the end?

VKislov: on the web this is an extra character. But the mobile phone doesn’t understand where to redirect otherwise

nomerdvadcatpyat: why we have 2 formats for push url? 
For example, in `sendTicketCommentNotifications` url used without `/` on the end: `${conf.SERVER_URL}/ticket/${ticket.id}`
Or in `sendBillingReceiptsAddedNotifications`

nomerdvadcatpyat (approved):

apps/condo/domains/marketplace/schema/Invoice.js
VKislov: Don't know. The current push URL without `/` at the end of the resident application could not be redirected. The developer said that the `/` at the end is needed

PR #4281, opened 2024-01-17
Fix/condo/doma 7815/add custom condo file field

abshnko:
new field -- now File.originalFilename will be converted to UTF-8 on resolve input stage


apps/condo/domains/ticket/schema/TicketFile.js
```diff
@@ -23,7 +23,7 @@ const TicketFile = new GQLListSchema('TicketFile', {
     fields: {
         file: {
             schemaDoc: 'File object with meta information and publicUrl',
-            type: File,
+            type: FileWithUTF8Name,
```
SavelevMatthew, line 26: You should propably left string here, like so:
```suggestion
            type: 'File',
```

SavelevMatthew: And then map this string into your file field... inside `@open-condo/keystone`

abshnko: seems like logic with custom fields requires type as an object (like for LocalizedText or DateInterval etc)
Otherwise it thinks that it's a normal `File` and doesn't use `FileWithUTF8Name` logic

abshnko: edit: updated File mapping to use FileWIthUTF8Name

packages/keystone/fields/FileWithUTF8Name/Implementation.js
Alllex202, line 1: please leave a comment here that this is the original implementation with custom logic for `originalFilename` only

Alllex202: And you can also mention ignoring the peculiarities of `mongoose` for `id` field  :)

dkoviazin (approved):

PR #4303, opened 2024-01-19
fix(eps): DOMA-8264 fixed eps receipt render

dkoviazin:


sitozzz (approved):
Why don't to use logger instead of console.error?

PR #4310, opened 2024-01-22
refactor(condo): DOMA-8030 remove AssigneeScope schema

nomerdvadcatpyat:
`AssigneeScope` it is a redundant schema left over from the development of property scopes.
Now it and the logic associated with it are not used anywhere, so we can remove it from code


apps/condo/migrations/20240122161345-0361_auto_20240122_1113.js
Alllex202, line 1: Why did you comment out all lines in this migration? Shouldn't we drop the table?

apps/condo/bin/fix-database-migration-state.js
```diff
@@ -163,6 +163,14 @@ ALTER TABLE "BillingIntegrationHistoryRecord" DROP COLUMN "label" CASCADE;
 ALTER TABLE "BillingIntegrationHistoryRecord" DROP COLUMN "partnerUrl" CASCADE;
 ALTER TABLE "BillingIntegrationHistoryRecord" DROP COLUMN "price" CASCADE;
 
+COMMIT;
+
+--
+-- 20240122161345-0361_auto_20240122_1113.js Remove redundant AssigneeScope schema
```
Alllex202, line 169: Migration number updated

Alllex202 (approved):

PR #4311, opened 2024-01-22
fix(condo): DOMA-8275 added maxResults 1000 to keystone lists

nomerdvadcatpyat:
If you make a query for schema with many objects, for example `allTickets`, without specifying `first` argument, instead of an `Your request exceeded server limits` error from the server, it returns a timeout error from database select `select \"t0\".* from \"public\".\"Ticket\" as \"t0\" where true and (\"t0\".\"deletedAt\" is null) - canceling statement due to statement timeout` or 502 error with falling pod.

The `maxResults` parameter is set to `first` argument if it passed.
Now we are already specify `maxTotalResults` (maximum number of returned objects, including nested ones) = 1000 in `prepareDefaultKeystoneConfig`, in this pr i set `maxResults` to 1000 too to prevent database queries when `first` not set or set > 1000.

Keystone `_itemsQuery` limits check:
изображение

### DIFF
--- a/packages/keystone/KSv5v6/v5/registerSchema.js
+++ b/packages/keystone/KSv5v6/v5/registerSchema.js
@@ -76,7 +76,10 @@ function registerKeystone5Schema (gqlSchemaObject, keystone, globalPreprocessors
     if (gqlSchemaObject._type === GQL_LIST_SCHEMA_TYPE) {
         gqlSchemaObject._register(globalPreprocessors, { addSchema })
         gqlSchemaObject._keystone = keystone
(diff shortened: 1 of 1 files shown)


sitozzz (commented):
Is it possible to test this?

sitozzz:
Can you query getAllTickets without specifying limit for example

nomerdvadcatpyat:
> Can you query getAllTickets without specifying limit for example?

Ok

1.1. Without `maxResults` and `first` arg:
изображение

1.2. With `maxResults` and without `first` arg
изображение


------

Behavior with `first` <= `maxTotalResults` (1000) and > `maxTotalResults`  did not change:

2.1. Without `maxResults`:
изображение
изображение

2.2 With `maxResults`:
изображение
изображение

sitozzz (approved):

Alllex202 (approved):

PR #4313, opened 2024-01-23
fix(condo): DOMA-8271 optimizing of news creating

Alllex202:
- Added and use bulk-operations for NewItemScopes
 - Optimized the number of scopes created if all units are selected
 
 Also: 
 - It was prohibited to update `organization` field for `NewsItem` schema


apps/condo/domains/news/access/NewsItemScope.js
```diff
@@ -37,33 +36,65 @@ async function canReadNewsItemScopes (attrs) {
...
+                if (!newsItem) return false
+                organizationIds = [get(newsItem, 'organization')]
             }
-            const newsItemScope = await getById('NewsItemScope', itemId)
-            newsItemId = newsItemScope.newsItem
-        }
+        } else if (isSoftDeleteOperation) {
+            if (isBulkRequest) {
+                if (!itemIds || !Array.isArray(itemIds)) return false
+                if (itemIds.length !== uniq(itemIds).length) return false
```
AleX83Xpert, line 75: Why is this condition needed?

nomerdvadcatpyat (approved):

apps/condo/domains/news/access/NewsItemScope.js
Alllex202: It will be strange if you delete the same object many times.
We can skip this check, but this means that we will do some extra operations when we can avoid them

apps/condo/domains/news/access/NewsItemScope.js
```diff
@@ -37,33 +36,65 @@ async function canReadNewsItemScopes (attrs) {
...
+                const newsItems = await find('NewsItem', {
+                    id_in: newsItemIds,
+                    deletedAt: null,
+                })
+                if (newsItemIds.length !== newsItems.length) return false
+                organizationIds = uniq(newsItems.map(item => get(item, 'organization')))
+            } else {
+                const newsItemId = get(originalInput, 'newsItem.connect.id')
+                const newsItem = await getById('NewsItem', newsItemId)
+                if (!newsItem) return false
```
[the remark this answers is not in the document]
Alllex202: fixed

AleX83Xpert (approved):

PR #4318, opened 2024-01-24
fix(condo): DOMA-8291 change cron task running time for notifyResidentsOnPaydayTask & sendSubmitMeterReadingsPushNotifications

VKislov:


sitozzz (commented):
What the reason of changing running time?

VKislov:
There are regions in Russia with a time zone of utc+10. 14 00 for them is already 00 00

sitozzz (approved):

Alllex202:
> There are regions in Russia with a time zone of utc+10. 14 00 for them is already 00 00

Why did this problem only appear now?
This is not a new feature

PR #4323, opened 2024-01-25
fix(condo): DOMA-8276 fix layout of filters

VKislov:
image
image
image


Alllex202 (approved):
You can fill full width of `DatePicker` if screen is small (optional)

PR #4327, opened 2024-01-25
hotfix(condo): add retrying for deliverMessage task

VKislov:
Before release, you need to delete the task with the old config in production. So that after the release there is no duplication

### DIFF
--- a/apps/condo/domains/notification/adapters/apple/AppleMessaging.js
+++ b/apps/condo/domains/notification/adapters/apple/AppleMessaging.js
@@ -13,6 +13,7 @@ const {
     APS_PUSH_TYPE_VOIP,
     APS_PUSH_TYPE_BACKGROUND,
     APS_RESPONSE_STATUS_SUCCESS,
+    RETRY_RESTRICTION,
 } = require('./constants')
 
 const logger = getLogger('AppleMessaging')
@@ -130,9 +131,9 @@ class AppleMessaging {
             const stream = this.#session.request(headers)
 
             stream.on('response', this.getResponseHandler(stream, resolve, reject))
-            stream.on('error', (error) => {
-                logger.error({ msg: 'sendPush errored', headers, options, payload, streamError: error })
-                return resolve(error)
+            stream.on('error', (err) => {
+                logger.error({ msg: 'sendPush errored', headers, options, payload, err })
+                return resolve(err)
             })
             stream.write(buffer)
             stream.end()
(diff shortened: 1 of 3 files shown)


apps/condo/domains/notification/adapters/apple/AppleMessaging.js
```diff
@@ -127,7 +129,14 @@ class AppleMessaging {
 
             logger.info({ msg: 'sendPush before request', headers, options, payload })
 
-            const stream = this.#session.request(headers)
+            let stream
+            for (let retryCounter = 0; retryCounter < 5; retryCounter++) {
```
AleX83Xpert, line 133: Looks like this code will send 5 push messages.
Every loop iteration you rewrite the `stream` variable. After the loop you have another "onResponse" callback. Which stream this callback is?

You need to wait the answer for the first request. If the answer is erroneous you should repeat the request.

AleX83Xpert (changes requested):

apps/condo/domains/notification/adapters/apple/AppleMessaging.js
```diff
@@ -127,11 +129,20 @@ class AppleMessaging {
...
+                stream = this.#session.request(headers)
```
pahaz, line 134: Why not to use await ?

apps/condo/domains/notification/adapters/apple/AppleMessaging.js
```diff
@@ -127,11 +129,20 @@ class AppleMessaging {
...
+                stream.on('error', (error) => {
+                    logger.error({ msg: 'sendPush errored', failCounter: retryCounter + 1, headers, options, payload, streamError: error })
```
pahaz, line 136: use `err` key for error! we need to save tranceback

apps/condo/domains/notification/adapters/apple/AppleMessaging.js
```diff
@@ -161,7 +161,12 @@ class AppleMessaging {
                 payload.aps.alert = { ...notification }
             }
 
-            const response = await this.sendPush(token, payload, options)
+            let response
+            for (let retryCounter = 0; retryCounter < 5; retryCounter++) {
```
AleX83Xpert, line 165: It's better to move `5` to some constant

apps/condo/domains/notification/adapters/apple/AppleMessaging.js
```diff
@@ -161,7 +161,12 @@ class AppleMessaging {
...
+                response = await this.sendPush(token, payload, options)
+                if (response instanceof Error) continue
```
AleX83Xpert, line 167: Maybe add some logs?
```suggestion
                if (response instanceof Error) {
                    logger.warn({ msg: '...' })
                    continue
                }
```

AleX83Xpert (approved):

apps/condo/domains/notification/adapters/apple/AppleMessaging.js
```diff
@@ -130,9 +131,9 @@ class AppleMessaging {
             const stream = this.#session.request(headers)
 
             stream.on('response', this.getResponseHandler(stream, resolve, reject))
-            stream.on('error', (error) => {
-                logger.error({ msg: 'sendPush errored', headers, options, payload, streamError: error })
-                return resolve(error)
+            stream.on('error', (err) => {
+                logger.error({ msg: 'sendPush errored', headers, options, payload, err })
+                return resolve(err)
```
pahaz, line 136: It looks so strange! you should use reject for such things

PR #4339, opened 2024-01-31
feature(condo): DOMA-8364 extend user ritesSet to support Property

dkoviazin:
We need service user to have access to organization Properties:
1. EPS will create properties for new flow
2. SPP will create properties when user accepts offer to publish accruals in condo
3. Registry will use current organization properties to normalize addresses and to highlight that some properties from accrual file are missing


apps/condo/domains/user/utils/directAccess/config.js
```diff
@@ -43,6 +43,9 @@ const DIRECT_ACCESS_AVAILABLE_SCHEMAS = {
         // Ticket domain
         'Ticket',
 
+        // Property domain
+        'Property',
```
SavelevMatthew, line 47: keep order of domains alphabetical)

SavelevMatthew (commented):
Any tests?)

SavelevMatthew:
You should also modify access checks for a changed model

SavelevMatthew:
> We need service user to have access to organization Properties:
> 
> 1. EPS will create properties for new flow
> 2. SPP will create properties when user accepts offer to publish accruals in condo
> 3. Registry will use current organization properties to normalize addresses and to highlight that some properties from accrual file are missing

1-2. In this case, aren't the service users also creating an organization? Then they already have access.
3. Registry cannot work without current user, so it's should use user token instead probably.

SavelevMatthew (changes requested):
Please add real-life tests / explain your decision :)

AleX83Xpert (approved):

PR #4341, opened 2024-01-31
feat(condo): DOMA-8321 updated form when sending news if one property

Alllex202:
<details>
	<summary>BEFORE</summary>

https://github.com/open-condo-software/condo/assets/56914444/afc3ea65-69c2-4dd1-a4a1-c6bb4018ad1c
</details>
<details>
	<summary>AFTER</summary>
	
https://github.com/open-condo-software/condo/assets/56914444/c6818eb4-1d3a-4594-bb35-34e0bb52b3d8
</details>


nomerdvadcatpyat (approved):

apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -87,8 +91,13 @@ export type BaseNewsFormProps = {
     OnCompletedMsg: (INewsItem) => ArgsProps | null,
     allNews: INewsItem[],
     actionName: ActionNameProps,
+    totalProperties: number
 }
 
+const OptionalBlock = styled.div<{ show?: boolean }>`
+  ${({ show }) => !show ? 'display: none;' : ''}
+`
+
```
sitozzz, line 100: You don't really need to use styled for that small functionality. Instead of this you can use React conditional rendering syntax

sitozzz: So there is no reason for CPU optimization, if you write this feature for that reason. If organization had only one property -> visibility of this block will be calculated once

Alllex202: We need to render both elements, but only one should be visible.
This is necessary in order not to rewrite the monstrous existing logic

sitozzz (approved):

PR #4344, opened 2024-02-01
feat(condo): INFRA-264 OIDC Enabled field and deletedAt fix

SavelevMatthew:


apps/condo/domains/user/oidc/adapter/OidcModelClientAdapter.js
```diff
@@ -184,7 +184,7 @@ class OidcModelClientAdapter {
      *
      */
     async find (id) {
-        const item = await OidcClient.getOne(this.context, { clientId: id })
+        const item = await OidcClient.getOne(this.context, { clientId: id, isEnabled: true, deletedAt: null })
```
AleX83Xpert, line 187: Should we send an error about disabled client?

AleX83Xpert (approved):

apps/condo/domains/user/schema/OidcClient.js
```diff
@@ -65,6 +65,17 @@ const OidcClient = new GQLListSchema('OidcClient', {
...
 
+        isEnabled: {
+            schemaDoc:
+                'A switch that allows you to disable some OIDC clients. ' +
+                'If an OIDC client is disabled, it cannot be used for OIDC authorization. ' +
+                'Used mainly by developers portal to create OIDC client before publishing an application, ' +
+                'and enable OIDC after publishing.',
+            type: 'Checkbox',
+            isRequired: true,
+            defaultValue: false,
```
pahaz, line 76: may be true?

pahaz (approved):

apps/condo/domains/user/schema/OidcClient.js
SavelevMatthew: Doesn't really matter, since there's only 2 ways of creating it. Via dev-portal, which will always pass this parameter. Or via admin panel, where you need to specify it explicitly.

apps/condo/domains/user/oidc/adapter/OidcModelClientAdapter.js
SavelevMatthew: It depends. Disabled client for OIDC is similar to no client, since it's not ready for use. Or disabled client is ready-to-use, but just turned off

PR #4347, opened 2024-02-01
feat(condo): DOMA-7262 skip discovering if user has 7+ residents in property

AleX83Xpert:


apps/condo/domains/resident/schema/DiscoverServiceConsumersService.js
```diff
@@ -477,7 +514,16 @@ const DiscoverServiceConsumersService = new GQLCustomSchema('DiscoverServiceCons
...
-                            ...serviceConsumerData,
+                            ...pick(serviceConsumerData, [
+                                'dv',
+                                'sender',
+                                'resident',
+                                'accountNumber',
+                                'organization',
+                                'billingAccount',
+                                'billingIntegrationContext',
+                                'acquiringIntegrationContext',
```
sitozzz, line 525: Why do you need to explicitly collect fields from serviceConsumerData?

AleX83Xpert: Yes, because `data` now contains some additional fields needed for this mutation, but erroneous for creating a service consumer.

dkoviazin (approved):

sitozzz (approved):

PR #4372, opened 2024-02-08
feat(condo): INFRA-210 enable query batching with BatchHttpLink option

sitozzz:
This PR enables apollo query batching feature.
For reasons of incompatibility with `apollo-upload-client` we need to split http links based on "has concrete request an attached file".
So there is two ways to deal with this problem:

1. Add special flag to mutation context and disable batching option for that request
2. Auto detect request type (you need to look for variables and attachments for each http request sent from client)

For now I choose 1 variant, because it is much faster than second one. But there is a problem - you need to manually pass `{ hasUpload: true }` to the mutation context

For more information you can read this issue https://github.com/jaydenseric/apollo-upload-client/issues/34

Edit:
For better DX we decide to choose second variant with request type auto-detection

### DIFF
--- a/packages/next/apollo.jsx
+++ b/packages/next/apollo.jsx
@@ -46,14 +46,21 @@ let createApolloClient = (initialState, ctx, apolloCacheConfig, apolloClientConf
         }
     }
 
+    const isObject = node => typeof node === 'object' && node !== null
+    const isFile = input => isOnClientSide && input instanceof File
+    const isBlob = input => isOnClientSide && input instanceof Blob
+
     const hasFiles = (data) => {
         let i = 0
         let found
-
         const keys = Object.keys(data)
 
         while (!found && i < keys.length) {
-            if (data[keys[i]] instanceof File || data[keys[i]] instanceof Blob) {
+            if (!isObject(data[keys[i]])) {
+                i++
+            } else if (isFile(data[keys[i]]) || isBlob(data[keys[i]])) {
+                found = true
+            } else if (hasFiles(data[keys[i]])) {
                 found = true
             } else {
                 i++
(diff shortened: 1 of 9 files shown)


sitozzz:
Sounds like I need to reinvestigate cypress tests and rid off the alias usages)

packages/next/apollo.jsx
```diff
@@ -43,20 +46,49 @@ let createApolloClient = (initialState, ctx, apolloCacheConfig, apolloClientConf
         }
     }
 
+    const hasFiles = (data) => {
+        let i = 0
+        let found
+
+        const keys = Object.keys(data)
+        try {
```
toplenboren, line 54: Looks cheeky

sitozzz: Thats was for debug. Removed try catch block

packages/next/apollo.jsx
```diff
@@ -43,20 +46,46 @@ let createApolloClient = (initialState, ctx, apolloCacheConfig, apolloClientConf
...
 
+    const hasFiles = (data) => {
+        let i = 0
+        let found
+
+        const keys = Object.keys(data)
+
+        while (!found && i < keys.length) {
+            if (data[keys[i]] instanceof File || data[keys[i]] instanceof Blob) {
+                found = true
```
SavelevMatthew, line 57: You can do `break` here :)

packages/next/apollo.jsx
```diff
@@ -43,20 +46,46 @@ let createApolloClient = (initialState, ctx, apolloCacheConfig, apolloClientConf
...
-        link: createUploadLink({
-            uri: apolloGraphQLUrl, // Server URL (must be absolute)
-            credentials: 'include',
-            fetchOptions: {
-                mode: 'cors',
-            },
-            fetch: (isOnClientSide && window.fetch) ? window.fetch : fetch,
-            headers: (ctx && ctx.req) ? ctx.req.headers : undefined,  // allow to use client cookies on server side requests
-        }),
+        link: ApolloLink.split(operation => hasFiles(get(operation, 'variables.data', {})), uploadLink, batchLink),
```
SavelevMatthew, line 88: what if I have a custom mutation with different name? variables.file or smth like that

packages/next/apollo.jsx
sitozzz: It will stop on the next iteration anyway)

packages/next/apollo.jsx
sitozzz: I don't get it. What the behaviour of "different name"? For the second point I agree with you. But if we look from the other side - for custom variables definition blob type search you should recursively search for blobs at all of the keys. So that's might be very complex operation.

packages/next/apollo.jsx
SavelevMatthew: missed that line, sorry :)

SavelevMatthew (approved):

PR #4378, opened 2024-02-12
fix(condo): DOMA-7829 fix invoice payment export. Fix export receipts payment

VKislov:


apps/condo/domains/acquiring/schema/ExportPaymentsService.test.js
```diff
@@ -80,6 +87,100 @@ describe('ExportPaymentsService', () => {
...
+        await createTestMultiPayment(admin, [invoicePayment], admin.user, acquiringIntegration)
+
+        const {
+            data: {
+                result: {
+                    status,
+                    linkToFile,
+                },
+            },
+        } = await admin.query(EXPORT_PAYMENTS_TO_EXCEL, {
```
toplenboren, line 111: export this to test utils

apps/condo/domains/acquiring/schema/ExportPaymentsService.test.js
```diff
@@ -80,6 +87,100 @@ describe('ExportPaymentsService', () => {
         })
     })
 
+    it('check payments export', async () => {
+        const locale = 'ru'
+        const { admin, billingReceipts, acquiringContext, acquiringIntegration, organization } = await makePayer()
```
toplenboren, line 92: best to export admin and regular data to beforeAll()

apps/condo/domains/acquiring/schema/ExportPaymentsService.test.js
```diff
@@ -80,6 +87,100 @@ describe('ExportPaymentsService', () => {
...
+        } = await admin.query(EXPORT_PAYMENTS_TO_EXCEL, {
+            data: {
+                dv: 1,
+                sender: { dv: 1, fingerprint: 'test-' + faker.random.alphaNumeric(8) },
+                where: { invoice: { organization: { id: organization.id } } },
+                sortBy: 'advancedAt_DESC',
+                timeZone: DEFAULT_ORGANIZATION_TIMEZONE,
+            },
+        })
+        const formatDate = (date) => dayjs(date).tz(DEFAULT_ORGANIZATION_TIMEZONE).format('DD.MM.YYYY HH:mm')
```
toplenboren, line 120: move to test utils

apps/condo/domains/acquiring/schema/ExportPaymentsService.test.js
```diff
@@ -80,6 +87,100 @@ describe('ExportPaymentsService', () => {
...
+            [
+                formatDate(receiptPayment.advancedAt),
+                receiptPayment.accountNumber,
+                receiptPayment.receipt.property.address,
+                receiptPayment.receipt.account.unitName,
+                acquiringIntegration.name,
+                '',
+                '',
+                i18n('payment.status.' + receiptPayment.status, { locale }),
+                Number(receiptPayment.amount).toFixed(2).toString(),
```
toplenboren, line 177: Shouldnt we use bigjs here?

apps/condo/domains/acquiring/schema/ExportPaymentsService.js
```diff
@@ -77,6 +77,7 @@ const ExportPaymentsService = new GQLCustomSchema('ExportPaymentsService', {
 
                 const timeZone = normalizeTimeZone(timeZoneFromUser) || DEFAULT_ORGANIZATION_TIMEZONE
                 const formatDate = (date) => dayjs(date).tz(timeZone).format(DATE_FORMAT)
+                const isInvoicePayments = !!get(where, 'invoice.organization.id')
```
toplenboren, line 80: best to specify fallback value as something that will be false always, here is an example:

`!!get(where, 'invoice.organization.id', null)`

apps/condo/domains/acquiring/schema/ExportPaymentsService.js
```diff
@@ -93,8 +94,12 @@ const ExportPaymentsService = new GQLCustomSchema('ExportPaymentsService', {
                     return {
                         date: formatDate(obj.advancedAt),
                         account: obj.accountNumber,
-                        address: get(obj, ['receipt', 'property', 'address'], ''),
-                        unitName: get(obj, ['receipt', 'account', 'unitName'], ''),
+                        address: isInvoicePayments ?
```
toplenboren, line 97: using giant ternary operators here makes it harder to read this code, please consider exporting it to variables

apps/condo/domains/acquiring/schema/ExportPaymentsService.test.js
VKislov: I think no. I repeated the behavior that is implemented in the export service

apps/condo/domains/acquiring/schema/ExportPaymentsService.test.js
VKislov: I agree, but this utility is used in many places, it’s not something new. Most likely, need to create an admin in `beforAll` and refactor it in the tests that use this utility. I'm not ready to do it in this pull request

apps/condo/domains/acquiring/schema/ExportPaymentsService.test.js
toplenboren: But you are getting amount which is bigJS field i think, number conversions in JS are tricky, why not use something will work correctly 100% time?

I fear that this test will fail randomly from time to time

apps/condo/domains/acquiring/schema/ExportPaymentsService.test.js
toplenboren: On the contrary, this functionality is rather new, and we have agreed that we are going to use beforeAll long before this functionality

toplenboren: So I really think that it is best for this PR to implement beforeAll

toplenboren (approved):

dkoviazin (approved):

PR #4388, opened 2024-02-14
chore(address-service): DOMA-8447 remove regexp while searching addre…

AleX83Xpert:
…ss sources


toplenboren (approved):

SavelevMatthew (approved):

dkoviazin (approved):

packages/keystone/fields/RegexplessText/index.js
```diff
@@ -0,0 +1,25 @@
...
+            [`${this.path}_not_i`]: value => b =>
+                value === null
+                    ? b.whereNotNull(dbPath)
+                    : b.whereRaw(`lower(${dbPath}) != lower(?)`, [value]).orWhereNull(dbPath),
+        }
+    }
+}
+
+module.exports = {
+    type: 'RegexplessText',
```
SavelevMatthew, line 17: Maybe just `Text`?

AleX83Xpert: `Text` associates with the build-in type. I'd like another name. Anyway, we may change this name if needed.

packages/keystone/fields/RegexplessText/index.js
```diff
@@ -0,0 +1,25 @@
+const { Text } = require('@keystonejs/fields')
+
+class RegexplessKnexFieldAdapter extends Text.adapters.knex {
+
+    equalityConditionsInsensitive (dbPath) {
+        return {
+            [`${this.path}_i`]: (value) => (b) => b.whereRaw(`lower("${dbPath.split('.').join('"."')}") = lower(?)`, [value]),
```
pahaz, line 7: If value null?

packages/keystone/fields/RegexplessText/index.js
pahaz: It’s really better to use Text. Some checks expect this type name.

packages/keystone/fields/RegexplessText/index.js
AleX83Xpert: looks like it will work as empty string
image

packages/keystone/fields/RegexplessText/index.js
AleX83Xpert: Ok :-/

PR #4391, opened 2024-02-14
feat(condo): DOMA-8441 Reuse opened connection to send a batch of VOIP pushes & use RedisGuard to prevent opening multiple connection with APNS

VKislov:
what's changed:

- [x] creating a session not at the time of sending a specific push (in sendPush) but at the time of preparing a batch (in sendAll). This allows you to send a bunch of push notifications using one connection, which is recommended by apple and allows you to fix the error of the connection being dropped by APNS due to too frequent connections to their server.
- [x] use RedisGuard to prevent multiple APNS connections from different workers from being opened at the same time. When multiple connections are open, APNS treats this as DDoS and closes all but the last one.

The second point is a temporary solution. There is an opinion that this will slow down voip push notifications. An alternative to blocking the connection is a dedicated server, which will maintain a constant connection and all pushes from the conda will be proxied through it. Another option is to create a separate queue and a separate worker who processes this queue. In this case, everything will work within Conda, but will not require a dedicated service. There is no clear understanding yet of how to implement this.

### DIFF
--- a/apps/condo/domains/notification/adapters/apple/AppleMessaging.js
+++ b/apps/condo/domains/notification/adapters/apple/AppleMessaging.js
@@ -138,7 +138,6 @@ class AppleMessaging {
                     })
                     stream.write(buffer)
                     stream.end()
-                    resolve()
                 })
                 .catch(err => reject(err))
         })
@@ -197,6 +196,7 @@ class AppleMessaging {
             }
         }
 
+        await this.#session.disconnect()
         return { responses, successCount, failureCount }
     }
 }
--- a/apps/condo/domains/notification/adapters/apple/AppleMessaging.js
+++ b/apps/condo/domains/notification/adapters/apple/AppleMessaging.js
@@ -137,6 +137,7 @@ class AppleMessaging {
             })
             stream.write(buffer)
             stream.end()
+            resolve()
         })
     }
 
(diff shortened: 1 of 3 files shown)


apps/condo/domains/notification/adapters/apple/AppleMessaging.js
```diff
@@ -128,19 +128,24 @@ class AppleMessaging {
...
+                    })
+                    stream.write(buffer)
+                    stream.end()
+                })
+                .catch(err => reject(err))
         })
     }
 
     async sendAll (notifications, isVoIP = false ) {
+        if (typeof this.#url !== 'string') throw new Error('config.url not provided to AppleMessaging instance')
```
SavelevMatthew, line 146: Should it be thrown in constructor instead?

apps/condo/domains/user/utils/serverSchema/guards.js
```diff
@@ -111,6 +111,11 @@ class RedisGuard {
         await this.redis.expire(`${this.lockPrefix}${actionFolder}${variable}`, ttl )
     }
 
+    async unlock (variable, action = '') {
```
SavelevMatthew, line 114: Why not just Redlock, like in webhooks?)

SavelevMatthew (approved):

PR #4398, opened 2024-02-16
feat(keystone): INFRA-265 rate limit observer proof of work

SavelevMatthew:


packages/keystone/rateLimiting/constants.js
```diff
@@ -0,0 +1,22 @@
+const DEFAULT_MUTATION_WEIGHT = 5
+const DEFAULT_QUERY_WEIGHT = 1
+const DEFAULT_MAX_TOTAL_RESULTS = 1000
+const DEFAULT_AUTHED_QUOTA = 5000
+const DEFAULT_NON_AUTHED_QUOTA = 100
+const DEFAULT_QUOTA_WINDOW = '1h'
+const DEFAULT_WHERE_COMPLEXITY_FACTOR = 2
+const DEFAULT_PAGE_LIMIT = 100
```
sitozzz, line 8: Why all these constants starts with `DEFAULT_` prefix? Is there any option to override it's values?
If so, I would prefer to see the following:
```suggestion
const MUTATION_WEIGHT = 5
const QUERY_WEIGHT = 1
const MAX_TOTAL_RESULTS = 1000
const AUTHED_QUOTA = 5000
const NON_AUTHED_QUOTA = 100
const QUOTA_WINDOW = '1h'
const WHERE_COMPLEXITY_FACTOR = 2
const PAGE_LIMIT = 100
```

packages/keystone/rateLimiting/plugin.js
```diff
@@ -0,0 +1,196 @@
+const ms = require('ms')
```
sitozzz, line 1: Does you really need package that converts to milliseconds? There are only two usage of it...

packages/keystone/rateLimiting/plugin.js
```diff
@@ -0,0 +1,196 @@
...
+
+                const { isAuthed, key } = extractQuotaKeyFromRequest(requestContext)
+                const maxQuota = isAuthed ? this.#authedQuota : this.#nonAuthedQuota
+
+                /** @type {string | null} */
+                const currentValue = await this.#redisClient.get(key)
+                const usedQuota = parseInt(currentValue) || 0
+
+                if (usedQuota + requestComplexity > maxQuota) {
+                    // TODO: Throw error instead here
```
sitozzz, line 180: Maybe add some log or metric hit instead of error for now?

sitozzz (commented):
I can't figure out how will you get weights adjustments without collecting special data or logs. You may use logger or `metrics` package

packages/keystone/rateLimiting/constants.js
SavelevMatthew: Yep, you can override all of it, by specifying `new ApolloRateLimitPlugin(keystone, { pageLimit: 10 })` or so

packages/keystone/rateLimiting/plugin.js
SavelevMatthew: In a perfect world (out of MVP) I probably want to override values on `per-query/mutation` level...
So if I do something like...
```
fields: ...
hooks: ...
rateLimit: {
    create: {
        window: "5min"
        quota: 25
    }
}
```

Plugin will generate separate reeds key for this mutation, that's the reason I've introduced this npm package. Quality of life

packages/keystone/rateLimiting/plugin.js
SavelevMatthew: It's already added. I've added all calculated info to `context.req`. And pino plugin will extract it later for Kibana to analyse

sitozzz (approved):

packages/keystone/rateLimiting/plugin.js
```diff
@@ -0,0 +1,196 @@
...
+                    return
```
pahaz, line 181: ?

packages/keystone/rateLimiting/plugin.js
```diff
@@ -0,0 +1,196 @@
...
+                }
+
+                if (currentValue === null) {
```
pahaz, line 184: It's not atomic

PR #4422, opened 2024-02-25
feat(condo): DOMA-8484 migrate reCaptcha v3 to reCaptcha Enterprise

Alllex202:


sitozzz (approved):

apps/condo/domains/user/utils/googleRecaptcha3.js
```diff
@@ -1,45 +1,86 @@
...
         })
 
         const result = await serverAnswer.json()
         logger.info({ msg: 'Captcha result', result })
 
         if (serverAnswer.ok) {
-            const isError = THROW_ERRORS_ON_LOW_CAPTCHA_SCORE && result.score < SAFE_CAPTCHA_SCORE
-            if (result.action !== action) {
-                logger.error(`Captcha actions mismatch ${result.action} - ${action}`)
+            if (!THROW_ERRORS_ON_LOW_CAPTCHA_SCORE) return { error: null }
```
nomerdvadcatpyat, line 67: why we need this logic?

apps/condo/domains/common/utils/request.js
```diff
@@ -0,0 +1,69 @@
...
+/**
+ *
+ * Default sets that used for detect request source by request headers.
+ * Can be used as an example and for development.
+ *
+ * If no header sets is specified for the source, then no request will match. (Ex. []. Equivalent to not indicating the source)
+ * If you specify an empty set of headers, then any request will match. (Ex. [{}])
+ *
+ * By defaults all requests are equal to requests from the site.
+ * You need to override this value in environment variables!
```
nomerdvadcatpyat, line 15: mb add it to `.env.example`?

apps/condo/domains/user/utils/googleRecaptcha3.js
Alllex202: Looks like legacy to allow requests to pass without checking ratings, but still executing a request to verify the token.
I think we can leave this

Alllex202: But I noticed that this check is very early, it should be just before the rating check

nomerdvadcatpyat (approved):

PR #4423, opened 2024-02-27
feat(condo): DOMA-8489 add findRaw server utility

ekabardinsky:


packages/keystone/schema.js
```diff
@@ -136,6 +136,24 @@ function transformByPreprocessors (preprocessors, schemaType, name, schema) {
     }, schema)
 }
 
+async function findRaw (schemaName, queryBuilderFn) {
+    if (isNil(queryBuilderFn)) throw new Error('Omit required parameter queryBuilderFn')
```
toplenboren, line 140:
```suggestion
    if (isNil(queryBuilderFn)) throw new Error('No queryBuilderFn')
```

toplenboren: Please check https://github.com/open-condo-software/condo/blob/eda528eecc7ddd5f028dc2bae888afafc0b46576/packages/keystone/adapterCache.js#L72

PR #4421, opened 2024-02-24
feat(condo): DOMA-8487 create pushes for decline incoming voip call

VKislov:
Partners must call a custom mutation `sendB2CAppPushMessage` with certain parameters and, using these parameters, send a push (regular, not voip) to all devices of a particular resident. When using types, the call is accepted by one of the resident devices and canceled by the caller. At the moment `DECLINED_BY_CALLER` and `RECEIVED_ON_DEVICE` are all types, later the list may expand.

PR #4436, opened 2024-03-01
feat(condo): DOMA-8519 expand sendB2CAppPushMessage to VoIP native support

VKislov:
Mutation types and push types have been expanded. The added fields are necessary for the mobile phone to work directly with the SIP server

type - call interface type, currently always sent 'sip'
address - server address for connection
login - user login to connect to the voip server
password - user password for connecting to the voip server
dtfmCommand - dtfm command for door open button
automaticallyAcceptIncomingCall - Whether to automatically accept an incoming call when you open the VoIP calling interface. Currently always true

### DIFF
--- a/apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.js
+++ b/apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.js
@@ -88,7 +88,7 @@ const SendB2CAppPushMessageService = new GQLCustomSchema('SendB2CAppPushMessageS
         },
         {
             access: true,
-            type: 'input SendB2CAppPushMessageData { body: String!, title: String, B2CAppContext: String }',
+            type: 'input SendB2CAppPushMessageData { body: String!, title: String, B2CAppContext: String, callId: String, voipType: String, voipAddress: String, voipLogin: String, voipPassword: String, voipDtfmCommand: String }',
         },
         {
             access: true,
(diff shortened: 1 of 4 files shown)


apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.js
Alllex202, line 1: where do we use `callId` ?

VKislov: its part of another pr(#4421). This one is asked to be pushed earlier. In principle, this field will not interfere in any way if it remains the same in this pull request.

apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.js
```diff
@@ -88,7 +88,7 @@ const SendB2CAppPushMessageService = new GQLCustomSchema('SendB2CAppPushMessageS
         },
         {
             access: true,
-            type: 'input SendB2CAppPushMessageData { body: String!, title: String, B2CAppContext: String }',
+            type: 'input SendB2CAppPushMessageData { body: String!, title: String, B2CAppContext: String, callId: String, voipType: String, voipAddress: String, voipLogin: String, voipPassword: String, voipDtfmCommand: String }',
```
Alllex202, line 91: It's better to add a new field `voipConfig` and type for this

apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.js
```diff
@@ -172,6 +173,14 @@ const SendB2CAppPushMessageService = new GQLCustomSchema('SendB2CAppPushMessageS
                     },
                 }
 
+                if (voipType) {
```
Alllex202, line 176: maybe check that the `type` is equal to `VOIP_INCOMING_CALL_MESSAGE`?

Alllex202 (approved):

dkoviazin (approved):

apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.js
VKislov: this is to determine the integration there may be. 'sip' or 'basIp'. So far only 'sip'

PR #4441, opened 2024-03-03
fix(condo): DOMA-8542 fixed case when billingReceiptFile was soft deleted

dkoviazin:


apps/condo/domains/billing/schema/RegisterBillingReceiptFileService.test.js
```diff
@@ -20,6 +20,7 @@ describe('RegisterBillingReceiptFileService', () => {
     beforeAll(async () => {
         billingTestUtils = new BillingTestUtils()
         await billingTestUtils.init()
+        process.env.ADDRESS_SERVICE_CLIENT_MODE = 'fake'
```
toplenboren, line 23: Seems really strange to set env vars from test code :-/ it creates unnecessary dependency

Why do we need it? Why can't address client be not fake here?

dkoviazin: :-)

PR #4446, opened 2024-03-05
feat(condo): DOMA-8440 add invoice number to payments export

VKislov:
- [x] export only selectedRows if it is not empty
- [x] fix gql to display unitName in invoice payment export
- [x] add invoice number to payments export

There is a need to display the invoice number in the upload of payments for invoices. I decided to expand the existing template rather than create a separate one for invoices. I chose this solution for two reasons. Firstly, adding a new template will complicate the mutation logic and make it difficult for other developers to understand what is happening. Secondly, Payment has such a field, but it is not required, so if the upload has an empty column, this is normal, since this field is present in the entity. However, there is also a minus. Unloading payments by receipts will always have an empty invoice column

### DIFF
--- a/apps/condo/domains/acquiring/schema/ExportPaymentsService.test.js
+++ b/apps/condo/domains/acquiring/schema/ExportPaymentsService.test.js
@@ -112,12 +112,13 @@ describe('ExportPaymentsService', () => {
         const data = await readXlsx(filename)
 
         expectDataFormat(data, [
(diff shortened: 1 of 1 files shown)


Alllex202 (approved):
You should also update tests for export

nomerdvadcatpyat (approved):

AleX83Xpert (approved):

PR #4453, opened 2024-03-06
feat(condo): DOMA-7635 updated bad words exclusions

Alllex202:
Now the same library is used for exclusions words as for bad words. But only in the opposite direction 🙃
Therefore, now we can add word masks so as not to write all forms of the word


toplenboren (approved):

dkoviazin (commented):
Maybe, use redis to store exclusions and edit it without PR?

Alllex202:
> Maybe, use redis to store exclusions and edit it without PR?

I don't think there is a need for this.
+ we will need to manually update the data on all stands - not very good. 
+ there will be some kind of non-obvious list of exclusions that we can only see in redis - this is strange.

We can move this list of exclusions to `.helm`

AleX83Xpert (approved):

PR #4460, opened 2024-03-07
fix(condo): DOMA-6732 fixed default values for unitType and sectionType

Alllex202:


nomerdvadcatpyat (approved):

apps/condo/domains/common/utils/serverSchema/resolveHelpers.js
```diff
@@ -0,0 +1,45 @@
...
+
+
+const getUnitTypeFieldResolveInput = ({ unitTypeFieldName = 'unitType', unitNameFieldName = 'unitName' } = {}) => ({ resolvedData, existingItem }) => {
+    const newItem = { ...existingItem, ...resolvedData }
+    const unitType = get(newItem, unitTypeFieldName)
+    const unitName = get(newItem, unitNameFieldName)
+    const existedUnitType = get(existingItem, unitTypeFieldName)
+
+    if (!unitName && unitType) {
+        return null
```
AleX83Xpert, line 13: Is it ok to reset `unitType` to null if no `unitName`?
What we should do if we need to send news items to all parkings?

apps/condo/domains/common/utils/serverSchema/resolveHelpers.js
```diff
@@ -0,0 +1,45 @@
...
+    } else if (!unitType && unitName) {
+        if (existedUnitType) {
+            return existedUnitType
+        }
+        return FLAT_UNIT_TYPE
```
AleX83Xpert, line 18:
```suggestion
        return existedUnitType || FLAT_UNIT_TYPE
```

AleX83Xpert, line 18: If we've decided to make 'flat' the default value, describe this decision within the field's schemaDoc.

apps/condo/domains/resident/schema/Resident.test.js
```diff
@@ -79,6 +80,78 @@ describe('Resident', () => {
             const [objCreated] = await createTestResident(adminClient, userClient.user, userClient.property, attrs)
             expect(objCreated.address).toEqual(userClient.property.addressMeta.value)
         })
+
+        describe('unitType and unitName', () => {
```
AleX83Xpert, line 84: Looks like much of similar tests within many files. Why don't to move them to the single place (domains/common/utils/testSchema/...)?

apps/condo/migrations/20240307165344-0372_auto_20240307_1153.js
```diff
@@ -0,0 +1,57 @@
...
+exports.up = async (knex) => {
+    await knex.raw(`
+        BEGIN;
+        
+        SET statement_timeout = '1500s';
+        
+        --
+        -- [CUSTOM] Set unitType to NULL in Tickets and Contacts which have unitName null and not null unitType 
+        --
+        UPDATE "Ticket"
```
AleX83Xpert, line 13: Why `Ticket` and `Contact` only? Don't we need to migrate the rest models to make them synced with new code rules?

apps/condo/domains/news/schema/NewsItemScope.js
```diff
@@ -47,32 +48,6 @@ const NewsItemScope = new GQLListSchema('NewsItemScope', {
                 create: false,
                 update: false,
             },
-            hooks: {
-                resolveInput: async ({ operation, resolvedData, fieldPath }) => {
```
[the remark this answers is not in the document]
Alllex202: Keystone processes hooks in the following order:
1) hooks from fields 
2) hooks from schema

Since a field `unitType` has transformations into `resolveInput` of field, and the value of the `type` field depends on it, this logic needs to be moved to the hooks of the schema

apps/condo/domains/common/utils/serverSchema/resolveHelpers.js
Alllex202: You can consider this to be default behavior.
You can redefine it to suit your needs.
NewsItem uses different logic for this field

apps/condo/domains/common/utils/serverSchema/resolveHelpers.js
Alllex202: ok, I add it to docs, but this is not a new change, it has always been the default unless otherwise stated

apps/condo/domains/resident/schema/Resident.test.js
Alllex202: I don't know how to write tests for hooks without using a schema...
But it is worth noting that not all schemas have the same tests, some have different logic

Alllex202: I can pass some fake arguments without using a schema, but that won't closely resemble real use

apps/condo/migrations/20240307165344-0372_auto_20240307_1153.js
Alllex202: I checked the database for bad data - they were found only in these schemas.
However, I can add migrations for other schemes :)

AleX83Xpert (approved):

nomerdvadcatpyat (approved):

PR #4479, opened 2024-03-16
WiP fix(condo) HOTFIX: add custom fetch.js

toplenboren:
Hey, need to get something to talk about


sitozzz (commented):
what is the need for this package?

toplenboren:
> what is the need for this package?

I think we need to implement our own wrapper around fetch to 

1. Get time elapsed => SLA on Bank related and other external fetch requests
2. Get 400s and other non OK status count 
3. Implement some security gates in the future (or anything like that)

I'm going to propose an ADR next week, need to add some code to the ADR, so we have something to talk about, hence the PR

packages/keystone/fetch.js
```diff
@@ -0,0 +1,41 @@
...
+    const executionContext = getExecutionContext()
+    const parentReqId = executionContext.reqId
+
+    const startTime = Date.now()
+
+    try {
+        const response = await nodeFetch(url, options)
+
+        const endTime = Date.now()
+        const elapsedTime = endTime - startTime
```
Alllex202, line 25: can we track `elapsedTime` not only for successful requests, but also for errors?

toplenboren: Nice suggestion :)

sitozzz (approved):

PR #4483, opened 2024-03-18
feat(global): INFRA-223 tasks queue priority

sitozzz:
telegram-cloud-photo-size-2-5458413884709919593-y

### DIFF
--- a/apps/condo/domains/news/tasks/publishSharedNewsItem.js
+++ b/apps/condo/domains/news/tasks/publishSharedNewsItem.js
@@ -74,4 +74,4 @@ async function publishSharedNewsItem (newsItemSharingId) {
     return await _publishSharedNewsItem(newsItem, newsItemSharing)
 }
 
-module.exports = createTask('publishSharedNewsItem', publishSharedNewsItem, {}, 'low')
+module.exports = createTask('publishSharedNewsItem', publishSharedNewsItem, 'low')
--- a/apps/condo/domains/news/tasks/publishSharedNewsItem.js
+++ b/apps/condo/domains/news/tasks/publishSharedNewsItem.js
@@ -74,4 +74,4 @@ async function publishSharedNewsItem (newsItemSharingId) {
     return await _publishSharedNewsItem(newsItem, newsItemSharing)
 }
 
-module.exports = createTask('publishSharedNewsItem', publishSharedNewsItem, { priority: 3 })
+module.exports = createTask('publishSharedNewsItem', publishSharedNewsItem, {}, 'low')
--- a/apps/condo/domains/notification/tasks/deliverMessage.js
+++ b/apps/condo/domains/notification/tasks/deliverMessage.js
@@ -226,5 +226,5 @@ async function deliverMessage (messageId) {
 }
 
 module.exports = {
-    deliverMessage: createTask('deliverMessage', deliverMessage, 'high'),
+    deliverMessage: createTask('deliverMessage', deliverMessage),
 }
--- a/apps/condo/domains/notification/tasks/deliverMessage.js
+++ b/apps/condo/domains/notification/tasks/deliverMessage.js
@@ -226,5 +226,5 @@ async function deliverMessage (messageId) {
 }
 
 module.exports = {
-    deliverMessage: createTask('deliverMessage', deliverMessage, {}, 'high'),
+    deliverMessage: createTask('deliverMessage', deliverMessage, 'high'),
 }
--- a/apps/condo/domains/notification/tasks/deliverMessage.js
+++ b/apps/condo/domains/notification/tasks/deliverMessage.js
@@ -189,7 +189,7 @@ async function deliverMessage (messageId) {
                 successCnt++
             } else {
                 const [isOk, deliveryMetadata] = await _sendMessageByAdapter(transport, adapter, messageContext, isVoIP)
-                logger.info({ msg: 'sendMessageByAdapter', isOk, messageId, deliveryMetadata })                
+                logger.info({ msg: 'sendMessageByAdapter', isOk, messageId, deliveryMetadata })
                 transportMeta.deliveryMetadata = deliveryMetadata
                 transportMeta.status = isOk ? MESSAGE_SENT_STATUS : MESSAGE_ERROR_STATUS
                 processingMeta.step = isOk ? MESSAGE_SENT_STATUS : MESSAGE_ERROR_STATUS
@@ -226,5 +226,5 @@ async function deliverMessage (messageId) {
 }
 
 module.exports = {
-    deliverMessage: createTask('deliverMessage', deliverMessage, { priority: 1 }),
+    deliverMessage: createTask('deliverMessage', deliverMessage, {}, 'high'),
 }
--- a/packages/keystone/tasks.js
+++ b/packages/keystone/tasks.js
@@ -10,6 +10,11 @@ const { getRedisClient } = require('./redis')
 const { getRandomString } = require('./test.utils')
 
 const TASK_TYPE = 'TASK'
+const DEFAULT_QUEUES = ['low', 'tasks', 'high']
+/* TODO: change this value to 'medium'
+   The 'tasks' name was initially defined as default queue name.
+   So for better migration experience we decided to leave this name as a default queue name */
+const DEFAULT_QUEUE_NAME = 'tasks'
 const WORKER_CONCURRENCY = parseInt(conf.WORKER_CONCURRENCY || '2')
 const IS_BUILD = conf['DATABASE_URL'] === 'undefined'
 // NOTE: If this is True, all tasks will be executed in the node process with setTimeout.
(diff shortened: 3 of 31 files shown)


apps/condo/domains/notification/tasks/deliverMessage.js
```diff
@@ -226,5 +226,5 @@ async function deliverMessage (messageId) {
 }
 
 module.exports = {
-    deliverMessage: createTask('deliverMessage', deliverMessage, { priority: 1 }),
+    deliverMessage: createTask('deliverMessage', deliverMessage, {}, 'high'),
```
SavelevMatthew, line 229: It's better to user medium for all messages, since it's default behaviour. And tweet some delays instead. Few delays to deliver important staff (like sms codes), and few delays to lower status (like message batch)

apps/condo/domains/news/tasks/publishSharedNewsItem.js
```diff
@@ -74,4 +74,4 @@ async function publishSharedNewsItem (newsItemSharingId) {
     return await _publishSharedNewsItem(newsItem, newsItemSharing)
 }
 
-module.exports = createTask('publishSharedNewsItem', publishSharedNewsItem, { priority: 3 })
+module.exports = createTask('publishSharedNewsItem', publishSharedNewsItem, {}, 'low')
```
SavelevMatthew, line 77: It's better to move options to last argument, so avoid spamming `{}` everywhere

packages/keystone/tasks.js
```diff
@@ -10,26 +11,20 @@ const { getRedisClient } = require('./redis')
 const { getRandomString } = require('./test.utils')
 
 const TASK_TYPE = 'TASK'
+/* TODO: change this value to 'medium'
```
SavelevMatthew, line 14: jira task?

apps/condo/domains/news/tasks/publishSharedNewsItem.js
sitozzz: done

packages/keystone/tasks.js
sitozzz: done that

apps/condo/domains/notification/tasks/deliverMessage.js
sitozzz: In fact, I don’t think this is possible without significant changes in the logic of how messages are sent

packages/keystone/tasks.js
```diff
@@ -38,23 +33,65 @@ const CRON_TASKS = new Map()
 const REMOVE_CRON_TASKS = []
 let isWorkerCreated = false
 
-function createTask (name, fn, opts = {}) {
+/**
+ * Create bull queue. For internal use only!
```
SavelevMatthew, line 37: How bout marking `_something` for internal functions?)

packages/keystone/tasks.js
```diff
@@ -38,23 +33,65 @@ const CRON_TASKS = new Map()
...
+
+/**
+ * Creates and register task. After that you can call it with delay or applyAsync methods
+ * @param name {String} name of the task
+ * @param fn {Function} function that will be executed at the worker side
+ * @param queue {String} task queue to which this function will be added. Can be used for change priority of the task execution
+ * @param opts {Object} bull task options
+ * @return {(function(): never)|*}
+ */
+function createTask (name, fn, queue = DEFAULT_QUEUE_NAME, opts = {}) {
```
SavelevMatthew, line 67: Should we check that queue name is correct here?

packages/keystone/tasks.js
```diff
@@ -38,23 +33,65 @@ const CRON_TASKS = new Map()
...
+ * Creates and register cron task. It will be called at the defined period of time
+ * @param name {String} name of cron task
+ * @param cron {String} cron style definition string. For example, 0 12 * * *
+ * @param fn {Function} function that will be executed at the worker side
+ * @param queue {String} task queue to which this function will be added. Can be used for change priority of the task execution
+ * @param opts {Object} bull task options
+ * @return {(function(): never)|*}
+ */
+function createCronTask (name, cron, fn, queue = DEFAULT_QUEUE_NAME, opts = {}) {
     if (typeof fn !== 'function') throw new Error('unsupported fn argument type. Function expected')
```
SavelevMatthew, line 87: Sam about queue name validation

packages/keystone/tasks.js
```diff
@@ -63,9 +100,19 @@ function removeCronTask (name, cron, opts = {}) {
     REMOVE_CRON_TASKS.push([name, taskOpts])
 }
 
-async function _scheduleRemoteTask (name, preparedArgs, preparedOpts) {
-    logger.info({ msg: 'Scheduling task', name, data: { preparedArgs, preparedOpts } })
-    const job = await taskQueue.add(name, { args: preparedArgs }, preparedOpts)
+async function _scheduleRemoteTask (name, preparedArgs, preparedOpts, queue = DEFAULT_QUEUE_NAME) {
+    logger.info({ msg: 'Scheduling task', name, queue, data: { preparedArgs, preparedOpts } })
```
SavelevMatthew, line 104: Maybe `meta` instead of `data` to avoid indexing all args later?

packages/keystone/tasks.js
```diff
@@ -252,53 +337,99 @@ async function createWorker (keystoneModule) {
     if (keystoneModule) {
         await prepareKeystoneExpressApp(keystoneModule)
     } else {
+        createTaskQueue(DEFAULT_QUEUE_NAME)
```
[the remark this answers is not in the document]
SavelevMatthew: Ah, I get it. If keystone is specified then queues will be created inside prepare keystone, otherwise here...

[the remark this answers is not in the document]
SavelevMatthew: Is there any worker without KS module?

packages/keystone/tasks.js
```diff
@@ -252,53 +337,99 @@ async function createWorker (keystoneModule) {
...
+                throw error
+            }
+        })
+
+        queue.on('failed', function (job) {
+            logger.info({ taskId: job.id, status: 'failed', queue: queueName, task: getTaskLoggingContext(job) })
+        })
+
+        queue.on('completed', function (job) {
+            logger.info({ taskId: job.id, status: 'completed', queue: queueName, task: getTaskLoggingContext(job), t0: job.finishedOn - job.timestamp, t1: job.processedOn - job.timestamp, t2: job.finishedOn - job.processedOn })
```
SavelevMatthew, line 388: Maybe we can rename t0 and t1, it will not be obvious to read (I guess)

packages/webhooks/tasks/sendModelWebhooks.js
```diff
@@ -17,5 +17,5 @@ async function sendModelWebhooks (modelName) {
 }
 
 module.exports = {
-    sendModelWebhooks: createTask('sendModelWebhooks', sendModelWebhooks, { priority: 3 }),
-}
\ No newline at end of file
+    sendModelWebhooks: createTask('sendModelWebhooks', sendModelWebhooks),
```
SavelevMatthew, line 20: Please add `low` here. Webhooks are spamming tasks and not that important..,

packages/webhooks/tasks/sendWebhook.js
```diff
@@ -151,5 +152,5 @@ async function sendWebhook (subscriptionId) {
 
 module.exports = {
     trySendData,
-    sendWebhook: createTask('sendWebHook', sendWebhook, { priority: 2 }),
+    sendWebhook: createTask('sendWebHook', sendWebhook),
```
SavelevMatthew, line 155: Same about `low` priority

SavelevMatthew: Hm...I have webhooks in multiple apps (some multi-queued, some are default only), maybe you should add wrapper similar to `getWebhookModels`, so you can specify queue as function args
image

packages/keystone/tasks.js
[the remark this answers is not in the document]
sitozzz: > Is there any worker without KS module?

For test proposes only

packages/webhooks/tasks/sendModelWebhooks.js
sitozzz: Good suggestion, I'll do it

packages/webhooks/tasks/sendWebhook.js
sitozzz: @SavelevMatthew I'm done that. But I would insist that you look at the solution again

packages/keystone/tasks.js
sitozzz: What kind of validation did you mean? There is actually two ways to register bull queues. And we picked up the way, that allow you to define queue at prepareKeystone stage.
That point means, that task creation will triggered much earlier than queues creation. So you can't check anything with this approach at the current context.

packages/keystone/tasks.js
sitozzz: I've changed `data` -> `meta`

packages/webhooks/schema/index.js
```diff
@@ -8,6 +13,19 @@ function getWebhookModels (schemaPath) {
     }
 }
 
+function getWebhookTasks (taskPriority = DEFAULT_QUEUE_NAME) {
+    if (WEBHOOK_TASKS.size === 0) {
```
SavelevMatthew, line 17: It's better to check names here `if (!WEBHOOK_TASKS.has(<task-name>)) // add task `

packages/webhooks/tasks/sendWebhook.spec.js
```diff
@@ -52,6 +55,7 @@ const SendWebhookTests = (appName, actorsInitializer, userCreator, userDestroyer
             const secondUser = await userCreator()
             deletedUser = await userDestroyer(actors.admin, secondUser)
             lastUser = await userCreator()
+            sendWebhook = getWebhookTasks()['sendWebhook']
```
SavelevMatthew, line 58: Why not just import? To init tasks somewhere else?

packages/webhooks/schema/index.js
```diff
@@ -8,6 +13,19 @@ function getWebhookModels (schemaPath) {
     }
 }
 
+function getWebhookTasks (taskPriority = DEFAULT_QUEUE_NAME) {
```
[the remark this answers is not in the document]
sitozzz: Ok, I'll place it inside tasks index

packages/webhooks/tasks/sendWebhook.spec.js
sitozzz: IDK)

SavelevMatthew (commented):
LGTM. But got some questions left:
1. When creating a task and specifying queue should it name be validated? (check that queue name is valid)
2. All this t0 / t1 / t2 logs are not understandable for average developer. Is there any standard here?
3. I still don't believe that deliverMessage must have high priority by default. Some marketing / product members like to throw a message-batch away for thousands of users. How about to put it to default (medium) queue.

packages/keystone/tasks.js
sitozzz: done

SavelevMatthew (approved):

PR #4484, opened 2024-03-18
feat(dev-api): INFRA-281 Emails 2FA and B2CAppAccessRight

SavelevMatthew:


apps/dev-api/domains/miniapp/schema/ImportB2CAppService.js
```diff
@@ -235,6 +236,69 @@ async function importAppBuilds ({ args, context }) {
...
+    }
+
+    if (productionAppId) {
+        await _importAppAccessRightFromEnvironment({
+            condoAppId: productionAppId,
+            environment: PROD_ENVIRONMENT,
+            args,
+            context,
+        })
+    }
```
sitozzz, line 299: Is there an option to a third environment? If no, you can easily do that stuff inside one block of the code. 
Or you want to "import environment" only for dev & production? And, for example, for test proposes there is no need to call this function here?

SavelevMatthew: You can have different users for production and development environment. So, importing app = import all its information from single of both environments including `B2CAppAccessRight`. It can be different (at least user ids in condo.d and v1 are not the same), so we need to import both.

sitozzz (approved):

PR #4487, opened 2024-03-19
feat(condo): DOMA-8348 create _internalSendNotificationNewMobileAppVersion service

VKislov:


apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,140 @@
...
+            access: true,
+            type: `enum App { ${APP_RESIDENT_KEY}, ${APP_MASTER_KEY} }`,
+        },
+        {
+            access: true,
+            type: `enum Platform { ${DEVICE_PLATFORM_ANDROID}, ${DEVICE_PLATFORM_IOS} }`,
+        },
+        {
+            access: true,
+            type: 'input _internalSendNotificationNewMobileAppVersionInput { dv: Int!, sender: JSON! platform: Platform!, app: App!, buildVersion: String!, title: String, body: String, organizationIds: [ID!] }',
```
SavelevMatthew, line 31: sender: SenderFieldInput!

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,140 @@
...
+    
+    mutations: [
+        {
+            access: access.can_internalSendNotificationNewMobileAppVersion,
+            schema: '_internalSendNotificationNewMobileAppVersion(data: _internalSendNotificationNewMobileAppVersionInput!): _internalSendNotificationNewMobileAppVersionOutput',
+            resolver: async (parent, args, context, info, extra = {}) => {
+
+                const { data: { dv, sender, platform, app, buildVersion, title, body, organizationIds } } = args
+
+                const { keystone } = await getSchemaCtx('MessageBatch')
```
SavelevMatthew, line 47: Are you sure you need this? You can just pass context from function argument:
`async (parent, args, context, info, extra)` <--- this one

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,140 @@
...
+                const adminContext = await keystone.createContext({ skipAccessControl: true })
+
+                const userType = app === RESIDENT ? RESIDENT : STAFF
+                let remoteClientKeys
+                //The case when a push is sent to all suitable devices, regardless of organization
+                if (isEmpty(organizationIds)) {
+                    remoteClientKeys = await loadListByChunks({
```
SavelevMatthew, line 54: Is there any estimations on list size? Thousands / millions rows? Maybe it can fail later with OOM error

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,140 @@
...
+                                }
+                            }
+
+                            return rcWithLowerBuild
+                        },
+                    })
+                }
+
+                //The case when a push is sent to suitable devices of a specific or several organizations
+                if (!isEmpty(organizationIds)) {
```
SavelevMatthew, line 80: A bit strange pattern...
```
if (condition) {
    // no return
}
if (!condition) {
    // no return
}
```

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,140 @@
...
+                    for (const organizationId of organizationIds) {
```
SavelevMatthew, line 81: Why not just `{ organization: { id_in: [<all-my-orgs>] } }`
and then `{ owner: { id_in } }`?

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,140 @@
...
+                            list: Resident,
+                            where: {
+                                organization: {
+                                    id: organizationId,
+                                },
+                            },
+                            chunkSize: 50,
+                        })
+
+                        const userIds = residents.map(r => r.user.id)
```
SavelevMatthew, line 93: It's better to move it to chunk processor....since loaded residents (all fields) can be too memory-heavy

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,140 @@
...
+                        })
+                    }
+                }
+
+                const messageBatch = await MessageBatch.create(adminContext, {
+                    dv,
+                    sender,
+                    targets: remoteClientKeys,
+                    title: title || 'title',
+                    message: body || 'body',
```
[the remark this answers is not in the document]
SavelevMatthew: Agree, why do you need this fallbacks? Just make args required :)

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.test.js
```diff
@@ -0,0 +1,116 @@
...
+        const userClient = await makeClientWithResidentAccessAndProperty()
+
+        const addressMetaWithFlat = cloneDeep(userClient.property.addressMeta)
+        addressMetaWithFlat.data.flat = '123'
+        addressMetaWithFlat.data.flat_type = 'кв.'
+        addressMetaWithFlat.value = addressMetaWithFlat.value + ', кв. 123'
+
+        const [objCreated] = await createTestResident(admin, userClient.user, userClient.property, {
+            address: addressMetaWithFlat.value,
+            unitName: faker.random.alphaNumeric(3),
```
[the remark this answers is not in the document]
SavelevMatthew: It should probably fail if we add some validations ;)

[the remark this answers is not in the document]
VKislov: this will not affect the logic of mutation. For mutation, all that is important is the presence of a resident and a record of the device

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.test.js
```diff
@@ -0,0 +1,116 @@
...
+        const admin = await makeLoggedInAdminClient()
+        const userClient = await makeClientWithResidentAccessAndProperty()
+        const userClient2 = await makeClientWithResidentAccessAndProperty()
+
+        const addressMetaWithFlat = cloneDeep(userClient.property.addressMeta)
+        addressMetaWithFlat.data.flat = '123'
+        addressMetaWithFlat.data.flat_type = 'кв.'
+        addressMetaWithFlat.value = addressMetaWithFlat.value + ', кв. 123'
+
+        const [objCreated] = await createTestResident(admin, userClient.user, userClient.property, {
```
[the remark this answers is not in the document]
VKislov: Yes, this test case checks the filtering of devices for a specific organization.

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
VKislov: support does not have access to some entities. So I pull out the keystone and create an admin context

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
AleX83Xpert, line 80: Why not just `else`?
```suggestion
                } else { //The case when a push is sent to suitable devices of a specific or several organizations
```

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,140 @@
...
+
+                            return rcWithLowerBuild
+                        },
+                    })
+                }
+
+                //The case when a push is sent to suitable devices of a specific or several organizations
+                if (!isEmpty(organizationIds)) {
+                    for (const organizationId of organizationIds) {
+                        const residents = await loadListByChunks({
```
AleX83Xpert, line 82: It's too heavy to store all residents models in the memory.
Is it possible to create several `MessageBatch` models within async `chunkProcessor`?

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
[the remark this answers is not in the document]
VKislov: It was planned to use standard text like: “an update is available, you can download it from the store.” But now it seems that it is better to make these parameters required and introduce them every time in mutations

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,151 @@
...
+} = require('@condo/domains/notification/constants/constants')
+const { MessageBatch, RemoteClient } = require('@condo/domains/notification/utils/serverSchema')
+const { Resident } = require('@condo/domains/resident/utils/serverSchema')
+const { RESIDENT, STAFF } = require('@condo/domains/user/constants/common')
+
+const _internalSendNotificationNewMobileAppVersionService = new GQLCustomSchema('_internalSendNotificationNewMobileAppVersionService', {
+    types: [
+        {
+            access: true,
+            type: `enum App { ${APP_RESIDENT_KEY}, ${APP_MASTER_KEY} }`,
```
SavelevMatthew, line 30: Too simple name. `App` can be related to miniapps / billing apps / mobiles apps. Maybe more clear name here?

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,151 @@
...
+        },
+        {
+            access: true,
+            type: `enum Platform { ${DEVICE_PLATFORM_ANDROID}, ${DEVICE_PLATFORM_IOS} }`,
```
SavelevMatthew, line 34: Same here. Maybe `MobileApp`, `MobilePlatform`?

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
SavelevMatthew: You're using server utils. Server utils can take context from mutation freely.

Here's the example:
https://github.com/open-condo-software/condo/blob/master/apps/dev-api/domains/miniapp/schema/PublishB2CAppService.js#L323

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,151 @@
...
+                                type: userType,
+                            },
+                        },
+                        chunkSize: 50,
+                        chunkProcessor: (/** @type {RemoteClient[]} */ chunk) => {
+                            const rcWithLowerBuild = []
+
+                            if (buildVersion) {
+                                for (const rc of chunk) {
+                                    if (get(rc, 'meta.Build', buildVersion) < buildVersion) rcWithLowerBuild.push(`rc:${rc.id}`)
```
SavelevMatthew, line 79: What if there's not meta.build or it's NaN?

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,151 @@
...
+
+                            return rcWithLowerBuild
+                        },
+                    })
+                } else {    //The case when a push is sent to suitable devices of a specific or several organizations
+                    const userIds = await loadListByChunks({
+                        context: adminContext,
+                        list: Resident,
+                        where: {
+                            organization: {
```
SavelevMatthew, line 91: Just to clarify. This will not work with any organisation except `MANAGING_COMPANY`. Are you ok with that?

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,151 @@
...
+                                id_in: organizationIds,
+                            },
+                        },
+                        chunkSize: 50,
+                        chunkProcessor: (/** @type {Resident[]} */ chunk) => {
+                            const userIds = []
+                            chunk.map(resident => {
```
SavelevMatthew, line 98: `map` should return value. You're not, so you creating redundant unused array of `undefined`

SavelevMatthew: So please use `for`-lop here (forEach is acceptable, but not recommended, since it slower)

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,151 @@
...
+                            return rcWithLowerBuild
+                        },
+                    })
+                }
+
+                const messageBatch = await MessageBatch.create(adminContext, {
+                    dv,
+                    sender,
+                    messageType: MOBILE_APP_UPDATE_AVAILABLE_MESSAGE_PUSH_TYPE,
+                    targets: remoteClientKeys,
```
SavelevMatthew, line 136: Are you ok for batch with giga-json? Since target can be 100k+ ids...

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.test.js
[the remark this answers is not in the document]
SavelevMatthew: Any changes in resident domain will break your test without no reason. That's bad practice.

For example, we'll add addressMeta validation match with address, your test swill be broken

apps/condo/domains/resident/constants/constants.js
```diff
@@ -100,7 +100,7 @@ const PAYMENT_CATEGORIES_META = [
     },
     {
         id: '7',
-        name: 'ТКО',
+        name: 'Мусор',
```
SavelevMatthew, line 103: Why this are present in current task?

apps/condo/domains/notification/constants/constants.js
```diff
@@ -950,6 +967,9 @@ const HUAWEI_APP_TYPE_BY_APP_ID = {
 const DEFAULT_TEMPLATE_FILE_EXTENSION = 'njk'
 const DEFAULT_TEMPLATE_FILE_NAME = `default.${DEFAULT_TEMPLATE_FILE_EXTENSION}`
 
+const DEFAULT_TITLE_FOR_NEW_MOBILE_APP_VERSION_PUSH_NOTIFICATION = 'Пора обновиться 😉'
```
SavelevMatthew, line 970: Hey, bro, we don't use this language here :)
Use `intl` please

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
VKislov: If the meta is empty, then get will return the value of the most current build and the push will not sent

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
VKislov: I think from the name of the mutation it’s clear that this is a mobile app

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
VKislov: I think from the name of the mutation it’s clear that this is a mobile app

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
VKislov: I didn't think about it, I'll ask PM if it's ok

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
VKislov: Yes. Butch has no restrictions as far as I know.

apps/condo/domains/resident/constants/constants.js
VKislov: because I made a commit related to another task. I don't see any problem with this
[fix(condo): DOMA-7567 change payment categories translations](https://github.com/open-condo-software/condo/pull/4487/commits/9689341512686e0624a5ffc36de235fdb5445b4b)

apps/condo/domains/notification/constants/constants.js
VKislov: I'm not sure that's true
https://github.com/open-condo-software/condo/blob/master/apps/condo/domains/resident/constants/constants.js

apps/condo/domains/notification/schema/MessageBatch.test.js
```diff
@@ -309,7 +309,7 @@ describe('MessageBatch', () => {
             const [remoteClientObj] = await createTestRemoteClient(admin, payload)
             const remoteClientId = remoteClientObj.id
             const remoteClient = `rc:${remoteClientId}`
-            const [customMessage] = await createTestMessageBatch(admin, { targets: [remoteClient] })
+            const [customMessage] = await createTestMessageBatch(admin, { messageType: CUSTOM_CONTENT_MESSAGE_TYPE, targets: [remoteClient] })
```
SavelevMatthew, line 312: Any extra tests for `MessageBatch.validateInput`

apps/condo/domains/notification/schema/_internalSendNotificationNewMobileAppVersionService.js
```diff
@@ -0,0 +1,167 @@
...
+                                type: userType,
+                            },
+                        },
+                        chunkSize: 50,
+                        chunkProcessor: (/** @type {RemoteClient[]} */ chunk) => {
+                            const rcWithLowerBuild = []
+
+                            if (buildVersion) {
+                                for (const rc of chunk) {
+                                    if (get(rc, 'meta.Build', buildVersion) < buildVersion) rcWithLowerBuild.push(`rc:${rc.id}`)
```
SavelevMatthew, line 75: Strings are compared lexicographically :)))
`'1.1.1(10)' < '1.1.1(2)' === true`

apps/condo/domains/resident/constants/constants.js
SavelevMatthew: It's a bit unclear. Good practice is 1 task = 1 PR, so it's easy to find changes after squash

apps/condo/domains/notification/helpers/compareMobileAppVersion.js
```diff
@@ -0,0 +1,28 @@
+function compareMobileAppVersions (ver1, ver2) {
```
SavelevMatthew, line 1: In compare functions it's common practice to name variables `(a, b)` (which is not readable imho) or `(lhs, rhs)`, which I think is good. (left hand side and right hand side)

SavelevMatthew (approved):

apps/condo/schema.graphql
```diff
@@ -74257,6 +74261,31 @@ type SetMessageStatusOutput {
   status: String!
 }
 
+enum MobileApp {
```
pahaz, line 74264: look at SyncRemoteClientInput, we use `appId: String` is it the same?

PR #4501, opened 2024-03-21
refactor(condo): DOMA-6931 refactored news sending

Alllex202:
# what's going on here?

### changes in the `NewsItem` schema:

_Before_: 
 - the field `sendAt` is set in when the news should be postponed

_After_: 
 - the `sendAt` field is now always set in (сan be `null` if not ready to send yet when `isPublished = false`), and indicates the date when the news sending should begin

---
### changes in tasks

_Before_: 
 - the `notifyResidentsAboutNewsItem` task used `setTimeout` and because of this we could not track the sending process, since the worker completed the task before the news sending process was completed
 
 _After_:
 - the `notifyResidentsAboutNewsItem` task has been refactored and now does not use `setTimeout`
 
 ---
 ### changes in accesses
 
 _Before_:
  - a dynamic filter was previously used to allow residents to read news
 
 _After_:
  - the filter for residents to read news has been simplified

---
### Other
 - Tests for sending news have been added
 - Fixed display of the date of sending news in CRM
 - Removed the button for editing if the news is already being sent out


apps/condo/domains/news/schema/NewsItem.test.js
```diff
@@ -1168,6 +1052,74 @@ describe('NewsItems', () => {
...
+                    await updateTestNewsItem(adminClient, newsItem.id, {
+                        deliverAt: faker.date.soon().toISOString(),
+                    })
+                }, (error) => {
+                    expect(error.errors[0].name).toEqual('UserInputError')
+                    expect(error.errors[0].message).toContain('Field "deliverAt" is not defined by type "NewsItemUpdateInput"')
+                })
+            })
+
+            describe('should be auto-calculate', () => {
```
toplenboren, line 1078:
```suggestion
            describe('should be auto-calculated', () => {
```

apps/condo/domains/news/schema/NewsItem.js
```diff
@@ -136,10 +136,40 @@ const NewsItem = new GQLListSchema('NewsItem', {
         },
 
         sendAt: {
```
toplenboren, line 138: what will happen with this value if user decides to click "No I do not want to send this newsItem"

Also, I think we should reuse 15 seconds from frontned part of our app. Otherwise we will change 15 to 30 in one place. and forget to change in other place

apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.js
```diff
@@ -90,31 +88,41 @@ async function sendNotifications (context, newsItem, taskId) {
...
-                    dateCreated: ['sendAt', 'publishedAt', 'updatedAt', 'createdAt'].reduce((result, field) => (result || get(newsItem, field)), null),
+        try {
+            await sendMessage(contextMessage, {
+                ...DV_SENDER,
+                to: { user: { id: resident.user.id } },
+                lang: conf.DEFAULT_LOCALE,
+                type: defineMessageType(newsItem),
+                meta: {
+                    dv: 1,
+                    title: truncate(newsItem.title, { length: TITLE_MAX_LEN, separator: ' ', omission: '...' }),
```
toplenboren, line 101: It might be better to use deidcated `…` char here, since it `might` (im not sure) take less space in the resulting message

Need to check whether the target system support the encodings though

apps/condo/domains/news/gql.js
```diff
@@ -12,7 +12,7 @@ const { ADDRESS_META_SUBFIELDS_TABLE_LIST } = require('@condo/domains/property/s
 
 const COMMON_FIELDS = 'id dv sender { dv fingerprint } v deletedAt newId createdBy { id name } updatedBy { id name } createdAt updatedAt'
 
-const NEWS_ITEM_FIELDS = `{ organization { id } number title body type validBefore sendAt sentAt isPublished publishedAt ${COMMON_FIELDS} compactScopes { count firstOnes { id unitType unitName property { address addressMeta { ${ADDRESS_META_SUBFIELDS_TABLE_LIST} } } } } }`
+const NEWS_ITEM_FIELDS = `{ organization { id } number title body type validBefore sendAt deliverAt sentAt isPublished publishedAt ${COMMON_FIELDS} compactScopes { count firstOnes { id unitType unitName property { address addressMeta { ${ADDRESS_META_SUBFIELDS_TABLE_LIST} } } } } }`
```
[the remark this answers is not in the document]
Alllex202: I will add a description to this pr to make it clearer what is changing

apps/condo/domains/news/schema/NewsItem.js
```diff
@@ -136,10 +137,40 @@ const NewsItem = new GQLListSchema('NewsItem', {
...
+            type: 'DateTimeUtc',
+            hooks: {
+                resolveInput: async ({ existingItem, resolvedData, fieldName }) => {
+                    const newItem = { ...existingItem, ...resolvedData }
+                    const prevIsPublished = get(existingItem, 'isPublished', false)
+                    const sendAt = get(newItem, 'sendAt', null)
+                    const isPublished = get(newItem, 'isPublished', false)
+                    const updatedIsPublished = prevIsPublished !== isPublished
+
+                    if (!isPublished || !updatedIsPublished) return get(resolvedData, fieldName)
```
AleX83Xpert, line 162: Should we set null if the newsItem was unpublished?

apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.js
```diff
@@ -90,31 +89,41 @@ async function sendNotifications (context, newsItem, taskId) {
...
-                    newsItemId: newsItem.id,
-                    organizationId: newsItem.organization.id,
-                    userId: resident.user.id,
-                    residentId: resident.id,
-                    userRelatedResidentsIds: get(residentsIdsByUser, resident.user.id, []).join(','),
-                    url: `${conf.SERVER_URL}/newsItem/${newsItem.id}`,
-                    validBefore: get(newsItem, 'validBefore', null),
-                    // The first truthy value will be returned, or null if no values are found.
-                    dateCreated: ['sendAt', 'publishedAt', 'updatedAt', 'createdAt'].reduce((result, field) => (result || get(newsItem, field)), null),
+        try {
```
AleX83Xpert, line 94: Will `Promise.all` decrease sending time?
I mean something like this
```js
Promise.all(residentsData.map((resident) => new Promise((resolve) => {
    sendMessage(...).then(() => resolve({ err: null })).catch((err) => resolve({ err: err.message }))
})))
```

apps/condo/domains/news/schema/NewsItem.js
Alllex202: Yes, we can

apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.js
Alllex202: Please don't send 100500 simultaneous requests :)

Alllex202: Each message will still be delivered separately one after another (in the worker, through the "deliverMessage" task)

apps/condo/domains/news/gql.js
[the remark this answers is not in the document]
toplenboren: Alright, got you

But this seems really counter intuitive.. Need docs or better naming

apps/condo/domains/news/schema/NewsItem.js
Alllex202: >  if user decides to click "No I do not want to send this newsItem"

This news item will be deleted (using `deletedAt`)

Alllex202: > Also, I think we should reuse 15 seconds from frontned part of our app. Otherwise we will change 15 to 30 in one place. and forget to change in other place

Fixed

apps/condo/domains/news/gql.js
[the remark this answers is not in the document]
Alllex202: Unfortunately, the name `sendTo` is already taken :(

apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.js
Alllex202: I think there is no need to change something that already works well :)

apps/condo/domains/news/gql.js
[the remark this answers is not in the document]
Alllex202: >  or better naming
Ok, I'll think about it

apps/condo/domains/news/schema/NewsItem.js
Alllex202: Updated

apps/condo/domains/news/gql.js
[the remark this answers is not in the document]
toplenboren: Just to be clear:

Why couldnt we set sendAt to `publishedAt + time to cancel sending 15 seconds - for news to be sent now` ?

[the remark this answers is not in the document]
Alllex202: Yeah, I wanted to do that, but...
Now the `sendAt` field is an indicator that the news was sent with a planned sending date in the future.
If it is full, then the news is planned; if it is empty, then the news is being sent right now.
I can’t yet say for sure whether we need this information now or whether this information may be useful in the future, but if we start always filling out the `sendAt` field, we will lose it.

However, we can add a separate flag for this.  (If needed)

I think it's worth discussing this point. if it doesn't affect anything, then why not

toplenboren (approved):

apps/condo/pages/news/[id]/index.tsx
```diff
@@ -83,6 +85,8 @@ const NewsItemCard: React.FC = () => {
     const ConfirmDeleteMessage = intl.formatMessage({ id: 'news.ConfirmDeleteMessage' })
     const CancelMessage = intl.formatMessage({ id: 'news.CancelMessage' })
 
+    const { publicRuntimeConfig: { newsItemsSendingDelay } } = getConfig()
```
Comment (nomerdvadcatpyat) [inline apps/condo/pages/news/[id]/index.tsx:88]: move this outside of component

nomerdvadcatpyat (approved):

PR #4510, opened 2024-03-22
WIP migrate to hcapthca

Alllex202:


apps/condo/pages/auth/forgot.tsx
```diff
@@ -47,7 +47,7 @@ function ResetPageView () {
     const REGISTER_PHONE_LABEL = <label style={ { alignSelf:'end' } }>{PhoneMsg}</label>
 
     const [form] = Form.useForm()
-    const { executeRecaptcha } = useGoogleReCaptcha()
+    const { executeCaptcha } = useHCaptcha()
```
toplenboren, line 50: Can we make HCaptcha optional? Our open-source users might not want to setup an additional dependency

Alllex202: Updated

Now on the front-end the captcha can be without configuration or disabled. If this is so, then instead of the token the text `condo-captcha-disabled` will be sent (this is just a placeholder, there could be any other text here, it does not affect anything, just the token field is required in the API)

Everything on the server remains the same. If the `DISABLE_CAPTCHA` flag is set, then the captcha check will not work. If for some reason you have not specified the configuration, the corresponding warnings and errors will be displayed. Therefore, if you want to disable the captcha, then this must be done explicitly through the `DISABLE_CAPTCHA` flag

apps/condo/domains/common/components/HCaptcha.tsx
```diff
@@ -0,0 +1,84 @@
...
+})
+
+const useHCaptcha = (): IHCaptchaContext => useContext(HCaptchaContext)
+
+const HCaptchaProvider: FC = ({ children }) => {
+    const intl = useIntl()
+    const requestFailedMessage = intl.formatMessage({ id: 'global.errors.requestFailed.title' })
+    const OkMessage = intl.formatMessage({ id: 'OK' })
+
+    const { publicRuntimeConfig: { hCaptcha: hCaptchaConfig, disableCaptcha } } = getConfig()
```
nomerdvadcatpyat, line 26: is `disableCaptcha` are equal than no `hCaptchaConfig`?

apps/condo/domains/user/utils/hCaptcha.js
```diff
@@ -0,0 +1,120 @@
...
+            if (!get(result, 'success', false)) {
+                const errorCode = get(result, 'error-codes')
+                const errorMessage = get(ERROR_MESSAGES, errorCode, 'Unknown error')
+                return { error: `Captcha is invalid: ${errorCode} (${errorMessage})` }
+            }
+
+            // if (!THROW_ERRORS_ON_LOW_CAPTCHA_SCORE) return { error: null }
+            // NOTE: Only works in Enterprise plan! (risk score: human 0 -> 1 bot)
+            // const riskScore = get(result, 'score', 1)
+            // if (riskScore >= SAFE_CAPTCHA_SCORE) return { error: `Low captcha score ${riskScore}` }
```
nomerdvadcatpyat, line 106: Is this comment needed?

apps/condo/lang/ru/ru.json
```diff
@@ -289,6 +289,7 @@
   "global.globalHints.canDisableHints.inProfile": "профиле",
   "global.globalHints.banner.more": "Как это работает",
   "global.errors.serverError.title": "Ошибка",
+  "global.errors.requestFailed.title": "Не удалось выполнить запрос",
```
nomerdvadcatpyat, line 292: mb write a more clear comment? something about captcha check

nomerdvadcatpyat (approved):

apps/condo/domains/common/components/HCaptcha.tsx
Alllex202: No. To turn off the captcha, you need to use the `DISABLE_CAPTCHA` flag

apps/condo/domains/user/utils/hCaptcha.js
Alllex202: Can be deleted

apps/condo/lang/ru/ru.json
Alllex202: This error is shown when the captcha token could not be created. It is unlikely that we should inform the user about this. You will be able to see a specific error in the console, and it definitely does not need to be shown to the user*

Moreover, the chance of seeing this error is extremely low.

------

`*` -  list of possible errors

Снимок экрана 2024-03-25 в 17 41 49

apps/condo/domains/common/components/HCaptcha.tsx
Alllex202: Now on the front-end the captcha can be without configuration or disabled. If this is so, then instead of the token the text condo-captcha-disabled will be sent (this is just a placeholder, there could be any other text here, it does not affect anything, just the token field is required in the API)

Everything on the server remains the same. If the DISABLE_CAPTCHA flag is set, then the captcha check will not work. If for some reason you have not specified the configuration, the corresponding warnings and errors will be displayed. Therefore, if you want to disable the captcha, then this must be done explicitly through the DISABLE_CAPTCHA flag

apps/condo/domains/user/utils/hCaptcha.js
Alllex202: deleted

PR #4512, opened 2024-03-25
fix(condo): INFRA-229 reduce sentry performance impact

sitozzz:


packages/keystone/sentry/ApolloSentryPlugin.js
```diff
@@ -1,5 +1,11 @@
 const Sentry = require('@sentry/node')
 
+const IGNORED_ERRORS = [
+    'No or incorrect authentication credentials',
+    'Your request exceeded server limits',
+    'No or incorrect authentication credentials',
```
nomerdvadcatpyat, line 6: remove duplicated string

PR #4533, opened 2024-03-27
feat(dev-portal): INFRA-243 stricter icon requirements

SavelevMatthew:
## Before

image

## After

https://github.com/open-condo-software/condo/assets/50069872/59b54866-b1fc-4a51-b727-d4fe927125d8


Alllex202 (approved):
What about the size limit in bytes? 
Can we add it to the description?

toplenboren:
It is best to provide templates for icon too. In png / svg format. So they can easily import it to their editing tool

SavelevMatthew:
> What about the size limit in bytes?
> Can we add it to the description?

I don't think so, 1mb is more than enough for any png in specified size and this limits are flexible. Size and color limits are strict, so we need to specify it explicitly

SavelevMatthew:
> It is best to provide templates for icon too. In png / svg format. So they can easily import it to their editing tool

Ty for awesome advice. Added a link to our style guides 👍

PR #4548, opened 2024-03-30
fix(condo): DOMA-8650 fixed case when address was not resolved correctly

dkoviazin:
1. Fixed find organization properties query in account resolver
2. Some times globalId was saved with unit in property
3. If dadata returns the correct value not in first place of array - wrong address was registered

### DIFF
--- a/apps/condo/domains/billing/schema/resolvers/accountResolver.js
+++ b/apps/condo/domains/billing/schema/resolvers/accountResolver.js
@@ -69,7 +69,7 @@ class AccountResolver extends Resolver {
                 const sameNumberAccount = this.accounts.find(({ number }) => number === accountNumber)
                 if (sameNumberAccount) {
                     const oldBillingProperty = await getById('BillingProperty', sameNumberAccount.property)
-                    const [organizationProperty] = await find('Property', { addressKey: oldBillingProperty.addressKey })
+                    const [organizationProperty] = await find('Property', { addressKey: oldBillingProperty.addressKey, deletedAt: null, organization: { id: get(this.billingContext, 'organization.id') } })
                     if (!organizationProperty) {
                         existingAccount = sameNumberAccount
                     } else {
--- a/apps/condo/domains/billing/schema/resolvers/propertyResolver.js
+++ b/apps/condo/domains/billing/schema/resolvers/propertyResolver.js
@@ -14,6 +14,7 @@ const {
 } = require('@condo/domains/billing/constants/registerBillingReceiptService')
 const { Resolver } = require('@condo/domains/billing/schema/resolvers/resolver')
 const { isValidFias, normalizePropertyGlobalId } = require('@condo/domains/billing/schema/resolvers/utils')
+const { UUID_REGEXP } = require('@condo/domains/common/constants/regexps')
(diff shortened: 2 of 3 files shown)


apps/condo/domains/billing/schema/resolvers/utils.js
```diff
@@ -36,6 +36,7 @@ const isPerson = (fullName) => {
 }
 
 const isValidFias = (fias = '') => FIAS_REGEXP.test(fias)
+const normalizePropertyGlobalId = (fiasCode) => fiasCode ? fiasCode.split(',') [0] : ''
```
pahaz, line 39: looks like magic :)

dkoviazin: added more letters :-)

apps/condo/domains/billing/schema/resolvers/propertyResolver.js
```diff
@@ -208,8 +216,8 @@ class PropertyResolver extends Resolver {
                 if (!updated.has(existingProperty.id)) {
                     const updateInput = this.buildUpdateInput({
                         importId: importIdInput,
-                        globalId,
-                        address: resultAddressKey !== existingProperty.addressKey ? address : null,
+                        globalId: normalizePropertyGlobalId(globalId),
+                        address: resultAddressKey !== existingProperty.addressKey ? this.addressFieldValue(address, resultAddressKey) : null,
```
AleX83Xpert, line 220: We are trying to set the address to null in case the address wasn't changed. Is it ok?

apps/condo/domains/billing/schema/resolvers/accountResolver.js
```diff
@@ -69,7 +69,7 @@ class AccountResolver extends Resolver {
                 const sameNumberAccount = this.accounts.find(({ number }) => number === accountNumber)
                 if (sameNumberAccount) {
                     const oldBillingProperty = await getById('BillingProperty', sameNumberAccount.property)
-                    const [organizationProperty] = await find('Property', { addressKey: oldBillingProperty.addressKey })
+                    const [organizationProperty] = await find('Property', { addressKey: oldBillingProperty.addressKey, deletedAt: null, organization: { id: get(this.billingContext, 'organization.id') } })
```
AleX83Xpert, line 72: Why not `getByCondition`?

apps/condo/domains/billing/schema/resolvers/utils.js
```diff
@@ -36,6 +36,14 @@ const isPerson = (fullName) => {
 }
 
 const isValidFias = (fias = '') => FIAS_REGEXP.test(fias)
+const normalizePropertyGlobalId = (fiasCode) => {
+    // FIAS CODE can contain information about unit separated with comma. We need only house part
```
AleX83Xpert, line 40: It's better to move this description to the function's jsdoc
```js
/**
```

AleX83Xpert (approved):

apps/condo/domains/billing/schema/resolvers/propertyResolver.js
dkoviazin: It will be compared with existing property. So, it's ok

apps/condo/domains/billing/schema/resolvers/utils.js
dkoviazin: Done

PR #4551, opened 2024-03-31
fix(condo): DOMA-8746 we need to build packages before any migrations

pahaz:


toplenboren (approved):

README.md
```diff
@@ -26,11 +26,13 @@ docker-compose up -d postgresdb redis
 # 4) create base .env file
 cp .env.example .env
 
-# 5) create databases for all apps and apps/.env files
+# 5) prepare local packages
+yarn build
```
SavelevMatthew, line 30: I don't think this is the right fix. Suggested in issues solution was workaround, not TODO.

SavelevMatthew (commented):
It's strange DX when you need to build icons to migrate DB. If you want to fix it properly - you should exclude NextJS at migration part. (Probably you can exclude any app except KS itself to obtain KS, but Next is enough to fix the issue)

toplenboren:
I'd say ship it

PR #4570, opened 2024-04-03
fix(condo): DOMA-8763 temporarily turn off MeterReading validation

abshnko:
basic validation will be moved to Meters miniapp, since in the condo domain it affects not only mobile app users, but also importing MeterReadings from xlsx on the platform which we don't want


dkoviazin (approved):

SavelevMatthew (commented):
Why temporary? No todos / notes / PR comments of smthing

If this a new logic (allow smaller readings), than add a test proving it...

Right now I see only removing business logic with "temporary" quote, so what's going on?

abshnko:
> Why temporary? No todos / notes / PR comments of smthing
> 
> If this a new logic (allow smaller readings), than add a test proving it...
> 
> Right now I see only removing business logic with "temporary" quote, so what's going on?

This is not new logic --  I added validation recently and it will be moved to a new miniapp, because it should not be here cause it affects the platform

PR #4575, opened 2024-04-04
feat(condo): INFRA-294 pass come contact information to resident 🔀

SavelevMatthew:


apps/condo/domains/resident/schema/Resident.js
```diff
@@ -46,12 +46,41 @@ const addressFieldHooks = {
...
+        deletedAt: null,
+        type: RESIDENT_USER_TYPE,
+    })
+
+    if (!user || !user.phone) {
+        return null
+    }
+
+    return await getByCondition('Contact', {
+        property: { id: propertyId },
```
AleX83Xpert, line 70: Do we need to add the filtration by organization?
```js
organization: item.organization
```

AleX83Xpert (approved):

apps/condo/domains/resident/schema/Resident.js
```diff
@@ -198,6 +227,30 @@ const Resident = new GQLListSchema('Resident', {
                 resolveInput: getUnitTypeFieldResolveInput(),
             },
         },
+        isVerifiedByManagingCompany: {
```
sitozzz, line 230: Why not just `isVerified`? It's to complex field name. Also we already have fields with same logic

SavelevMatthew: It will not be as clear for API users as current name.

Most likely, soon information about not only the **managing company**, but also about **the service providers** will be included here. This naming emphasizes the link to the managing company, as I cannot guarantee that in some future we will not have to add similar fields for service providers.

PR #4577, opened 2024-04-05
feat(condo): DOMA-8761 added the ability to use two sbbоl configs with different client_id

VKislov:


Alllex202:
Some tests for new logic?

dkoviazin (approved):

Alllex202 (approved):

PR #4580, opened 2024-04-05
feat(global): INFRA-298 remove virtual fields from history records

SavelevMatthew:


nomerdvadcatpyat (approved):

apps/condo/migrations/20240405125421-0378_remove_bankcontractoraccounthistoryrecord_relatedtransactions_and_more.js
```diff
@@ -0,0 +1,266 @@
...
+
+exports.up = async (knex) => {
+    await knex.raw(`
+    BEGIN;
```
sitozzz, line 6: Maybe add a statement timeout here? That's a lot of data to remove

SavelevMatthew: @reactpro said, that its ok, since drop column = mark at as hidden. So it's fast. And then garbage collector clears it in background

PR #4591, opened 2024-04-09
feat(condo): INFRA-300 Message prioritization 📧📈

SavelevMatthew:


apps/condo/domains/notification/schema/SendMessageService.js
```diff
@@ -76,6 +83,19 @@ async function checkSendMessageMeta (type, meta, context) {
...
+ * Gets condo tasks queue on which to start message delivery
+ * @param {string} type Message type
+ * @returns {'medium' | 'high' | 'low'} condo task queue
+ */
+function getMessageQueue ({ type }) {
+    // TODO(VKislov): separate queue for VoIP here?
+    const defaultQueue = MESSAGE_DELIVERY_PRIORITY_TO_TASK_QUEUE_MAP[MESSAGE_DELIVERY_DEFAULT_PRIORITY]
+    const messagePriority = get(MESSAGE_DELIVERY_OPTIONS, [type, 'priority'], MESSAGE_DELIVERY_DEFAULT_PRIORITY)
+
+    return get(MESSAGE_DELIVERY_PRIORITY_TO_TASK_QUEUE_MAP, messagePriority, defaultQueue)
```
sitozzz, line 96: There is already a default queue in the `tasks` package. Why don't you to reuse it?

apps/condo/domains/notification/constants/constants.js
```diff
@@ -651,9 +651,15 @@ const MESSAGE_TYPES = Object.keys(MESSAGE_META)
...
 const MESSAGE_DELIVERY_STRATEGY_AT_LEAST_ONE_TRANSPORT = 'atLeastOneTransport'
 const MESSAGE_DELIVERY_STRATEGY_ALL_TRANSPORTS = 'allTransports'
+
 const MESSAGE_DELIVERY_DEFAULT_PRIORITY = 'default'
 const MESSAGE_DELIVERY_SLOW_PRIORITY = 'slow'
 const MESSAGE_DELIVERY_FAST_PRIORITY = 'fast'
+const MESSAGE_DELIVERY_PRIORITY_TO_TASK_QUEUE_MAP = {
+    [MESSAGE_DELIVERY_FAST_PRIORITY]: 'high',
+    [MESSAGE_DELIVERY_DEFAULT_PRIORITY]: 'medium',
+    [MESSAGE_DELIVERY_SLOW_PRIORITY]: 'low',
```
AleX83Xpert, line 661: Why not just priority as an integer value (a higher value is more prioritized)?
What if we will need something between medium and high?
How these 3 queues will work?

apps/condo/domains/notification/constants/constants.js
```diff
@@ -651,9 +651,15 @@ const MESSAGE_TYPES = Object.keys(MESSAGE_META)
 
 const MESSAGE_DELIVERY_STRATEGY_AT_LEAST_ONE_TRANSPORT = 'atLeastOneTransport'
 const MESSAGE_DELIVERY_STRATEGY_ALL_TRANSPORTS = 'allTransports'
+
 const MESSAGE_DELIVERY_DEFAULT_PRIORITY = 'default'
 const MESSAGE_DELIVERY_SLOW_PRIORITY = 'slow'
 const MESSAGE_DELIVERY_FAST_PRIORITY = 'fast'
+const MESSAGE_DELIVERY_PRIORITY_TO_TASK_QUEUE_MAP = {
+    [MESSAGE_DELIVERY_FAST_PRIORITY]: 'high',
+    [MESSAGE_DELIVERY_DEFAULT_PRIORITY]: 'medium',
```
sitozzz, line 660: Actually I just found a bug. I forgot to remap queue name provided from `applyAsync` method with `TASK_QUEUE_REMAPPING`... 
So that's line doesn't work and fail with error in runtime. 
Anyway, you have two options right now:

1. Change default task priority constant to default task name from package (also you can reuse constant value exported from that package)
```suggestion
    [MESSAGE_DELIVERY_DEFAULT_PRIORITY]: DEFAULT_QUEUE_NAME,
```
2. Add remapping logic same as a code inside `registerTaskQueues` method

apps/condo/domains/notification/schema/SendMessageService.js
```diff
@@ -76,6 +83,19 @@ async function checkSendMessageMeta (type, meta, context) {
     }
 }
 
+/**
+ * Gets condo tasks queue on which to start message delivery
+ * @param {string} type Message type
+ * @returns {'medium' | 'high' | 'low'} condo task queue
+ */
+function getMessageQueue ({ type }) {
+    // TODO(VKislov): separate queue for VoIP here?
```
sitozzz, line 92: Jira task?

apps/condo/domains/notification/schema/SendMessageService.js
```diff
@@ -152,7 +172,7 @@ const SendMessageService = new GQLCustomSchema('SendMessageService', {
                     ? messageWithSameUniqKey
                     : await Message.create(context, messageAttrs)
 
-                if (!messageWithSameUniqKey) await deliverMessage.delay(message.id)
+                if (!messageWithSameUniqKey) await deliverMessage.applyAsync([message.id], getMessageQueue({ type }))
```
AleX83Xpert, line 175: What about changing the `deliverMessage()` signature?

SavelevMatthew: `delay` takes args like so:
```
function delay(...args) {
   console.log(args)
}
```
`applyAsync` has the following signature:
```
function applyAsync(args, queue, bullOptions) {
    console.log(args)
}
```

So `delay(arg1, arg2, arg3)` is equal to `applyAsync([arg1, arg2, arg3])`, but `applyAsync` can take extra options, that's the main difference

apps/condo/domains/notification/schema/fields/MessageType.js
```diff
@@ -0,0 +1,28 @@
...
+    return Math.pow(2, pow)
+}
+
+function getMessageTypeField ({ isRequired, ...extraAttrs }) {
+    return {
+        type: 'Select',
+        options: MESSAGE_TYPES,
+        graphQLReturnType: 'MessageType',
+        isRequired,
+        kmigratorOptions: { null: !isRequired, max_length: getMaxLength() },
```
AleX83Xpert, line 19: Dynamic `max_length` means possible unplanned migration. Is it ok?

apps/condo/domains/notification/constants/constants.js
SavelevMatthew: That's our architectural decision. Before we used to specify numeric priority. Which is (probably) powered by [priority queue](https://en.wikipedia.org/wiki/Priority_queue) data structure. This solution usually requires `O(logN)` to insert element and not even supported well outside of Redis provider.

The other popular way is to use N named queues. So you can create common worker to get tasks from all of them. Or worker for a specific queue (for example, dedicated worker for `high`-queue tasks, which will process data even if all other workers are full)

apps/condo/migrations/20240409162533-0380_alter_message_type_alter_messageappblacklist_type_and_more.js
```diff
@@ -0,0 +1,66 @@
...
+
+exports.up = async (knex) => {
+    await knex.raw(`
+    BEGIN;
+--
+-- Alter field type on message
+--
+ALTER TABLE "Message" ALTER COLUMN "type" TYPE varchar(128) USING "type"::varchar(128);
```
AleX83Xpert, line 10: Maybe increase the timeout before this query?

AleX83Xpert (commented):
Why we can not use [bull's priorities](https://optimalbits.github.io/bull/#prioritized)?

apps/condo/domains/notification/schema/SendMessageService.js
SavelevMatthew: There's small difference in logic:
1. Your suggestion is - use default app queue as default.
2. Mine is - use **same queue** as **default priority uses**

For now they're the same, since they're both pointing to `medium/default`. But IMHO the second option is more accurate in terms of possible future changes

sitozzz:
> Why we can not use [bull's priorities](https://optimalbits.github.io/bull/#prioritized)?

Because of how this mechanism works under the hood. In short, this is due to the fact that priority in the queue only works with redis. Also that's really expensive operation for redis instance.

apps/condo/domains/notification/schema/SendMessageService.js
sitozzz: Ok, sounds reasonable

apps/condo/domains/notification/schema/fields/MessageType.js
SavelevMatthew: It's not that dynamic. Its growth algorithm is similar to list capacity growth. The idea is to limit length from some power of 2. So now max_len of all constants is `73`, so the closest power of 2 to it is `128`. So it will be static (until you create some type with `> 128` symbols :)

packages/keystone/tasks.js
```diff
@@ -100,7 +100,9 @@ function removeCronTask (name, cron, opts = {}) {
     REMOVE_CRON_TASKS.push([name, taskOpts])
 }
 
-async function _scheduleRemoteTask (name, preparedArgs, preparedOpts, queue = DEFAULT_QUEUE_NAME) {
+async function _scheduleRemoteTask (name, preparedArgs, preparedOpts, queueName = DEFAULT_QUEUE_NAME) {
+    const queue = TASK_QUEUE_REMAPPING.hasOwnProperty(queueName) ? TASK_QUEUE_REMAPPING[queueName] : queueName
```
AleX83Xpert, line 104:
```suggestion
    const queue = get(TASK_QUEUE_REMAPPING, queueName, queueName)
```

AleX83Xpert (approved):

PR #4604, opened 2024-04-11
fix(condo) HOTFIX: add ability to track requests between microservices

toplenboren:


packages/keystone/fetch.js
```diff
@@ -9,14 +9,33 @@ const logger = getLogger('fetch')
...
+    //
+    if (!skipSettingHeaders) {
+        if (!options.headers) {
+            options.headers = {}
+        }
+
+        options.headers['X-Request-Id'] = parentReqId || parentTaskId
+        options.headers['reqId'] = parentReqId
+        options.headers['taskId'] = parentTaskId
+    }
```
sitozzz, line 38: What if no parentReqId or parentTaskId was provided?

toplenboren: what do you think?

sitozzz: I suppose you should set value explicitly. In this case we can filter requests without that prop

sitozzz (approved):

PR #4637, opened 2024-04-18
hotfix: revert migration type conversion

SavelevMatthew:


apps/condo/migrations/20240418134550-0387_alter_message_type_alter_messageappblacklist_type_and_more.js
```diff
@@ -0,0 +1,66 @@
...
+
+exports.up = async (knex) => {
```
toplenboren, line 4: Dont we need to add a timeout here too? :D

toplenboren (approved):
lgtm!

AleX83Xpert (approved):

apps/condo/migrations/20240418134550-0387_alter_message_type_alter_messageappblacklist_type_and_more.js
SavelevMatthew: Yep, can be helpful for `condo.d`

PR #4638, opened 2024-04-18
fix(condo): DOMA-8684 fixed contact editor validator for name

Alllex202:
## Before

https://github.com/open-condo-software/condo/assets/56914444/022821d5-df68-44cd-93e1-041ec0589eb5

## After

https://github.com/open-condo-software/condo/assets/56914444/65ccde3b-e54b-4323-9077-64342187f2df

### DIFF
--- a/apps/condo/domains/contact/components/ContactsEditor/NewContactFields.tsx
+++ b/apps/condo/domains/contact/components/ContactsEditor/NewContactFields.tsx
@@ -2,7 +2,6 @@ import styled from '@emotion/styled'
 import { AutoComplete, Col, Form, FormInstance, FormItemProps, InputProps, Row, RowProps } from 'antd'
 import get from 'lodash/get'
 import isEmpty from 'lodash/isEmpty'
-import isString from 'lodash/isString'
 import { Rule } from 'rc-field-form/lib/interface'
 import React, { CSSProperties, useCallback, useEffect, useMemo, useState } from 'react'
 
@@ -90,7 +89,7 @@ const NewContactFields: React.FC<INewContactFieldsFieldsProps> = ({
     const handleChangeContact = (field) => (fieldValue) => {
         const newValue = {
             ...value,
-            [field]: isString(fieldValue) ? fieldValue.trim() : fieldValue,
+            [field]: typeof fieldValue === 'string' ? fieldValue.trim() : fieldValue,
         }
         setValueAndTriggerOnChange(newValue)
     }
--- a/apps/condo/domains/contact/components/ContactsEditor/NewContactFields.tsx
+++ b/apps/condo/domains/contact/components/ContactsEditor/NewContactFields.tsx
@@ -1,9 +1,10 @@
 import styled from '@emotion/styled'
-import { AutoComplete, Col, Form, FormItemProps, InputProps, Row, RowProps } from 'antd'
+import { AutoComplete, Col, Form, FormInstance, FormItemProps, InputProps, Row, RowProps } from 'antd'
 import get from 'lodash/get'
 import isEmpty from 'lodash/isEmpty'
+import isString from 'lodash/isString'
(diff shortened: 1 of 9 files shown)


toplenboren (approved):

apps/condo/domains/contact/components/ContactsEditor/NewContactFields.tsx
```diff
@@ -35,12 +37,13 @@ interface INewContactFieldsFieldsProps {
...
     newContactNameFormItemProps?: FormItemProps
     disabled?: boolean
+    form: FormInstance,
 }
 
 const FIELD_WRAPPER_COL = { span: 24 }
 const FIELD_ROW_GUTTER: RowProps['gutter'] = [16, 0]
 const RADIO_COL_SMALL_SCREENS_STYLE: CSSProperties = { position: 'relative', top: '62px' }
-const RADIO_COL_LARGE_SCREENS_STYLE: CSSProperties = { height: '48px' }
+const RADIO_COL_LARGE_SCREENS_STYLE: CSSProperties = { height: '48px', marginTop: '52px' }
```
SavelevMatthew, line 46: It's a bit strange value for margin. We usually use 24, 32, 40, 60 px. Check, that it is correct please

apps/condo/domains/contact/components/ContactsEditor/NotResidentFields.tsx
```diff
@@ -3,11 +3,14 @@ import { AutoComplete, Col, Form, Row, RowProps } from 'antd'
 import { OptionProps } from 'antd/lib/mentions'
 import { get } from 'lodash'
 import debounce from 'lodash/debounce'
+import isString from 'lodash/isString'
```
SavelevMatthew, line 6: cmon, dude :)
Sometimes I think we should add `isEven` and `isOdd` npm packages after lines like this 😄

apps/condo/domains/property/components/BasePropertyForm/usePropertyValidations.tsx
```diff
@@ -1,8 +1,10 @@
 import { get } from 'lodash'
```
SavelevMatthew, line 1: can u fix this import please? since you have touched old file :)

apps/condo/domains/contact/components/ContactsEditor/NotResidentFields.tsx
Alllex202: It's easier to understand `isString(someValue)` then `typeof someValue === 'string'` :)

apps/condo/domains/contact/components/ContactsEditor/NewContactFields.tsx
Alllex202: This is necessary so that the elements are aligned

Alllex202: Look at the video, everything is fine there)
And this is what it looks like now:
Снимок экрана 2024-04-19 в 15 50 34

apps/condo/domains/contact/components/ContactsEditor/NotResidentFields.tsx
SavelevMatthew: For me second one is the way to go, since I know what exactly is happening. Lodash functions can be tricky sometimes despite of simple names and do not what you exactly expect from its name.

Like this one:
image

Did you expect isObject(new String('')) to be true? Many of our developers using it to check `Record<string, unknown>` with it, but according to the code none of them knows that it checks js-objects and not like datatype-objects

Alllex202: > Did you expect isObject(new String('')) to be true?

Yes, because I read the documentation😄

nomerdvadcatpyat (approved):

PR #4656, opened 2024-04-23
refactor(condo): DOMA-7753 added types of keys for translations

Alllex202:
- [x] Add types of keys for translations (Added in `condo`, `callcenter`, `dev-portal`, `document-next` and `pass` apps)
- [x] Migrate `@open-condo/next` package from `js` to `ts` (Without this, key types will not work)
- [x] Fix after the above changes
- [x] Run tests `*.spec.ts` in ci
- [x] Fix tests
- [x] Update submodules to master

### Preview: 
_Checking that translation keys for application exist. If key is missing then an error will be thrown.
Autocompletion works correctly in `VS Code`, but not good in `WebStorm`_

https://github.com/open-condo-software/condo/assets/56914444/26ecf5cf-36d1-47f5-9597-b6e54342a859


packages/next/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@open-condo/next",
   "version": "1.0.0",
-  "main": "./index.js",
```
SavelevMatthew, line 4: exports?

packages/next/tsconfig.json
```diff
@@ -0,0 +1,29 @@
+{
```
SavelevMatthew, line 1: should be inherited from `@open-condo/ts-configs`?

packages/next/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@open-condo/next",
```
SavelevMatthew, line 2: Should it be built like icons / ui / bridge? Or it's more like folder with ts files for now?) If so, should it specified in transpileModules of apps?

packages/next/package.json
Alllex202: `index.js` was empty...

packages/next/package.json
Alllex202: It was just a folder with files, everything remains as before :)

SavelevMatthew: Technically it was valid JS package, since it met all the requirements and does not require the transpilation:
1. package-name
2. deps
3. main for file imports

Now it's not a package anymore, since ts packages are requiring transpiling to generate cjs / esm, which other apps can consume. also there's no  main / export fields anymore, which make it more similar to folder.

Alllex202: > Technically it was valid JS package

returned it to a state that satisfies these points

> Should it be built like icons / ui / bridge?

I don't think this needs to be built. We do not plan to publish this package. It will be the like `@open-condo/codegen`

SavelevMatthew: > I don't think this needs to be built. We do not plan to publish this package...

I don't think so. There's a task for generating next and ks apps from CLI. I want next app to support Next 13+, intl and Apollo-client (with cache / SSR / proper codegen) out of the box. So, the intl is there. But I'm not really sure we should maintain `@open-condo/next` package. Apollo can be installed with ~100 lines of code, and intl with ~20. And we don't need all this `WithAnotherWrapper` anymore...

So, what do you think?

apps/miniapp/global.d.ts
```diff
@@ -0,0 +1,24 @@
...
+
+// NOTE: Combine all keys together
+const translations = [
+    en, ru,
+]
+const translationsCondo = [
+    enCondo, ruCondo,
+]
+
+export type MessagesKeysType = keyof typeof translations[number] | keyof typeof translationsCondo[number]
```
SavelevMatthew, line 15: 👀 was it always like that? I thought each miniapp should have its own set of translations

toplenboren (approved):

toplenboren:
Hey, any idea on why WS code completion is lame?

apps/miniapp/global.d.ts
SavelevMatthew: Mb TODO here?

packages/next/package.json
SavelevMatthew: Maybe todo to rename it to legacy-next or smth?

SavelevMatthew: Docs about transpiling?

Alllex202: > Maybe todo to rename it to legacy-next or smth?

Added todo

> Docs about transpiling?

Added docs

Alllex202:
@AleX83Xpert Please check `SbbolMetersImporter` utils. Test `SbbolMetersImporter › Rows transformation must work` failed

AleX83Xpert:
> @AleX83Xpert Please check `SbbolMetersImporter` utils. Test `SbbolMetersImporter › Rows transformation must work` failed

@Alllex202, done

apps/condo/jest.config.js
```diff
@@ -45,6 +45,9 @@ module.exports = {
             setupFilesAfterEnv: [`${__dirname}/jest.setupSpec.js`],
             // NOTE: need to pass uuid export syntax through babel
             transformIgnorePatterns: ['/node_modules/(?!(uuid|nanoid|msgpackr)/)'],
+            moduleNameMapper: {
+                '@open-condo/(.*)': '<rootDir>/../../packages/$1',
```
SavelevMatthew, line 49: Why do we need it? Yarn is not enough?

SavelevMatthew (approved):

apps/condo/jest.config.js
Alllex202: it helps `jest` find imports from `@open-condo`. 
Without this, some tests on `ts` could fail with such an error when importing `@open-condo`
Снимок экрана 2024-05-28 в 13 18 24

PR #4664, opened 2024-04-25
refactor(condo): DOMA-8830 update readOnly access in Ticket schema fields

nomerdvadcatpyat:
Made correct access to fields, which filled only on server-side


Alllex202 (approved):
How about new tests? :)

sitozzz (approved):

toplenboren (approved):

PR #4667, opened 2024-04-25
feat(ui): DOMA-8920 added IconButton

Alllex202:
## Preview
https://github.com/open-condo-software/condo/assets/56914444/f41a6fdd-0d6c-46a7-b774-77edb31304a6


packages/ui/src/components/IconButton/iconButton.tsx
```diff
@@ -0,0 +1,69 @@
...
+            [`${ICON_BUTTON_CLASS_PREFIX}-pressed`]: pressed,
+            [`${ICON_BUTTON_CLASS_PREFIX}-focus`]: focus,
+            [`${ICON_BUTTON_CLASS_PREFIX}-${mergedSize}`]: mergedSize,
+        },
+        className,
+    )
+
+    const wrappedContent = children
+        ? <span className={`${ICON_BUTTON_CLASS_PREFIX}-content`}>{children}</span>
+        : null
```
nomerdvadcatpyat, line 38: I think it's better to pass icon from prop (like `icon={PlusCircle}`, `size` of icon = `size` of IconButton) and disable children prop. Now you can pass any node in children

Alllex202: You can also pass any node to `icon` prop.

Your icon can be custom: svg, font icon, icon from other packages, because `Icon` size and `Button` size it's two different params and not depends each other

Alllex202:
> Mb do it like in Input: Button – common button and Button.IconButton (or Button.Icon) – this component?

yes, probably worth changing

nomerdvadcatpyat (approved):

PR #4685, opened 2024-05-02
fix(global): INFRA-313 update node-fetch dep

sitozzz:


AleX83Xpert (approved):

toplenboren:
Do we need node-fetch?

toplenboren:
We should only use cross-fetch thorough our custom fetch

toplenboren (approved):
lgtm tho

PR #4694, opened 2024-05-06
feat(condo): DOMA-8508 new submodule (the meters mini-app)

AleX83Xpert:


.gitmodules
```diff
@@ -50,3 +50,6 @@
 [submodule "apps/resident-app"]
 	path = apps/resident-app
 	url = git@github.com:open-condo-software/condo-resident-app.git
+[submodule "apps/meters"]
+	path = apps/meters
```
SavelevMatthew, line 54: I guess the name is too generous...That's probably not the last app dedicated to meters domain. How about some clarification here?

AleX83Xpert: Any suggestions?
Maybe `meters-sync`?

SavelevMatthew (approved):

PR #4703, opened 2024-05-06
fix(banking): BankAccount access should check is resident deleted or not

pahaz:


apps/condo/domains/banking/access/BankAccount.js
```diff
@@ -48,7 +48,7 @@ async function canReadBankAccounts ({ authentication: { item: user } }) {
             ],
         }
     } else if (user.type === RESIDENT) {
-        const residents = await find('Resident', { user: { id: user.id } })
+        const residents = await find('Resident', { user: { id: user.id }, deletedAt: null })
```
AleX83Xpert, line 51: Any test?

PR #4713, opened 2024-05-08
fix(condo): DOMA-8864 make settings tab available for RSO

abshnko:
make settings tab available for RSO so that they have MobileConfig settings only for MeterReadings

image


SavelevMatthew (approved):

nomerdvadcatpyat (approved):

SavelevMatthew:
Use SP (Service provider) instead of RSO, just to clarify architectural terminology :)

PR #4714, opened 2024-05-08
fix(condo): DOMA-8589 remove expensive access checks for BillingReceipt

abshnko:
remove because the request is too heavy


dkoviazin (approved):
You can remove OR = as you have only one condition remain

PR #4722, opened 2024-05-14
feat(address-service): added the ability to add an address via Google provider in the language of the platform

paulo-rossy:
The Google provider's API endpoint doesn't function correctly for adding addresses, especially when adding a house. Although the address can be searched in English, the added house appears in a different language. 

To address this issue, we incorporated logic from GoogleSuggestionProvider.js (https://github.com/open-condo-software/condo/blob/master/apps/address-service/domains/common/utils/services/suggest/providers/GoogleSuggestionProvider.js#L11-L14). 

By including the language parameter, the address now appears in the language of the platform.


pahaz (approved):

AleX83Xpert (approved):
Thanks for your contribution.

SavelevMatthew (commented):
Hi. Thanks again for your contribution to our project! Overall, it looks good, but there's one detail I'm currently taking into account with the team: should the locale be extracted from the request (as it is for translation) or be fixed for a specific provider (as you imlemented)? This shouldn't take too long and I'll be back with an answer soon.

SavelevMatthew:
> Hi. Thanks again for your contribution to our project! Overall, it looks good, but there's one detail I'm currently taking into account with the team: should the locale be extracted from the request (as it is for translation) or be fixed for a specific provider (as you imlemented)? This shouldn't take too long and I'll be back with an answer soon.

We decided that our approach with locale should be reworked a bit later, as most likely locale should sprout from the geo-location of the organization, so the current solution looks ok as we don't have that logic at the moment.
Thanks again for your contribution 🥳 ❤️

SavelevMatthew (approved):

PR #4727, opened 2024-05-14
fix(keystone): INFRA-296 move gql related logging to pino auto-loggin…

sitozzz:
…g context

### DIFF
--- a/packages/keystone/logging/getKeystonePinoOptions.js
+++ b/packages/keystone/logging/getKeystonePinoOptions.js
@@ -1,6 +1,7 @@
+const serializers = require('pino-std-serializers')
+
 const { getLogger } = require('./getLogger')
 const { getReqLoggerContext } = require('./getReqLoggerContext')
-
 const logger = getLogger('http')
 const IGNORE_PATH = new Set(['/api/features', '/api/version', '/favicon.ico'])
 
@@ -13,6 +14,12 @@ function getKeystonePinoOptions () {
         },
         customProps: (req, res) => {
             return getReqLoggerContext(req)
+        },
+        serializers: {
+            res: (res) => {
+                const errors = res.raw.req.errors
+                return { gqlErrors: errors, ...serializers.req(res) }
+            },
         },
     }
 }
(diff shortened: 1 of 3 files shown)


packages/keystone/logging/getKeystonePinoOptions.js
```diff
@@ -1,6 +1,7 @@
+const serializers = require('pino-std-serializers')
```
SavelevMatthew, line 1: This should it in package.json of keystone package

packages/keystone/logging/getKeystonePinoOptions.js
```diff
@@ -14,6 +15,12 @@ function getKeystonePinoOptions () {
         customProps: (req, res) => {
             return getReqLoggerContext(req)
         },
+        serializers: {
+            res: (res) => {
+                const errors = res.raw.req.errors
```
SavelevMatthew, line 20: Should it be wrapped with get or optional chain in case of no errors or smth like this?

packages/keystone/logging/getKeystonePinoOptions.js
```diff
@@ -14,6 +15,12 @@ function getKeystonePinoOptions () {
...
+                return { gqlErrors: errors, ...serializers.req(res) }
```
SavelevMatthew, line 21: Hm, is it correct behaviour? (You're overriding` serializers.res`, but using `.req` here)

packages/keystone/logging/getReqLoggerContext.js
```diff
@@ -35,7 +49,7 @@ function getReqLoggerContext (req) {
         }
     }
 
-    return { reqId, sessionId, user, ip, fingerprint, complexity }
+    return omitBy({ reqId, sessionId, user, ip, fingerprint, complexity, operationName, variables, query, operationId }, isNil)
```
SavelevMatthew, line 52: Is omitting previously existing properties on object safe?

packages/keystone/logging/getKeystonePinoOptions.js
sitozzz: Good point, do it right here

packages/keystone/logging/getKeystonePinoOptions.js
sitozzz: Yup, sounds reasonable

PR #4734, opened 2024-05-15
Feat/condo/doma-7317/update qr code logic

AleX83Xpert:
New logic for mutations, which is used to scan receipts QR codes.
to-do:
- [ ] try to add captcha (maybe in separated task)

### DIFF
--- a/apps/condo/domains/acquiring/schema/CreatePaymentByLinkService.js
+++ b/apps/condo/domains/acquiring/schema/CreatePaymentByLinkService.js
@@ -59,7 +59,7 @@ const CreatePaymentByLinkService = new GQLCustomSchema('CreatePaymentByLinkServi
         },
         {
             access: true,
-            type: 'type CreatePaymentByLinkOutput { multiPaymentId: ID!, amount: String!, explicitFee: String!, totalAmount: String!, address: String!, addressMeta: AddressMetaField!, unitType: String!, unitName: String!, accountNumber: String!, period: String! }',
+            type: 'type CreatePaymentByLinkOutput { multiPaymentId: ID!, amount: String!, explicitFee: String!, totalAmount: String!, integrationHostUrl: String!, address: String!, addressMeta: AddressMetaField!, unitType: String!, unitName: String!, accountNumber: String!, period: String! }',
         },
     ],
 
@@ -225,6 +225,7 @@ const CreatePaymentByLinkService = new GQLCustomSchema('CreatePaymentByLinkServi
                     amount: multiPayment.amountWithoutExplicitFee,
                     explicitFee: multiPayment.explicitServiceCharge,
                     totalAmount: multiPayment.amount,
+                    integrationHostUrl: get(acquiringContexts, [0, 'integration', 'hostUrl']),
                     address: normalizedAddress.address,
                     addressMeta: {
                         dv: 1,
--- a/apps/condo/domains/acquiring/gql.js
+++ b/apps/condo/domains/acquiring/gql.js
@@ -72,7 +72,7 @@ const RecurrentPayment = generateGqlQueries('RecurrentPayment', RECURRENT_PAYMEN
 
 const PAYMENT_BY_LINK_MUTATION = gql`
     mutation createPaymentByLink ($data: CreatePaymentByLinkInput!) {
-        result: createPaymentByLink(data: $data) { multiPaymentId amount explicitFee totalAmount address addressMeta { ${ADDRESS_META_SUBFIELDS_QUERY_LIST} } unitType unitName accountNumber period }
+        result: createPaymentByLink(data: $data) { multiPaymentId amount explicitFee totalAmount integrationHostUrl address addressMeta { ${ADDRESS_META_SUBFIELDS_QUERY_LIST} } unitType unitName accountNumber period }
     }
 `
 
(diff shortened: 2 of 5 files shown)


apps/condo/domains/acquiring/schema/CreatePaymentByLinkService.js
```diff
@@ -161,97 +132,92 @@ const CreatePaymentByLinkService = new GQLCustomSchema('CreatePaymentByLinkServi
...
-                    // find acquiring context and routing number from older receipt
-                    const billingIntegrationContext = await getById('BillingIntegrationOrganizationContext', lastBillingReceipt.context.id)
-                    /** @type {AcquiringIntegrationContext[]} */
-                    const acquiringContexts = await AcquiringIntegrationContext.getAll(context, {
-                        organization: { id: billingIntegrationContext.organization, deletedAt: null },
-                        status: ACQUIRING_CONTEXT_FINISHED_STATUS,
-                        deletedAt: null,
-                    })
+                /** @type {TCompareQRResolvers} */
+                const resolvers = {
```
pahaz, line 158: Is it really should be here? maybe extract it to utility?

apps/condo/domains/acquiring/schema/CreatePaymentByLinkService.js
```diff
@@ -161,97 +132,92 @@ const CreatePaymentByLinkService = new GQLCustomSchema('CreatePaymentByLinkServi
...
+                    onNoReceipt: async () => {
```
pahaz, line 159: you can extract it as local function

AleX83Xpert: I've extracted the `payForLastBillingReceipt` function to avoid code duplication. But `onNoReceipt` is used only once. All these functions should be defined within mutation cause they use many local variables, mutate a local variable, and call from utilities.

apps/condo/domains/acquiring/schema/CreatePaymentByLinkService.js
AleX83Xpert: I think yes. Because resolvers are defined for mutation purposes, but conditions for these resolvers are defined within utilities.

apps/condo/domains/billing/schema/ValidateQRCodeService.js
```diff
@@ -65,43 +87,50 @@ const ValidateQRCodeService = new GQLCustomSchema('ValidateQRCodeService', {
...
-            type: 'type ValidateQRCodeOutput { qrCodeFields: JSON! }',
+            type: 'type ValidateQRCodeLastReceiptDataOutput { id: ID!, period: String!, toPay: String! }',
+        },
+        {
+            access: true,
+            type: 'type ValidateQRCodeFeesOutput { explicitServiceCharge: String, explicitFee: String }',
+        },
+        {
+            access: true,
+            type: 'type ValidateQRCodeOutput { qrCodeFields: JSON!, lastReceiptData: ValidateQRCodeLastReceiptDataOutput, explicitFees: ValidateQRCodeFeesOutput!, amount: String! }',
```
sitozzz, line 98: Is there any option to serialize qrCodeFields?

AleX83Xpert: Maybe, but this field is decoded fron input string. Input string content may be different for different qr-codes.

apps/condo/domains/billing/schema/ValidateQRCodeService.js
```diff
@@ -111,15 +140,72 @@ const ValidateQRCodeService = new GQLCustomSchema('ValidateQRCodeService', {
...
+                    explicitServiceCharge: String(explicitFee),
+                    explicitFee: '0',
+                } : {
+                    explicitServiceCharge: '0',
+                    explicitFee: String(explicitFee),
+                }
+
+                return {
+                    qrCodeFields,
+                    lastReceiptData: foundReceipt ? pick(foundReceipt, ['id', 'period', 'toPay']) : null,
```
sitozzz, line 205: why do you need to filter result object by these properties?

AleX83Xpert:
I'd like to return only these fields

sitozzz (approved):

apps/condo/domains/billing/schema/ValidateQRCodeService.js
AleX83Xpert: I'd like to return only these fields

apps/condo/domains/billing/schema/ValidateQRCodeService.js
```diff
@@ -65,43 +87,50 @@ const ValidateQRCodeService = new GQLCustomSchema('ValidateQRCodeService', {
...
-                // TODO(AleX83Xpert): maybe decode requisitesStr
-                // Need to test the result of scanning from mobile devices
-                // https://encoding.spec.whatwg.org/#koi8-r
-                // https://encoding.spec.whatwg.org/#windows-1251
-                // const encodingTag = get(matches, ['groups', 'encodingTag'])
-                // const encoding = get(['windows-1251', 'utf-8', 'koi8-r'], encodingTag, 'utf-8')
+                const currentUserId = get(context, ['authedItem', 'id'])
+                if (!currentUserId) {
+                    const ip = context.req.ip
+                    await redisGuard.checkCustomLimitCounters(
```
Alllex202, line 113: Why are you skipping this check for an authorized user?
They can spam too :)

AleX83Xpert: If so, we can find these users and maybe spread limit for registered users too.

PR #4752, opened 2024-05-20
feat(condo): INFRA-317 organizations caching proof of concept

SavelevMatthew:
## Idea
The idea of this mechanism is to remove unnecessary many-to-many joins to save computational time as well as Postgres IOPS. To achieve this I've decided to store information about user organizations in Redis cache with short enough, but long enough TTL. The goal of data structure is to be as minimal as possible with read-checks as priority...

## Cache record structure:

```json
{
  "dv": 1, // data version in case we decided to change format
  "organizations": {
    "7ca218ce-a594-461a-bd4e-efaddeda47b9": {
      "permissions": {
        "canRead...": true,
        "canManage...": false,
        "....": true,
      },
      "childOrganizations": ["3008f093-60da-41ff-965a-41c25025b7e9", "4cfd9835-be88-4650-861a-ba049e8e2d3e"] 
    }
  }
}
```

 1. Filter directly employed organizations to read is fast (just traverse organizations, check permissions and return keys)
 2. Filter related (child organizations) to read is fast (return `childOrganizations` when traversing)
 3. Filter directly employed and related is fast too (return both)
 4. Check access for direct organization is fast (lookup by key and check permissions)
 5. Check access for related organizations is not so fast (get ids from read, convert to set, and apply has N times)
 6. Check access for both is not so fast (because of 5th).

But this structure allow fast reads and relatively fast writes...We can add `child_id` -> `parent_id` data to speed up 5th point, but this will increase cache size (and network is slow), so I'm not sure about that

## TODO

- [x] Remove `checkUserPermissionsInOrganizations`
- [x] Remove `checkPermissionInUserOrganizationOrRelatedOrganization`
- [x] Remove `checkPermissionsInUserOrganizationOrRelatedOrganization`
- [x] Remove `checkOrganizationPermission`
- [x] Remove `checkUserBelongsToOrganization`
- [x] Remove `checkUserBelongsToRelatedOrganization`
- [x] Remove `checkRelatedOrganizationPermission`
- [x] Remove `queryOrganizationEmployeeFromRelatedOrganizationFor`
- [x] Remove `queryOrganizationEmployeeFor`
- [x] Remove `checkPermissionsInUserOrganizationsOrRelatedOrganizations`

## Important files to check

- domains/organization/schema/* (cache invalidation in hooks)
- domains/organization/utils/* (cache logic)

## Something to optimise further:

- [x] Go to redis once for all subqueries

## Some measurement
- Ticket K6 Test

| Variant             | p90 (ms) | p95 (ms) | p99 (ms) |
|---------------------|----------|----------|----------|
| No cache (run 1)    | 263      | 276      | 397      |
| No cache (run 2)    | 245      | 296      | 1190     |
| No cache (run 3)    | 260      | 277      | 453      |
|                     |          |          |          |
| Cache (run 1)       | 267      | 283      | 575      |
| Cache (run 2)       | 265      | 294      | 448      |
| Cache (run 3)       | 252      | 280      | 373      |
|                     |          |          |          |
| Cache + ctx (run 1) | 238      | 250      | 844      |
| Cache + ctx (run 2) | 223      | 234      | 710      |
| Cache + ctx (run 3) | 209      | 230      | 390      |


apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -1,191 +1,297 @@
...
 
-const { getByCondition, find } = require('@open-condo/keystone/schema')
+const conf = require('@open-condo/config')
+const { getRedisClient } = require('@open-condo/keystone/redis')
+const { find } = require('@open-condo/keystone/schema')
 
+const _redisClient = getRedisClient('default', 'cache')
+// NOTE: larger = better, but it can affect "after migration" state, where roles are changed via SQL
+const CACHE_TTL_IN_MS = 60 * 60 * 1000  // 1 hour in ms
+const DISABLE_USER_ORGANIZATION_CACHING = get(conf, 'DISABLE_USER_ORGANIZATION_CACHING', 'false').toLowerCase() === 'true'
```
sitozzz, line 12: IMO configurable ttl could be kinda useful feature
Maybe join these two lines into single config?

apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -1,191 +1,297 @@
...
+ * @property {Record<string, UserOgranizationInfo>} organizations - record where keys are organization ids and values are info it
+ */
+
+/**
+ * Obtains user organizations info via caching or sub-querying
+ * @param {{ id: string }} user - user object
+ * @returns {Promise<UserOgranizationsCache>}
+ * @private
+ */
+async function _getUserOrganizations (user) {
```
sitozzz, line 89: I don't think that it's necessary to mark function as 'private' here. It's used only in concrete module and not exported

apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -1,191 +1,297 @@
...
-    if (!uniqOrganizationIds.every(isUUID)) return false
+    const permissionsToCheck = Array.isArray(permissions) ? permissions : [permissions]
+    const organizationIds = []
 
-    for (const id of organizationIds) {
-        const hasAccess = await checkRelatedOrganizationPermission(userId, id, permission)
-        if (!hasAccess) return false
+    for (const info of Object.values(userOrganizationsInfo.organizations)) {
+        if (_checkPermissionsInOrganization(info.permissions, permissionsToCheck)) {
+            organizationIds.push(...info.childOrganizations)
```
sitozzz, line 189: `Array.concat` will show better performance here

apps/condo/domains/analytics/access/TicketReportService.js
```diff
@@ -1,15 +0,0 @@
-const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
-
-const { queryOrganizationEmployeeFor } = require('@condo/domains/organization/utils/accessSchema')
-
-async function canReadTicketReportWidgetData ({ authentication: { item: user } }) {
```
pahaz, line 5: Why it’s deleted?

apps/condo/domains/banking/utils/accessSchema.js
```diff
@@ -20,27 +22,21 @@ async function canManageBankEntityWithOrganization ({ authentication: { item: us
...
-                deletedAt: null,
-            })
-            if (items.length !== itemIds.length) return false
-            organizationIds = uniq(items.map(item => item.organization))
-        } else {
-            if (!itemId) return false
-            const item = await getById(listKey, itemId)
-            organizationIds = [item.organization]
-        }
+        const ids = itemIds || [itemId]
```
pahaz, line 29: May be we need some helper for that 5 loc ? (Optional)

apps/condo/domains/contact/access/Contact.js
```diff
@@ -29,12 +30,11 @@ async function canReadContacts (args) {
...
 
+    const permittedOrganizations = await getEmployedOrRelatedOrganizationsByPermissions(user, 'canReadContacts')
+
     return {
         organization: {
-            OR: [
-                queryOrganizationEmployeeFor(user.id, 'canReadContacts'),
-                queryOrganizationEmployeeFromRelatedOrganizationFor(user.id, 'canReadContacts'),
-            ],
+            id_in: permittedOrganizations,
```
pahaz, line 37: Could you please check that if the ‘ permittedOrganizations’ is empty we will not have any access.

apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -1,239 +1,308 @@
...
+    }
 
-    if (!employeeRole) return false
-    return permissions.every((permission) => !!employeeRole[permission]) || false
-}
+    const employees = await find('OrganizationEmployee', employeesFilter)
 
-async function checkRelatedOrganizationPermission (userId, organizationId, permission) {
-    return checkRelatedOrganizationPermissions(userId, organizationId, [permission])
+    if (employees.length) {
```
pahaz, line 53: It looks like we should use delete always.

apps/condo/domains/contact/access/Contact.js
SavelevMatthew: 1. It's checked by many tests where employee with some permission can and without cannot
2. Empty `id_in` works as you expect it to work

image
image

apps/condo/domains/organization/utils/accessSchema.js
pahaz: Someone can add it to export. I’d you really don’t want it it’s ok to mark it as private

apps/condo/domains/organization/utils/accessSchema.js
SavelevMatthew: No, `ioredis` will throw an error on empty list delete attempt

apps/condo/domains/analytics/access/TicketReportService.js
SavelevMatthew: Not used anywhere in code. (Found duplicate in ticket domain, which is actually used, so removed legacy code)

apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -1,239 +1,308 @@
...
-    if (!userId || !isArray(organizationIds) || isEmpty(organizationIds)) return false
+async function _getUserOrganizations (user) {
+    const cacheKey = _getUserOrganizationsCacheKey(user)
+
+    if (!DISABLE_USER_ORGANIZATION_CACHING) {
+        const cachedDataString = await _redisClient.get(cacheKey)
 
-    const uniqOrganizationIds = uniq(organizationIds)
+        if (cachedDataString !== null) {
+            return JSON.parse(cachedDataString)
```
pahaz, line 96: It’s also good to check the structure

apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -1,239 +1,308 @@
...
+
+    const newCacheEntry = { dv: 1, organizations: {} }
+
+    const userEmployees = await find('OrganizationEmployee', {
+        deletedAt: null,
+        organization: { deletedAt: null },
+        user: { id: user.id },
+        isAccepted: true,
+        isBlocked: false,
+        isRejected: false,
```
pahaz, line 108: Do we really need to check that it’s Rejected is false? As I remember we have such legacy  cases. But I’m not sure

apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -1,239 +1,308 @@
...
+    })
 
-    if (!uniqOrganizationIds.every(isUUID)) return false
+    const userRoleIds = userEmployees.map(employee => employee.role)
+    const userRoles = await find('OrganizationEmployeeRole', {
+        id_in: userRoleIds,
+        // TODO(SavelevMatthew): INFRA-317 Add this check
```
pahaz, line 114: @Alllex202  check it

apps/condo/domains/organization/utils/accessSchema.js
SavelevMatthew: Idk, all tests are passed multiple times...So I guess there's no more cases like that (otherwise someone would fix this behaviuor by test)

apps/condo/domains/organization/utils/accessSchema.js
SavelevMatthew: R u sure? concat usually create new array of size arr1 + arr2, and copy the whole sequence...Why push is expanding current array

SavelevMatthew: https://dev.to/uilicious/javascript-array-push-is-945x-faster-than-array-concat-1oki

apps/condo/domains/organization/utils/accessSchema.js
SavelevMatthew: I don't think this is necessary at this point. Shape matching usually is not the fastest part of processing. And it's not user-input, so additional validation is redundant. The only way you can mess it up is by redefining it manually in redis, but if you do so - you're probably doing something wrong

apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -1,239 +1,308 @@
...
+    if (!DISABLE_USER_ORGANIZATION_CACHING) {
+        const cachedDataString = await _redisClient.get(cacheKey)
 
-    const uniqOrganizationIds = uniq(organizationIds)
+        if (cachedDataString !== null) {
+            return JSON.parse(cachedDataString)
+        }
+    }
+
+    const newCacheEntry = { dv: 1, organizations: {} }
```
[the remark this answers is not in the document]
SavelevMatthew: I'm not sure about shape/structure (read comment above please), but I can check dv I guess...

The only reason I skipped it for now is I've got one and the only version of data structure

apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -1,239 +1,308 @@
...
+    const userOrganizationsInfo = await _getUserOrganizations(user)
 
-    if (!uniqOrganizationIds.every(isUUID)) return false
+    const permissionsToCheck = Array.isArray(permissions) ? permissions : [permissions]
+    const organizationIds = []
 
-    for (const id of organizationIds) {
-        const hasAccess = await checkRelatedOrganizationPermission(userId, id, permission)
-        if (!hasAccess) return false
+    for (const [id, info] of Object.entries(userOrganizationsInfo.organizations)) {
```
Alllex202, line 165: Probably at beginning of function it is worth checking that `user `was transferred, and if they are empty, then return false.

This logic was in previous utilities, and we won’t have to request data in vain

apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -1,239 +1,308 @@
...
+ * so use it when you don't need related organizations
+ * @param {{ id: string }} user - user object
+ * @param {Array<string> | string} organizationIds - organizations to checks (can be passed as array of IDs or a single ID)
+ * @param {Array<string> | string} permissions - permissions to check (can be passed as array or a single string)
+ * @returns {Promise<boolean>}
+ */
+async function checkPermissionsInEmployedOrganizations (user, organizationIds, permissions) {
+    const userOrganizationsInfo = await _getUserOrganizations(user)
+
+    const organizationsToCheck = Array.isArray(organizationIds) ? uniq(organizationIds) : [organizationIds]
```
Alllex202, line 229: ^^ same for `user` and `organizations`

and for other utils

apps/condo/domains/acquiring/access/PaymentsFilterTemplate.js
```diff
@@ -5,16 +5,18 @@
...
     if (!user) return throwAuthenticationError()
     if (user.deletedAt) return false
 
     if (user.isAdmin) return {}
 
+    const permittedOrganizations = await getEmployedOrganizationsByPermissions(user, [])
+
     return {
-        employee: { organization: { ...queryOrganizationEmployeeFor(user.id) } },
+        employee: { organization: { id_in: permittedOrganizations }, user: { id: user.id }, deletedAt: null },
```
Alllex202, line 19: Isn't it enough for us to just use `{ organization: { id_in: permittedOrganizations } }`?

In `getEmployedOrganizationsByPermissions` we have already checked that the user is an employee of the organization. Is not it?

apps/condo/domains/organization/utils/accessSchema.js
sitozzz: Yeah, my bad. My test had an issue, so I get a wrong result

apps/condo/domains/banking/access/BankTransaction.js
```diff
@@ -28,11 +24,12 @@ async function canReadBankTransactions ({ authentication: { item: user }, contex
         }
     }
 
+    const permittedOrganizations = await getEmployedOrRelatedOrganizationsByPermissions(user, [])
```
Alllex202, line 27: What about function `getEmployedOrRelatedOrganizations`?

It seems that calling of `getEmployedOrRelatedOrganizationsByPermissions` without permissions is common

You have added similar functions `checkUserEmploymentInOrganizations`/`checkUserEmploymentOrRelationToOrganization`, why not add `getEmployeed...`/`getRelatedOrganizations` as well?

apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -1,239 +1,308 @@
...
+    const permissionsToCheck = Array.isArray(permissions) ? permissions : [permissions]
+
+    return organizationsToCheck.every(orgId =>
+        userOrganizationsInfo.organizations.hasOwnProperty(orgId) &&
+        _checkPermissionsInOrganization(
+            userOrganizationsInfo.organizations[orgId].permissions,
+            permissionsToCheck
+        )
+    )
 }
```
Alllex202, line 239: Сheck that `organizationsToCheck`/`organizationIds` is not empty.
It is important. If there is an empty array, then access will be allowed (`[].every(cb)` return true); this should not happen. He should be banned

apps/condo/domains/meter/access/MeterReadingFilterTemplate.js
```diff
@@ -5,15 +5,17 @@
...
 async function canReadMeterReadingFilterTemplates ({ authentication: { item: user } }) {
     if (!user) return throwAuthenticationError()
     if (user.deletedAt) return false
     if (user.isAdmin) return {}
 
+    const permittedOrganizations = await getEmployedOrganizationsByPermissions(user, 'canReadMeters')
+
     return {
-        employee: { organization: { ...queryOrganizationEmployeeFor(user.id, 'canReadMeters') } },
+        employee: { organization: { id_in: permittedOrganizations }, user: { id: user.id }, deletedAt: null },
```
Alllex202, line 18: ^^ same

apps/condo/domains/organization/schema/OrganizationEmployee.js
```diff
@@ -236,6 +237,14 @@ const OrganizationEmployee = new GQLListSchema('OrganizationEmployee', {
         afterChange: async ({ context, operation, existingItem, updatedItem }) => {
             const isSoftDeleteOperation = operation === 'update' && !existingItem.deletedAt && Boolean(updatedItem.deletedAt)
 
+            // NOTE: Creating / updating employee must reset users cache
+            const updatedUserId = updatedItem.user
+            const existingUserId = get(existingItem, 'user')
+            await resetUserEmployeesCache(updatedUserId)
+            if (existingUserId && existingUserId !== updatedUserId) {
+                await resetUserEmployeesCache(existingUserId)
+            }
```
Alllex202, line 246: Do we allow to change `OrganizatonEmployee.user`? O_o

apps/condo/domains/ticket/access/TicketFilterTemplate.js
```diff
@@ -5,15 +5,17 @@
...
 async function canReadTicketFilterTemplates ({ authentication: { item: user } }) {
     if (!user) return throwAuthenticationError()
     if (user.deletedAt) return false
     if (user.isAdmin) return {}
 
+    const permittedOrganizations = await getEmployedOrganizationsByPermissions(user, 'canReadTickets')
+
     return {
-        employee: { organization: { ...queryOrganizationEmployeeFor(user.id, 'canReadTickets') } },
+        employee: { organization: { id_in: permittedOrganizations }, user: { id: user.id }, deletedAt: null },
```
Alllex202, line 18: ^^

apps/condo/domains/acquiring/access/PaymentsFilterTemplate.js
SavelevMatthew: `employee: { organization: { id_in: permittedOrganizations } }` - give me **all** employees from my organizations
`employee: { organization: { id_in: permittedOrganizations }, user: { id: user.id } }` - give me **all my** employees from my organizations

apps/condo/domains/meter/access/MeterReadingFilterTemplate.js
SavelevMatthew: Check comment above

apps/condo/domains/organization/utils/accessSchema.js
SavelevMatthew: Why? It was unclear behaviour before :)

Why do you assume that empty array is a bad thing? 
For example I can do request like this:
```graphql
mutation {
    createSomething (data: {
        relation: { create: { ...relationPayload } }
    }) {
        id
    }
}
```

So your typical `get(originalInput, 'relation.connect.id')` will be null, and `[]` is valid to resolve in true, since I do not connect any ogranization

apps/condo/domains/organization/utils/accessSchema.js
SavelevMatthew: I'm not sure, that I should do it. The reason behind empty organizations I've described below and what about user.

The goal of this util is to check if user have access to specified organizations, JSDoc added the interface that user is required parameter. So you must provide a user :)

(In some other typed language like Go (or even Typescript) you'll be punished for not passing a user, so if you create signature and passing something else - that's your fault)

Maybe I should add user-guard, but I'm not sure that I should do it. My point is that by moving more guards and logics to utils you're skipping some corner cases in access functions themself. 
(Because everyone will assume that your util is Swiss army knife :) )

apps/condo/domains/organization/utils/accessSchema.js
SavelevMatthew: Added separate variable for it, JSON configs is a bit messy imho...

sitozzz (approved):

apps/condo/domains/acquiring/access/PaymentsFilterTemplate.js
Alllex202: ok my mistake

apps/condo/domains/organization/utils/accessSchema.js
Alllex202: If you do not use an organization when checking access, then most likely you do not need these utilities)

Alllex202: You want the developer to check that he is not passing in an empty organization(s). But the developer will expect that the utility will check everything itself))

apps/condo/domains/organization/utils/accessSchema.js
Alllex202: > (In some other typed language like Go (or even Typescript) you'll be punished for not passing a user, so if you create signature and passing something else - that's your fault)

JS does not have such checks, so developers should do this) 
Just in case, I would add a check so as not to worry about it)

apps/condo/domains/news/access/NewsItemScope.js
```diff
@@ -55,51 +52,42 @@ async function canManageNewsItemScopes ({ authentication: { item: user }, origin
...
+                },
+                deletedAt: null,
+            })
+
+            if (newsItems.some(item => !item.organization)) return false
+            organizationIds = uniq(newsItems.map(newsItem => newsItem.organization))
         }
 
-        return await checkPermissionsInUserOrganizationsOrRelatedOrganizations(user.id, organizationIds, 'canManageNewsItems')
+        return await checkPermissionsInEmployedOrRelatedOrganizations(context, user, organizationIds, 'canManageNewsItems')
```
Alllex202, line 90: check that `organizationIds` is not empty

apps/condo/domains/banking/utils/accessSchema.js
```diff
@@ -20,27 +22,21 @@ async function canManageBankEntityWithOrganization ({ authentication: { item: us
...
+
+        const items = await find(listKey, {
+            id_in: ids,
+            deletedAt: null,
+        })
+        organizationIds = uniq(items.map(item => item.organization))
     }
 
-    return await checkPermissionsInUserOrganizationsOrRelatedOrganizations(user.id, organizationIds, permission)
+    return await checkPermissionsInEmployedOrRelatedOrganizations(context, user, organizationIds, permission)
```
Alllex202, line 39: ^^ same

apps/condo/domains/news/access/NewsItemScope.js
SavelevMatthew: This line is already filtering properly
```js
 if (newsItems.some(item => !item.organization)) return false
```

Alllex202: `if ([].some(item => !item.organization)) return false` will be skip

SavelevMatthew: added additional check

apps/condo/domains/organization/utils/accessSchema.js
```diff
@@ -1,239 +1,330 @@
...
     }
+    return uniq(organizationIds)
+}
 
-    return employee.deletedAt === null
+/**
+ * Checks if user is employed in all listed organizations and has all correct permissions in it.
+ * Both organizations and permissions can be single elements if passed as strings instead of arrays
+ * This utils is faster than filtering organization ids from corresponding get<> function,
+ * so use it when you don't need related organizations
```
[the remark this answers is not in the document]
SavelevMatthew: added additional info to JSDoc

PR #4765, opened 2024-05-23
fix(condo): DOMA-9080 Add resource to conditions for searching of meters

AleX83Xpert:


Alllex202 (approved):
Should any tests be added?

PR #4769, opened 2024-05-24
fix(condo): INFRA-333 running docker containers as in a previous cli

Gladioluss:


docker-compose.yml
```diff
@@ -21,12 +21,20 @@ services:
       - ./.env:/app/.env
     networks:
       - app-network
+
   postgresdb:
+    image: alpine
+    depends_on:
+      - postgresdb-master
```
SavelevMatthew, line 28: ?

Why does it depends on master / replica if its standalone?

SavelevMatthew (approved):

PR #4786, opened 2024-05-30
feat(condo): DOMA-9047 add localization for water removal category

abshnko:
...and little automatic fixes according to es lint


dkoviazin (approved):

dkoviazin (commented):
You need to add migration with unique uuid for new category that will be the same for all servers

PR #4789, opened 2024-05-31
fix(keystone): DOMA-8928 fixed types for Select type from Keystone

Alllex202:
## Breaking changes:

1) Now gql-type for fields with the type "Select" does not depend on the property `datatype` and will always be `enum`.
But... if for some reason you still want to specify your type, then you need to use the property `graphQLReturnType`, for example:
```
myField: {
    type: 'Select',
    dataType: 'string',
    graphQLReturnType: 'String', // gql-type for this fields will be 'String', not generated enum
}
```

2) Now field types will need to be specified as strings, not as the imported type.
Using keystone types will result in a error

-----------------

Todo: 
- [x] Fix gql-types for Select fields. There should always be an `enum` regardless of `datatype`
- [x] Specify field types as strings instead of imported types
- [x] Prohibit import usage for schema field types
- [x] Removed duplicates in `ISO_CODES`
- [x] Fix typescript errors
- [x] Remove unused files from `@open-condo/keystone`
- [x] Update all submodules
- [x] Throw an error in `convertStringToTypes` when trying to use an imported field type
- [x] Update submodules to master


packages/keystone/KSv5v6/v5/registerSchema.js
```diff
@@ -33,28 +56,34 @@ function applyKeystoneV5AdminFixes (schema) {
...
-        AuthedRelationship,
-        HiddenRelationship,
         DateInterval,
+        File: FileWithUTF8Name,
+        Json,
+        HiddenRelationship,
+        LocalizedText,
+        Select,
+        SignedDecimal,
+        Text,
```
SavelevMatthew, line 86: Should we throw a error now in non-string cases few lines below?
image

packages/keystone/fields/Select/implementation.js
```diff
@@ -20,7 +22,13 @@ class Select extends KSSelect.implementation {
         if (this.graphQLReturnType) {
             return []
         } else {
-            return super.getGqlAuxTypes()
+            return [
```
SavelevMatthew, line 25: I guess something wrong with alignment here?
image

SavelevMatthew (approved):
Awesome PR 😍

Finally someone did it! Thanks
🥇 🥇 🥇 🥇 🥇

apps/condo/domains/banking/schema/BankTransaction.js
```diff
@@ -75,36 +73,37 @@ const BankTransaction = new GQLListSchema('BankTransaction', {
...
         },
 
         importRemoteSystem: {
             schemaDoc: 'Name of external system, from that this record was copied',
-            type: Select,
+            type: 'Select',
             options: IMPORT_REMOTE_SYSTEM_VALUES,
             // Cannot use `enum` datatype because a number is presented in first character of a value "1CClientBankExchange".
             dataType: 'string',
+            graphQLReturnType: 'String',
```
pahaz, line 106: ?

Alllex202: `enum` cannot be generated because one of the values ​​starts with a digit (`1CClientBankExchange`).
Such values ​​are not allowed for `enum`

packages/keystone/KSv5v6/v5/registerSchema.js
Alllex202: added

sitozzz (approved):

packages/keystone/fields/Select/implementation.js
Alllex202: It's possible that this doesn't affect anything, but just in case I'll leave it as it was done in the original implementation)

PR #4797, opened 2024-06-04
fix(condo): DOMA-9181 fix image and text of card on guide page

nomerdvadcatpyat:


apps/condo/public/onboarding/guide/aboutApp/en/extraIncome/cardImage.webp
Alllex202, line 1: + 200KB for one image :)
other picture from this page are smoller in size

Alllex202 (approved):

PR #4813, opened 2024-06-11
feat(condo): added validation of the Spanish NIF

paulo-rossy:


apps/condo/domains/organization/utils/tin.utils.js
```diff
@@ -29,10 +29,15 @@ const validateTinRU = tinValue => {
     return false
 }
 
+const validateTinES = tinValue => {
```
dkoviazin, line 32: It's more complicated

https://regex101.com/r/eV1fA0/1

https://www.oecd.org/tax/automatic-exchange/crs-implementation-and-assistance/tax-identification-numbers/Spain-TIN.pdf

toplenboren (approved):

PR #4814, opened 2024-06-11
fix(condo): DOMA-9253 fix errors displaying while uploading meters

AleX83Xpert:


apps/condo/domains/meter/utils/metersImporters/AbstractMetersImporter.ts
```diff
@@ -149,12 +149,19 @@ export abstract class AbstractMetersImporter {
                         } else {
                             const mutationError = errors[errorIndex++]
 
-                            const messageForUser = get(mutationError, ['extensions', 'messageForUser'])
+                            const messageForUser = get(
+                                mutationError,
+                                ['extensions', 'messageForUser'], // Error thrown from the mutation itself
+                                get( // Error thrown by models created within the mutation
+                                    mutationError,
+                                    ['originalError', 'errors', 0, 'extensions', 'messageForUser'],
```
sitozzz, line 157: Is it possible to provide 'fallback' error message in case of no `messageForUser` provided?

AleX83Xpert: Maybe, but this is temporary code. For now we moving this logic to backend. Error processing will be changed.

PR #4816, opened 2024-06-11
hotfix(condo): HOTFIX fixed gql type error

Alllex202:


apps/condo/domains/ticket/utils/clientSchema/search.ts
```diff
@@ -92,7 +92,7 @@ const GET_ALL_TICKETS_QUERY = gql`
 `
 
 const GET_ALL_CONTACTS_QUERY = gql`
```
sitozzz, line 94: Why there is no test for that query?

PR #4823, opened 2024-06-13
feat(condo-restapi): DOMA-8986 added external request id middleware in condo-restapi

nomerdvadcatpyat:
Added middleware that added in the response the `rquid` header sent in the request


apps/condo-restapi
sitozzz, line 1: Javascript object is case sensitive! If you want to safely extract some value from headers you should use express api for that.
You can achieve same result via code like that:
```
req.get(EXTERNAL_REQUEST_ID_HEADER)
```

nomerdvadcatpyat: Case sensitivity doesn't seem to matter here. The task specifies that the header will be `rquid`, as expected in the code.
But ok, I’ll do it using the express get method

sitozzz: If someone set header like rquID, your code won't work. req.get is simple & recommended solution for values extraction from request object

sitozzz (approved):

toplenboren (approved):

PR #4840, opened 2024-06-18
feat(condo): DOMA-9224 added isVerified checkbox to contact import modal

nomerdvadcatpyat:
image


apps/condo/pages/contact/index.tsx
```diff
@@ -162,6 +204,8 @@ const ContactTableContent: React.FC<ContactPageContentProps> = (props) => {
                                             rowValidator={contactValidator}
                                             objectCreator={contactCreator}
                                             domainName='contact'
+                                            exampleModalAdditionalContent={IsVerifiedCheckbox}
```
sitozzz, line 207: Why this prop called `exampleModalAdditionalContent`? I suppose it is not an example content

sitozzz:
I think I get it. This is related to the layout and location of the element in the component. Is there any option to split logic of "extra content" and optional "example content"?

apps/condo/pages/contact/index.tsx
nomerdvadcatpyat: because it is a additional content in a import example modal (first step of import modal – import example)

sitozzz (approved):

Alllex202 (approved):

PR #4841, opened 2024-06-18
feat(condo): Doma-9300 fetch with retries

dkoviazin:
We need more control for online interactions. Such as: abort request by timeout and make retries

### DIFF
--- a/packages/keystone/fetch.spec.js
+++ b/packages/keystone/fetch.spec.js
@@ -68,7 +68,7 @@ describe('Fetch with retries', () => {
         expect(fetch).toHaveBeenCalledTimes(2)
     })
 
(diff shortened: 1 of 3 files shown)


sitozzz (approved):

packages/keystone/fetch.spec.js
```diff
@@ -0,0 +1,85 @@
+jest.mock('node-fetch', () => jest.fn())
```
toplenboren, line 1: I think it is better to test this agains express app

Check out initTestExpressApp

toplenboren (approved):

packages/keystone/fetch.spec.js
dkoviazin: In spec files much easier to mock node_module

PR #4848, opened 2024-06-19
refactor(condo): DOMA-9195 add the ability to choose transport for MessageBatch

VKislov:
1. Refactoring allows you to choose a delivery method for specified targets. 
2. Adding the ability to send SMS to the user’s phone by indicating the user ID in the target


VKislov:
A more concise solution to this problem was proposed in #4882

PR #4851, opened 2024-06-19
feat(condo): DOMA-9289: added ticket domain in B2BAppAccessRightSet

nomerdvadcatpyat:


apps/condo/domains/ticket/access/TicketComment.js
Alllex202, line 1: should we allow b2b apps to read and post comments to residents?

nomerdvadcatpyat: update accesses for comments and comment files

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -923,27 +935,59 @@ describe('B2BApp permissions for service user', () => {
...
+        await expectToThrowAccessDeniedErrorToCount(async () => {
+            await TicketComment.count(serviceUser, {})
+        })
+        await expectToThrowAccessDeniedErrorToObjects(async () => {
+            await TicketComment.getOne(serviceUser, {
+                ticket: { id: ticket.id },
+            })
+        })
+        await expectToThrowAccessDeniedErrorToObjects(async () => {
+            await TicketComment.getAll(serviceUser, {})
```
Alllex202, line 989: What about tests for `TicketFile.get...` and `TicketCommentFile.get...`?

Alllex202 (approved):
make sure that TicketChange is displayed correctly in a interface after a ticket has been changed by a service user

toplenboren (approved):

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
nomerdvadcatpyat: they exists

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -975,59 +1048,108 @@ describe('B2BApp permissions for service user', () => {
...
-
             const countWithPermissions = await Ticket.count(serviceUser, {})
-            expect(countWithPermissions).toBe(2)
-            const contactWithPermissions = await Ticket.getOne(serviceUser, { id: ticket.id })
-            expect(contactWithPermissions.id).toBe(ticket.id)
+            expect(countWithPermissions).toEqual(2)
+
+            const ticketWithPermissions = await Ticket.getOne(serviceUser, { id: ticket.id })
+
+            expect(ticketWithPermissions.id).toEqual(ticket.id)
```
SavelevMatthew, line 1066: I guess checking something like `contact.phone` / `contact.email` and `property.user` might be helpful

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -975,59 +1048,108 @@ describe('B2BApp permissions for service user', () => {
...
-            expect(countByUser2).toBe(2)
-
             const countWithPermissions = await TicketComment.count(serviceUser, {})
-            expect(countWithPermissions).toBe(2)
-            const contactWithPermissions = await TicketComment.getOne(serviceUser, { id: ticketComment.id })
-            expect(contactWithPermissions.id).toBe(ticketComment.id)
+            expect(countWithPermissions).toEqual(2)
+
+            const ticketCommentWithPermissions = await TicketComment.getOne(serviceUser, { id: ticketComment.id })
+            expect(ticketCommentWithPermissions.id).toEqual(ticketComment.id)
```
SavelevMatthew, line 1095: Maybe add field check, similar to ticket?

apps/condo/domains/ticket/access/TicketComment.js
```diff
@@ -59,7 +74,29 @@ async function canReadTicketComments ({ authentication: { item: user }, context
...
+const checkManageCommentAccess = async (args) => {
+    const { authentication: { item: user }, originalInput, operation, itemId, context } = args
+
+    if (user.type === SERVICE) {
+        const hasAccess = await canManageObjectsAsB2BAppServiceUser(args)
+        if (!hasAccess) return false
+
+        // service user can't create ticket comment with resident type or update type to resident
+        const resolvedCommentType = get(originalInput, 'type')
+        if (resolvedCommentType === RESIDENT_COMMENT_TYPE) {
```
SavelevMatthew, line 86: Broken logic right here, if it's bulk request or there's not type in input, you'll get `undefined` and this check will be passed

SavelevMatthew: This code is a bit unclear to read. Since create `return` is in the end, while update logic is mixed somewhere in the middle... How about something like this:

```
if (operation === 'create') {
    const ticketsInput = Array.isArray(originalInput) ? originalInput : [originalInput]
    if (ticketsInput.some(ticketData => ticketData.type && ticketData.type === RESIDENT_COMMENT_TYPE)) {
        return false
    }

    return true
} else if (operation === 'update') {
    const ids = itemIds || [itemId]
    const uniqueIds = uniq(ids)
    const comments = await find('TicketComment', { id_in: uniqueIds, deletedAt: null, user: { id: user.id } })
    
    return comments.length === uniqueIds.length
}

return false
```

SavelevMatthew: Example above is more readable imho and support bulk operations (in case we add bulk in canManageAsB2B utils (and we definitely will later))

apps/condo/domains/ticket/access/TicketCommentFile.js
```diff
@@ -51,7 +66,38 @@ async function canReadTicketCommentFiles ({ authentication: { item: user }, cont
...
+            const ticketComment = await getByCondition('TicketComment', {
+                id: ticketCommentFromOriginalInput,
+                deletedAt: null,
+            })
+            if (!ticketComment) return false
+
+            return ticketComment.createdBy === user.id
+        } else if (operation === 'update') {
+            const ticketCommentFile = await getByCondition('TicketCommentFile', {
+                id: itemId,
```
SavelevMatthew, line 89: What do you expect here? If `itemId` is `undefined` (in case of bulk requests), filter should probably be dropped?

SavelevMatthew: How about implementing something similar I've shared above?

apps/condo/domains/ticket/access/TicketComment.js
nomerdvadcatpyat: > Broken logic right here, if it's bulk request or there's not type in input, you'll get undefined and this check will be passed

If it bulk request, in `checkManageCommentAccess` function `originalInput` will be `originalInput[i]` from request

`checkManageCommentAccess` calls below in the code in `canManageTicketComments`, which use for access check to `TicketComment` model. Same for TicketCommentFile.

I will only add a check that bulk update is denied (it's not worked before, but not explicit)

SavelevMatthew (approved):

PR #4858, opened 2024-06-24
feat(resident-app): DOMA-9347 redesigned tabs to look more mobile-friendly 💅

SavelevMatthew:


packages/ui/src/components/Tabs/style.less
```diff
@@ -51,6 +51,29 @@
     }
   }
 
+  & > .condo-tabs-nav,
+  & > div > .condo-tabs-nav {
+    .condo-tabs-nav-wrap::before {
+      width: @condo-global-spacing-40;
+      box-shadow:
+        inset @condo-global-spacing-40 0
```
toplenboren, line 59: maybe move this property to variable?

toplenboren (approved):

packages/ui/src/components/Tabs/style.less
SavelevMatthew: It's different in terms of orientation (left / right shadows use different signs), so I can replace `@condo-global-spacing-40` with something like `@shadow-width`, but for now it's not necessary imho

PR #4860, opened 2024-06-24
hotfix: reduce code duplication via news form functions reuse

sitozzz:


apps/condo/domains/news/components/NewsForm/OldBaseNewsForm.tsx
```diff
@@ -80,169 +87,6 @@ export const FULL_WIDTH_STYLE: React.CSSProperties = { width: '100%' }
 const SMALL_VERTICAL_GUTTER: [Gutter, Gutter] = [0, 24]
 const EXTRA_SMALL_VERTICAL_GUTTER: [Gutter, Gutter] = [0, 10]
 const BIG_HORIZONTAL_GUTTER: [Gutter, Gutter] = [50, 0]
-const ALL_SQUARE_BRACKETS_OCCURRENCES_REGEX = /\[[^\]]*?\]/g
```
toplenboren, line 83: That way if we change something in BaseNewsForm -> OldBaseNewsForm will fail

sitozzz: Ok, just ignored it

Alllex202 (approved):

PR #4867, opened 2024-06-25
fix(condo): remove Yandex Metrica if there is no runtime configuration

paulo-rossy:
I believe it is possible to omit adding a metric component if the corresponding variables, such as yandexMetricaID, are not provided


toplenboren (approved):

pahaz (approved):

SavelevMatthew:
I'm happy to announce that we now know how to handle PRs and CI/CDs from forks 🥳  Please rebase your branch on the current master, after that the new CI will work and we will be able to accept your PRs ✅

PR #4879, opened 2024-06-27
chore(condo): DOMA-9298 update meter fields within unified mutation

AleX83Xpert:


apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
```diff
@@ -342,13 +342,24 @@ const RegisterMetersReadingsService = new GQLCustomSchema('RegisterMetersReading
...
                     }
-
-                    if (foundMeter) {
-                        meterId = foundMeter.id
-                    } else {
-                        try {
+                    try {
+                        if (foundMeter) {
+                            meterId = foundMeter.id
+                            await Meter.update(context, foundMeter.id, {
```
dkoviazin, line 348: Looks like it will update meter every time. If integration sends data every day - then lot's of history records will be created

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
```diff
@@ -369,10 +380,10 @@ const RegisterMetersReadingsService = new GQLCustomSchema('RegisterMetersReading
                                 controlReadingsDate: rawControlReadingsDate ? toISO(rawControlReadingsDate) : dayjs().toISOString(),
                             })
                             meterId = createdMeter.id
-                        } catch (e) {
-                            resultRows.push(e)
-                            continue
                         }
+                    } catch (e) {
+                        resultRows.push(e)
```
pahaz, line 385: Why you just ignore this error? No logs?

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
AleX83Xpert: Good comment, thanks!
Fixed. Also the test was added.

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
AleX83Xpert: This error is not ignored. The client should process it. I think we do not need to log thousands lines from huge registries.

dkoviazin (approved):

PR #4881, opened 2024-06-27
feat(condo): DOMA-9003 prevent creation of duplicated meter readings

AleX83Xpert:


apps/condo/domains/meter/schema/RegisterMetersReadingsService.test.js
```diff
@@ -928,4 +928,48 @@ describe('RegisterMetersReadingsService', () => {
             expect(meter.controlReadingsDate).toBe(output ? dayjs(output).toISOString() : output)
         })
     })
+
+    test('prevent to create readings duplicates', async () => {
+        const [o10n] = await createTestOrganization(adminClient)
```
VKislov, line 933: I’m not sure exactly, but it seems @SavelevMatthew  wrote that we should not use the abbreviation o10n since it is not a generally accepted abbreviation like i18n -> internationalization. And for other o10n developers it is not obvious that this is an organization

VKislov (approved):

PR #4883, opened 2024-06-27
fix(global): INFRA-348 run test dbs as docker compose services

sitozzz:


pahaz (approved):

.github/workflows/_nodejs.condo.core.tests.yml
```diff
@@ -17,24 +17,37 @@ jobs:
     runs-on: ${{ inputs.runs-on }}
 
     steps:
+      - name: install deps
+        run: |
+          sudo apt update && sudo apt install -y git
+
+      - name: setup docker compose
+        uses: KengoTODA/actions-setup-docker-compose@v1
```
SavelevMatthew, line 25: R u sure we can trust this one? Seems like it's not that popular

.github/workflows/_nodejs.condo.core.tests.yml
```diff
@@ -17,24 +17,37 @@ jobs:
...
+          fetch-depth: 0
+          ssh-key: ${{ secrets.SSH_DOCK_SERVER_PRIVATE_KEY }}
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
 
-      - name: run redis for tests
+      - name: run dbs
         run: |
-          docker run -p="127.0.0.1:6379:6379" -d ${{ secrets.DOCKER_REGISTRY }}/doma/utils/redis:6.2
+          cat << EOF > .env
+          POSTGRES_IMAGE=${{ secrets.DOCKER_REGISTRY }}/doma/utils/postgres:13.2
```
SavelevMatthew, line 46: Can we pass only registry here? (Don't like hard-coded versions here)

docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: '3.8'
 x-postgres-common:
   &postgres-common
-  image: postgres:13.8
+  image: ${POSTGRES_IMAGE:-postgres:13.8}
```
SavelevMatthew, line 4: Seems like it's going twice?
`${POSTGRES_IMAGE:-postgres:13.8}`
`${{ secrets.DOCKER_REGISTRY }}/doma/utils/postgres:13.2:-postgres:13.8`

SavelevMatthew: Why versions mismatched?

SavelevMatthew: And can I run it locally without envs?

.github/workflows/_nodejs.condo.core.tests.yml
sitozzz: Here you can pass whatever you want) Not only hard-coded version. By default the image described in the docker-compose file will be used

docker-compose.yml
sitozzz: No it's not like that.
According to official docker docs `:-` is special marker for "default value" if no custom env variable was provided by user

sitozzz: So, if you don't set any env's, you have `postgres:13.8` as your image. Otherwise image value would be same as `POSTGRES_IMAGE`

.github/workflows/_nodejs.condo.core.tests.yml
sitozzz: I think we can. But your truth, that's not a really popular solution to resolve installation issue. So, I can rollback to manual plugin installation process, if this is necessary

.github/workflows/_nodejs.condo.core.tests.yml
```diff
@@ -17,24 +17,36 @@ jobs:
...
+          fetch-depth: 0
+          ssh-key: ${{ secrets.SSH_DOCK_SERVER_PRIVATE_KEY }}
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
 
-      - name: run redis for tests
+      - name: run dbs
         run: |
-          docker run -p="127.0.0.1:6379:6379" -d ${{ secrets.DOCKER_REGISTRY }}/doma/utils/redis:6.2
+          cat << EOF > .env
+          POSTGRES_IMAGE=${{ secrets.DOCKER_REGISTRY }}/doma/utils/
```
SavelevMatthew, line 45: Registry?

.github/workflows/_nodejs.condo.core.tests.yml
```diff
@@ -17,24 +17,36 @@ jobs:
...
+          REDIS_IMAGE=${{ secrets.DOCKER_REGISTRY }}/doma/utils/
+          EOF
+
+          docker-compose up -d redis postgresdb-master postgresdb-replica
```
SavelevMatthew, line 49: just Postgres?

SavelevMatthew (approved):

.github/workflows/_nodejs.condo.core.tests.yml
sitozzz: Yup, my bad)

.github/workflows/_nodejs.condo.core.tests.yml
sitozzz: Fixed

PR #4886, opened 2024-06-27
HOTFIX: add try->catch to addressservice fetch

toplenboren:


yarn.lock
```diff
@@ -1599,6 +1599,97 @@ __metadata:
   languageName: unknown
   linkType: soft
```
pahaz, line 1601: Lock!

PR #4888, opened 2024-06-27
fix(condo) HOTFIX: Add try-catch block to address service

toplenboren:


SavelevMatthew (approved):

AleX83Xpert (approved):
Looks like we need to implement same solution everywhere.
What about to write some logs [here](https://github.com/open-condo-software/condo/blob/master/packages/keystone/fetch.js#L80)?

apps/address-service/domains/common/utils/services/suggest/providers/DadataSuggestionProvider.js
```diff
@@ -171,36 +171,42 @@ class DadataSuggestionProvider extends AbstractSuggestionProvider {
...
+                 * In the case of 403, we may check if our limit exceeded
+                 * If yes, we should create a separate token rotator for dadata provider (search & suggest)
+                 * @link https://dadata.ru/api/stat/
+                 */
+            }
 
-        this.logger.warn({ msg: 'CallToDaData error', status, url, data: body, reqId: this.req.id })
+            this.logger.warn({ msg: 'CallToDaData error', status, url, data: body, reqId: this.req.id })
+        } catch (err) {
+            this.logger.warn({ msg: 'CallToDaData failed to fetch', err: err.message, url, data: body, reqId: this.req.id })
```
pahaz, line 208:
```suggestion
            this.logger.warn({ msg: 'CallToDaData failed to fetch', err, url, data: body, reqId: this.req.id })
```

toplenboren:
> Looks like we need to implement same solution everywhere.
> What about to write some logs [here](https://github.com/open-condo-software/condo/blob/master/packages/keystone/fetch.js#L80)?

Like this Idea, but regardless of fetch implementation you should be able to collect logs and do something on failure

dkoviazin (approved):

PR #4889, opened 2024-06-28
fix(condo): DOMA-9421 fix error throwing and processing if some cells…

AleX83Xpert:
… are nil

### DIFF
--- a/apps/condo/domains/meter/utils/taskSchema/DomaMetersImporter.js
+++ b/apps/condo/domains/meter/utils/taskSchema/DomaMetersImporter.js
@@ -22,13 +22,18 @@ class DomaMetersImporter extends AbstractMetersImporter {
         return isEqual(columnsNames, normalizedColumns)
     }
 
+    /**
+     * @inheritDoc
+     * @return {RegisterMetersReadingsReadingInput}
+     */
     transformRow (row) {
+        /** @type {string[]} */
         const errors = []
         if (!get(this, ['mappers', 'resourceId', row[4]])) {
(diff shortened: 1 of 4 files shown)


apps/condo/domains/meter/utils/taskSchema/DomaMetersImporter.js
```diff
@@ -39,7 +44,7 @@ class DomaMetersImporter extends AbstractMetersImporter {
         return {
             address: row[0],
             addressInfo: {
-                unitType: this.mappers.unitType[row[2].toLowerCase()],
+                unitType: this.mappers.unitType[String(row[2]).toLowerCase()],
                 unitName: row[1],
             },
             accountNumber: row[3],
```
sitozzz, line 50: Is there or somewhere else in this class any mechanism of checking row length and corresponding data types? IMO looks quite dangerous

AleX83Xpert: This class receives `string[][]` data got from incoming file. But there was a erroneous file with cell below the main table. See description in the jira task.

AleX83Xpert: I do not think we should control string length here. It's better incapsulate it within models.

sitozzz: Ok, sounds reasonable

sitozzz (approved):

PR #4891, opened 2024-06-28
feat(condo): DOMA-9391 add read query for ResidentBillingVirtualReceipt

ekabardinsky:


apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -353,4 +353,5 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
 
 module.exports = {
     BillingReceipt,
+    DEFAULT_CATEGORY,
```
AleX83Xpert, line 356: It's better to move this var to external constant, because you have execute all file.

apps/condo/domains/billing/access/AllResidentBillingVirtualReceiptsService.js
```diff
@@ -0,0 +1,20 @@
+/**
+ * Generated by `createservice billing.AllResidentBillingVirtualReceiptsService --type queries`
+ */
+
+const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
+
+async function canAllResidentBillingVirtualReceipts ({ authentication: { item: user } }) {
```
AleX83Xpert, line 7: Maybe allow only for residents?

AleX83Xpert (approved):

apps/condo/domains/billing/access/AllResidentBillingVirtualReceiptsService.js
```diff
@@ -0,0 +1,20 @@
...
+    if (!user) return throwAuthenticationError()
+    if (user.deletedAt) return false
+
+    return true
```
sitozzz, line 11: It would be nice to somehow limit access to data. At least for the user group to which this method applies

apps/condo/domains/billing/schema/BillingReceipt.js
ekabardinsky: We already have such constant, will replace it

apps/condo/domains/billing/access/AllResidentBillingVirtualReceiptsService.js
ekabardinsky: This is going to work exactly as allResidentBillingReceipt query

apps/condo/domains/billing/access/AllResidentBillingVirtualReceiptsService.js
ekabardinsky: This is going to work exactly as allResidentBillingReceipt query

PR #4898, opened 2024-06-29
fix(condo): DOMA-9362 fixed currency code for ruble in sbbol integration

Alllex202:
- Fixed currency code for ruble in sbbol integration
 - Removed code RUR


apps/condo/domains/common/constants/currencies.js
```diff
@@ -181,11 +180,15 @@ const ISO_CODES = [
...
 
 const ISO_CODES_FOR_SBBOL = {
-    '643': 'RUB',
-    '810': 'RUR',
+    '810': 'RUB',
     '840': 'USD',
     '978': 'EUR',
 }
+
+const ALTERNATIVE_CURRENCY_CODES_FROM_SBBOL = {
```
SavelevMatthew, line 188: I'm not sure this one should be place here....

Why `SBBOL` in common and on open-source part?

Alllex202: > Why SBBOL in common ...

You're right, it shouldn't be here. Moved it to integration folder.

> Why SBBOL ... on open-source part?

This is how it happened historically. I won't change anything in this regard in this pr :)

nomerdvadcatpyat (approved):

PR #4905, opened 2024-07-02
feat(condo): DOMA-9434 use existed organization in ticket load test

nomerdvadcatpyat:
Added an ORGANIZATION_ID env, used employee from it in test for test case when user works in organization with a lot of data


Alllex202 (approved):

apps/condo/k6/src/ticket.test.ts
```diff
@@ -156,10 +185,22 @@ export async function checkFrontend (data) {
...
     const page = context.newPage()
 
+    const organizationForCheckFrontend = data.organizationForCheckFrontend
+    let cookie
+    let organizationLinkId
+
+    if (organizationForCheckFrontend.cookie && organizationForCheckFrontend.organizationLinkId) {
+        cookie = organizationForCheckFrontend.cookie
+        organizationLinkId = organizationForCheckFrontend.organizationLinkId
+    } else {
```
sitozzz, line 195: Why do you need extra token for that organization? Is there a "special" user?

nomerdvadcatpyat: It's a employee user with `property` ticket visibility type

sitozzz (approved):

PR #4917, opened 2024-07-04
feat(condo): DOMA-9367 added ServiceProblemsAlert component

nomerdvadcatpyat:
Added a notification about not working functionality in the platform. Configurable using feature flags.

image


apps/condo/domains/common/components/ServiceProblemsAlert.tsx
```diff
@@ -0,0 +1,47 @@
...
+    const serviceProblemsAlertConfig = useFlagValue(SERVICE_PROBLEMS_ALERT)
+
+    const title = useMemo(() => get(serviceProblemsAlertConfig, 'title'), [serviceProblemsAlertConfig])
+    const description = useMemo(() => get(serviceProblemsAlertConfig, 'description'), [serviceProblemsAlertConfig])
+    const pages = useMemo(() => get(serviceProblemsAlertConfig, 'pages', []), [serviceProblemsAlertConfig])
+
+    // do not show alert for some time if user closed it
+    const expires = useMemo(() => new Date(new Date().getTime() + SERVICE_PROBLEMS_ALERT_COOKIE_EXPIRES_IN_SECONDS), [])
+    const handleClose = useCallback(() =>  cookie.set(SERVICE_PROBLEMS_ALERT_COOKIE_KEY, true, { expires }), [expires])
+    const alreadyClosed = useMemo(() => cookie.get(SERVICE_PROBLEMS_ALERT_COOKIE_KEY), [])
```
Alllex202, line 28: Do we need logic with closing this alert?

Maybe it would be better to always show it while the service has problems?

Alllex202, line 28: Do we need logic with closing this alert?

Maybe it would be better to always show it while the service has problems?

apps/condo/domains/common/components/ServiceProblemsAlert.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const ServiceProblemsAlert = () => {
+    const router = useRouter()
+
+    const { useFlagValue } = useFeatureFlags()
+    const serviceProblemsAlertConfig = useFlagValue(SERVICE_PROBLEMS_ALERT)
+
+    const title = useMemo(() => get(serviceProblemsAlertConfig, 'title'), [serviceProblemsAlertConfig])
+    const description = useMemo(() => get(serviceProblemsAlertConfig, 'description'), [serviceProblemsAlertConfig])
+    const pages = useMemo(() => get(serviceProblemsAlertConfig, 'pages', []), [serviceProblemsAlertConfig])
```
Alllex202, line 23: can there be different messages for different pages?

abshnko (approved):

apps/condo/domains/common/components/ServiceProblemsAlert.tsx
nomerdvadcatpyat: We have decided to remove this logic

apps/condo/domains/common/components/ServiceProblemsAlert.tsx
nomerdvadcatpyat: So far, no, we will not complicate the logic of the config of this alert.
If there is such a case, we can add it in the future

toplenboren:
Hey) Let's remove the border radius property. It looks better with square borders

Or lets add padding, so that top border radius is visible too.

apps/condo/domains/common/components/ServiceProblemsAlert.tsx
```diff
@@ -0,0 +1,36 @@
...
+    const showOnPage = useMemo(() => pages.length === 0 || pages.some(page => router.pathname.includes(page)),
+        [pages, router.pathname])
+    const isEmptyAlert = !title && !description
+
+    if (isEmptyAlert || !showOnPage) {
+        return null
+    }
+
+    return (
+        <Alert
```
toplenboren, line 30: as a hack we can render this on a yellow background :-)

toplenboren (approved):
Please remove border radius or make it non visible :-) Otherwise its good

Alllex202 (approved):

apps/condo/domains/common/components/ServiceProblemsAlert.tsx
nomerdvadcatpyat: I've added a `banner` prop to Alert component

PR #4942, opened 2024-07-09
refactor(condo): INFRA-155 rid off node env test usages

sitozzz:


apps/address-service/index.js
```diff
@@ -63,7 +63,7 @@ module.exports = {
                 getAddressProviderLimitHealthCheck(),
             ],
         }),
-        conf.NODE_ENV === 'test' ? undefined : new OIDCKeystoneApp(),
+        conf.DISABLE_OIDC_AUTH ? undefined : new OIDCKeystoneApp(),
```
nomerdvadcatpyat, line 66: do we need add new envs in `.env.example` ?

nomerdvadcatpyat (approved):

run_condo_domain_tests.sh
```diff
@@ -32,7 +32,7 @@ fi
 
 node bin/prepare.js -f condo -r condo
 
-export NODE_ENV=test
+export USE_LOCAL_FEATURE_FLAGS=true
```
Alllex202, line 35: Should we add `DISABLE_NEXT_APP =true`?

Alllex202 (approved):

run_condo_domain_tests.sh
sitozzz: Yup, I forget to add it here

apps/address-service/index.js
sitozzz: I don't think so. This flag should be used in test environment only. But .env.example used to show how to startup project

PR #4948, opened 2024-07-10
feat(condo): DOMA-9628 add field Inn to user component

tolmachev21:
Было: 
image

Стало: 
image

### DIFF
--- a/apps/condo/domains/user/components/UserOrganizationsList.tsx
+++ b/apps/condo/domains/user/components/UserOrganizationsList.tsx
@@ -48,6 +48,7 @@ interface IOrganizationEmployeeItem {
 const OrganizationEmployeeItem: React.FC<IOrganizationEmployeeItem> = (props) => {
     const intl = useIntl()
     const OrganizationMessage = intl.formatMessage({ id: 'pages.condo.property.field.Organization' })
+    const TinMessage = intl.formatMessage({ id: 'pages.condo.property.field.Tin' })
     const PositionMessage = intl.formatMessage({ id: 'employee.Position' })
     const RoleMessage = intl.formatMessage({ id: 'employee.Role' })
 
@@ -84,11 +85,19 @@ const OrganizationEmployeeItem: React.FC<IOrganizationEmployeeItem> = (props) =>
                 </Col>
                 <Col lg={5} xs={10}>
                     <Typography.Text type='secondary'>
+                        {TinMessage}
+                    </Typography.Text>
+                </Col>
+                <Col lg={18} xs={12} offset={1}>
+                    <NotDefinedField value={get(employee, ['organization', 'tin'])} />
+                </Col>
+                <Col lg={5} xs={10}>
+                    <Typography.Text type='secondary'>
                         {PositionMessage}
                     </Typography.Text>
                 </Col>
                 <Col lg={18} xs={12} offset={1}>
-                    <NotDefinedField value={get(employee, ['position'])}/>
+                    <NotDefinedField value={get(employee, ['position'])} />
                 </Col>
                 <Col lg={5} xs={10}>
                     <Typography.Text type='secondary'>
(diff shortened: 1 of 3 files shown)


abshnko (commented):
you should fix your yarn.lock, it shouldn't be updated

apps/condo/domains/user/components/UserOrganizationsList.tsx
```diff
@@ -48,6 +48,7 @@ interface IOrganizationEmployeeItem {
 const OrganizationEmployeeItem: React.FC<IOrganizationEmployeeItem> = (props) => {
     const intl = useIntl()
     const OrganizationMessage = intl.formatMessage({ id: 'pages.condo.property.field.Organization' })
+    const InnMessage = intl.formatMessage({ id: 'pages.condo.property.field.Inn'})
```
SavelevMatthew, line 51: In our domain notation we use word `tin` instead of `inn`:

image

Look, you've already got your answer few lines below :)

apps/condo/domains/user/components/UserOrganizationsList.tsx
```diff
@@ -82,13 +83,21 @@ const OrganizationEmployeeItem: React.FC<IOrganizationEmployeeItem> = (props) =>
                         )}
                     />
                 </Col>
+                <Col lg={5} xs={10}>
+                    <Typography.Text type='secondary'>
+                        {InnMessage}
+                    </Typography.Text>
+                </Col>
+                <Col lg={18} xs={12} offset={1}>
+                    <NotDefinedField value={get(employee, ['organization', 'tin'])} />
```
SavelevMatthew, line 92: Why its `NotDefinedField`? I thought every organization must have one...

SavelevMatthew: https://github.com/open-condo-software/condo/blob/master/apps/condo/domains/organization/schema/Organization.js#L108

apps/condo/domains/user/components/UserOrganizationsList.tsx
Alllex202, line 51: It would be better to call it `tin` :)
```suggestion
    const TinMessage = intl.formatMessage({ id: 'pages.condo.property.field.Tin' })
```

apps/condo/lang/en/en.json
```diff
@@ -631,6 +631,7 @@
   "pages.condo.property.field.Address.notFound.modal.contacts.description.bot": "Telegram-bot",
   "pages.condo.property.field.Name": "Name",
   "pages.condo.property.field.Organization": "Organization",
+  "pages.condo.property.field.Inn": "Taxpayer Identification Number",
```
SavelevMatthew, line 634: Same about naming ^

yarn.lock
```diff
@@ -2753,6 +2753,16 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/code-frame@npm:^7.24.7":
```
pahaz, line 2756: You don't need to change eny dependencies

apps/condo/lang/en/en.json
toplenboren: Best to just leave it as `TIN` it would not look good to render giant string in UI :-)

apps/condo/domains/user/components/UserOrganizationsList.tsx
toplenboren: NotDeinedField is used everywhere in this component. I'd have left it like this tbh, maybe we can even rename it `OrganizationInfoField` (it is in org domain)

yarn.lock
```diff
@@ -2753,6 +2753,16 @@ __metadata:
   languageName: node
```
toplenboren, line 2753: Hey)

This file contains versions, hashes and soruces for every dependency you have. It should not be changed if dependency list was not changed

(So if no package.json was changed -> this file should not be changed too)

I suggest you do this:

1. git fetch and rebase on master
2. check all submodules are inited (`git submodule update --init --recursive ) from root
3. manually delete `node_modules` folder and reinstall dependencies

toplenboren: Result should be like this:

Your `yarn.lock` is 100% like `yarn.lock` on server

pahaz (approved):

SavelevMatthew (approved):

Alllex202 (approved):

abshnko (approved):

PR #4961, opened 2024-07-12
hotfix: update knex dependency to resolve security warning

sitozzz:


toplenboren (approved):

package.json
```diff
@@ -90,7 +90,7 @@
     "es5-ext": "0.10.53",
     "debuglog": "1.0.1",
     "debug": "4.3.4",
-    "knex": "0.95.15",
+    "knex": "2.4.0",
```
toplenboren, line 93: We also need to double check `adapterCache` with this change

toplenboren (approved):

package.json
sitozzz: Yup, that's a great suggest. But do we have tests of the caching mechanism?

SavelevMatthew (approved):
🤞

package.json
toplenboren: yea

PR #4975, opened 2024-07-18
fix(condo): DOMA-9622 print localized errors to the result file after meters import finished

AleX83Xpert:


apps/condo/domains/meter/utils/taskSchema/AbstractMetersImporter.js
```diff
@@ -176,13 +176,26 @@ class AbstractMetersImporter {
                             if (messageForUser) {
                                 rowErrors.push(messageForUser)
                             } else {
-                                const mutationErrorMessages =
+                                let mutationErrorMessages =
                                     get(mutationError, ['originalError', 'errors', 0, 'data', 'messages'], []) || []
+
+                                // errors thrown to mutation from models
+                                if (mutationErrorMessages.length === 0) {
+                                    mutationErrorMessages = (get(mutationError, ['originalError', 'errors', 0, 'originalError', 'errors'], []) || []).reduce((result, error) => {
```
sitozzz, line 184: Why do you need another default value as empty array. get function already had an fallback value

AleX83Xpert: null is valid value for `get()`, but I need an array here.
image

PR #4982, opened 2024-07-19
feat(condo): DOMA-8525 allow to restrict send meter readings

AleX83Xpert:
- [x] Allow meter to be automatic without master b2b app
- [x] Allow to set `Meter.isAutomatic` via `registerMetersReadings` mutation
- [x] Allow to set isAutomatic by user using meters import template


apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
```diff
@@ -405,6 +407,7 @@ const RegisterMetersReadingsService = new GQLCustomSchema('RegisterMetersReading
                                 commissioningDate: toISO(get(reading, ['meterMeta', 'commissioningDate'])),
                                 sealingDate: toISO(get(reading, ['meterMeta', 'sealingDate'])),
                                 controlReadingsDate: rawControlReadingsDate ? toISO(rawControlReadingsDate) : dayjs().toISOString(),
+                                isAutomatic: get(reading, ['meterMeta', 'isAutomatic']),
```
SavelevMatthew, line 410: What if value is not provided?

apps/condo/domains/meter/utils/taskSchema/DomaMetersImporter.spec.js
```diff
@@ -70,6 +70,7 @@ describe('DomaMetersImporter', () => {
                 '',
                 '',
                 '',
+                'y',
```
SavelevMatthew, line 73: y?

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
AleX83Xpert: Same behavior as with place `field` - ignoring. Improved a test for this case.

apps/condo/domains/meter/utils/taskSchema/DomaMetersImporter.spec.js
AleX83Xpert: For now, any not empty value means `true`.

apps/condo/domains/meter/utils/taskSchema/importerResolver.js
```diff
@@ -74,14 +75,15 @@ function getColumnNames (format, locale) {
...
-        { name: ReadingSubmissionDateMessage, type: 'custom', required: true },
+        { name: ReadingSubmissionDateMessage, type: 'date', required: true },
         { name: VerificationDateMessage, type: 'date', required: false },
         { name: NextVerificationDateMessage, type: 'date', required: false },
         { name: InstallationDateMessage, type: 'date', required: false },
         { name: CommissioningDateMessage, type: 'date', required: false },
         { name: SealingDateMessage, type: 'date', required: false },
         { name: ControlReadingsDate, type: 'date', required: false },
         { name: PlaceColumnMessage, type: 'string', required: false },
+        { name: AutomaticColumnMessage, type: 'string', required: false },
```
SavelevMatthew, line 86: Should it be bool type instead?

apps/condo/domains/meter/utils/taskSchema/DomaMetersImporter.spec.js
SavelevMatthew: like `false` is true too?

Or where I can find the behaviour?

apps/condo/domains/meter/utils/taskSchema/importerResolver.js
AleX83Xpert: I'd like to abstract from any excel types. Empty cell means false, filled cell means true.

apps/condo/domains/meter/utils/taskSchema/DomaMetersImporter.spec.js
AleX83Xpert: There is only one behavior: empty cell means false.
> like false is true too?

Yes, "false" is not empty string. As other strings like "no", "-", "нет", "да", "+", ...

apps/condo/domains/meter/utils/taskSchema/importerResolver.js
AleX83Xpert: Hm, as I see this settings not used since processing was moved to backend.

dkoviazin (approved):

apps/condo/domains/meter/utils/taskSchema/importerResolver.js
AleX83Xpert: Removed

apps/condo/domains/meter/utils/taskSchema/DomaMetersImporter.spec.js
SavelevMatthew: I don't think its user-friendly...

What if organisation will fill everything with "yes/no"

AleX83Xpert: Refactored. Localized strings yes/no now used.

apps/condo/domains/meter/utils/taskSchema/DomaMetersImporter.spec.js
```diff
@@ -70,6 +75,7 @@ describe('DomaMetersImporter', () => {
                 '',
                 '',
                 '',
+                'да',
```
SavelevMatthew, line 78: You should specify locale explicitly. Right now it's a random test

SavelevMatthew: If you run it outside of `ru` region - its failed.
If you run it on device with different local - its failed

SavelevMatthew:
```
const { Yes } = require('@app/condo/lang/.../')
```

AleX83Xpert: I use the same mappers for testing (without lang files).
See class ImporterWrapper at the beginning of the file.

AleX83Xpert: Translations is not using in this test. It's possible to use "угу" and "нене" instead.

SavelevMatthew: can we replace it with `en` words, since it's the language of open_source.

Or you can generate `yes/no` values (strings) from faker, to test that is working as expected

AleX83Xpert: Ok, done

apps/condo/domains/meter/utils/taskSchema/DomaMetersImporter.js
```diff
@@ -64,6 +69,7 @@ class DomaMetersImporter extends AbstractMetersImporter {
                 commissioningDate: row[15],
                 sealingDate: row[16],
                 controlReadingsDate: row[17],
+                isAutomatic: isEmpty(cell19Value) ? undefined : this.mappers.isAutomatic[cell19Value],
```
SavelevMatthew, line 72: Incorrect comparison.

In line 40:
```
const cell19Value = String(null).toLowerCase() // "null"
```

```
isEmpty("null") // false -> this.mappers.isAutomatic['null']
```

AleX83Xpert: this.mappers.isAutomatic['null'] -> undefined -> ignored of backend as empty value

apps/condo/domains/meter/utils/taskSchema/DomaMetersImporter.js
```diff
@@ -64,6 +69,7 @@ class DomaMetersImporter extends AbstractMetersImporter {
                 commissioningDate: row[15],
                 sealingDate: row[16],
                 controlReadingsDate: row[17],
+                isAutomatic: get(this, ['mappers', 'isAutomatic', cell19Value]),
```
SavelevMatthew, line 72:
```suggestion
                isAutomatic: row[19] ? get(this, ['mappers', 'isAutomatic', cell19Value]) : undefined,
```

SavelevMatthew: If you have mapping like this: 
```json
{
  "": true
}
```

In case of empty string, you will throw error on line 42, but update meter on line 72. Stay consistent

SavelevMatthew (approved):

PR #4984, opened 2024-07-19
fix(condo): DOMA-9737 fixed context for getting of feature flag in task of close completed tickets

Alllex202:
When executing a task, there is no need to pass the Keystone context.  if you pass it, then the data about feature flags will not be up-to-date

### DIFF
--- a/apps/condo/domains/ticket/tasks/closeCompletedTickets.js
+++ b/apps/condo/domains/ticket/tasks/closeCompletedTickets.js
@@ -27,7 +27,7 @@ const closeCompletedTickets = async (defaultLimit = 100) => {
     // Therefore, you should limit updating tickets for each organization at one time if there are a lot of them.
     // With the help of the feature flag, you can manage this restriction.
     const limitByOrganization = await featureToggleManager.getFeatureValue(
-        context, MAX_COUNT_COMPLETED_TICKET_TO_CLOSE_FOR_ORGANIZATION_TASK, defaultLimit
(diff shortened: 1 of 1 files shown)


sitozzz (approved):

apps/condo/domains/ticket/tasks/closeCompletedTickets.js
```diff
@@ -27,7 +27,7 @@ const closeCompletedTickets = async (defaultLimit = 100) => {
     // Therefore, you should limit updating tickets for each organization at one time if there are a lot of them.
     // With the help of the feature flag, you can manage this restriction.
     const limitByOrganization = await featureToggleManager.getFeatureValue(
-        context, MAX_COUNT_COMPLETED_TICKET_TO_CLOSE_FOR_ORGANIZATION_TASK, defaultLimit
+        null, MAX_COUNT_COMPLETED_TICKET_TO_CLOSE_FOR_ORGANIZATION_TASK, defaultLimit
```
pahaz, line 30: why you don't need to send context?

Alllex202: We really need caching of flags when they are requested within the same request so that the values are the same.   But when we request feature flags in tasks, then in the case of passing the Keystone context, we will always receive outdated data, except for the very first request, since the Keystone context retains the value of the feature flags.

toplenboren:
Looks really non obvious. Perhaps we should modify feature flags lib? So that evereyone knows :-) that you should not pass context in tasks :-)

PR #4987, opened 2024-07-20
feat(condo): DOMA-9739 disabling notifications for specific appIds

VKislov:
I googled that the error occurs when incorrect data is used for the team_id, key_id and private_key variables.
We use exactly the correct data, but if the connection is raised to push the old mobile application, and then in the same connection we configure another application(new mob app), then apns throws an error, because they opened a connection with it for one application, and are trying to send a push to another, That's why they swear that the provider is invalid. If the hypothesis is correct and this is the reason, then you need to make a separate sending queue for each application so that the configs do not change within the same open connection. The easiest way to check is this: if old application -> return 0. If errors disappear in voice pushes, then this hypothesis is correct.

### DIFF
--- a/apps/condo/domains/notification/adapters/appleAdapter.js
+++ b/apps/condo/domains/notification/adapters/appleAdapter.js
@@ -184,7 +184,7 @@ class AppleAdapter {
                     appId: get(appIds, pushToken),
                 }
 
-            target.push(pushData)
+            if (!APPS_WITH_DISABLED_NOTIFICATIONS.includes(data.app)) target.push(pushData)
 
             if (!pushContext[pushType]) pushContext[pushType] = pushData
         })
@@ -234,7 +234,7 @@ class AppleAdapter {
                     logger.error({ msg: 'Unknown appId. Config was not found', appId })
                     continue
                 }
-                if (APPS_WITH_DISABLED_NOTIFICATIONS.includes(appId)) continue
+
                 const currentNotificationsBatch = notificationsSortedByAppId[appId]
                 const app = new AppleMessaging(currentConfig)
 
--- a/apps/condo/domains/notification/adapters/appleAdapter.js
+++ b/apps/condo/domains/notification/adapters/appleAdapter.js
@@ -10,6 +10,7 @@ const {
     APPLE_CONFIG_ENV,
     PUSH_TYPE_DEFAULT,
     PUSH_TYPE_SILENT_DATA,
+    APPS_WITH_DISABLED_NOTIFICATIONS_ENV,
 } = require('@condo/domains/notification/constants/constants')
 const { EMPTY_APPLE_CONFIG_ERROR, EMPTY_NOTIFICATION_TITLE_BODY_ERROR } = require('@condo/domains/notification/constants/errors')
 
@@ -17,6 +18,7 @@ const AppleMessaging = require('./apple/AppleMessaging')
 const { APS_RESPONSE_STATUS_SUCCESS } = require('./apple/constants')
 
 const APPLE_CONFIG = conf[APPLE_CONFIG_ENV] ? JSON.parse(conf[APPLE_CONFIG_ENV]) : null
+const APPS_WITH_DISABLED_NOTIFICATIONS = conf[APPS_WITH_DISABLED_NOTIFICATIONS_ENV] ? JSON.parse(conf[APPS_WITH_DISABLED_NOTIFICATIONS_ENV]) : []
 const DEFAULT_PUSH_SETTINGS = {
     aps: {
         'mutable-content': 1,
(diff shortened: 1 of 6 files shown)


apps/condo/domains/notification/adapters/appleAdapter.js
```diff
@@ -232,6 +232,7 @@ class AppleAdapter {
                     logger.error({ msg: 'Unknown appId. Config was not found', appId })
                     continue
                 }
+                if (appId === 'ai.doma.clients') continue
```
toplenboren, line 235: please explain the change in PR desciption

+ move this to env

toplenboren (changes requested):

apps/condo/domains/notification/adapters/appleAdapter.js
toplenboren: it is also best to add this not only to apple adapter

toplenboren (changes requested):

apps/condo/domains/notification/adapters/appleAdapter.js
pahaz, line 235: It looks like hardcore

apps/condo/domains/notification/adapters/firebaseAdapter.spec.js
```diff
@@ -195,6 +195,40 @@ describe('Firebase adapter utils', () => {
         expect(pushContext.data._body).toEqual(pushData.notification.body)
     })
 
+    it('dont send push notification to blocked app', async () => {
```
toplenboren, line 198:
```suggestion
    it('doesnt send push notification to blocked app', async () => {
```

apps/condo/domains/notification/adapters/firebaseAdapter.js
```diff
@@ -178,7 +180,7 @@ class FirebaseAdapter {
                     ...extraPayload,
                 }
 
-            target.push(pushData)
+            if (!APPS_WITH_DISABLED_NOTIFICATIONS.includes(data.app)) target.push(pushData)
```
toplenboren, line 183: So, if there are no configuration, then:

APPS_WITH_DISABLED_NOTIFICATIONS will be null

and this command will be null.includes(data.app) which will be unhandled exception

apps/condo/domains/notification/adapters/firebaseAdapter.spec.js
```diff
@@ -195,6 +195,40 @@ describe('Firebase adapter utils', () => {
...
+        const tokens = [PUSH_FAKE_TOKEN_SUCCESS]
+        const pushData = {
+            tokens,
+            notification: {
+                title: 'Doma.ai',
```
toplenboren, line 203: Please do not use Doma.ai in open source part! We have open source engineers!

apps/condo/domains/notification/adapters/firebaseAdapter.spec.js
```diff
@@ -195,6 +195,40 @@ describe('Firebase adapter utils', () => {
...
+                body: `${dayjs().format()} Condo greets you!`,
+            },
+            data: {
+                app : 'ai.doma.clients',
```
toplenboren, line 207: ^^^

toplenboren (changes requested):

apps/condo/domains/notification/adapters/firebaseAdapter.js
toplenboren: !

apps/condo/domains/notification/adapters/firebaseAdapter.spec.js
```diff
@@ -195,6 +195,40 @@ describe('Firebase adapter utils', () => {
         expect(pushContext.data._body).toEqual(pushData.notification.body)
     })
 
+    it('doesnt send push notification to blocked app', async () => {
```
toplenboren, line 198: to app with disabled notifications

toplenboren (changes requested):

apps/condo/domains/notification/adapters/firebaseAdapter.js
toplenboren: Why have you implemented this stuff using two different instructions? 

In Apple adapter you use continue, here you use includes 

+ What about Huawei adapter?

VKislov: These are two different adapters, they have the same essence but are implemented differently in some logic. This task is not about refactoring and bringing all adapters to a single image. This is a task about testing a hypothesis, which should fix a certain error.

toplenboren: What about huawei? 

I do not see a blocker in writing code in same way

toplenboren (approved):

dkoviazin (approved):

PR #4991, opened 2024-07-21
fix(greendom) HOTFIX: Do not create NewsItemSharing if step with it was skipped

toplenboren:


toplenboren:
Bypassing branch protections due to the fix being limited only to one file and tested by @toplenboren

PR #4992, opened 2024-07-21
feat(condo) DOMA-9326 Let users delete published and sent news items

toplenboren:
So, if a news item has been sent to mobile app, you can now delete it.

1. You can now set validBefore on common news items

image

2. If news has been sent you can 'delete' it 

image

image


apps/condo/lang/ru/ru.json
```diff
@@ -1989,6 +1989,9 @@
   "notification.messages.SEND_DAILY_STATISTICS.email.moreIncidents": "и еще {more} отключений",
   "news.ConfirmDeleteTitle": "Удалить новость?",
   "news.ConfirmDeleteMessage": "Восстановить ее будет нельзя. Но можно создать заново.",
+  "news.DeprecateTitle": "Сделать неактуальной",
+  "news.ConfirmDeprecateTitle": "Удалить уже отправленную новость",
```
toplenboren, line 1993: Сделать новость неактуальной sounds better here

apps/condo/lang/ru/ru.json
```diff
@@ -1989,6 +1989,9 @@
...
+  "news.ConfirmDeprecateMessage": "Новость пропадет с главного экрана мобильного приложения. Push-уведомления удалены не будут. Новость можно будет отправить заново",
```
toplenboren, line 1994: Новость останется в ленте sounds better here

toplenboren: Новость останется в push-уведомлении is better here

SavelevMatthew:
Should "delete" button on page be as dangerous as in modal?
image

SavelevMatthew:
image

"from resident**s**", and I would probably add term "Sent" to "push notifications"

apps/condo/domains/news/schema/NewsItem.js
```diff
@@ -260,9 +260,11 @@ const NewsItem = new GQLListSchema('NewsItem', {
             }
 
             if (operation === 'update') {
-                if (sentAt) {
+                // You can update only validBefore field if news item was already sent
+                if (sentAt && (Object.keys(resolvedData) > 1 || isEmpty(get(resolvedData, 'validBefore') ))) {
```
Alllex202, line 264: Should we use `resolvedData` or `originalInput` here?

Alllex202: what about dv and sender?

apps/condo/pages/news/[id]/index.tsx
```diff
@@ -363,6 +374,20 @@ const NewsItemCard: React.FC = () => {
                                                 <Button type='primary'>{ResendTitle}</Button>
                                             </Link>
                                         ),
+                                        (isSent && (!newsItem.validBefore || dayjs(newsItem.validBefore) > dayjs()) && newsItemSharings.length === 0) && (
```
Comment (Alllex202) [inline apps/condo/pages/news/[id]/index.tsx:377]: Is this condition necessary here? `newsItemSharings.length === 0`

Comment (Alllex202) [reply apps/condo/pages/news/[id]/index.tsx:377]: For example, I want to hide a news item from the resident app, even if I sent it through other channels. Why can't I do that?

Comment (toplenboren) [reply apps/condo/pages/news/[id]/index.tsx:377]: This is a great point. Right now product solution is to remove that, but i'm not sure how to provide the best ux here

apps/condo/domains/news/schema/NewsItem.js
VKislov: From the front-line utilities, the dv and sender are entered automatically, it seems.

VKislov (approved):

apps/condo/domains/news/schema/NewsItem.js
Alllex202, line 264: You have an incorrect condition

`Object.keys(resolvedData) > 1` ==> `[el1, el2, el3] > 1` ==> `false` - It will always be false

So you can update any field together with `validBefore`

Alllex202: You should check that `originalInput` only has `dv`, `sender` and `validBefore`


Something like
```suggestion
const originalInputFieldNames = Object.keys(originalInput).filter(fieldName => !['dv', 'sender'].includes(fieldName))
if (originalInputFieldNames.length !== 1 || originalInputFieldNames[0] !== 'validBefore' || isEmpty(get(resolvedData, 'validBefore')))
```

Alllex202: > dv and sender are entered automatically, it seems.

That's right, so they need to be filtered and take `originalInput` instead of `resolvedData`.
`resolvedData` can contain a lot of unnecessary stuff

toplenboren: Thanks, missed that

pahaz (approved):

abshnko (approved):

PR #5001, opened 2024-07-23
fix(condo): Added missing csv-parse package

dkoviazin:


apps/condo/package.json
```diff
@@ -91,6 +91,7 @@
     "body-parser": "^1.19.1",
     "carbone": "3.2.3",
     "classnames": "^2.2.6",
+    "csv-parse": "^5.5.6",
```
sitozzz, line 94: It is not a version, that used inside another apps!

sitozzz (changes requested):
I think we should use same version across all of the apps if it possible. If it is not - please comment why we should not do that

apps/condo/package.json
dkoviazin: Fixed

sitozzz (approved):
Now looks good

PR #5004, opened 2024-07-23
feat(condo): DOMA-9715 added checkUserExistence query

Alllex202:
Снимок экрана 2024-07-23 в 21 38 06


apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,130 @@
...
+        {
+            access: access.canCheckUserExistence,
+            schema: 'checkUserExistence(data: CheckUserExistenceInput!): CheckUserExistenceOutput',
+            doc: {
+                summary: 'Using an action token with a verified phone number,' +
+                    ' checks whether a user with the specified type (resident, staff, service) is registered.\n' +
+                    'As a result, information about the completion of some important fields (name, email, phone, password) is also returned.',
+                errors: ERRORS,
+            },
+            resolver: async (parent, args, context) => {
```
toplenboren, line 80: what about guard?

apps/condo/domains/user/schema/CheckUserExistenceService.test.js
```diff
@@ -0,0 +1,265 @@
...
+                const [confirmPhoneAction] = await createTestConfirmPhoneAction(adminClient, {
+                    isPhoneVerified: true,
+                    phone: deletedUserAttrs.phone,
+                })
+                const [result] = await checkUserExistenceByTestClient(anonymousClient, {
+                    confirmActionToken: confirmPhoneAction.token,
+                    userType: deletedUserAttrs.type,
+                })
+                expect(result).toEqual({
+                    userIsExist: false,
```
toplenboren, line 210: but you woulrnt be able to register when user was soft deleted, will you not?

apps/condo/domains/user/schema/CheckUserExistenceService.test.js
```diff
@@ -0,0 +1,265 @@
...
+                    nameIsSet: false,
+                    emailIsSet: false,
+                    phoneIsSet: false,
+                    passwordIsSet: false,
+                })
+            })
+        })
+
+        test('confirmPhoneAction should not be marked as used after checkUserExistence has been executed', async () => {
```
toplenboren, line 219: does it mean, that I can reuse one confirmPhoneAction to trigger only this handle?

apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,130 @@
...
+        message: 'Invalid format of "sender" field value',
+        correctExample: '{ dv: 1, fingerprint: \'example-fingerprint-alphanumeric-value\'}',
+    },
+}
+
+const CheckUserExistenceService = new GQLCustomSchema('CheckUserExistenceService', {
+    types: [
+        {
+            access: true,
+            type: 'input CheckUserExistenceInput { dv: Int!, sender: JSON!, confirmActionToken: String!, userType: UserTypeType! }',
```
SavelevMatthew, line 62:
```suggestion
            type: 'input CheckUserExistenceInput { dv: Int!, sender: SenderFieldInput!, confirmActionToken: String!, userType: UserTypeType! }',
```

SavelevMatthew: Token must be passed as` ConfirmPhoneActionWhereUniqueIUnput` or as `ID`

SavelevMatthew:
```suggestion
            type: 'input CheckUserExistenceInput { dv: Int!, sender: SenderFieldInput!, confirmActionToken: ID!, userType: UserTypeType! }',
```

apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,130 @@
...
+                const user = users[0]
+                const result = {
+                    userIsExist: false,
+                    nameIsSet: false,
+                    emailIsSet: false,
+                    phoneIsSet: false,
+                    passwordIsSet: false,
+                }
+
+                if (!user || user.deletedAt) return result
```
SavelevMatthew, line 113: Correct me if I'm wrong, but deletedAt is used as "block mechanism". (User cannot relogin / reregister, if deletedAt and his phone  is set to non-null

SavelevMatthew: It's important for us to keep this behaviour, can we add test for this?

SavelevMatthew (changes requested):
I cannot pass this PR without guard, so I'll block it for now and will return after guards addition

apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * List of possible errors, that this custom schema can throw
+ * They will be rendered in documentation section in GraphiQL for this custom schema
+ */
+const ERRORS = {
+    MULTIPLE_USERS_FOUND: {
+        query: 'checkUserExistence',
+        code: NOT_FOUND,
+        type: 'MULTIPLE_USERS_FOUND',
+        message: 'Multiple users found',
```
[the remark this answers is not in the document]
Alllex202: I don't find any errors here that should be shown to the user

apps/condo/domains/user/schema/CheckUserExistenceService.test.js
Alllex202: with one `confirmPhoneAction` you can call `checkUserExistence` and then something else, like `signInUser`

apps/condo/domains/user/schema/CheckUserExistenceService.js
Alllex202: > Token must be passed as` ConfirmPhoneActionWhereUniqueIUnput` or as `ID`

Token is not `ConfirmPhoneActionWhereUniqueIUnput`. It is simple string, type changed to `ID`.

Description from GraphQL docs
> The ID scalar type represents a unique identifier, often used to refetch an object or as the key for a cache. The ID type is serialized in the same way as a String; however, defining it as an ID signifies that it is not intended to be human‐readable.

So there is almost no difference

apps/condo/domains/user/schema/CheckUserExistenceService.js
Alllex202: > Correct me if I'm wrong, but deletedAt is used as "block mechanism". (User cannot relogin / reregister, if deletedAt and his phone is set to non-null

Yes, a user who was soft deleted" cannot sign in or register again

apps/condo/domains/user/schema/CheckUserExistenceService.test.js
Alllex202: Yes, a user who was soft deleted cannot register again

apps/condo/domains/user/schema/CheckUserExistenceService.js
Alllex202: > It's important for us to keep this behaviour, can we add test for this?

Added tests

apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,133 @@
...
+        message: 'Invalid format of "sender" field value',
+        correctExample: '{ dv: 1, fingerprint: \'example-fingerprint-alphanumeric-value\'}',
+    },
+}
+
+const CheckUserExistenceService = new GQLCustomSchema('CheckUserExistenceService', {
+    types: [
+        {
+            access: true,
+            type: 'input CheckUserExistenceInput { dv: Int!, sender: SenderFieldInput!, confirmActionToken: ID!, userType: UserTypeType! }',
```
SavelevMatthew, line 63: What about service users? You block them in other mutations

apps/condo/domains/user/access/CheckUserExistenceService.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Generated by `createservice user.CheckUserExistenceService --type queries`
+ */
+
+async function canCheckUserExistence () {
+    return true
```
SavelevMatthew, line 6: (optional, just strange idea) Seems strange, but we can limit usage to only unauthorised users. Like reverse default logic

apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,133 @@
...
+                    {
+                        token: confirmActionToken,
+                        expiresAt_gte: new Date().toISOString(),
+                        completedAt: null,
+                        isPhoneVerified: true,
+                    },
+                )
+                if (!action) throw new GQLError(ERRORS.TOKEN_NOT_FOUND, context)
+
+                const phoneFromAction = normalizePhone(action.phone) || null
```
SavelevMatthew, line 101: Should we check it here? Action must validateInput by itself...

SavelevMatthew: The only case I can find it reasonable, if ConfirmAction will support landline, and mutations / not

SavelevMatthew: But its definitely not the case :)

apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,133 @@
...
+                if (!phoneFromAction) throw new GQLError(ERRORS.INVALID_PHONE_NUMBER, context)
+
+                const users = await find('User', { type: userType, phone: action.phone })
+                if (users.length > 1) throw new GQLError(ERRORS.MULTIPLE_USERS_FOUND, context)
```
SavelevMatthew, line 105: Is it possible?
image

apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,133 @@
...
+
+                const user = users[0]
```
SavelevMatthew, line 107: C-like language will kill your process in this case :)

What if it is `[]`?

apps/condo/domains/user/schema/CheckUserExistenceService.js
Alllex202: Not all mutations are blocked for service users. So I decided to allow this mutation for them

apps/condo/domains/user/access/CheckUserExistenceService.js
Alllex202: What is this for?
I have never come across such logic anywhere

apps/condo/domains/user/schema/CheckUserExistenceService.js
Alllex202: > Should we check it here? Action must validateInput by itself...

You are right, I will remove normalization from here :)

apps/condo/domains/user/schema/CheckUserExistenceService.js
Alllex202: Actually it's not possible, I'll remove this check

apps/condo/domains/user/schema/CheckUserExistenceService.js
Alllex202: Nothing terrible will happen))

Снимок экрана 2024-07-25 в 18 53 12

But as I wrote above, I will change this check a little

apps/condo/domains/user/schema/CheckUserExistenceService.test.js
SavelevMatthew: I still not quite sure this one is right...

apps/condo/domains/user/schema/CheckUserExistenceService.js
toplenboren: JS))

apps/condo/domains/user/constants/common.js
```diff
@@ -19,6 +19,7 @@ const CONFIRM_PHONE_ACTION_EXPIRY = 3600 // 1 hour
 const CONFIRM_PHONE_SMS_MAX_RETRIES = 10
 const MAX_SMS_FOR_IP_BY_DAY = 10
 const MAX_SMS_FOR_PHONE_BY_DAY = 10
+const MAX_ANONYMOUS_REQUESTS_FOR_IP_BY_DAY = 10
```
toplenboren, line 22:
```suggestion
const MAX_ANONYMOUS_REQUESTS_FOR_IP_PER_DAY = 10
```

apps/condo/domains/user/constants/errors.js
```diff
@@ -80,6 +81,12 @@ const GQL_ERRORS = {
         type: SMS_FOR_IP_DAY_LIMIT_REACHED,
         message: 'Too many sms requests from this ip address. Try again tomorrow',
     },
+    DAILY_REQUEST_LIMIT_FOR_IP_REACHED: {
+        code: 'BAD_USER_INPUT',
+        type: DAILY_REQUEST_LIMIT_FOR_IP_REACHED,
+        message: 'Too many requests from this ip address. Try again tomorrow',
```
toplenboren, line 87: I'm happy we are giving hackers friendly error messages :-)

apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,116 @@
...
+                summary: 'Using an action token with a verified phone number,' +
+                    ' checks whether a user with the specified type (resident, staff, service) is registered.\n' +
+                    'As a result, information about the completion of some important fields (name, email, phone, password) is also returned.',
+                errors: ERRORS,
+            },
+            resolver: async (parent, args, context) => {
+                const { data } = args
+                const { confirmActionToken, userType } = data
+
+                await checkDayAnonymousRequestLimitCounters(context, 'checkUserExistence', context.req.ip)
```
toplenboren, line 72:
```suggestion
                await checkDailyAnonymousRequestLimitCounters(context, 'checkUserExistence', context.req.ip)
```

apps/condo/domains/user/schema/CheckUserExistenceService.test.js
toplenboren: But can I use one confirmPhoneAction to call this particular query? Maybe from different IPs (VPNs and such)

apps/condo/domains/user/utils/serverSchema/requestLimitHelpers.js
```diff
@@ -0,0 +1,25 @@
...
+const { MAX_ANONYMOUS_REQUESTS_FOR_IP_BY_DAY } = require('@condo/domains/user/constants/common')
+const { GQL_ERRORS } = require('@condo/domains/user/constants/errors')
+const { RedisGuard } = require('@condo/domains/user/utils/serverSchema/guards')
+
+
+const REDIS_GUARD = new RedisGuard()
+const MAX_ANONYMOUS_REQUESTS_BY_DAY = Number(conf['MAX_ANONYMOUS_REQUESTS_FOR_IP_BY_DAY']) || MAX_ANONYMOUS_REQUESTS_FOR_IP_BY_DAY
+const IP_WHITE_LIST = conf.IP_WHITE_LIST ? JSON.parse(conf.IP_WHITE_LIST) : []
+
+const checkDayAnonymousRequestLimitCounters = async (context, prefix, rawIp) => {
```
toplenboren, line 13: I'm not sure that guard by ip is sufficient. What if I just switch VPNs? Use different proxies?

toplenboren: I'd have added smth with confirm phone action. So that even if I change IPs I would not be able to use this mutation more then N times

toplenboren: Phone numbers are a bit harder to get :-(

apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,116 @@
...
+
+                if (!confirmActionToken) throw new GQLError(ERRORS.TOKEN_NOT_FOUND, context)
+
+                const action = await ConfirmPhoneAction.getOne(context,
+                    {
+                        token: confirmActionToken,
+                        expiresAt_gte: new Date().toISOString(),
+                        completedAt: null,
+                        isPhoneVerified: true,
+                    },
```
nomerdvadcatpyat, line 84: `deletedAt` check?

apps/condo/domains/user/schema/CheckUserExistenceService.test.js
Alllex202: > But can I use one confirmPhoneAction to call this particular query? Maybe from different IPs (VPNs and such)

You can.
Similarly, you can call query "getPhoneByConfirmPhoneActionToken"

Alllex202: > I still not quite sure this one is right...

We shouldn't ask the user to confirm their phone number 10 times to register, only 1... So when using a token in this query, we don't mark it as used

In this case we can allow it

apps/condo/domains/user/utils/serverSchema/requestLimitHelpers.js
Alllex202: also added guard by phone

apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,130 @@
...
+
+                const phoneFromAction = normalizePhone(action.phone) || null
+                if (!phoneFromAction) throw new GQLError(ERRORS.INVALID_PHONE_NUMBER, context)
+
+                const users = await find('User', { type: userType, phone: action.phone })
+                if (users.length > 1) throw new GQLError(ERRORS.MULTIPLE_USERS_FOUND, context)
+
+                const user = users[0]
+                const result = {
+                    userIsExist: false,
```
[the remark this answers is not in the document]
SavelevMatthew: Can we remove other `Is` to be consistent? Like...:

```
{
  "userExists": true,
  "phoneSet": false
  "emailSet": true
}
```

SavelevMatthew (approved):

apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,119 @@
...
+                        deletedAt: null,
+                    },
+                )
+                if (!action) throw new GQLError(ERRORS.TOKEN_NOT_FOUND, context)
+
+                await checkDailyRequestLimitCountersByPhone(context, 'checkUserExistence', action.phone)
+
+                const users = await find('User', { type: userType, phone: action.phone })
+
+                const user = users.length === 1 ? users[0] : null
```
SavelevMatthew, line 93: why not just `getByCondition` above?)

apps/condo/domains/user/schema/CheckUserExistenceService.js
```diff
@@ -0,0 +1,119 @@
...
+        message: 'Wrong value for data version number',
+    },
+    WRONG_SENDER_FORMAT: {
+        query: 'checkUserExistence',
+        variable: ['data', 'sender'],
+        code: BAD_USER_INPUT,
+        type: WRONG_FORMAT,
+        message: 'Invalid format of "sender" field value',
+        correctExample: '{ dv: 1, fingerprint: \'example-fingerprint-alphanumeric-value\'}',
+    },
```
nomerdvadcatpyat, line 43: looks like common errors, mb move it to common consts?

nomerdvadcatpyat (approved):

apps/condo/domains/user/schema/CheckUserExistenceService.js
Alllex202: fixed

apps/condo/domains/user/schema/CheckUserExistenceService.js
Alllex202: fixed

apps/condo/domains/user/schema/CheckUserExistenceService.js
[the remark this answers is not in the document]
Alllex202: > Can we remove other Is to be consistent? 

updated

PR #5014, opened 2024-07-25
feat(condo) DOMA-9795 Add canReadPayments and canReadBillingReceipts 

toplenboren:


VKislov (approved):

apps/condo/domains/acquiring/access/Payment.js
```diff
@@ -61,10 +65,13 @@ async function canManagePayments ({ authentication: { item: user }, operation, i
...
 }
 
-async function canReadPaymentsSensitiveData ({ authentication: { item: user }, existingItem, context }) {
+async function canReadPaymentsSensitiveData ({ authentication: { item: user }, existingItem, context, listKey }) {
     if (!user || user.deletedAt) return false
     if (user.isSupport || user.isAdmin) return true
 
+    const hasDirectAccess = await canDirectlyReadSchemaObjects(user, listKey)
+    if (hasDirectAccess) return true
+
```
Alllex202, line 74: Is it ok that we give direct access to sensitive data?

toplenboren: I do not really see caveats here, if you have direct access for list you should be, imo, able to see all fields

Alllex202: I compared the accesses, they seem similar, so ok

Alllex202 (approved):

PR #5017, opened 2024-07-25
fix(condo): DOMA-9780 fix translations so that mob app could take translations directly from the entity

VKislov:
This is necessary so that the mobile app does not store translations on its own and they are always synchronized between web/ios/android platforms


dkoviazin (approved):

apps/condo/lang/en/en.json
```diff
@@ -1745,13 +1745,13 @@
...
-  "meterResource.HotWater.measure": "cubic meter",
-  "meterResource.Electricity.measure": "kw",
-  "meterResource.HeatSupply.measure": "gcal",
-  "meterResource.GasSupply.measure": "cubic meter",
-  "meterResource.ColdAir.measure": "gcal",
-  "meterResource.Drainage.measure": "cubic meter",
+  "meterResource.ColdWater.measure": "m³",
+  "meterResource.HotWater.measure": "m³",
+  "meterResource.Electricity.measure": "kWh",
+  "meterResource.HeatSupply.measure": "Gcal",
```
SavelevMatthew, line 1751: why do you use small `k` and big `G`?

SavelevMatthew (commented):
Can you explain, why this is necessary for mobile? They're not ready for label with more than 3 characters or what?

Alllex202 (approved):

toplenboren (approved):

pahaz (approved):

VKislov:
> Can you explain, why this is necessary for mobile? They're not ready for label with more than 3 characters or what?

1) in figma it's like this.
2) This is necessary so that the mobile app does not store translations on its own and they are always synchronized between web/ios/android platforms

PR #5034, opened 2024-07-30
fix(condo): DOMA-9746 deeplink to mini-app after registration

tolmachev21:


apps/condo/domains/organization/components/OrganizationSelect.tsx
```diff
@@ -78,8 +80,12 @@ export const InlineOrganizationSelect: React.FC = () => {
     const { setIsVisible: showCreateOrganizationModal, ModalForm: CreateOrganizationModalForm } = useCreateOrganizationModalForm({
         onFinish: async (createdOrganization) => {
             const organizationType = get(createdOrganization, 'type')
+            const isValidNextUrl = next && !Array.isArray(next) && isSafeUrl(next)
```
toplenboren, line 83: doesnt isSafeUrl check `next` type already?

apps/condo/domains/organization/components/OrganizationSelect.tsx
```diff
@@ -78,8 +80,12 @@ export const InlineOrganizationSelect: React.FC = () => {
...
 
-            if (organizationType === MANAGING_COMPANY_TYPE) {
+            // After registration and created first organization we want redirect user to nextUrl
```
toplenboren, line 85:
```suggestion
            // Redirects user to `next` url after signup
```

apps/condo/domains/organization/components/OrganizationSelect.tsx
```diff
@@ -78,8 +80,12 @@ export const InlineOrganizationSelect: React.FC = () => {
...
+            if (isValidNextUrl && !hasInvites && !link && !filteredEmployees.length) {
+                await router.push(next)
+            } else if (organizationType === MANAGING_COMPANY_TYPE) {
```
toplenboren, line 88: Why wouldn't it work for management companies?

apps/condo/domains/organization/components/OrganizationSelect.tsx
```diff
@@ -78,8 +80,12 @@ export const InlineOrganizationSelect: React.FC = () => {
     const { setIsVisible: showCreateOrganizationModal, ModalForm: CreateOrganizationModalForm } = useCreateOrganizationModalForm({
         onFinish: async (createdOrganization) => {
             const organizationType = get(createdOrganization, 'type')
+            const isValidNextUrl = next && !Array.isArray(next) && isSafeUrl(next)
 
-            if (organizationType === MANAGING_COMPANY_TYPE) {
+            // After registration and created first organization we want redirect user to nextUrl
+            if (isValidNextUrl && !hasInvites && !link && !filteredEmployees.length) {
```
toplenboren, line 86: - why !hasInvites blocks redirect to `next` ?

- why `!link` blocks redirect?

tolmachev21: We need to redirect to nextUrl only after registration and first creation of the organization. Therefore there is a check for lint, filteredEmployees, hasInvites, isValidNextUrl

apps/condo/pages/auth/register.tsx
```diff
@@ -31,12 +31,17 @@ const RegisterPage: AuthPage = () => {
...
     const router = useRouter()
+    const { query: { next }  } = router
 
     const { token, isConfirmed, tokenError, setToken, setTokenError } = useContext(RegisterContext)
     const [step, setStep] = useState('inputPhone')
 
     const handleFinish = useCallback(async () => {
-        await router.push('/')
+        if (next) {
+            await router.push(`/?next=${next}`)
```
SavelevMatthew, line 41: Strange, next must only available on auth pages

SavelevMatthew (changes requested):
Seems strange to me, why all pages do accept `next` parameter now?

apps/condo/domains/organization/components/OrganizationSelect.tsx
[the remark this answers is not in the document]
tolmachev21: I am understand this, but after registration and creation of the organization, although we get a page with /nextUrl, however after a few seconds we are redireced on /tour. So we need to change the code in the index. How I can change this behaviore?

apps/condo/domains/organization/components/OrganizationSelect.tsx
nomerdvadcatpyat: If user has invites/employees, then he is not be redirected to `next` link.
Are you sure that it is correct logic?
I think if the `next` parameter is specified, then the user should be redirected there.

SavelevMatthew (approved):

PR #5038, opened 2024-07-30
refactor(condo): DOMA-8306 refactored mutation "getNewsItemsRecipientsCounters"

Alllex202:
**Problem**: `getNewsItemsRecipientsCounters` crashes with large data

**Solution**: changed `loadListByChunks` to `allItemsQueryByChunks` so that flat objects are returned, without nesting and virtual fields

_Сan also think about caching queries, but I haven't done it yet._

### DIFF
--- a/apps/condo/domains/news/constants/common.js
+++ b/apps/condo/domains/news/constants/common.js
@@ -4,7 +4,7 @@ const conf = require('@open-condo/config')
 
 const SENDING_DELAY_SEC = Number(get(conf, 'NEWS_ITEMS_SENDING_DELAY_SEC', 15))
 const NEWS_SENDING_TTL_IN_SEC = Number(get(conf, 'NEWS_ITEM_SENDING_TTL_SEC', 60 * 30))
-const LOAD_RESIDENTS_CHUNK_SIZE = 20
+const LOAD_RESIDENTS_CHUNK_SIZE = 50
 
 module.exports = {
     SENDING_DELAY_SEC,
--- a/apps/condo/domains/news/schema/GetNewsItemsRecipientsCountersService.js
+++ b/apps/condo/domains/news/schema/GetNewsItemsRecipientsCountersService.js
@@ -3,9 +3,8 @@
  */
 const { filter, pick, uniq, uniqBy } = require('lodash')
 
-const { GQLCustomSchema } = require('@open-condo/keystone/schema')
+const { GQLCustomSchema, allItemsQueryByChunks } = require('@open-condo/keystone/schema')
 
-const { loadListByChunks } = require('@condo/domains/common/utils/serverSchema')
 const access = require('@condo/domains/news/access/GetNewsItemsRecipientsCountersService')
 const { queryFindResidentsByOrganizationAndScopes } = require('@condo/domains/news/utils/accessSchema')
 const {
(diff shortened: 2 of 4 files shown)


toplenboren (approved):
Good stuff!

apps/condo/domains/news/schema/GetNewsItemsRecipientsCountersService.test.js
```diff
@@ -22,10 +22,13 @@ const { buildPropertyMap } = require('@condo/domains/property/utils/testSchema/f
 const { createTestResident } = require('@condo/domains/resident/utils/testSchema')
 const { makeClientWithNewRegisteredAndLoggedInUser, createTestUser } = require('@condo/domains/user/utils/testSchema')
 
+
 let adminClient, staffClientYes
 let dummyO10n
 describe('GetNewsItemsRecipientsCountersService', () => {
-    beforeAll(async () => {
+    jest.setTimeout(60000)
```
AleX83Xpert, line 29: It's better to set previous value after this test.

apps/condo/domains/news/constants/common.js
```diff
@@ -4,7 +4,7 @@ const conf = require('@open-condo/config')
 
 const SENDING_DELAY_SEC = Number(get(conf, 'NEWS_ITEMS_SENDING_DELAY_SEC', 15))
 const NEWS_SENDING_TTL_IN_SEC = Number(get(conf, 'NEWS_ITEM_SENDING_TTL_SEC', 60 * 30))
-const LOAD_RESIDENTS_CHUNK_SIZE = 20
+const LOAD_RESIDENTS_CHUNK_SIZE = 50
```
nomerdvadcatpyat, line 7: We decrease this const from 50 to 20 before. For fix out of memory errors when fetch residents?
If yes, are you sure that now it's will be work fine?

apps/condo/domains/news/schema/GetNewsItemsRecipientsCountersService.js
```diff
@@ -3,20 +3,18 @@
...
 const {
     countUniqueUnitsFromResidents,
     getUnitsFromProperty,
     queryConditionsByUnits,
 } = require('@condo/domains/news/utils/serverSchema/recipientsCounterUtils')
-const { Property } = require('@condo/domains/property/utils/serverSchema')
 
 
-const CHUNK_SIZE = 20
+const CHUNK_SIZE = 50
```
nomerdvadcatpyat, line 17: Or it was be this const :)

nomerdvadcatpyat (approved):

apps/condo/domains/news/schema/GetNewsItemsRecipientsCountersService.test.js
Alllex202: What is the previous value? :) Where can I get it?

Alllex202: To run tests locally there is a limit of 60 seconds, and in CI it is 15. Therefore it is not clear to which value we should set

AleX83Xpert: I'm afraid that this value will set for all not started tests everywhere since this line being executed (in current process). If not, it's ok.

AleX83Xpert: Maybe it's possible to use something like that: https://jestjs.io/docs/api#testname-fn-timeout (3rd argument as timeout)

Alllex202: I didn't know about this, thanks)
Increased timeout for one test only

AleX83Xpert (approved):

PR #5045, opened 2024-07-31
fix(condo): DOMA-9843 fix marketSetting formValuePreprocessor

VKislov:
image
"Variable \"$data\" got invalid value { dv: 1, sender: { dv: 1, fingerprint: \"ZOEdK5UUn9Ru\" }, organization: { connect: [Object] }, isPaymentCashTypeBlocked: false, residentAllowedPaymentTypes: [\"online\", \"cash\"] }; Field \"isPaymentCashTypeBlocked\" is not defined by type \"MarketSettingCreateInput\"."

### DIFF
--- a/apps/condo/domains/marketplace/components/MarketSetting/MarketSettingForm.tsx
+++ b/apps/condo/domains/marketplace/components/MarketSetting/MarketSettingForm.tsx
@@ -2,6 +2,7 @@ import { MarketSetting as MarketSettingType } from '@app/condo/schema'
 import { Col, Form, Row } from 'antd'
 import { Gutter } from 'antd/es/grid/row'
 import get from 'lodash/get'
+import omit from 'lodash/omit'
 import uniq from 'lodash/uniq'
 import { useRouter } from 'next/router'
 import React, { useMemo } from 'react'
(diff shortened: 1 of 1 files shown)


apps/condo/domains/marketplace/components/MarketSetting/MarketSettingForm.tsx
```diff
@@ -74,6 +74,7 @@ export const MarketSettingForm: React.FC<IMarketSettingForm> = ({ marketSetting,
                 const residentAllowedPaymentTypes = [INVOICE_PAYMENT_TYPE_ONLINE]
                 if (!values[IS_CASH_PAYMENT_TYPE_BLOCKED]) residentAllowedPaymentTypes.push(INVOICE_PAYMENT_TYPE_CASH)
                 values.residentAllowedPaymentTypes = uniq(residentAllowedPaymentTypes)
+                values.isPaymentCashTypeBlocked = undefined
```
toplenboren, line 77: ?

SavelevMatthew (changes requested):
Why do you pass field that not even exist?
And why do you mask it undefined?

apps/condo/domains/marketplace/components/MarketSetting/MarketSettingForm.tsx
Alllex202, line 77:
```suggestion
                values[IS_CASH_PAYMENT_TYPE_BLOCKED]= undefined
```

Alllex202 (approved):

VKislov:
> Why do you pass field that not even exist? And why do you mask it undefined?

Because it comes from Form.Item, but it is not represented in the mutation input, so I overwrite it in the mutation preprocessor

SavelevMatthew (approved):

PR #5053, opened 2024-08-02
fix(condo): INFRA-296 temp rollback http logger back

sitozzz:


packages/keystone/logging/getKeystonePinoOptions.js
```diff
@@ -2,12 +2,15 @@ const { getLogger } = require('./getLogger')
...
 const logger = getLogger('http')
+const IGNORE_PATH = new Set(['/api/features', '/api/version', '/favicon.ico'])
 
 function getKeystonePinoOptions () {
     // NOTE(pahaz): https://github.com/pinojs/pino-http#pinohttpopts-stream
     return {
         logger,
-        autoLogging: false,
+        autoLogging: {
+            ignore: (req) => IGNORE_PATH.has(req.url),
```
pahaz, line 12: it's better to ignore 2XX, 3XX, 4XX statuses and log others results

toplenboren (approved):

packages/keystone/logging/getKeystonePinoOptions.js
sitozzz: I don't think this is possible, because of that module didn't allow to use response object here https://github.com/pinojs/pino-http#pinohttpopts-stream

VKislov (approved):

PR #5054, opened 2024-08-02
feat(condo): DOMA-9868 temporary ability to create heap snapshots

AleX83Xpert:
### How to use
1. detect process id with
```bash
ps -ef|grep keystone
```
2. send signal
```bash
kill -SIGPIPE <pid>
```
3. go to pod's console
4. search for "Snapshot file URL:"
5. download file

### DIFF
--- a/packages/keystone/KSv5v6/v5/prepareKeystone.js
+++ b/packages/keystone/KSv5v6/v5/prepareKeystone.js
@@ -1,4 +1,6 @@
+const fs = require('fs')
 const process = require('node:process')
+const path = require('path')
 const v8 = require('v8')
 
 const { AdminUIApp } = require('@keystonejs/app-admin-ui')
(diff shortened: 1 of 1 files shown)


packages/keystone/KSv5v6/v5/prepareKeystone.js
```diff
@@ -241,6 +247,44 @@ process.on('unhandledRejection', (err, promise) => {
...
+        console.timeEnd(labelUploadSnapshot)
+
+        const { filename } = fileInfo
+        const url = ExportFileAdapter.publicUrl({ filename })
+
+        console.log(`Snapshot file info: ${JSON.stringify(fileInfo)}`)
+        console.log(`Snapshot file URL: ${url}`)
+    } catch (err) {
+        console.error(err)
+    }
```
sitozzz, line 285: Should it also fire `process.exit` at the end of that callback?

AleX83Xpert: No, we do not need to kill process. try-catch is just for preventing process failing.

sitozzz (approved):

toplenboren (approved):

Alllex202 (approved):

packages/keystone/KSv5v6/v5/prepareKeystone.js
AleX83Xpert: This callback work only if someone execute `kill -SIGPIPE <pid>` in console. Node process will not be killed.

PR #5060, opened 2024-08-05
refactor(condo): DOMA-9873 move file adapter to packages

AleX83Xpert:


packages/keystone/fileAdapter/fileAdapter.js
```diff
@@ -5,10 +5,9 @@ const express = require('express')
 const { isEmpty, get } = require('lodash')
```
sitozzz, line 5: It'll be better to add this package to "exports" section of the `package.json`. In case if we have plans to upgrade minimum requirement version of nodejs

AleX83Xpert: Done

packages/keystone/package.json
```diff
@@ -12,6 +12,7 @@
     "./preprocessors": "./preprocessors/index.js",
     "./rateLimiting": "./rateLimiting/index.js",
     "./sentry": "./sentry/index.js",
+    "./fileAdapter/": "./fileAdapter/",
```
sitozzz, line 15: This notation should point on the concrete executable file)
```suggestion
    "./fileAdapter/": "./fileAdapter/file Adapter.js",
```

sitozzz (approved):

packages/keystone/package.json
sitozzz: There is two options. You can export every executable that you need to use outside packages module. Or you can export only `index.js` file with reexport of all of the modules. That's up to you)

pahaz (approved):

PR #5061, opened 2024-08-05
feature(condo): DOMA-9798 user can delete himself from DomaAI

tolmachev21:
Added a new button: 
image

With a confirm modal: 
image

The delete handler uses the ResetUserService mutation. And after deleting the user, signout() is called. The user cannot log in with the same credentials.


apps/condo/lang/ru/ru.json
```diff
@@ -2392,6 +2392,9 @@
   "profile.ChangePasswordModalTitle": "Изменение пароля",
   "profile.ChangePassword": "Сбросить пароль",
   "profile.PasswordResetTimeout": "(сбросить ещё раз можно через 00: {countdown} секунд)",
+  "profile.Delete.CurrentUser": "Удалить профиль",
+  "profile.Delete.ConfirmMessage": "Вы уверены, что хотите удалить профиль? Восстановить профиль будет невозможно",
```
Alllex202, line 2396: Is it worth mentioning here that the organization's data will not be deleted?

apps/condo/domains/user/components/UserProfileForm.tsx
```diff
@@ -1,24 +1,29 @@
+import { useMutation } from '@apollo/client'
```
Alllex202, line 1:
```suggestion
import { useMutation } from '@open-condo/next/apollo'
```

apps/condo/domains/user/components/UserProfileForm.tsx
```diff
@@ -83,6 +92,36 @@ export const UserProfileForm = () => {
         return router.push(RESET_PASSWORD_URL)
     }, [router])
 
+    const handleDeleteButtonClick = useCallback(async () => {
+        const profileExtraData = {
+            dv: 1,
+            sender: getClientSideSenderInfo(),
+        }
+
+        const saveName = false
```
Alllex202, line 101: I think you should keep the name.

apps/condo/domains/user/components/UserProfileForm.tsx
```diff
@@ -83,6 +92,36 @@ export const UserProfileForm = () => {
...
+        const userId = {
+            id: user.id,
+        }
```
Alllex202, line 104: Why put this in a separate variables?

apps/condo/domains/user/components/UserProfileForm.tsx
```diff
@@ -83,6 +92,36 @@ export const UserProfileForm = () => {
...
+            variables: { data },
+            onCompleted: () => {
+                signout()
+            },
+            onError: (e) => {
+                console.log(e)
+                console.error(e.friendlyDescription)
+                throw e
+            },
+            OnCompletedMsg: null,
```
Alllex202, line 119: why is it null?

apps/condo/domains/user/components/UserProfileForm.tsx
```diff
@@ -167,6 +206,16 @@ export const UserProfileForm = () => {
                                                 >
                                                     {CancelLabel}
                                                 </Button>,
+                                                <DeleteButtonWithConfirmModal
+                                                    key='deleteProfile'
```
Alllex202, line 210:
```suggestion
                                                    key='deleteUser'
```

apps/condo/lang/ru/ru.json
tolmachev21: Hmmm, I think, that name of button "Delete Profile", says users about delete only his profile user. However, I can easily fix this.

Alllex202: You can do it as you see fit :)

apps/condo/domains/user/components/UserProfileForm.tsx
nomerdvadcatpyat: @Alllex202, Why do you think that?

Alllex202: > Why do you think that?

If we don't save the name, it will look bad in the interface.

tolmachev21: Wow, I didn't even think about it

pahaz:
It's better also to ask user to type his name or phone to chnfirm this action! Or it's better to confirm it by write his password! Or confirm by phone code!

apps/condo/domains/user/components/UserProfileForm.tsx
```diff
@@ -83,6 +92,30 @@ export const UserProfileForm = () => {
...
+        const data = { user: { id: user.id }, saveName: true, ...profileExtraData }
+
+        return runMutation({
+            mutation: resetUserMutation,
+            variables: { data },
+            onCompleted: () => {
+                signout()
+            },
+            onError: (e) => {
+                console.log(e)
```
pahaz, line 110: some debug detected!

PR #5064, opened 2024-08-06
fix(condo): DOMA-9882 test for sending news to the right resident

tolmachev21:
Test for the latest prodaction bug sending messages to wrong users

### DIFF
--- a/apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.spec.js
+++ b/apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.spec.js
@@ -306,14 +306,11 @@ describe('notifyResidentsAboutNewsItem', () => {
 
             // The message should not be sent to the second user with a different address, but the same unitName and unitType 
             await waitFor(async () => {
-                const messages1 = await Message.getAll(adminClient, messageWhere1)
-                const messages2 = await Message.getAll(adminClient, messageWhere2)
+                const messages1 = await Message.getOne(adminClient, messageWhere1)
+                const messages2 = await Message.getOne(adminClient, messageWhere2)
 
                 expect(messages1).toBeDefined()
-                expect(messages1).toHaveLength(1)
-
-                expect(messages2).toBeDefined()
-                expect(messages2).toHaveLength(0)
+                expect(messages2).toBeUndefined()
             })
         })
 
--- a/apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.spec.js
+++ b/apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.spec.js
@@ -243,6 +243,80 @@ describe('notifyResidentsAboutNewsItem', () => {
             })
         })
 
+        test('newsItem message is sent to the desired property and unitName', async () => {
+            const residentClient1 = await makeClientWithResidentUser()
+            const residentClient2 = await makeClientWithResidentUser()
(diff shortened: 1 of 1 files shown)


apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.spec.js
```diff
@@ -243,6 +243,80 @@ describe('notifyResidentsAboutNewsItem', () => {
...
+            const messageWhere2 = { user: { id: residentClient2.user.id }, type: NEWS_ITEM_COMMON_MESSAGE_TYPE }
+
+            // Publish NewsItem to make it sendable
+            await publishTestNewsItem(adminClient, newsItem.id)
+            await notifyResidentsAboutNewsItem(newsItem.id)
+
+            // The message should not be sent to the second user with a different address, but the same unitName and unitType 
+            await waitFor(async () => {
+                const messages1 = await Message.getAll(adminClient, messageWhere1)
+                const messages2 = await Message.getAll(adminClient, messageWhere2)
```
nomerdvadcatpyat, line 310: you can use `.getOne` here and check `expect(message1).toBeDefined()` and `expect(message2).toBeUndefined()` below

nomerdvadcatpyat (approved):

apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.spec.js
tolmachev21: Okey, I am use getOne, but message2 receives [], which will not pass the test toBeUndefinded()

tolmachev21: Ow, method getAll return only object, not array. Okey, I will learn it, thanks^

Alllex202 (approved):

apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.spec.js
```diff
@@ -243,6 +243,77 @@ describe('notifyResidentsAboutNewsItem', () => {
...
+            await publishTestNewsItem(adminClient, newsItem.id)
+            await notifyResidentsAboutNewsItem(newsItem.id)
+
+            // The message should not be sent to the second user with a different address, but the same unitName and unitType 
+            await waitFor(async () => {
+                const messages1 = await Message.getOne(adminClient, messageWhere1)
+                const messages2 = await Message.getOne(adminClient, messageWhere2)
+
+                expect(messages1).toBeDefined()
+                expect(messages2).toBeUndefined()
```
toplenboren, line 313: There might be a case when:

1. messages1 is successfully sent (it is defined)
2. messages2 is GOING to be sent, but the sending process just havent started yet (it is undefined)

In this case your test is going to be successful

toplenboren: But in essesnse it should fail

Alllex202: Yes, it is

Do you have any idea how to solve this case?

Alllex202: As a suggestion, we can in the function of sending news count the number of created messages and return this value. And check this count in test

PR #5067, opened 2024-08-06
fix(condo): DOMA-9881 test to prevent notifications with three dots

tolmachev21:
There was a breakdown in this task [DOMA-9829](https://jira.in.doma.ai/browse/DOMA-9829) 
Now there is a test that prevents this behavior.


apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.spec.js
```diff
@@ -215,6 +220,14 @@ describe('notifyResidentsAboutNewsItem', () => {
                 expect(message1).toBeDefined()
                 expect(message1.id).toMatch(UUID_RE)
 
+                // Check that push notifications do not contain dots instead of title and body
```
toplenboren, line 223: Best to just check content for notifications,

apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.spec.js
```diff
@@ -215,6 +220,14 @@ describe('notifyResidentsAboutNewsItem', () => {
...
                 expect(message1.id).toMatch(UUID_RE)
 
+                // Check content for notifications
+                expect(message1).toEqual(expect.objectContaining({
+                    meta: expect.objectContaining({
+                        title: expect.not.stringMatching(THREE_DOT_FOR_TEST_REGEXP),
+                        body: expect.not.stringMatching(THREE_DOT_FOR_TEST_REGEXP),
+                    }),
+                }))
+
```
Alllex202, line 230: Maybe it would be better to declare the variables `MESSAGE_TITLE_MAX_LEN` and `MESSAGE_BODY_MAX_LEN` in this file with tests too?
Thus we can't update the text length without changing the tests

Alllex202: And in this case you don't have to add this strange check

toplenboren: I'd have rewritten the check too.. You need to chekck pattern like this [a-z]*... and maybe use startsWith, it will check that 

1. You actually are sending some content in message
2. Message is exactly what you'd expect (original title and description starts with shortened message w/o ...)

apps/condo/domains/news/tasks/notifyResidentsAboutNewsItem.spec.js
```diff
@@ -227,8 +233,9 @@ describe('notifyResidentsAboutNewsItem', () => {
                         })],
                     }),
                     meta: expect.objectContaining({
-                        title: truncate(newsItemTitle, { length: MESSAGE_TITLE_MAX_LEN, separator: ' ', omission: '...' }),
-                        body: truncate(newsItemBody, { length: MESSAGE_BODY_MAX_LEN, separator: ' ', omission: '...' }),
+                        // Check content for notifications
+                        title: expect.stringMatching(CHECK_CONTENT_MESSAGE_REGEXP),
+                        body: expect.stringMatching(CHECK_CONTENT_MESSAGE_REGEXP),
```
Alllex202, line 238: I think this is a bad test. If I change `MESSAGE_TITLE_MAX_LEN` or `MESSAGE_BODY_MAX_LEN` to 1, it will pass. If I change it to 1000, it will fail

I would like the test to not let me write bad code.
It would be good if the file with tests had its own variables `MESSAGE_TITLE_MAX_LEN` and `MESSAGE_BODY_MAX_LEN` declared. This way we can be sure that nothing is broken in the sending logic

Alllex202 (approved):

PR #5070, opened 2024-08-07
chore(condo): DOMA-9803 (submodule) increase memory, decrease pods count

AleX83Xpert:
It's needed to have an ability to create heap dump to investigate memory leaks issues


SavelevMatthew (changes requested):
no custom acquiring today, sorry

SavelevMatthew:
Contains some strange tasks. Reverted and merged here: https://github.com/open-condo-software/condo/pull/5072

PR #5073, opened 2024-08-07
feat(condo): INRA-397 move history record to external db

sitozzz:


toplenboren (approved):

pahaz (changes requested):
It look like to early for such changes. It's not backward safe at the moment

PR #5083, opened 2024-08-09
refactor(miniapp): DOMA-9911 updated the miniapp according to the new template

nomerdvadcatpyat:
Just remove old `miniapp` and generate it with `createapp` again 

image


Alllex202:
7 thousand lines just to show one label, ok 😂

Do you really need to update the migration?

nomerdvadcatpyat:
> Do you really need to update the migration?

Probably no. I will delete this change

.sonarcloud.properties
```diff
@@ -1,2 +1,2 @@
 sonar.cpd.exclusions=**/*.js,**/*.ts,**/OldBaseNewsForm.tsx
-sonar.exclusions=**/*.test.js,**/*.spec.js
```
Alllex202, line 2: Are you sure you should do this?

nomerdvadcatpyat: Yes, i need to disable duplicates check in miniapp app because it's very similar as miniapp template

Alllex202: But this will disable all checks.. Оr not?

Is it possible to skip just duplicates checks?

Alllex202: Try using `sonar.cpd.exclusions `
https://docs.sonarsource.com/sonarcloud/advanced-setup/analysis-scope/#scope-of-duplication-detection

Alllex202 (approved):

toplenboren (approved):

abshnko (approved):

pahaz (approved):

apps/miniapp/middlewares/oidc.js
```diff
@@ -0,0 +1,191 @@
...
+                await req.session.save()
+
+                return res.redirect(nextUrl)
+            } catch (err) {
+                await req.session.destroy()
+                return next(err)
+            }
+        })
+
+        app.set('trust proxy', true)
```
pahaz, line 183: it should be inside prepareKeystoneApp

PR #5088, opened 2024-08-12
fix(condo): INFRA-424 notify about tooManyReturned exponentially

pahaz:


apps/condo/domains/common/utils/serverSchema/index.js
```diff
@@ -40,26 +58,22 @@ class GqlWithKnexLoadList {
...
-                    functionName: 'GqlWithKnexLoadList.load',
-                    schemaName: this.listKey,
-                    data: {
-                        limit: 1000,
-                        GqlWithKnexLoadListArgs: { singleRelations: this.singleRelations, multipleRelations: this.multipleRelations, where: this.where, fields: this.fields },
-                    },
-                })
-                haveNotifiedAboutTooManyObjs = true
-            }
+            logTooManyReturnedIfRequired(tooManyReturnedMemory, all, {
```
toplenboren, line 69: You should also introduce this change to `loadListByChunks` it is in the same file

SavelevMatthew (approved):

apps/condo/domains/common/utils/serverSchema/index.js
```diff
@@ -21,6 +21,24 @@ const logger = getLogger('common/utils/serverSchema.js')
 // Tested on tickets export for 24755 tickets: without knex  71220.902ms, with knex: 16094.841ms )
 // TODO(zuch): find out how to make 1 request for 1. and 2.
 
+function logTooManyReturnedIfRequired (tooManyReturnedMemory, allObjects, { limit, functionName, schemaName, data }) {
+    if (tooManyReturnedMemory.length < 5) return  // trying to notify only if have not notidied yet. Max 4 times
```
toplenboren, line 25:
```suggestion
    if (tooManyReturnedMemory.length < 5) return  // trying to notify only if have not notified yet. Max 4 times
```

apps/condo/domains/common/utils/serverSchema/index.js
```diff
@@ -21,6 +21,24 @@ const logger = getLogger('common/utils/serverSchema.js')
...
+
+    let realLimit = limit  // notify on: limit, limix * 10, limit * 100, limit * 1000
+    for (let i = 0; i < tooManyReturnedMemory.length; i++) realLimit *= 10
```
toplenboren, line 28: I think it would be better to just count these limits on file load and then use here, like

const TOO_MANY_RETURNED_LIMITS = [10, 10*10, 10*100, 10*1000] 

and then use in logic using your function

This way we wont have to count these on every function call

apps/condo/domains/common/utils/serverSchema/index.js
```diff
@@ -40,26 +58,22 @@ class GqlWithKnexLoadList {
         let skip = 0
         let newchunk = []
         let all = []
-        let haveNotifiedAboutTooManyObjs = false
+        let tooManyReturnedMemory = []
```
toplenboren, line 61: It seems confusing to me

Maybe: tooManyReturnedLogsSentCounter? And why can't we just store number, like

1. If we have logged only 1 time, this var will be 1
2. If we have logged 2 times, this var will be 2 

etc

toplenboren (approved):

apps/condo/domains/common/utils/serverSchema/index.js
```diff
@@ -21,6 +22,24 @@ const logger = getLogger('common/utils/serverSchema.js')
...
 // TODO(zuch): find out how to make 1 request for 1. and 2.
 
+function logTooManyReturnedIfRequired (tooManyReturnedLimitCounters, allObjects, { functionName, schemaName, data }) {
+    if (!Array.isArray(tooManyReturnedLimitCounters)) throw new Error('logTooManyReturned: wrong argument type')
+    if (tooManyReturnedLimitCounters.length <= 0) return  // trying to notify only if have any counter
+
+    const realLimit = tooManyReturnedLimitCounters[0]
+
+    if (allObjects && Array.isArray(allObjects) && allObjects.length > realLimit) {
+        logger.warn({
```
sitozzz, line 32: Where is request id? There is no option now to determine request source

apps/condo/domains/common/utils/serverSchema/index.js
```diff
@@ -21,6 +22,24 @@ const logger = getLogger('common/utils/serverSchema.js')
 // Tested on tickets export for 24755 tickets: without knex  71220.902ms, with knex: 16094.841ms )
 // TODO(zuch): find out how to make 1 request for 1. and 2.
 
+function logTooManyReturnedIfRequired (tooManyReturnedLimitCounters, allObjects, { functionName, schemaName, data }) {
+    if (!Array.isArray(tooManyReturnedLimitCounters)) throw new Error('logTooManyReturned: wrong argument type')
+    if (tooManyReturnedLimitCounters.length <= 0) return  // trying to notify only if have any counter
+
+    const realLimit = tooManyReturnedLimitCounters[0]
+
```
sitozzz, line 30: You can use execution context for that purpose

PR #5118, opened 2024-08-19
feat(eps): DOMA-9975 EPS does not fill up implicitFee field for acqui…

ekabardinsky:
…ringPaymentLog model


SavelevMatthew (commented):
Why url is empty string and not null / undefined?

AleX83Xpert (commented):
Are you sure you need always first element of `implicitFeeDistributionSchema` field?
image

ekabardinsky:
@AleX83Xpert Yes, since eps aic always contains a single element for implicitFeeDistributionSchema

AleX83Xpert (approved):

abshnko (approved):

PR #5124, opened 2024-08-20
feat(condo): DOMA-9940 meter reading billing status

AleX83Xpert:


abshnko (approved):

dkoviazin (approved):

apps/condo/domains/meter/schema/MeterReading.js
```diff
@@ -110,6 +115,27 @@ const MeterReading = new GQLListSchema('MeterReading', {
             kmigratorOptions: { null: false, on_delete: 'models.PROTECT' },
         },
 
+        billingStatus: {
```
pahaz, line 118: It looks like we need some access check here?

pahaz (approved):

toplenboren (approved):

apps/condo/domains/meter/schema/MeterReading.js
AleX83Xpert: Maybe we should allow access only for service users. ¯\\\_(ツ)_/¯

PR #5126, opened 2024-08-20
fix(condo): DOMA-10006 add throttling

toplenboren:


pahaz (approved):

apps/condo/domains/news/components/RecipientCounter.tsx
```diff
@@ -237,11 +238,13 @@ export const RecipientCounter: React.FC<RecipientCounterProps> = ({ newsItemScop
         },
     )
 
+    const throttledGetCounters = useCallback(throttle(getCounters, 1500), [throttle])
```
SavelevMatthew, line 241: Why `getCounters` not in deps array? To avoid callback rerender?

SavelevMatthew: btw, throttle is imported function, its not necessary to put it in reps

apps/condo/domains/news/components/RecipientCounter.tsx
```diff
@@ -266,6 +269,9 @@ export const RecipientCounter: React.FC<RecipientCounterProps> = ({ newsItemScop
 
     const isLoadingCounters = isCountersLoading || !counters
 
+    console.log('Recipient counter', { isLoadingCounters, isCountersLoading, counters })
```
SavelevMatthew, line 272: ?

apps/condo/domains/news/components/RecipientCounter.tsx
SavelevMatthew: And why 1500? I'm almost sure we can reproduce error consistently with this value

PR #5133, opened 2024-08-22
feat(condo): DOMA-10026 Support #ADR35 | X-Target header for load balancing

toplenboren:
Hey!

Speaking of #ADR35

I propose this API:

1. Declare an environment variable like this:

```
FETCH_X_TARGET_CONFIG = { _default: 'priority', 'v1.condo.ai':'lalala' }
```

2. All requests to all services will have `x-target` header set to 'priority' All requests to v1.condo.ai will have `x-target` header set to `lalala`

What do you think?


nomerdvadcatpyat (approved):

packages/keystone/fetch.js
```diff
@@ -15,6 +15,12 @@ const logger = getLogger('fetch')
 const FETCH_COUNT_METRIC_NAME = 'fetch.count'
 const FETCH_TIME_METRIC_NAME = 'fetch.time'
 
+/**
+ * Should be: { [hostname: str]:[x-target: str] }
+ * @example { v1.condo.ai: 'priority' }
```
pahaz, line 20:
```suggestion
 * @example { v1.condo.ai: 'example-app' }
```

pahaz (approved):

SavelevMatthew (commented):
I guess we should discuss the proposal at tech. talks before it can be merged. Don't want to block PR right now, but it seems like this change is not reasonable without client-side utils

packages/keystone/fetch.js
```diff
@@ -53,7 +59,20 @@ async function fetchWithLogger (url, options, extraAttrs) {
...
-    const { skipTracingHeaders } = extraAttrs
+    if (!options.headers) {
+        options.headers = {}
+    }
+
+    const originalHeaders = pickBy(options.headers)
+
+    const { skipTracingHeaders, skipXTargetHeader } = extraAttrs
+
+    if (!skipXTargetHeader && !options.headers['X-Target']) {
```
SavelevMatthew, line 70: Why do you need skip? I cannot imagine case, when you don't follow the convention

toplenboren: When you are sending requests to dadata or other extrenal services

packages/keystone/fetch.js
```diff
@@ -15,6 +15,12 @@ const logger = getLogger('fetch')
 const FETCH_COUNT_METRIC_NAME = 'fetch.count'
 const FETCH_TIME_METRIC_NAME = 'fetch.time'
 
+/**
+ * Should be: { [hostname: str]:[x-target: str] }
+ * @example { v1.condo.ai: 'example-app' }
+ */
+const FETCH_X_TARGET_CONFIG = JSON.parse(conf.FETCH_X_TARGET_CONFIG || '{}')
```
SavelevMatthew, line 22: It should probably the same for all domains, or have default option

SavelevMatthew: Motivation is that I want to go to all internal services (even for newly created one) with specific header by default

toplenboren:
Hey, @SavelevMatthew I just added `_default` option now, can you please re-review this PR ^^

packages/keystone/fetch.js
toplenboren: Agreed, wait a day or two

packages/keystone/fetch.js
```diff
@@ -15,6 +15,15 @@ const logger = getLogger('fetch')
 const FETCH_COUNT_METRIC_NAME = 'fetch.count'
 const FETCH_TIME_METRIC_NAME = 'fetch.time'
 
+/**
+ * Should be: { [hostname: str]:[x-target: str] }
+ *
+ * @example { _default: 'group0', v1.condo.ai: 'group1' }
+ *
+ * All requests will have X-Target=group0. Requests to v1.condo.aoo will have X-Target=group1
```
SavelevMatthew, line 23:
```suggestion
 * All requests will have X-Target=group0. Requests to v1.condo.ai will have X-Target=group1
```

SavelevMatthew (approved):

abshnko (approved):

PR #5151, opened 2024-08-28
fix(condo): DOMA-9475 allow update unitName on Resident

YEgorLu:
Problem: 
- Can not change wrong unitName
- When you add resident, delete it and add new one with same unitName, but in different case, old resident reactivates and keeps old unitName

Solution: 
Allow updates on unitName

### DIFF
--- a/apps/condo/domains/resident/schema/RegisterResidentService.test.js
+++ b/apps/condo/domains/resident/schema/RegisterResidentService.test.js
@@ -431,6 +431,24 @@ describe('RegisterResidentService', () => {
         expect(restoredResident.property.id).toEqual(property.id)
     })
 
+
+    it('restore deleted Resident for the same address and unitName updates unitName', async () => {
+        const userClient = await makeClientWithResidentAccessAndProperty()
+        const testUnitName = faker.random.numeric(3) + faker.random.alpha(5).toUpperCase()
+
+        const [resident, attrs] = await registerResidentByTestClient(userClient, {
+            unitName: testUnitName,
+        })
+        const [softDeletedResident] = await Resident.softDelete(userClient, resident.id)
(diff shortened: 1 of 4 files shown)


YEgorLu:
> Seems legit, but I guess we should decide first, are unitNames sensitive to registry or not. If so, this PR should not be merged, so I'll block it for now. Otherwise LGTM

The problem is, that user can add address like "Washington street, 15A" (mobile app sends case sensitive request), then delete it, and then add address "Washington street, 15a". In that case soft deleted resident will be restored with old unitName and still look like "Washington street, 15A", but we (users) want "Washington street, 15a". With address in wrong case users can't transmit meter readings.

toplenboren:
UnitNames are not case sensitive

apps/condo/domains/resident/schema/RegisterResidentService.test.js
```diff
@@ -431,6 +431,24 @@ describe('RegisterResidentService', () => {
         expect(restoredResident.property.id).toEqual(property.id)
     })
 
+
+    it('restore deleted Resident for the same address and unitName updates unitName', async () => {
```
toplenboren, line 435: Isnt it DOMA-9474?

YEgorLu: They look like duplicates, highlighting ability to change unitName to be able to transmit the readings, but 9475 has more info

toplenboren: They are not duplicates, they are different 

74 is about fixing deletion/restore
75 is about fixing Resident.organizationFeatures

toplenboren (approved):

PR #5163, opened 2024-09-02
feat(condo): DOMA-10042 handler for suggesting providers

YEgorLu:
Added handler for suggesting providers by tin or name


dkoviazin (commented):
You need to use Service Provider instead of provider

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
```diff
@@ -0,0 +1,97 @@
...
+        {
+            access: access.canSuggestProvider,
+            schema: 'suggestProvider (data: SuggestProviderInput!): [SuggestProviderOutput]',
+            resolver: async (parent, args, context = {}) => {
+                if (context.authedItem.type === RESIDENT) {
+                    await checkLimits(context.authedItem.id)
+                }
+
+                const { data: { tinOrName } } = args
+                if (!tinOrName) {
```
pahaz, line 74: You must check the length! At least 3 symbols

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
```diff
@@ -0,0 +1,97 @@
...
+                }
+
+                const serviceProviders = await Organization.getAll(context, {
+                    OR: [
+                        { tin_starts_with: tinOrName.trim() },
+                        { name_contains: tinOrName.trim() },
+                    ],
+                    deletedAt: null,
+                    type: SERVICE_PROVIDER_TYPE,
+                }, { sortBy: ['name_ASC'] }) // 100 default limit
```
pahaz, line 85: It looks like full-scan. We don't have a name index.

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
```diff
@@ -0,0 +1,97 @@
...
+        {
+            access: true,
+            type: 'type SuggestProviderOutput { tin: String!, name: String! }',
+        },
+    ],
+    
+    queries: [
+        {
+            access: access.canSuggestProvider,
+            schema: 'suggestProvider (data: SuggestProviderInput!): [SuggestProviderOutput]',
```
pahaz, line 67: It looks like we need to move this inside address-service. Architecture meet required

apps/condo/domains/billing/utils/testSchema/mixins/organization.js
```diff
@@ -21,6 +22,11 @@ const OrganizationTestMixin = {
         await createTestOrganizationEmployee(this.clients.admin, this.organization, this.clients.employee[type].user, role)
     },
 
+    async updateOrganization (updateInput) {
+        const [updatedOrganization] = await updateTestOrganization(this.clients.admin, this.organization.id, updateInput)
```
SavelevMatthew, line 26: Why everything is created / updated from admin perspective? Its not close to production environment, where there's no admins at all

SavelevMatthew: @dkoviazin ^

apps/condo/domains/resident/constants/constants.js
```diff
@@ -168,6 +168,9 @@ const MAX_RESIDENT_DISCOVER_CONSUMERS_BY_WINDOW_SEC = 10
 const RESIDENT_FIND_ORGANIZATIONS_WINDOW_SEC = 24 * 60 * 60 // seconds
 const MAX_RESIDENT_FIND_ORGANIZATIONS_BY_WINDOW_SEC = 100
 
+const RESIDENT_SUGGEST_SERVICE_PROVIDER_WINDOW_SEC = 24 * 60 * 60 // seconds
```
SavelevMatthew, line 171: Please be more informative with time comments, so we can avoid extra calculation while reading

```suggestion
const RESIDENT_SUGGEST_SERVICE_PROVIDER_WINDOW_SEC = 24 * 60 * 60 // 1 day in seconds
```

apps/condo/domains/resident/constants/constants.js
```diff
@@ -168,6 +168,9 @@ const MAX_RESIDENT_DISCOVER_CONSUMERS_BY_WINDOW_SEC = 10
...
+const MAX_RESIDENT_SUGGEST_SERVICE_PROVIDER_BY_WINDOW_SEC = 100
```
SavelevMatthew, line 172:
```suggestion
const MAX_RESIDENT_SUGGEST_SERVICE_PROVIDER_CALLS_PER_WINDOW = 100
```

apps/condo/domains/resident/gql.js
```diff
@@ -72,6 +72,12 @@ const REGISTER_RESIDENT_SERVICE_CONSUMERS_MUTATION = gql`
     }
 `
 
+const SUGGEST_PROVIDER_QUERY = gql`
+    query suggestProviderByTin ($data: SuggestProviderInput!) {
+        result: suggestProvider(data: $data) { tin, name }
```
SavelevMatthew, line 77: We usually avoid commas in gql types as far as I remember, so...
```suggestion
        result: suggestProvider(data: $data) { tin name }
```

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
```diff
@@ -0,0 +1,97 @@
...
+const access = require('@condo/domains/resident/access/SuggestServiceProviderService')
+const { MAX_RESIDENT_SUGGEST_SERVICE_PROVIDER_BY_WINDOW_SEC, RESIDENT_SUGGEST_SERVICE_PROVIDER_WINDOW_SEC } = require('@condo/domains/resident/constants')
+const { RESIDENT } = require('@condo/domains/user/constants/common')
+const { RedisGuard } = require('@condo/domains/user/utils/serverSchema/guards')
+
+const redisGuard = new RedisGuard()
+
+const checkLimits = async (uniqueField) => {
+    await redisGuard.checkCustomLimitCounters(
+        `suggest-service-provider-by-tin-${uniqueField}`,
```
SavelevMatthew, line 23: Also strange naming convention, we use underscore for multi-words and `:` for parts delimiter

SavelevMatthew: Also note that:
1. redis guard by default has its own prefix guard_counter:
2. unique field must be following model name and be as close to beginning as possible

SavelevMatthew, line 23: so the correct redis keys should look something like this:
`guard_counter:user:${id}:suggest_service_provider`

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
```diff
@@ -0,0 +1,97 @@
...
+    const organizationsForSuggest = new Set(
+        acquiringIntegrationContexts.map(context => context.organization.id)
+            .concat(meterResourceOwners.map(owner => owner.organization.id))
+    )
+
+    return organizations.filter(organization => organizationsForSuggest.has(organization.id))
+}
+
+
+const SuggestServiceProviderService = new GQLCustomSchema('SuggestServiceProviderByTinService', {
```
SavelevMatthew, line 52: Inconsistent naming. In half of your names `ByTin` is present, in other half - not

SavelevMatthew: You should remove byTin suffix, since search logic can be changed in the future

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
```diff
@@ -0,0 +1,97 @@
...
+    types: [
+        {
+            access: true,
+            type: 'input SuggestProviderInput { tinOrName: String! }',
```
SavelevMatthew, line 56:
```suggestion
            type: 'input SuggestProviderInput { search: String! }',
```

SavelevMatthew: Prepare it for expansions :))
tinOrName -> tinOrNameOrColor -> tinOrNameOrColorOrSomethingElse

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
```diff
@@ -0,0 +1,97 @@
...
+
+                const { data: { tinOrName } } = args
+                if (!tinOrName) {
+                    return []
+                }
+
+                const serviceProviders = await Organization.getAll(context, {
+                    OR: [
+                        { tin_starts_with: tinOrName.trim() },
+                        { name_contains: tinOrName.trim() },
```
SavelevMatthew, line 81: Should it be case insensitive, probably so

SavelevMatthew:
```suggestion
                        { name_contains_i: tinOrName.trim() },
```

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
SavelevMatthew, line 85: Didn't get a comment about 100? where is the limit?

Please pass `first: 100` explicitly. Don't depend on any other logic

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
```diff
@@ -0,0 +1,97 @@
...
+                }
+
+                const { data: { tinOrName } } = args
+                if (!tinOrName) {
+                    return []
+                }
+
+                const serviceProviders = await Organization.getAll(context, {
+                    OR: [
+                        { tin_starts_with: tinOrName.trim() },
```
SavelevMatthew, line 80: Why different searching behaviour? One is `starts_with`, other is `contains`?

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
SavelevMatthew: You should also sort results by tin instead of name

SavelevMatthew: We also discussed feature, that organisations with different tin, but the same name must be grouped together.

Lets take 2 organizations:
`{ name: 'my company', tin: '123' }` and `{ name: 'company', tin: '123' }`

By typing "my" in search I should get first one, BUT if it has other aliases (different name, same TIN) it should also be returned.

So mobile clients can group it when register service consumers

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
```diff
@@ -0,0 +1,97 @@
...
+ * Generated by `createservice organization.SuggestServiceProviderByTinService --type queries`
+ */
+
+const { pick, uniqBy } = require('lodash')
+
+const { GQLCustomSchema } = require('@open-condo/keystone/schema')
+
+const { CONTEXT_FINISHED_STATUS } = require('@condo/domains/acquiring/constants/context')
+const { AcquiringIntegrationContext } = require('@condo/domains/acquiring/utils/serverSchema')
+const { MeterResourceOwner } = require('@condo/domains/meter/utils/serverSchema')
```
SavelevMatthew, line 11: Please avoid using server utils, when possible. They're doing full gql request emulation, but skipping access.

So your request becomes very heavy by all relations from domain/X/gql.js

SavelevMatthew: Use `find` / `getByCondition` and so on

apps/condo/domains/resident/schema/SuggestServiceProviderService.test.js
```diff
@@ -0,0 +1,241 @@
...
+
+    let utils
+
+    beforeAll(async () => {
+        utils = new TestUtils([ResidentTestMixin, MeterTestMixin])
+        await utils.init()
+        await utils.updateOrganization({ type: SERVICE_PROVIDER_TYPE })
+    })
+
+    describe('Permissions check', () => {
```
SavelevMatthew, line 65: Please refer other domains for structuring the tests, flat list is hard to navigate

SavelevMatthew:
```
describe('<Util name>`, () => {
    beforeAll(async () => {
        // prepare all actors and other staff here
    })
    describe('Execute', () => {
        test('<A> can execute', async () => {
            // ...
        })
        test('<B> can execute', async () => {
            // ...
        })
        test('<C> cannot execute', async () => {
            // ...
        })
    })
    describe('Logic', () => {
        describe('guards', () => {
            
        })
        // Other logic / return / behaviour test, grouped by common pattern / type
    })
})
```

SavelevMatthew (changes requested):
Overall looks OK, got some code-style issues, which can be easily resolved...

But I've also got some serious issues, like:
1. Ordering logic
2. Redis keys naming convention 
3. TIN duplicates display
4. Server utils usage

So I cannot pass it for now, but code itself is fine

apps/condo/domains/billing/utils/testSchema/mixins/organization.js
dkoviazin: Because, there is a limitation on editing some fields for organization. For example, tin
And we already have tests for updating Organization - so there is no need to test it again in utils

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
YEgorLu: Check other usages of checkCustomLimitCounters in domains folder. There they use `-` as delimiter and puts unique key at the end. Maybe pass this now, and refactor all at once later?

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
YEgorLu: We now discussed that it's better to stay as it is (returns only unique tins). In suggested case handler should return one record with tin: '123' and any name

Also i don't understand why should we sort results by tin. Analogues have name sort, it is more convenient  for eyes to search from results, that sorted alphabetically. Tin is more like random string\number, so what convenience can it give?

apps/condo/domains/billing/utils/testSchema/mixins/organization.js
SavelevMatthew: Yep, but mixin pattern is bad because of that

SavelevMatthew: in my head mixins is added on top of actor, so actor should call all by himself

dkoviazin: we need an utils to fast create Organization, Property, etc....
we use them only for that
then we test new mutations on the old way

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
dkoviazin: It will work only for organizations that exists only in our data base

SavelevMatthew (approved):

SavelevMatthew (changes requested):
Redis Keys must be changed

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
```diff
@@ -0,0 +1,102 @@
...
+    MAX_RESIDENT_SUGGEST_SERVICE_PROVIDER_CALLS_BY_WINDOW_SEC,
+    RESIDENT_SUGGEST_SERVICE_PROVIDER_WINDOW_IN_SEC,
+    MAX_SUGGEST_SERVICE_PROVIDER_ITEMS_COUNT,
+} = require('@condo/domains/resident/constants')
+const { RESIDENT } = require('@condo/domains/user/constants/common')
+const { RedisGuard } = require('@condo/domains/user/utils/serverSchema/guards')
+
+const redisGuard = new RedisGuard()
+
+const checkLimits = async (uniqueField) => {
```
SavelevMatthew, line 22:
```suggestion
const checkLimits = async (userId) => {
```

apps/condo/domains/resident/schema/SuggestServiceProviderService.js
```diff
@@ -0,0 +1,102 @@
...
+    await redisGuard.checkCustomLimitCounters(
+        `suggest-service-provider-by-tin-${uniqueField}`,
```
SavelevMatthew, line 24:
```suggestion
        `suggest_service_provider:user:${userId}`,
```

apps/condo/domains/resident/schema/SuggestServiceProviderService.test.js
```diff
@@ -0,0 +1,206 @@
...
+        describe('Duplicates', () => {
+
+            test('should merge organizations with same tin', async () => {
+                const utils2 = new TestUtils([AcquiringTestMixin])
+                await utils2.init()
+                await utils2.updateOrganization({
+                    tin: utils.organization.tin,
+                    type: SERVICE_PROVIDER_TYPE,
+                })
+                await utils2.updateAcquiringContext({ status: CONTEXT_FINISHED_STATUS })
```
dkoviazin, line 192: Context already in Finished status

apps/condo/domains/resident/schema/SuggestServiceProviderService.test.js
```diff
@@ -0,0 +1,206 @@
...
+    })
+
+    describe('Permissions check', () => {
+
+        describe('Can not execute', () => {
+
+            test('anonymous: can not execute', async () => {
+                await expectToThrowAuthenticationErrorToResult(async () => {
+                    await suggestServiceProviderByTestClient(utils.clients.anonymous, {
+                        search: String(generateTin(RUSSIAN_COUNTRY)),
```
dkoviazin, line 56: no need to use RUSSIAN_COUNTRY, you can pass any string, as we only check for access rights

dkoviazin (approved):

PR #5166, opened 2024-09-02
proof of work: fix(condo): DOMA-10099 change counters logic

toplenboren:
# Update

## We have changed the logic of counting recipients

– if resident unitName does not exist in property map – this resident will still be counted

## Why

As it was discussed in ADR-36, we should remove all loadListByChunks-like operations, here I switch from allItemsQuery for Residents table (which is quite big) to raw sql query (just like in analytics domain)

### Known issues: 
- Raw SQL query will not be fixed, If needed I will implement LRU cache decorator
- Should we handle unitTypes here? I'll ask around.
- Usage of `distinct(count` + lack of index on `unitName` table will still be quite expensive for large users, we should take a close look to the timeout errors

### DIFF
--- a/apps/condo/domains/news/schema/GetNewsItemsRecipientsCountersService.js
+++ b/apps/condo/domains/news/schema/GetNewsItemsRecipientsCountersService.js
@@ -3,16 +3,28 @@
  */
 const { filter, pick, uniq, uniqBy } = require('lodash')
 
-const { GQLCustomSchema, allItemsQueryByChunks } = require('@open-condo/keystone/schema')
+const { GQLError, GQLErrorCode: { BAD_USER_INPUT } } = require('@open-condo/keystone/errors')
+const { GQLCustomSchema, allItemsQueryByChunks, getById } = require('@open-condo/keystone/schema')
 
 const access = require('@condo/domains/news/access/GetNewsItemsRecipientsCountersService')
 const {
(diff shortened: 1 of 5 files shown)


pahaz (approved):

dkoviazin (approved):

apps/condo/domains/news/schema/GetNewsItemsRecipientsCountersService.js
```diff
@@ -57,24 +56,28 @@ const GetNewsItemsRecipientsCountersService = new GQLCustomSchema('GetNewsItemsR
                             propertiesCount += chunk.length
                             for (const property of chunk) {
                                 unitsCount += (property.unitsCount + property.uninhabitedUnitsCount)
-                                orConditions.push(...queryConditionsByUnits(property))
+
+                                if (!unitsByProperty[property.id]) {
+                                    unitsByProperty[property.id] = []
+                                }
+
+                                unitsByProperty[property.id] = [...unitsByProperty[property.id], ...getUnitsFromProperty(property).map(u => u.unitName)]
```
SavelevMatthew, line 64: unitType?

toplenboren: Just added that + todo about addressKey

pahaz (approved):

PR #5184, opened 2024-09-04
fix(ci): DOMA-10147 add Open README.md test

pahaz:


toplenboren (approved):

.github/workflows/opensource.yml
```diff
@@ -0,0 +1,45 @@
...
+    types:
+      - opened
+      - reopened
+      - synchronize
+    branches:
+      - main
+  workflow_dispatch:
+
+
+jobs:
```
SavelevMatthew, line 17: You should probably add condition, so this job is only runs on open-condo-platform...

Seems useless to test in condo PR state of main of open-condo-platform

PR #5191, opened 2024-09-05
Feat(condo): DOMA-10067 normalizing unit name

YEgorLu:
added function for unit name normalization and used it in allResidentBillingReceipts.

Problem in allResidentBillingReceipts was that user can have unit name like '12a', and billing receipt says '12A' and we needed way to say that they are equal

There are many cases when users describe unitName strangely, tests may not have covered all of them, so the plan is to extend normalization in the future if needed

Added slugify package to make translation of symbols from other languages to english more convenient.


packages/clients/address-service-client/utils/normalizeUnitName.js
SavelevMatthew, line 1: It's common practice to use functions instead of anonymous arrow function when making utils

packages/clients/address-service-client/utils/normalizeUnitName.spec.js
```diff
@@ -0,0 +1,139 @@
...
+        ['770 М', '770-m'],
+        ['770-M', '770-m'],
+        ['770-m', '770-m'],
+        ['Особняк Оксфорд', 'osobnyak-oksford'],
+        ['к', 'k'],
+        ['нет', 'net'],
+        ['0-1', '0-1'],
+        ['3-8-13\n', '3-8-13'],
+        ['050-2/1', '050-2-1'],
+        ['кл.кл. 3 (19Б кв. 14)', '3-19-b-14'],
```
SavelevMatthew, line 60: where is `kl`?

SavelevMatthew: Who is responsible for list of abbreviations?

SavelevMatthew (changes requested):
This feature needs to be discussed first, I've got a lot of questions related to the decisions you've and @dkoviazin  made

packages/clients/address-service-client/utils/normalizeUnitName.js
pahaz: it's easier to debug and understan the traces

packages/clients/package.json
```diff
@@ -9,8 +9,10 @@
         "clients"
     ],
     "dependencies": {
+        "@open-condo/cli": "workspace:^",
         "@open-condo/config": "workspace:^",
-        "dotenv": "^16.3.1"
+        "dotenv": "^16.3.1",
```
pahaz, line 14: Why? you should use `@open-condo/config`

dkoviazin: old dependancies was "dotenv": "^16.3.1" 
current PR only adds comma

PR #5193, opened 2024-09-05
feat(condo): add ux property gate before news item is created

toplenboren:
## Problem

We cannot create news before we create at least one property

## Solution

Redirect user to property creation screen, and after creating of property redirect them back to news page

## 1

image

## 2 

image

## 3

image


apps/condo/domains/property/components/PropertyForm/UpdatePropertyForm.tsx
```diff
@@ -7,27 +7,34 @@ import { useOrganization } from '@open-condo/next/organization'
...
 import { Button } from '@condo/domains/common/components/Button'
 import { Loader } from '@condo/domains/common/components/Loader'
+import { isSafeUrl } from '@condo/domains/common/utils/url.utils'
 import { Property } from '@condo/domains/property/utils/clientSchema'
 
 import BasePropertyForm from '../BasePropertyForm'
 
 interface IUpdatePropertyForm {
     id: string
+    next?: string,
```
Alllex202, line 17: Why are you passing this value through several components? Why not just get it from URL?

apps/condo/domains/property/components/PropertyForm/UpdatePropertyForm.tsx
```diff
@@ -7,27 +7,34 @@ import { useOrganization } from '@open-condo/next/organization'
...
     const intl = useIntl()
     const ApplyChangesLabel = intl.formatMessage({ id: 'ApplyChanges' })
     const { push } = useRouter()
     const { organization } = useOrganization()
     const { refetch, obj: property, loading, error } = Property.useObject({ where: { id } })
     const initialValues = Property.convertToFormState(property)
-    const action = Property.useUpdate({}, (property) => push(`/property/${property.id}`))
+    const action = Property.useUpdate({}, async (property) => {
+        let redirectUrl = `/property/${property.id}`
+        if (next && isSafeUrl(next)) { redirectUrl = next }
```
Alllex202, line 34: You have already checked that the URL is safe

apps/condo/domains/property/components/PropertyForm/UpdatePropertyForm.tsx
```diff
@@ -7,27 +7,34 @@ import { useOrganization } from '@open-condo/next/organization'
 
 import { Button } from '@condo/domains/common/components/Button'
 import { Loader } from '@condo/domains/common/components/Loader'
+import { isSafeUrl } from '@condo/domains/common/utils/url.utils'
```
Alllex202, line 10: It seems that you can redirect the user to any site. This should be fixed (it can be done in another task, but please don't forget about it)

Alllex202:
1) How about the task number? :)

2) We should have the same and obvious behavior. Let's agree on what kind of behavior we should do everywhere and create a task for this.
There are currently 3 different behaviors for adding a house (Perhaps there is a similar story with other entities):
   - (new) redirect to the house creation page
   - link in the select to create a house
   - empty select

toplenboren:
> * How about the task number? :)
> * We should have the same and obvious behavior. Let's agree on what kind of behavior we should do everywhere and create a task for this.
>   There are currently 3 different behaviors for adding a house (Perhaps there is a similar story with other entities):
>   
>   * (new) redirect to the house creation page
>   * link in the select to create a house
>   * empty select

Yes, good catch :-) will do just that

PR #5196, opened 2024-09-06
feat(condo): DOMA-10067 basic unitName normalization

YEgorLu:
Made normalizeUnitName function in accordance with what was agreed upon

Placed it in billing domain because at first it will be used here, maybe move it somewhere when making plugin for unitNameKey

Didn't find good slugify package, others does not work with \ plan to remove CommonJS imports. One package, that we could use, has difficulties with shared charMap state. They have instruction on how to reset it's state, but it isn't working for me. So i copied package and added methods that allow resetting state and extending local-based state. These methods are needed in current task and allow to reuse package. I need different charMaps for different locales to address collisions between translations of symbols in different locales.


packages/slugify/slugify.d.ts
```diff
@@ -0,0 +1,21 @@
+declare const slugify: {
```
SavelevMatthew, line 1: This package must be written in Typescript, and then built properly if you want to publish it

SavelevMatthew: But I'm not sure we need this one

packages/slugify/slugify.js
```diff
@@ -0,0 +1,78 @@
+const _charMap = require('@open-condo/slugify/config/charMap')
+const _locales = require('@open-condo/slugify/config/locales')
+    
+
+;(function (name, root, factory) {
```
SavelevMatthew, line 5: ;(

YEgorLu: removed

packages/slugify/slugify.d.ts
YEgorLu: removed

apps/condo/domains/billing/utils/unitName.utils.js
```diff
@@ -0,0 +1,49 @@
...
+    replacement: DELIMITER_IN_RESULT,
+    locale,
+    strict: true,
+    lower: true,
+})
+
+const normalizeUnitName = (unitName, locale = conf.DEFAULT_LOCALE) => {
+    locale = locale.toLowerCase()
+
+    const steps = [
```
pahaz, line 35: looks like a const

apps/condo/domains/billing/utils/unitName.utils.spec.js
```diff
@@ -0,0 +1,126 @@
+const  { normalizeUnitName } = require('./unitName.utils')
+
+const TEST_CASES = {
```
pahaz, line 3: As for me it looks like we can remove all `-` in results (optional)

pahaz (approved):

apps/condo/domains/billing/utils/unitName.utils.js
YEgorLu: moved to global const

apps/condo/domains/billing/utils/unitName.utils.spec.js
YEgorLu: remove delimiter? can not understand why. If you mean normalizeUnitName fn result, some delimiter required to see, where is a word, where is a separate symbol

YEgorLu:
kept npm slugify package, removed dictionaries, changed one regex for sonarcloud

This solves one bug. Some more research required to solve problem more globally

dkoviazin (approved):

PR #5204, opened 2024-09-10
fix(condo): DOMA-10091 growthbook features keeping

YEgorLu:
We had case when growthbook's server was not responding. For that case TTL for featureflags was removed, we still update features once per minute, but not removing old one. 
Also extended refetch interval for /features api route, because client spamming fetch queries too much


packages/featureflags/featureToggleManager.js
```diff
@@ -101,6 +101,23 @@ class FeatureToggleManager {
...
     }
+
+    async _getFeaturesFromCache () {
+        const cachedFeatureFlags = await this.redis.get(this._redisKey)
+        if (cachedFeatureFlags) return JSON.parse(cachedFeatureFlags)
+        return null
+    }
+
+    async _loadFeaturesFromGrowthBook () {
+        const result = await fetch(this._url)
```
dkoviazin, line 112: You need to keep checks and wrap it to try-catch
as JSON can be broken

dkoviazin (approved):

PR #5213, opened 2024-09-10
hotfix(condo): use custom build script to control memory

pahaz:


bin/run-keystone-build.js
```diff
@@ -0,0 +1,48 @@
...
+
+const CWD = process.cwd()
+
+async function main () {
+    const index = path.resolve('./index.js')
+
+    console.log(chalk.gray.bold(`KEYSTONE 🚀 prepare keystone instance. index=${index}`))
+    const { keystone, apps, distDir } = await prepareKeystoneInstance(index, {
+        excludeApps: [
+            'HealthCheck', 'RequestCache', 'AdapterCache', 'VersioningMiddleware', 'OIDCMiddleware',
```
SavelevMatthew, line 25: We should not hard-code this ones, this means each new middleware should be added here.

That's bad. You should handle it gracefully

pahaz: It's just got speed up. Really you can skip new. It's just a small speed up.
And you really don't need to add all or new

bin/run-keystone-build.js
```diff
@@ -0,0 +1,48 @@
...
+
+const { prepareKeystoneInstance } = require('@open-condo/keystone/prepareKeystoneApp')
+
+const CWD = process.cwd()
+
+async function main () {
+    const index = path.resolve('./index.js')
+
+    console.log(chalk.gray.bold(`KEYSTONE 🚀 prepare keystone instance. index=${index}`))
+    const { keystone, apps, distDir } = await prepareKeystoneInstance(index, {
```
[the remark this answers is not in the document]
pahaz: No, it's a keystone logic. Each up should use it for their purpose. https://github.com/keystonejs/keystone-5/blob/main/packages/keystone/bin/commands/build.js#L43

PR #5240, opened 2024-09-16
fix(condo): INFRA-519 added dependency on "babel-loader" in packege.json

Alllex202:


apps/condo/package.json
```diff
@@ -191,6 +191,7 @@
     "@types/react-infinite-scroller": "^1.2.3",
     "antd-dayjs-webpack-plugin": "^1.0.6",
     "babel-jest": "^29.7.0",
+    "babel-loader": "^9.1.3",
```
pahaz, line 194: may be we need to use resolve?

Alllex202: I updated `yarn.lock`, now all versions within one major use one version of the package. I think this will be enough

toplenboren (approved):

PR #5242, opened 2024-09-16
fix(condo): DOMA-10216 increasing number of char in incident

tolmachev21:
In filed "Whats happend" in incident log there was a 500 character limit. Now there is a 1500 character limit. This change is in both callcenter and condo


Alllex202:
there's something wrong with your pr naming...

PR #5256, opened 2024-09-18
chore(global): INFRA-522 rollback replica test setup

sitozzz:


sitozzz:
That task will be done in the other PR

PR #5258, opened 2024-09-18
hotfix(locales): default locale doesn't work

pahaz:
Case in code: `exactReqLocale(…..) || conf.DEAFULT`

### DIFF
--- a/packages/locales/extractReqLocale.js
+++ b/packages/locales/extractReqLocale.js
@@ -14,7 +14,7 @@ const ACCEPT_LOCALES = ['ru', 'en']
  * @returns {string | null}
  */
 const extractReqLocale = (req) => {
-    if (!req) return null
+    if (!req) return undefined
     try {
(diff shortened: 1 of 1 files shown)


packages/locales/extractReqLocale.js
```diff
@@ -14,7 +14,7 @@ const ACCEPT_LOCALES = ['ru', 'en']
  * @returns {string | null}
  */
 const extractReqLocale = (req) => {
-    if (!req) return null
+    if (!req) return undefined
```
SavelevMatthew, line 17: That's a breaking change. Many apps uses it, and can compare `if (locale !== null)`, Also see return type above: `string | null`, undefined return us not declared

SavelevMatthew (changes requested):
> Case in code: `exactReqLocale(…..) || conf.DEAFULT`

null is falsey value, so `null || 'something` will return 'something'

image

So what exactly does not work?

PR #5265, opened 2024-09-19
feat(condo): DOMA-9648 add new templates in schema news item template

tolmachev21:
We want to have two new fields in the model NewsItemTemplate: "name": string, "category"?: select.

Migration also update old row, in which empty "name" will be equal to "title"


apps/condo/migrations/20240923115003-0426_newsitemtemplate_category_newsitemtemplate_name_and_more.js
```diff
@@ -0,0 +1,55 @@
...
+ALTER TABLE "NewsItemTemplate" ADD COLUMN "category" varchar(50) NULL;
+--
+-- Add field name to newsitemtemplate
+--
+ALTER TABLE "NewsItemTemplate" ADD COLUMN "name" text DEFAULT 'Temp' NOT NULL;
+ALTER TABLE "NewsItemTemplate" ALTER COLUMN "name" DROP DEFAULT;
+--
+-- [CUSTOM] Set "name" = "title" in old records
+--
+UPDATE "NewsItemTemplate" SET name = title WHERE name = 'Temp';
```
Alllex202, line 19: what's going on here? You just added a "name" field

tolmachev21: Because "temp" is the default temporary value for old records in the database. We want to set all old records to name = title

Alllex202: So why do you need this `temp` value?
For you, old templates are all those created before the release.
That is, all records are old for you and all need to be renamed. Isn't that right?

tolmachev21: I need to add another column to the database, which should be mandatory. Also for the old values ​​I need the value name = title, i.e. it is dynamic. It is possible to set such a value only after the static value ("Temp") is set, and then updated to a new one by overwriting the temporary value and setting a new, dynamic value.

Is this bad? Are there any other options? :-)

Alllex202: Ok, now I get it :)

Please write this in the comments to the migration

Alllex202 (commented):
Don't you need to update the logic on the pages?

Alllex202:
You should run `yarn maketypes` for update schema

apps/condo/domains/news/constants/newsCategory.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * Generated by `createschema news.OrganizationNewsItem 'organization:Relationship:Organization:CASCADE; title:Text; body:Text; type:Select:common,emergency'`
+ */
+
+const NEWS_CATEGORY_WATER = 'water'
+const NEWS_CATEGORY_HEATING = 'heating'
+const NEWS_CATEGORY_ELECTRICITY = 'electricity'
+const NEWS_CATEGORY_ALL_UTILITIES = 'allUtilities'
+const NEWS_CATEGORY_ELECTRICITY_AND_WATER = 'electricityAndWater'
+const NEWS_CATEGORY_LIGHT = 'light'
```
toplenboren, line 10: We do not need Light category – please check categories from News Sharing Greendom

apps/condo/domains/news/constants/newsCategory.js
```diff
@@ -0,0 +1,53 @@
+/**
+ * Generated by `createschema news.OrganizationNewsItem 'organization:Relationship:Organization:CASCADE; title:Text; body:Text; type:Select:common,emergency'`
+ */
+
+const NEWS_CATEGORY_WATER = 'water'
+const NEWS_CATEGORY_HEATING = 'heating'
+const NEWS_CATEGORY_ELECTRICITY = 'electricity'
+const NEWS_CATEGORY_ALL_UTILITIES = 'allUtilities'
+const NEWS_CATEGORY_ELECTRICITY_AND_WATER = 'electricityAndWater'
```
toplenboren, line 9: Lets make it simple

ELEVATORS_CATEGORY ...
ELECTRICITY_CATEGORY ...
ELECTRICITY_ELEVATORS_CATEGORY = 'ELECTRICTY_AND_ELEVATORS'
UTILITIES_CATEGORY = 'UTILITIES' // Электроэнергия, лифты, вода

apps/condo/domains/news/schema/NewsItemTemplate.js
```diff
@@ -27,6 +28,19 @@ const NewsItemTemplate = new GQLListSchema('NewsItemTemplate', {
             kmigratorOptions: { null: true, on_delete: 'models.CASCADE' },
         },
 
+        name: {
+            schemaDoc: 'Name of template the news item',
```
toplenboren, line 32: Add an example to all your schemaDocs. It should be obvious to user how to use this (User may be not engineer, but product manager)

apps/condo/migrations/20240923115003-0426_newsitemtemplate_category_newsitemtemplate_name_and_more.js
toplenboren: Can you not use an UPDATE? I'm personally OK with this (you also do not really need to filter on 'Temp') but maybe there are more elegant solution

toplenboren (changes requested):

apps/condo/migrations/20240923115003-0426_newsitemtemplate_category_newsitemtemplate_name_and_more.js
tolmachev21: I will leave this solution because I don't know any other way to set a dynamic value except UPDATE. I will also leave the filtering by "Temp" because I don't think it can do any harm.

apps/condo/domains/news/schema/NewsItemTemplate.test.js
```diff
@@ -86,6 +86,8 @@ describe('NewsItemTemplate', () => {
                 expect(obj.createdAt).toMatch(DATETIME_RE)
                 expect(obj.updatedAt).toMatch(DATETIME_RE)
                 expect(obj.organization).toBeNull()
+                expect(obj.name).toEqual(attrs.name)
+                expect(obj.category).toBeNull()
```
toplenboren, line 90: best to add a test that checks that you can create smth with category

apps/condo/lang/en/en.json
```diff
@@ -1688,6 +1688,19 @@
   "news.fields.templates": "Templates",
   "news.fields.profanityInTitle.error": "Profanity in title detected",
   "news.fields.profanityInBody.error": "Profanity in body detected",
+  "news.template.category.water": "Water",
+  "news.template.category.heating": "Heating",
+  "news.template.category.electricity": "Electricity",
+  "news.template.category.utilities": "Electricity, water, elevators",
+  "news.template.category.electricityAndElevators": "Electricity, elevators",
```
toplenboren, line 1695: Lets make it just Electricity and elevators

apps/condo/lang/en/en.json
```diff
@@ -1688,6 +1688,19 @@
   "news.fields.templates": "Templates",
   "news.fields.profanityInTitle.error": "Profanity in title detected",
   "news.fields.profanityInBody.error": "Profanity in body detected",
+  "news.template.category.water": "Water",
+  "news.template.category.heating": "Heating",
+  "news.template.category.electricity": "Electricity",
+  "news.template.category.utilities": "Electricity, water, elevators",
```
toplenboren, line 1694: All utilities

toplenboren (approved):

Alllex202 (approved):

apps/condo/domains/news/utils/testSchema/index.js
```diff
@@ -132,13 +133,17 @@ async function createTestNewsItemTemplate (client, extraAttrs = {}) {
     if (!client) throw new Error('no client')
     const sender = { dv: 1, fingerprint: faker.random.alphaNumeric(8) }
 
+    const name = faker.lorem.words(1)
+    const category = ALL_NEWS_CATEGORIES[Math.floor(Math.random() * ALL_NEWS_CATEGORIES.length)]
```
Alllex202, line 137: better use `faker.helpers.arrayElement([1,2,3])`

apps/condo/domains/news/utils/testSchema/index.js
```diff
@@ -153,12 +158,16 @@ async function updateTestNewsItemTemplate (client, id, extraAttrs = {}) {
     if (!id) throw new Error('no id')
     const sender = { dv: 1, fingerprint: faker.random.alphaNumeric(8) }
 
+    const name = faker.lorem.words(1)
+    const category = ALL_NEWS_CATEGORIES[Math.floor(Math.random() * ALL_NEWS_CATEGORIES.length)]
```
Alllex202, line 162: same^^

PR #5267, opened 2024-09-20
feat(condo): INFRA-469 registerBillingReceipt optimisations

ekabardinsky:


apps/condo/k6/src/registerBillingReceipt.test.ts
```diff
@@ -0,0 +1,155 @@
+import { faker } from '@faker-js/faker/locale/ru'
+import dayjs from 'dayjs'
+import { check } from 'k6'
+import crypto from 'k6/crypto'
+import encoding from 'k6/encoding'
+
```
dkoviazin, line 6: You need to create 5000 properties before checking out will the mutation works faster or slower

apps/condo/domains/billing/schema/resolvers/propertyResolver.js
```diff
@@ -49,19 +50,34 @@ class PropertyResolver extends Resolver {
...
         }
         return property
     }
 
     async init () {
+        // We need to load properties cache in case we do not use addressService for address normalize
+        // Let's select first 5000 properties - will take 15 mb approx to store them
+        // Exception is isCottageVillage - we have to load all properties for such organisations
+        const firstCondition = this.isCottageVillage ? {} : { first: 5000 }
+        this.organizationProperties = await itemsQuery('Property', {
```
dkoviazin, line 73: we do not need to load such cache. In case of small organizations - we will have 2-10 properties. In case of service providers - we will have 30K-100K properties

apps/condo/k6/src/registerBillingReceipt.test.ts
ekabardinsky: Good point. Will try

apps/condo/domains/billing/schema/resolvers/propertyResolver.js
ekabardinsky: 99 percent of organizations has less than 5000 properties. So such cache going to be work for 99% percent cases

ekabardinsky: I just run load test with pre created 1000 properties. Looks like this caching going to increate time to proceed a request. Will revert this change

dkoviazin (approved):

apps/condo/k6/src/registerBillingReceipt.test.ts
```diff
@@ -0,0 +1,155 @@
...
+export const options = {
+    tags: { testid: 'registerBillingReceipt', serverUrl: __ENV.BASE_URL },
+    scenarios: {
+        registerBillingReceipt: {
+            exec: 'registerBillingReceipt',
+            executor: 'constant-vus',
+            duration: DURATION,
+            // executor: 'per-vu-iterations',
+            // iterations: 1,
```
nomerdvadcatpyat, line 28: remove comments

apps/condo/k6/src/registerBillingReceipt.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    let accountCounter = 0
+    for (let i = 0; i < TOTAL_RECEIPT_NUMBER; i++) {
+        receipts.push(createJSONReceipt({ bankAccount, address: addressWithUnit, accountNumber: `${accountNumberPrefix}${accountIndex}` }))
+        accountCounter++
+
+        if (accountCounter >= RECEIPT_PER_ACCOUNT_NUMBER) {
+            accountCounter = 0
+            accountIndex++
+        }
+    }
```
nomerdvadcatpyat, line 126: I think it's better to prepare data in `setup` function

apps/condo/domains/billing/schema/resolvers/propertyResolver.js
```diff
@@ -49,9 +50,17 @@ class PropertyResolver extends Resolver {
         if (loaded) {
             return loaded
         }
+
+        const knownMissingProperty = this.missingOrganizationPropertiesIndex[addressKey]
+        if (knownMissingProperty) {
+            return loaded
```
nomerdvadcatpyat, line 56: It's ok that if we check not existed organization property, then create this property and recheck `getOrganizationProperty`, then it returns that this property does not exist?

apps/condo/k6/src/registerBillingReceipt.test.ts
ekabardinsky: I don't think this is required to have such complex set up function.
In other hand - setup function supose

apps/condo/k6/src/registerBillingReceipt.test.ts
ekabardinsky: 10x

apps/condo/domains/billing/schema/resolvers/propertyResolver.js
ekabardinsky: This logic are working in other way. Firstly it check all properties, calc missing, create missing. For case, when some property are omit - it going to check and check the same property again and again. So, if we already know that such property are omit - why we have to check it again?

nomerdvadcatpyat (approved):

PR #5269, opened 2024-09-23
feat(keystone): INFRA-446 extend Keystone to support GQL execution context 🧩

SavelevMatthew:


packages/keystone/KSv5v6/v5/keystone.js
```diff
@@ -0,0 +1,61 @@
...
+            wrappedResolvers.Mutation = Object.fromEntries(
+                Object.entries(Mutation).map(([mutationName, resolver]) => {
+                    return [mutationName, _getContextResolver(resolver)]
+                })
+            )
+        }
+
+        return wrappedResolvers
+    }
+}
```
sitozzz, line 57: That method called once at the application start, right?

packages/keystone/databaseAdapters/ReplicaKnexAdapter.js
```diff
@@ -54,6 +57,18 @@ class ReplicaKnexAdapter extends KnexAdapter {
 
         //override runner method
         this.knex.client.runner = (builder) => {
+            const graphqlContext = graphqlCtx.getStore()
+
+            const operationType = get(graphqlContext, ['info', 'operation', 'operation'])
```
sitozzz, line 62: Is that a default path for operation definition?

packages/keystone/KSv5v6/v5/keystone.js
SavelevMatthew: Yep. `keystone.prepare -> createApolloServer -> getResolvers`
image
image

packages/keystone/databaseAdapters/ReplicaKnexAdapter.js
```diff
@@ -54,6 +57,18 @@ class ReplicaKnexAdapter extends KnexAdapter {
...
+
+            // NOTE: any DB calls inside mutation => master, any other calls routed by SQL operation.
+            // Probably we want to pass function (builder => { replicaSet: string, master: boolean }) later in constructor,
+            // but for now "hard-coded" solution seems fine
+            if (operationType === 'mutation') {
+                return this.knexWrite.client.runner(builder)
+            }
+
+            // TODO(SavelevMatthew): think about removing this block later
```
sitozzz, line 71: You should also remember about transactions, such as migrations, index creations and etc

packages/keystone/databaseAdapters/ReplicaKnexAdapter.js
sitozzz: So what if user called an 'mutation' inside which there will be a query. A similar question can be asked in reverse order

SavelevMatthew: As far as I checked - yes
image

packages/keystone/databaseAdapters/ReplicaKnexAdapter.js
SavelevMatthew: Yep, I'm talking about `queryContext.useMaster`. If I've understood its purpose correctly, then its goal is to route request based on JS-flag. So, block above does the same. Right now I left it as is, since I'm not sure (more routing = good or bad), but we'll figure it out later.

I'm also not planning to remove`checkUseMasterSingle`, since its goal to parse raw SQL (such as migrations / indexes and so on), not business (JS) logic. So we should keep it 100%

packages/keystone/databaseAdapters/ReplicaKnexAdapter.js
SavelevMatthew: Docs about resolver: https://www.apollographql.com/docs/apollo-server/data/resolvers/#resolver-arguments
Info lead to this type definitions: https://github.com/graphql/graphql-js/blob/f851eba93167b04d6be1373ff27927b16352e202/src/type/definition.ts#L891-L902
And then it leads to here: https://github.com/graphql/graphql-js/blob/16.x.x/src/language/ast.ts#L319

So it's 100% correct (default apollo implementation, not modified by KS)

packages/keystone/executionContext.js
```diff
@@ -57,5 +57,6 @@ function _internalGetAsyncLocalStorage (name = 'default') {
 
 module.exports = {
     _internalGetExecutionContextAsyncLocalStorage,
+    _internalGetAsyncLocalStorage,
```
pahaz, line 60: why is the `_internalGetExecutionContextAsyncLocalStorage` is not enought?

SavelevMatthew: Since execution context in my head = request context, it contains `reqId` and other stuff.

`gqlContext` -> GQL call stack, if request contains multiple queries = its multiple contexts, if mutation calls another mutation / query -> that's a call stack (right now not implemented, since not required)...

So I guess its just different context from definition...

pahaz: @toplenboren could you please check

packages/keystone/databaseAdapters/ReplicaKnexAdapter.js
toplenboren: you should add tasknumber here

packages/keystone/executionContext.js
toplenboren: Because @SavelevMatthew wants custom context, not the one that we use for reqId storage

Why couldnt we just squeeze context into `executionContext`? It would also help in traces and logs. I mean, just add one more param besides `reqId` `execId` `taskId`... like `graphQLOperation`

pahaz (approved):

PR #5284, opened 2024-09-25
feat(condo): INFRA-524 optimise register meters

nomerdvadcatpyat:
- moved meter and meter readings selects from for loop to top -> select it 1 time instead n
- wrote tests for case with duplicates in one input data

### DIFF
--- a/apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
+++ b/apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
@@ -317,6 +317,32 @@ const RegisterMetersReadingsService = new GQLCustomSchema('RegisterMetersReading
                 const endFindMetersTime = Date.now()
                 // console.log('time to find meters', endFindMetersTime - findMetersTime)
 
+
+                console.log('before map')
+                const readingsWithValidDates = readings.filter(reading => isDateValid(reading.date))
+                console.log('readingsWithValidDates', readingsWithValidDates)
+                // Собираем все meterId и даты до цикла
+                const meterDatePairs = readingsWithValidDates.map(reading => ({
+                    meterNumber: reading.meterNumber.trim(),
+                    date: toISO(reading.date),
+                }))
+                console.log('after map', meterDatePairs)
+
+                const meterReadings = await MeterReading.getAll(context, {
+                    meter: { id_in: meters.map(meter => meter.id) }, // используем все найденные meters
+                    date_in: uniq(meterDatePairs.map(pair => pair.date)), // используем все даты
+                })
+                console.log('meterReadings', meterReadings)
+                // Строим объект для быстрого поиска дубликатов по meterId и дате
+                const meterReadingMap = meterReadings.reduce((acc, reading) => {
+                    const key = `${reading.meter.id}-${reading.date}`
+                    acc[key] = reading
+                    return acc
+                }, {})
+                console.log('meterReadingMap', meterReadingMap)
+
+                console.log('readings', readings)
+
                 for (const reading of readings) {
                     // console.log('--- ITER ---')
                     const meterNumber = reading.meterNumber.trim()
@@ -343,10 +369,13 @@ const RegisterMetersReadingsService = new GQLCustomSchema('RegisterMetersReading
                         continue
                     }
 
+                    console.log('reading.date', reading.date)
                     if (!isDateValid(reading.date)) {
                         resultRows.push(new GQLError(ERRORS.INVALID_DATE(i18n('meter.import.column.meterReadingSubmissionDate', { locale })), context))
                         continue
                     }
+
+                    const dateISO = toISO(reading.date)
 
                     const property = properties.find((p) => p.addressKey === addressKey)
 
@@ -450,18 +479,23 @@ const RegisterMetersReadingsService = new GQLCustomSchema('RegisterMetersReading
                     // console.log('create or update meter time', endCreateOrUpdate - startCreateOrUpdate)
 
                     try {
-                        const findDuplicatesTime = Date.now()
-                        // Вытащить find по date из цикла. Тут по values искать в переданных MeterReadings
-                        // Затем мапить созданные и добавлять в массив тот
-                        const duplicates = await MeterReading.getAll(context, {
-                            meter: { id: meterId },
-                            date: toISO(reading.date),
-                            ...values,
-                        })
-                        const findDuplicatesEndTime = Date.now()
-                        // console.log('find duplicates time', findDuplicatesEndTime - findDuplicatesTime)
+                        // const findDuplicatesTime = Date.now()
+                        // // Вытащить find по date из цикла. Тут по values искать в переданных MeterReadings
+                        // // Затем мапить созданные и добавлять в массив тот
+                        // const duplicates = await MeterReading.getAll(context, {
+                        //     meter: { id: meterId },
+                        //     date: toISO(reading.date),
+                        //     ...values,
+                        // })
+                        // const findDuplicatesEndTime = Date.now()
+                        // // console.log('find duplicates time', findDuplicatesEndTime - findDuplicatesTime)
 
-                        if (duplicates.length === 0) {
+
+                        // Проверяем на дубликат, используя заранее созданную карту meterReadingMap
+                        const key = `${meterId}-${dateISO}`
+                        const duplicateReading = meterReadingMap[key]
+
+                        if (!duplicateReading) {
                             const createMeterReadingTime = Date.now()
                             // Поменять на MeterReadingId
                             const createdMeterReading = await MeterReading.create(context, {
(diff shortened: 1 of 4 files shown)


apps/condo/domains/meter/schema/RegisterMetersReadingsService.test.js
```diff
@@ -975,6 +975,69 @@ describe('RegisterMetersReadingsService', () => {
...
+                    id: property.id,
+                    address: property.address,
+                    addressKey: property.addressKey,
+                }),
+                unitType: duplicateReadings[0].addressInfo.unitType,
+                unitName: duplicateReadings[0].addressInfo.unitName,
+                accountNumber: duplicateReadings[0].accountNumber,
+                number: duplicateReadings[0].meterNumber,
+            }),
+        })]))
```
Alllex202, line 1000: Do you expect the data length to be 1 or not?
This code will not check the length, you should do it explicitly

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
```diff
@@ -298,13 +317,36 @@ const RegisterMetersReadingsService = new GQLCustomSchema('RegisterMetersReading
...
+                const meters = await find('Meter', {
+                    organization,
+                    number_in: meterNumbers,
+                    deletedAt: null,
+                })
+
+                const readingsWithValidDates = readings.filter(reading => isDateValid(reading.date))
+                const meterDatePairs = readingsWithValidDates.map(reading => ({
+                    meterNumber: reading.meterNumber.trim(),
+                    date: toISO(reading.date),
```
SavelevMatthew, line 330: Should you check, that all dates are valid? (`toISO` can return `undefined`, so filter is skipped), or String is not valid date

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
```diff
@@ -298,13 +317,36 @@ const RegisterMetersReadingsService = new GQLCustomSchema('RegisterMetersReading
...
+                }))
+                const meterReadings = await MeterReading.getAll(context, {
+                    meter: { id_in: meters.map(meter => meter.id) },
```
SavelevMatthew, line 333: Seems a bit strange (or am I missing something).

If I have 500 different meters and dates, that's turn query into:
(Give me all readings, where meter is one of 500 and date is one of 500), so it can be 500 x 500 = 250_000 rows here

SavelevMatthew: Also getAll will fail for >1000 objects

apps/condo/domains/meter/schema/RegisterMetersReadingsService.test.js
```diff
@@ -975,6 +975,69 @@ describe('RegisterMetersReadingsService', () => {
...
+            property: { id: property.id },
+        })
+        expect(meters).toHaveLength(1)
+        expect(meters[0].number).toBe(duplicateReadings[0].meterNumber)
+
+        const metersReadings = await MeterReading.getAll(adminClient, { meter: { id_in: map(meters, 'id') } })
+        expect(metersReadings).toHaveLength(1)
+    })
+
+    test('update meter if it present in input two times with different values', async () => {
```
SavelevMatthew, line 1013: Specify in test name what behaviour do you expect. First one registered / second one / latest by date?

apps/condo/k6/src/registerMetersReadings.test.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    const createdOrganization = createOrganization({ token })
+    const organizationId = createdOrganization.json('data.obj.id')
+
+    getOrganizationEmployeeId({ token, organizationId })
+    const createdPropertyResponse = createProperty({ token, organizationId })
+    const property = createdPropertyResponse.json('data.obj')
+
+    const readings = []
+    for (let i = 0; i < 250; i++) {
```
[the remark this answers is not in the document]
nomerdvadcatpyat: update to 500

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
```diff
@@ -415,20 +459,18 @@ const RegisterMetersReadingsService = new GQLCustomSchema('RegisterMetersReading
...
-                        const duplicates = await MeterReading.getAll(context, {
-                            meter: { id: meterId },
-                            date: toISO(reading.date),
-                            ...values,
-                        })
+                        const key = `${meterId}-${dateISO}`
+                        const duplicateReading = meterReadingByDate[key]
 
-                        if (duplicates.length === 0) {
+                        if (!duplicateReading) {
```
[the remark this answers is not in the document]
nomerdvadcatpyat: No, because if there is no duplicate, reading will be created and set in `meterReadingByDate`. And for create reading, we need a existed meter, which created above

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
nomerdvadcatpyat: In line 327 we have filter readings with valid dates

apps/condo/domains/meter/schema/RegisterMetersReadingsService.test.js
nomerdvadcatpyat: It should be 2 same rows in the data

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
nomerdvadcatpyat: > Also getAll will fail for >1000 objects

rewrote to find

nomerdvadcatpyat: > Seems a bit strange (or am I missing something).
> 
> If I have 500 different meters and dates, that's turn query into: (Give me all readings, where meter is one of 500 and date is one of 500), so it can be 500 x 500 = 250_000 rows here

250k readings it's very strange case: it's mean that we have 500 different dates in import file and for every 500 meters we have a reading for every of 500 dates.
I think we can set a `first` here, for example 5000. What do you think?

Alllex202: > I think we can set a `first` here

If we take only the first 5000 (or any other), it means that we can skip the duplicate check for some target reading

Alllex202: > 500 x 500 = 250_000 rows here

Worst thing is that in the worst case we won't be able to turn it off. And it's a normal mutation, not a deferred task

Alllex202 (approved):

apps/condo/k6/src/registerMetersReadings.test.ts
```diff
@@ -0,0 +1,106 @@
...
+        registerMetersReadings: {
+            exec: 'registerMetersReadings',
+            executor: 'constant-vus',
+            duration: DURATION,
+            vus: 1,
+        },
+    },
+    thresholds: {
+        http_req_failed: ['rate<0.01'],
+        http_req_duration: ['med<5000'],
```
SavelevMatthew, line 30: Why 5000 is here? Seems like a lot, doesn't it?

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
```diff
@@ -298,13 +317,43 @@ const RegisterMetersReadingsService = new GQLCustomSchema('RegisterMetersReading
...
+                    deletedAt: null,
+                })
+
+                const readingsWithValidDates = readings.filter(reading => isDateValid(reading.date))
+                const plainMeterReadings = await find('MeterReading', {
+                    meter: { id_in: meters.map(meter => meter.id) },
+                    date_in: uniq(readingsWithValidDates.map(reading => toISO(reading.date))),
+                })
+                const meterReadings = plainMeterReadings.map(reading => {
+                    const meter = meters.find(meter => meter.id === reading.meter)
```
SavelevMatthew, line 333: `find` has O(N) complexity, and plain is limited of `500`, so its `500` x `25000` operations. 

Can be simplified by converting `Array<Meter>` to `Record<id, Meter>`. So it's `25000` for conversion and `500` for map. 25500 total cost

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
nomerdvadcatpyat: > Should we add index here, or is it fast enough?

I don't think an index is needed here. `registerMetersReadings` is a rare request

apps/condo/k6/src/registerMetersReadings.test.ts
nomerdvadcatpyat: decrease to 1s

SavelevMatthew (approved):

PR #5288, opened 2024-09-25
feat(condo): DOMA-10102 add column minFloor to PropertyMap in import excel file 

tolmachev21:
We want users to be able to specify a minimum floor when loading houses from excel file.

Before:

image

Now:
Screenshot 2024-09-26 at 00 10 06


apps/condo/public/import/property/ru/property-import-example.xlsx
tolmachev21, line 1: Need to update files when minFloor column name is approved (Don't forget!)

apps/condo/domains/property/hooks/useImporterFunctions.tsx
```diff
@@ -14,12 +14,16 @@ import { Property } from '@condo/domains/property/utils/clientSchema'
...
+const createPropertyUnitsMap = (units, sections, floors, minFloor) => {
     const unitsOnFloor = Math.ceil(units / (floors * sections))
     if (!unitsOnFloor) {
         return
     }
 
+    if (!minFloor) {
+        minFloor = 1
+    }
+
```
toplenboren, line 26: You can set it in arguments https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

apps/condo/domains/property/hooks/useImporterFunctions.tsx
```diff
@@ -70,6 +75,7 @@ export const useImporterFunctions = (): [Columns, RowNormalizer, RowValidator, O
         { name: UnitLabel, type: 'number', required: true },
         { name: SectionLabel, type: 'number', required: true },
         { name: FloorLabel, type: 'number', required: true },
+        { name: minFloorLabel, type: 'number', required: false },
```
toplenboren, line 78: why not simply make it required = true? For backwards compatibility?

apps/condo/domains/property/hooks/useImporterFunctions.tsx
```diff
@@ -34,8 +38,8 @@ const createPropertyUnitsMap = (units, sections, floors) => {
         mapEditor.addSection({
             name,
             unitsOnFloor,
-            minFloor: 1,
-            maxFloor: floors,
+            minFloor,
+            maxFloor: floors + (minFloor - 1),
```
toplenboren, line 42: So, for example, I have a building which is 26 floors and first living floor is 2nd

Judging by your code...

minFloor = 2
maxFloor = 26 + (2 - 1) = 27

It is not really obvious why the number of floors changes

apps/condo/public/import/property/ru/property-import-example.xlsx
toplenboren: I think you should also update /en

apps/condo/domains/property/hooks/useImporterFunctions.tsx
tolmachev21: I I knew about this, but the TC complains that I am passing a pure number type when it could be a string or Date type

apps/condo/domains/property/hooks/useImporterFunctions.tsx
```diff
@@ -7,15 +7,19 @@ import { useIntl } from '@open-condo/next/intl'
...
+import { Columns, RowNormalizer, RowValidator, ObjectCreator, TableRow } from '@condo/domains/common/utils/importer'
 import { MapEdit } from '@condo/domains/property/components/panels/Builder/MapConstructor'
 import { validHouseTypes } from '@condo/domains/property/constants/property'
 import { Property } from '@condo/domains/property/utils/clientSchema'
 import { searchProperty } from '@condo/domains/ticket/utils/clientSchema/search'
 
 
-const createPropertyUnitsMap = (units, sections, floors) => {
-    const unitsOnFloor = Math.ceil(units / (floors * sections))
+const createPropertyUnitsMap = (units, sections, floors, minFloor) => {
```
Alllex202, line 17: If I specify a negative floor, will everything work correctly?

apps/condo/domains/property/hooks/useImporterFunctions.tsx
```diff
@@ -34,7 +38,7 @@ const createPropertyUnitsMap = (units, sections, floors) => {
         mapEditor.addSection({
             name,
             unitsOnFloor,
-            minFloor: 1,
+            minFloor,
             maxFloor: floors,
```
Alllex202, line 42: don't you need to change `maxFloor` depending on `minFloor`?

Alllex202 (approved):

toplenboren (approved):

PR #5289, opened 2024-09-26
feat(condo): DOMA-10287 caching receiptId by it's hash if it has importId and skipping checks for it next time

YEgorLu:
Some clients attacking us with many receipts in RegisterBillingsReceiptService every few hours. For example they can load every time 1000 receipts for update, but data changed only in 5 receipts.

Now we wasting resources to check updates on every receipt, so instead we hashing it and storing in cache.

Most likely they will keep spamming same receipts till the end of period, so cache ttl is 30 days.


apps/condo/domains/billing/constants/registerBillingReceiptService.js
```diff
@@ -5,6 +5,7 @@ const { NOT_FOUND, WRONG_FORMAT, WRONG_VALUE } = require('@condo/domains/common/
 const RECEIPTS_LIMIT = 1000
 const ADDRESS_SERVICE_NORMALIZE_CHUNK_SIZE = 10
 const PROPERTY_SCORE_TO_PASS = 95
+const CONTROL_SUM_TTL_IN_MS = 1000 * 60 * 60 * 24 * 30 // 30 days
```
pahaz, line 8: use more context naming for that variable

apps/condo/domains/billing/schema/RegisterBillingReceiptsService.js
```diff
@@ -26,10 +27,14 @@ const { BillingReceipt } = require('@condo/domains/billing/utils/serverSchema')
 const {
     BillingIntegrationOrganizationContextForRegisterBillingReceiptsService: BillingContextApi,
 } = require('@condo/domains/billing/utils/serverSchema')
+const { ReceiptInputCache } = require('@condo/domains/billing/utils/serverSchema/receiptInputCache')
 
 const appLogger = getLogger('condo')
 const registerReceiptLogger = appLogger.child({ module: 'register-billing-receipts' })
 
+const redisClient = getRedisClient('register-billing-receipts', 'cache')
```
pahaz, line 35: you can use `default` redis client. I don't think that we will move it to another redis server.

apps/condo/domains/billing/schema/RegisterBillingReceiptsService.js
```diff
@@ -117,10 +122,22 @@ const RegisterBillingReceiptsService = new GQLCustomSchema('RegisterBillingRecei
...
                 // errors are critical and receipt will be skipped, problems can be fixed later
-                let receiptIndex = Object.fromEntries(receiptsInput.map((receiptInput, index) =>
-                    ([index, { ...receiptInput, error: null, problems: [] }])
-                ))
+                let receiptIndex = {}
                 let errorsIndex = {}
+                let unchangedIndex = {}
+                const cache = new ReceiptInputCache(redisClient, getRedisKey, billingContextId)
+
+                for (let index = 0; index < receiptsInput.length; index++) {
```
pahaz, line 130: it looks like some internal cache logic...

When adding caching, I would expect to see a simpler structure in the code, like:

```javascript
const cache = new XXX(…);
…
if (cache.isChanged(receipt)) {
 …
}
```

However, it seems there are quite a few changes here. Perhaps we should reconsider the approach to code organization?

apps/condo/domains/billing/schema/RegisterBillingReceiptsService.test.js
```diff
@@ -554,4 +558,59 @@ describe('RegisterBillingReceiptsService', () => {
         })
     })
 
+    describe('Cache', () => {
+        const REDIS_CLIENT_NAME = 'register-billing-receipt'
+        const redisClient = getRedisClient(REDIS_CLIENT_NAME, 'cache')
```
pahaz, line 563: You can't use redis Client inside test.js files!

You should use only API calls inside tests. If you want some redis checks you need to use .spec.js file

apps/condo/domains/billing/utils/testSchema/mixins/billing.js
```diff
@@ -62,38 +62,19 @@ const BillingTestMixin = {
     },
 
     randomNumber (numDigits) {
-        const min = 10 ** (numDigits - 1)
-        const max = 10 ** numDigits - 1
-        return faker.datatype.number({ min, max })
+        return randomNumber(numDigits)
```
pahaz, line 65: is this funtion a really part of MinIn? look like helper

PR #5295, opened 2024-09-27
feat(condo): INFRA-538 add new generateServerUtils implementation wit…

ekabardinsky:
…h support of deprecated one

### DIFF
--- a/apps/condo/domains/acquiring/schema/RegisterMultiPaymentForVirtualReceiptService.js
+++ b/apps/condo/domains/acquiring/schema/RegisterMultiPaymentForVirtualReceiptService.js
@@ -140,7 +140,7 @@ const RegisterMultiPaymentForVirtualReceiptService = new GQLCustomSchema('Regist
 
                 const acquiringIntegration = await AcquiringIntegration.getOne(context, {
                     id: acquiringContext.integration,
-                })
+                }, 'id minimumPaymentAmount maximumPaymentAmount hostUrl deletedAt')
 
                 if (acquiringIntegration.deletedAt) {
                     throw new GQLError({
--- a/packages/codegen/generate.gql.js
+++ b/packages/codegen/generate.gql.js
@@ -245,4 +245,11 @@ module.exports = {
     generateGqlQueries,
     generateQueryWhereInput,
     generateQuerySortBy,
+    genGetAllGQL,
+    genGetCountGQL,
+    genCreateGQL,
+    genCreateManyGQL,
+    genUpdateGQL,
+    genUpdateManyGQL,
+    genDeleteGQL,
 }
--- a/packages/codegen/generate.gql.js
+++ b/packages/codegen/generate.gql.js
@@ -9,7 +9,7 @@ function getModelForms (key) {
     return [MODEL, MODELS]
 }
 
-function genGetAllGQL (key, fields, prefix = '') {
+function generateGetAllGQL (key, fields, prefix = '') {
     const [MODEL, MODELS] = getModelForms(key)
     const capitalizedPrefix = prefix.length ? prefix.charAt(0).toUpperCase() + prefix.slice(1) : ''
     const queryName = prefix ? `${prefix}All${MODELS}` : `all${MODELS}`
@@ -23,7 +23,7 @@ function genGetAllGQL (key, fields, prefix = '') {
     `
 }
 
-function genGetCountGQL (key, prefix = '') {
+function generateGetCountGQL (key, prefix = '') {
     const [MODEL, MODELS] = getModelForms(key)
     const metaQueryName = `${prefix}_all${MODELS}Meta`
     const capitalizedPrefix = prefix.length ? prefix.charAt(0).toUpperCase() + prefix.slice(1) : ''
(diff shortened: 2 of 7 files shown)


packages/codegen/generate.server.utils.js
```diff
@@ -330,6 +344,304 @@ function generateServerUtils (gql) {
...
+     * @param extraAttrs - can hold additional data for an update operation
+     * @returns {Promise<*>} - model stored object
+     */
+    async function softDelete (context, id, fields, extraAttrs = {}) {
+        const attrs = {
+            deletedAt: 'true',
+            ...extraAttrs,
+        }
+        return await update(context, id, attrs, fields)
+    }
```
Alllex202, line 585: What about options?
```suggestion
    async function softDelete (context, id, fields, extraAttrs = {}, options = {}) {
        const attrs = {
            deletedAt: 'true',
            ...extraAttrs,
        }
        return await update(context, id, attrs, fields, options)
    }
```

packages/codegen/generate.server.utils.js
```diff
@@ -330,6 +344,304 @@ function generateServerUtils (gql) {
...
+     */
+    async function softDeleteMany (context, ids, fields, extraAttrs = {}) {
+        const data = ids.map(id => ({
+            id,
+            data: {
+                deletedAt: 'true',
+                ...extraAttrs,
+            },
+        }))
+        return await updateMany(context, data, fields)
```
Alllex202, line 603: ^^ same
```suggestion
        return await updateMany(context, data, fields, options)
```

packages/codegen/generate.server.utils.js
```diff
@@ -142,9 +152,13 @@ async function execGqlWithoutAccess (context, { query, variables, errorMessage =
     return (dataPath) ? get(data, dataPath) : data
 }
 
-function generateServerUtils (gql) {
+/** @deprecated use generateServerUtils with schemaName parameter */
+function generateServerUtilsDeprecated (gql) {
```
Alllex202, line 156:
```suggestion
function _generateServerUtilsDeprecated (gql) {
```

apps/condo/domains/acquiring/schema/RegisterMultiPaymentForVirtualReceiptService.js
```diff
@@ -140,7 +140,7 @@ const RegisterMultiPaymentForVirtualReceiptService = new GQLCustomSchema('Regist
 
                 const acquiringIntegration = await AcquiringIntegration.getOne(context, {
                     id: acquiringContext.integration,
-                })
+                }, 'id minimumPaymentAmount maximumPaymentAmount hostUrl deletedAt')
```
SavelevMatthew, line 143: I guess its better to move this to some const on the top of the file, like we do with `ERRORS`.

packages/codegen/generate.gql.js
```diff
@@ -245,4 +245,11 @@ module.exports = {
     generateGqlQueries,
     generateQueryWhereInput,
     generateQuerySortBy,
+    genGetAllGQL,
```
SavelevMatthew, line 248: Should it be consistent with other names? (`gen` vs `generate`)

apps/condo/domains/acquiring/schema/RegisterMultiPaymentForVirtualReceiptService.js
pahaz: It's better to keep it close to use. It's easier to review. Moving it to another file will decrease readability because you need to look at another file.

packages/codegen/generate.gql.js
ekabardinsky: Gen and Generate prefix has differences in implementation. genGetAllGQL suppose to be close to genGetCountGQL/genCreateGQL/genUpdateGQL and other.

apps/condo/domains/acquiring/schema/RegisterMultiPaymentForVirtualReceiptService.js
ekabardinsky: I prefer to keep fields along with getOne/getAll and other server methods at the same line of code.
I believe we have to not move it to top of the same file and not move it to another file - since it more readable to see exactly fields you are selecting from a model. More over - we going to never reuse those constants in future to not mix up different queries together

SavelevMatthew: > It's better to keep it close to use. It's easier to review. Moving it to another file will decrease readability because you need to look at another file.

I meant not another file, but the top of existing one. Like we do with ERRORS. So you can avoid horizontal scroll

packages/codegen/generate.gql.js
SavelevMatthew: Yep, I mean why not to use `generate` on every name? Because now you have half of exported members to have `generate` prefix, and other half is `gen`, so its inconsistent

apps/condo/domains/acquiring/schema/RegisterMultiPaymentForVirtualReceiptService.js
Alllex202: I would leave the inline style, it makes the code easier to review.
You can extract some long query into a constant, but only to improve readability

packages/codegen/generate.gql.js
ekabardinsky: make sense, will change

SavelevMatthew (approved):

PR #5298, opened 2024-09-30
fix(condo): DOMA-9422 meter readings import does not detect bad dates

YEgorLu:
Explicitly giving error if user tries to import dates as YYYY-DD-MM (month is not in center), because everywhere months places in center and it impossible to detect where month is placed

Found that excel cells in Date format parsed as days from epoch with specific microsoft logic (like 4512, 4512.2818273), so made it's parsing in normal date form

Refactored RegisterMeterReadingsService: moved helper functions and errors to different files


apps/condo/domains/meter/utils/date.utils.js
```diff
@@ -0,0 +1,76 @@
...
+    const sanitizedDateStr = dateStr.replace(SYMBOLS_TO_CUT_FROM_DATES_REGEXP, '')
+    if (isEmpty(sanitizedDateStr)) {
+        return undefined
+    }
+
+    return sanitizedDateStr
+}
+
+function monthIsLoweOrEqualTwelve (dateStr) {
+    const [dateData] = dateStr.split(/T|\s/, 1)
```
SavelevMatthew, line 27: ??

SavelevMatthew: just use `dayjs` for that, it should support array of formats

SavelevMatthew: don't you hard-coded (hand-written) regex for dates, it'll be hard to support it later

SavelevMatthew (commented):
Please note, that scaling team changed `RegisterMeterReadings` logic here:
https://github.com/open-condo-software/condo/pull/5284

So you should probably await, while PR is above is merged, and then submit yours.

Date utils seems strange btw, do you need some help on it?

apps/condo/domains/meter/utils/date.utils.js
YEgorLu: Why not datejs

Some users send dates like 2024.28.05, expecting us to recognize, that 28 in the middle is day, and 05 at the end is month. That looks okey to detect. But they also want us to understand, that 2024.05.01 is not 5th february, it's 1st may, so it seems impossible to understand correctly. I checked date formats in tables, everywhere in date formats month is placed in the middle of date. Our supported formats have month only in the middle, if day present. So i decided that it's better to throw error, than trying to understand what user wants.

Yes, dayjs with array of formats is already being used, but in case of 2024.28.05 it thinks, that 28 is a month and converts overflow to year (2024.28.05 -> 2026.01.05 or something like that).

So i think that explicit check for month is required, and errors on wrong month position are required

apps/condo/domains/meter/constants/registerMetersReadingsService.js
```diff
@@ -0,0 +1,90 @@
...
+    'YYYY-MM-DDTHH:mm:ss.SSS[Z]', // The result of dayjs().toISOString()
+    `${ISO_DATE_FORMAT} HH:mm:ss`, `${EUROPEAN_DATE_FORMAT} HH:mm:ss`, // Up to seconds
+    `${ISO_DATE_FORMAT} HH:mm`, `${EUROPEAN_DATE_FORMAT} HH:mm`, // Up to minutes
+    ISO_DATE_FORMAT, EUROPEAN_DATE_FORMAT, // No time
+    'YYYY-MM', 'MM-YYYY', 'YYYY.MM', 'MM.YYYY', // Some exotic cases
+]
+const READINGS_LIMIT = 500
+
+const ERRORS = {
+    TOO_MUCH_READINGS: (sentCount) => ({
```
pahaz, line 27: Please, don't use functions. We already have interpolation and you don't need to use functions

apps/condo/domains/meter/constants/registerMetersReadingsService.js
```diff
@@ -0,0 +1,90 @@
+const { GQLErrorCode: { BAD_USER_INPUT } } = require('@open-condo/keystone/errors')
```
pahaz, line 1: why do you want to move it to another file? usually we keep it close to service

apps/condo/domains/meter/utils/date.utils.js
pahaz: We should accept only ISO standard dates. `2024.28.05` should return error it's not ISO format

dkoviazin (approved):

apps/condo/domains/common/utils/date.js
```diff
@@ -1,8 +1,15 @@
...
+const customParseFormat = require('dayjs/plugin/customParseFormat')
+const { isEmpty, isString } = require('lodash')
+
+dayjs.extend(customParseFormat)
 
 const DATE_FORMAT = 'YYYY-MM-DD'
 const DATE_FORMAT_Z = 'YYYY-MM-DDTHH:mm:ss.SSSZZ'
 
+const SYMBOLS_TO_CUT_FROM_DATES_REGEXP = /[^-TZ_:.,( )/\d]/gi // -_:.,()/ recognizable by dayjs https://day.js.org/docs/en/parse/string-format
+const UTC_END_REGEXP = /(Z|(\.\d{3}[+-](\d{2}:?\d{2})))$/ // Z | +5000 | +50:00
```
SavelevMatthew, line 11: Whats this used for? We're trying to avoid manual string parsing. I get the "remove trash" part from above, but what's the use of this one?

apps/condo/domains/common/utils/date.js
```diff
@@ -36,10 +43,65 @@ const getStartDates = (dateRaw) => {
  */
 const getCurrTimeStamp = () => Math.floor(Date.now() / 1000)
 
+/**
+ * Clear bad user input
+ * @param dateStr raw user input
+ * @returns {null|string}
+ */
+function clearDateStr (dateStr) {
```
SavelevMatthew, line 51: Naming is a bit unconvincing. 

I expect it to work like that:
```
const clearedInput = clearInput(input)
const isValid = isValidInput(clearedInput)
```

SavelevMatthew: `null` in return type is bad, since you cannot predict it from function name. You expect to clear `string`, not to override it with `null`. 

That's bad pattern, which is acceptable in-place (in specific mutation file), but not in common utils

apps/condo/domains/common/utils/date.js
```diff
@@ -36,10 +43,65 @@ const getStartDates = (dateRaw) => {
...
+ * @param {{formats?: Array<string>, strict?: boolean}} options
+ * @return {boolean}
+ */
+function isDateStrValid (dateStr, options) {
+    if (!dateStr) {
+        return false
+    }
+
+    const params = {
+        formats: undefined,
```
SavelevMatthew, line 77: why exactly this options?

apps/condo/domains/common/utils/date.js
```diff
@@ -36,10 +43,65 @@ const getStartDates = (dateRaw) => {
...
+        strict: false,
+        ...options,
+    }
+
+    let dateFunc = dayjs
+    if (isDateStrInUTCFormat(dateStr)) {
```
SavelevMatthew, line 83: That's something bad. Why do you even need this if?

SavelevMatthew: Seems like you're patching specific local problem here. And don't even know which...

User input MUST NOT serve as guideline for parsing function.

apps/condo/domains/common/utils/date.spec.js
```diff
@@ -35,4 +37,116 @@ describe('Date helper tests', () => {
...
+        })
+
+        it('should return null for non-string input (object)', () => {
+            const input = { date: '2021-05-04' }
+            expect(clearDateStr(input)).toBeNull()
+        })
+
+        it('should return null for a string with only invalid characters', () => {
+            const input = '!!!@#@#$$%'
+            expect(clearDateStr(input)).toBeNull()
```
SavelevMatthew, line 74:
```suggestion
            expect(clearDateStr(input)).toEqual('')
```

SavelevMatthew: I give you a string, you Promise to clean it. Why you return me nothing? I expect cleared string

SavelevMatthew (changes requested):
UTC detection is sooo strange for me. Seems like local hack, which should not be in common domain

apps/condo/domains/common/utils/date.js
YEgorLu: I though that if it's in common utils, it should accept any parameters. Now i moved it to specific utils and force needed params

apps/condo/domains/common/utils/date.js
YEgorLu: We have to do these hacks since dayjs couldn't do everything himself.
Moved this from common to specific utils thou

YEgorLu:
I get that this date parsing logic is specific and does not belong to common domain, so i moved it in meter domain in specific utils
There was an idea, to move date parsing in meterImportTask, where files being parsed, then we could restrict specific format of dates in RegisterMetersReadings. But we have integration with 1C, where they instructed to send any dates, like in Excel right into RegisterMetersReadings mutation. So we have to keep date parsing in here and accept multiple formats of dates

YEgorLu:
Decided to restrict date formats in RegisterMetersReadings with 2 types: UTC, which we want, and YYYY-MM-DD which we wrote in docs for integration.

Learned that all dates must be validated. Before we threw error only on invalid submission date, it was a mistake, so adding validation for all dates, not only submission.

Now if we parsing file, validating and parsing dates to UTC on task level, before mutation.

apps/condo/domains/meter/constants/registerMetersReadingsService.js
```diff
@@ -0,0 +1,61 @@
...
+    ...DATE_TIME_FORMATS,
+    ...DATE_FORMATS,
+
+    'YYYY-MM-DDTHH:mm:ss.SSS[Z]', // The result of dayjs().toISOString()
+    'YYYY-MM-DDTHH:mm:ss.SSSZ',
+    'YYYY-MM-DDTHH:mm:ss.SSS',
+    'YYYY-MM-DDTHH:mm:ssZZ',
+    'YYYY-MM-DDTHH:mm:ssZ',
+    'YYYY-MM-DDTHH:mm:ss',
+].sort((a, b) => b.length - a.length) // Order matters! see "Differences to moment" https://day.js.org/docs/en/parse/string-format
```
SavelevMatthew, line 30: Why do you need all code above?

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
```diff
@@ -185,31 +102,39 @@ function transformToPlainObject (input) {
...
-        'installationDate',
-        'commissioningDate',
-        'sealingDate',
-        'controlReadingsDate',
-    ]
-
-    return fieldsToUpdate.reduce((result, field) => {
-        if (result) {
-            return result
+const tryToISO = (dateStr) => {
```
SavelevMatthew, line 109: Why do you redefine utils just to call it?

apps/condo/domains/meter/constants/registerMetersReadingsService.js
```diff
@@ -0,0 +1,61 @@
...
+
+const READINGS_LIMIT = 500
+
+const DATE_FIELD_PATHS = [
```
SavelevMatthew, line 34: Maybe use date fields from schema file instead?

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
```diff
@@ -185,31 +102,39 @@ function transformToPlainObject (input) {
     return result
 }
 
+const isDateStrValid = (dateStr) => {
+    return isDateStrValidUtils(dateStr, { utc: true, formats: [DATE_FORMAT] })
```
SavelevMatthew, line 106: Why iso is not passed

apps/condo/domains/meter/constants/registerMetersReadingsService.js
YEgorLu: don't want to manually write all formats

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
YEgorLu: If you are talking about YYYY-MM-DDTHH:mm:ss.SSSZ format in formats array, because parsing logic for utc format is different and passing this format to dayjs in strict mode has no use

apps/condo/domains/meter/constants/registerMetersReadingsService.js
YEgorLu: I assume you mean that in MeterReading and Meter schema we setting "isRequired" to fields and could use them. But there is a cool circular dependency happening through meter/tasks. So maybe keep it as is

apps/condo/domains/common/utils/importDate.utils.spec.js
```diff
@@ -0,0 +1,91 @@
...
+
+    describe('isDateStrValid', () => {
+        it('should return true for valid date strings with default formats', () => {
+            validDateStrings.forEach(dateStr => {
+                expect(isDateStrValid(dateStr)).toBe(true)
+            })
+        })
+
+        it('should return false for invalid date strings', () => {
+            invalidDateStrings.forEach(dateStr => {
```
SavelevMatthew, line 51: same

apps/condo/domains/common/utils/importDate.utils.spec.js
```diff
@@ -0,0 +1,91 @@
...
+            expect(tryToISO('2024-01-01T00:00:00.000Z')).toBe('2024-01-01T00:00:00.000Z')
+            expect(tryToISO('2024-01-01')).toBe(dayjs('2024-01-01T00:00:00').toISOString()) // Default to UTC start
+        })
+
+        it('should return undefined for invalid date strings', () => {
+            const invalidDatesOrNotForFormat = [
+                '[]',
+                '2024-01-01',
+            ]
+            invalidDatesOrNotForFormat.forEach(dateStr => {
```
SavelevMatthew, line 79: test.each()

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
```diff
@@ -92,127 +91,57 @@ const ERRORS = {
...
- */
-function normalizeMeterValue (value) {
-    if (!(isString(value) || isNumber(value) || isUndefined(value))) return null
-    if (isUndefined(value)) return value
-
-    const transformedValue = String(value).trim().replaceAll(',', '.')
-    if (isEmpty(transformedValue)) return undefined
-    return String(Number(transformedValue))
+function tryToISO (dateStr) {
+    return tryToISOUtils(dateStr, [DATE_FORMAT])
```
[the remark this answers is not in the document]
SavelevMatthew: You're checking validness in 2 formats, but converting only 1. Why?

apps/condo/domains/common/utils/importDate.utils.js
SavelevMatthew, line 1: I guess this one should be called just `import.utils.js`, so we can put more utils in here, related to phone / strings and so on...

Or we should group them into folder. Flat list like so is hard I guess:
```
importDate.utils.js
importPhone.utils.js
importString.utils.js
...
```

apps/condo/domains/meter/constants/registerMetersReadingsService.js
SavelevMatthew: Does this mean, that after adding a new field to meterReading, I should also add it in paths, and in translations...

I guess it's not critical here and can be skipped, since it's not frequently-changed model, but it seems strange for me. I've already got required fields and names in schema. why do I need to redefine new ones?

apps/condo/domains/meter/constants/registerMetersReadingsService.js
```diff
@@ -0,0 +1,27 @@
...
+    { path: 'meterMeta.verificationDate', nullable: true },
+    { path: 'meterMeta.nextVerificationDate', nullable: true },
+    { path: 'meterMeta.installationDate', nullable: true },
+    { path: 'meterMeta.commissioningDate', nullable: true },
+    { path: 'meterMeta.sealingDate', nullable: true },
+    { path: 'meterMeta.controlReadingsDate', nullable: true },
+]
+
+const DATE_FIELD_PATH_TO_TRANSLATION = {
+    'date': 'meter.import.column.meterReadingSubmissionDate',
```
SavelevMatthew, line 14: Why can't we just map it like so: 
```
(field) =>  `meter.import.column.${field}.name`
```

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
```diff
@@ -68,12 +65,14 @@ const ERRORS = {
...
         message: 'Invalid meter values',
         messageForUser: 'api.meter.registerMetersReadings.INVALID_METER_VALUES',
+        messageInterpolation: { valuesList: '"{column}"="{errorValue}"' },
     },
     MULTIPLE_METERS_FOUND: {
         code: BAD_USER_INPUT,
         type: MULTIPLE_METERS_FOUND,
         message: 'Multiple meters found',
         messageForUser: 'api.meter.registerMetersReadings.MULTIPLE_METERS_FOUND',
+        messageInterpolation: { count: '??' },
```
SavelevMatthew, line 75: ??)

SavelevMatthew, line 75: is this intended behaviour?

apps/condo/domains/meter/tasks/importMeters.spec.js
```diff
@@ -18,16 +21,26 @@ const {
...
 
 const { keystone } = index
 const dvAndSender = { dv: 1, sender: { dv: 1, fingerprint: faker.datatype.uuid() } }
 
+
+const MOCK_FOLDER = 'mock-files'
+
+const readMockFile = (fileName) => {
+    // nosemgrep: javascript.lang.security.audit.path-traversal.path-join-resolve-traversal.path-join-resolve-traversal
+    return fs.readFileSync(path.join(__dirname, MOCK_FOLDER, fileName))
```
SavelevMatthew, line 41: https://nodejs.org/api/os.html#ostmpdir

apps/condo/domains/meter/utils/meter.utils.js
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * @param value
+ * @return {undefined|string|null}
+ */
+function normalizeMeterValue (value) {
+    if (!(isString(value) || isNumber(value) || isUndefined(value))) return null
+    if (isUndefined(value)) return value
+
+    const transformedValue = String(value).trim().replaceAll(',', '.')
```
SavelevMatthew, line 12: What's this used for? If it's import util, move it to them....

SavelevMatthew: If it's API util, should not modify user data. Only validate them. Since writing numbers is locale-specific. API must only accept `xyz.xyz` format

apps/condo/domains/meter/utils/meter.utils.js
SavelevMatthew, line 1: Right now as a developer, what's this used for. If it's for API, then all utils are wrong. If it's for import, move it to import utils

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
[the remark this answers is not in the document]
YEgorLu: because utc date string already in utc

apps/condo/domains/meter/constants/registerMetersReadingsService.js
YEgorLu: we would need to map meterMeta.verificationDate to VerificationDate beforehand and so on
we would need to map date to meterReadingSubmissionDate beforehand

So we would just add more mappings

apps/condo/domains/meter/utils/meter.utils.js
YEgorLu: made another task for that. now marked with @deprecated and TODO

apps/condo/domains/meter/utils/meter.utils.js
YEgorLu: made another task for that. now marked with @deprecated and TODO

apps/condo/domains/common/utils/import/importDate.utils.js
SavelevMatthew, line 1: You have 2 words `date`, `import` and `utils` in your path. 

Why not just:
`apps/condo/domains/common/utils/import/date.js`
or
``apps/condo/domains/common/utils/import/dateField.js``

apps/condo/domains/common/utils/import/index.js
```diff
@@ -0,0 +1,5 @@
+const importDateUtils = require('@condo/domains/common/utils/import/importDate.utils')
+
+module.exports = {
+    ...importDateUtils,
```
SavelevMatthew, line 4: Its' a bad practice to have dynamic exports, since you can export something wrong from your module, and it will be chained here.

Good practice to explicitly specify, what you want others to use

apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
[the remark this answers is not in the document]
SavelevMatthew: And what? You have an util which accepts string in any of specified formats and converts it to UTC, right?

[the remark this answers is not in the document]
SavelevMatthew: So, from my perspective it should not work on other formats....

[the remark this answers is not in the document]
SavelevMatthew: In your case you expect it to work with UTC, but your interface (passing args) does not

packages/keystone/file/utils.js
```diff
@@ -1,11 +1,12 @@
...
+const { createReadStream } = require('fs')
 const path = require('path')
 const { Duplex } = require('stream')
 
 const INVISIBLE_CHARS_REGEXP = new RegExp('[\u200B-\u200D\uFEFF]', 'g')
 
 const clearString = (cell) => {
-    if (!cell) {
+    // number 0 should be parsed as '0', not ''
+    if (!cell && typeof cell !== 'number') {
```
SavelevMatthew, line 9: Where is it used? Maybe you should check for null / undefined, instead of number?

SavelevMatthew: Right now seems like you're trying to solve specific problem...

What if `false` or other false values goes here? Will it be more `if`s?

YEgorLu: right now only meters import using that code, so there goes strings and numbers, so these changes should be enough

made checks for null / undefined, so 'false' as boolean can be parsed.
But it's hard right now to imagine getting something other than number or string from file

apps/condo/domains/meter/constants/registerMetersReadingsService.js
YEgorLu: If you add date field in schema it does not mean that it automatically starts parsing in registerMetersReadings or import, does not mean that translations in /lang appeared by themselves. So you will check custom mutation and task to modify logic, see that translations / validations depending on these variables and will modify them.

So i don't thinks there is any issue, maybe this even will make you change only 2 lines here, instead of 4 lines in different files

SavelevMatthew (approved):
It's been a long road that turned out long and not without regexps, but thanks for your patience. Good job 🏅 

Remember, the API should always be strictly standardized, and the data processing should be handled by frontend or bffs

apps/condo/domains/common/utils/import/date.js
```diff
@@ -0,0 +1,143 @@
+const dayjs = require('dayjs')
+const customParseFormat = require('dayjs/plugin/customParseFormat')
+const utc = require('dayjs/plugin/utc')
+const { isEmpty, isString } = require('lodash')
+
+const { DEFAULT_DATE_PARSING_FORMATS } = require('@condo/domains/common/constants/import')
+
+dayjs.extend(customParseFormat)
```
pahaz, line 8: Is it better to move it to _app ?

PR #5299, opened 2024-09-30
hotfix(codegen): INFRA-560 fixed useAllObjects

Alllex202:
In `apollo/client` version 3.11.3 (we have 3.11.8) the the logic of `fetchMore` was changed, because of this we began to receive an error: _You must provide an updateQuery function when using fetchMore with a no-cache fetch policy._
[Link to documentation](https://www.apollographql.com/docs/react/pagination/core-api/#using-fetchmore-with-queries-that-set-a-no-cache-fetch-policy)

### DIFF
--- a/packages/codegen/generate.hooks.ts
+++ b/packages/codegen/generate.hooks.ts
@@ -4,6 +4,7 @@ import { FetchMoreQueryOptions } from '@apollo/client/core/watchQueryOptions'
 import { TypedDocumentNode } from '@graphql-typed-document-node/core'
 import dayjs from 'dayjs'
 import { DocumentNode } from 'graphql'
+import get from 'lodash/get'
 import isFunction from 'lodash/isFunction'
 import { useCallback, useEffect, useRef, useState } from 'react'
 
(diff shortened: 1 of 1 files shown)


packages/codegen/generate.hooks.ts
```diff
@@ -379,7 +380,11 @@ export function generateReactHooks<
     }
 
     function useAllObjects (variables: QueryVariables, options?: QueryHookOptions<IUseObjectsQueryReturnType<GQLObject>, QueryVariables>) {
-        const { objs, count, error, loading, refetch: _refetch, fetchMore, stopPolling } = useObjects(variables, options)
+        const fetchPolicy = get(options, 'fetchPolicy')
+        const { objs, count, error, loading, refetch: _refetch, fetchMore, stopPolling } = useObjects(variables, {
+            ...options,
+            fetchPolicy: fetchPolicy && fetchPolicy !== 'no-cache' ? fetchPolicy : 'network-only',
```
sitozzz, line 386: Network only is default policy?

sitozzz: Oh, I miss description and doc link...

sitozzz: Anyway I can't understand why you compare user input of fetchPolicy with `no-cache` value?

Alllex202: We use no-cache by default, but "fetchMore" now works differently with "no-cache". So to restore functionality of this hook, we had to make it "network-only".
To achieve the expected behavior of the hook, I need to make sure that "no-cache" does not get here

SavelevMatthew (approved):

abshnko (approved):

sitozzz (approved):

AleX83Xpert (approved):

PR #5309, opened 2024-10-02
feat(condo): DOMA-10141 add column address and unitName to table in marketplace

tolmachev21:
Problem: 
Currently we do not have an address and unit column in the table on the marketplace page.
However, we would like to be able to sort invoices by address and unit, because users do not remember the invoice number and the only way to identify an invoice is by address and unit.
Therefore, these two columns were added with the ability to filter in them

View before: 
Screenshot 2024-10-02 at 21 05 30

View now (locally) _**without address-service provider**_: 
Screenshot 2024-10-04 at 16 59 30


nomerdvadcatpyat (approved):

apps/condo/domains/marketplace/hooks/useMarketplaceInvoicesFilters.tsx
```diff
@@ -41,6 +47,13 @@ export function useMarketplaceInvoicesFilters (): Array<FiltersMeta<InvoiceWhere
     const EndDateMessage = intl.formatMessage({ id: 'pages.condo.meter.EndDate' })
     const NumberMessage = intl.formatMessage({ id: 'ticketsTable.Number' })
     const SelectMessage = intl.formatMessage({ id: 'Select' })
+    const EnterAddressMessage = intl.formatMessage({ id: 'pages.condo.meter.EnterAddress' })
+    const AddressMessage = intl.formatMessage({ id: 'field.Address' })
+    const EnterUnitNameMessage = intl.formatMessage({ id: 'pages.condo.ticket.filters.EnterUnitName' })
```
Alllex202, line 52: ticket?

apps/condo/domains/marketplace/hooks/useMarketplaceInvoicesFilters.tsx
```diff
@@ -41,6 +47,13 @@ export function useMarketplaceInvoicesFilters (): Array<FiltersMeta<InvoiceWhere
     const EndDateMessage = intl.formatMessage({ id: 'pages.condo.meter.EndDate' })
     const NumberMessage = intl.formatMessage({ id: 'ticketsTable.Number' })
     const SelectMessage = intl.formatMessage({ id: 'Select' })
+    const EnterAddressMessage = intl.formatMessage({ id: 'pages.condo.meter.EnterAddress' })
```
Alllex202, line 50: meter?

apps/condo/domains/marketplace/hooks/useMarketplaceInvoicesTableColumns.tsx
```diff
@@ -78,11 +88,36 @@ export const useMarketplaceInvoicesTableColumns = ({ filtersMeta }) => {
             filterDropdown: getFilterDropdownByKey(filtersMeta, 'ticket'),
             filterIcon: getFilterIcon,
         },
+        {
+            title: AddressTitle,
+            key: 'property',
+            dataIndex: 'property',
+            width: '15%',
```
Alllex202, line 95: Why is the total width of your columns 125%? It was 100%

apps/condo/domains/marketplace/hooks/useMarketplaceInvoicesTableColumns.tsx
```diff
@@ -78,11 +88,36 @@ export const useMarketplaceInvoicesTableColumns = ({ filtersMeta }) => {
...
+            render: renderAddress,
+            sorter: true,
```
Alllex202, line 97: In fact, sorting by addresses will not work as expects (alphabetically). But ok, let it remain

tolmachev21: There is sorting in the ticket table. Maybe someone will use it

Alllex202 (approved):

PR #5243, opened 2024-09-16
feat(condo): DOMA-10226 add full stacked gql error traces

pahaz:
Improve developer experience. And show what the real original error. And provide a full stack trace with internal and intermediate errors. 
Screenshot 2024-10-03 at 13 11 40

```
    GQLError: [error] Create Meter internal error
        at _throwIfError (~/packages/codegen/generate.server.utils.js:67:15)
        at _throwIfError (~/packages/codegen/generate.server.utils.js:140:5)
        at runMicrotasks (<anonymous>)
        at processTicksAndRejections (node:internal/process/task_queues:96:5)
        at Object.create (~/packages/codegen/generate.server.utils.js:213:16)
        at resolver (~/apps/condo/domains/meter/schema/RegisterMetersReadingsService.js:397:50)
    ^- Caused By: Error: Unable to connect a Meter.resource<MeterResource>
        at ~/node_modules/@keystonejs/keystone/lib/ListTypes/utils.js:44:23
        at runMicrotasks (<anonymous>)
        at processTicksAndRejections (node:internal/process/task_queues:96:5)
        at List._resolveRelationship (~/node_modules/@keystonejs/keystone/lib/ListTypes/list.js:617:35)
        at ~/node_modules/@keystonejs/keystone/lib/ListTypes/list.js:741:26
        at List._nestedMutation (~/node_modules/@keystonejs/keystone/lib/ListTypes/list.js:683:35)
        at List._createSingle (~/node_modules/@keystonejs/keystone/lib/ListTypes/list.js:733:12)
        at List.createMutation (~/node_modules/@keystonejs/keystone/lib/ListTypes/list.js:713:12)
        at List.createMutation (~/packages/keystone/requestCache.js:151:16)
        at execGqlWithoutAccess (~/packages/codegen/generate.server.utils.js:131:30)
        at Object.create (~/packages/codegen/generate.server.utils.js:213:16)
        at resolver (~/apps/condo/domains/meter/schema/RegisterMetersReadingsService.js:397:50)
    ^- Caused By: Error: Unable to connect a Meter.resource<MeterResource>
        at throwWithErrors (~/node_modules/@keystonejs/fields/dist/fields.cjs.dev.js:1949:17)
        at resolveNestedSingle (~/node_modules/@keystonejs/fields/dist/fields.cjs.dev.js:2183:7)
        at runMicrotasks (<anonymous>)
        at processTicksAndRejections (node:internal/process/task_queues:96:5)
[...]

PR #5314, opened 2024-10-03
feat(global): DOMA-10386 add script to test translations

YEgorLu:
Recently there were changes in many translation keys, so we want to somehow test, that there is no problems with them

There is an idea to write script, which would scan files and tell if there are invalid translation keys in code, or unused translations in (ru|en).json

I used @babel/parser, @babel/travarse to make and traverse AST tree for .js, .ts, .jsx, .tsx. Looking for occurrences like { messageForUser: <key> } ; i18n(<key>) ; itntl.formatMessage({ id: <key> })
There is restrictions that <key> is string, not variable, and keys (messageForUser, id) must be written explicitly, not by spreading objects or something else. otherwise it's more work on script. As of i seen they usually being used like that.

@babel/parse @babel/traverse already being used implicitly in code packages, so i don't think there is a problem using them

### DIFF
--- a/bin/test-translations.js
+++ b/bin/test-translations.js
@@ -1,0 +1,285 @@
+const path = require('path')
+const fs = require('fs')
+const parser = require('@babel/parser')
+const traverse = require('@babel/traverse').default
+
+/*
+    Inspects .ts, .js, .tsx, .jsx files in all apps inside folders "domain", "pages"
+    Searching for occurrences like { messageForUser: <key> } ; i18n(<key>) ; itntl.formatMessage({ id: <key> })
+    Does not work if <key> isn't string
+    Saves output in ./bin/.local/test.json in format {
+        [pathToApp]: {
+            {filePath, lang, translationKey, message}
+        }
+    }
+    where message =
+    'Code references undefined translation' - code translates key, which not present in en.json | ru.json
+    ||
+    'Unused translation' - translation exists in ru.json | en.json, but isn't used in code
+ */
+
+const APPS_TO_EXCLUDE = [
+    'address-service'
+]
+
+const FILE_EXTENSIONS_TO_INCLUDE = [
+    'js',
+    'ts',
+    'jsx',
+    'tsx',
+]
+
+function getDirectoriesNames (source) {
+    return fs.readdirSync(source, {withFileTypes: true})
+        .filter(dirent => dirent.isDirectory())
+        .map(dirent => dirent.name)
+}
+
+function getLanguageJSONPaths (source) {
+    const subFiles = fs.readdirSync(source, {withFileTypes: true})
+    let enJson = subFiles.find(v => v.name === 'en.json')
+    let ruJson = subFiles.find(v => v.name === 'ru.json')
+
+    const throwIfNotBothFilesDefined = (ruJson, enJson) => {
+        if (ruJson === !enJson) {
+            throw new Error(`Does not have both translation files at ${source}`)
+        }
+    }
+
+    const pathToTranslation = (source, name) => {
+        return path.join(source, name)
+    }
+
+    const formatResult = (ruJson, enJson) => {
+        return {
+            en: path.join(enJson.parentPath, enJson.name),
+            ru: path.join(ruJson.parentPath, ruJson.name),
+        }
+    }
+
+    if (!enJson && !ruJson) {
+        const enJsonFolderPath = path.join(source, 'en')
+        const ruJsonFolderPath = path.join(source, 'ru')
+        const enJsonFolder = fs.readdirSync(enJsonFolderPath, {withFileTypes: true})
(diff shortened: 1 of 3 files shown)


bin/test-translations.js
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+
+    const pathToTranslation = (source, name) => {
+        return path.join(source, name)
+    }
+
+    const formatResult = (ruJson, enJson) => {
+        return {
+            en: path.join(enJson.parentPath, enJson.name),
```
SavelevMatthew, line 55: Hard-code is not allowed

bin/test-translations.js
```diff
@@ -0,0 +1,285 @@
...
+
+function getDirectoriesNames (source) {
+    return fs.readdirSync(source, {withFileTypes: true})
+        .filter(dirent => dirent.isDirectory())
+        .map(dirent => dirent.name)
+}
+
+function getLanguageJSONPaths (source) {
+    const subFiles = fs.readdirSync(source, {withFileTypes: true})
+    let enJson = subFiles.find(v => v.name === 'en.json')
```
SavelevMatthew, line 40: same here

bin/test-translations.js
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+    where message =
+    'Code references undefined translation' - code translates key, which not present in en.json | ru.json
+    ||
+    'Unused translation' - translation exists in ru.json | en.json, but isn't used in code
+ */
+
+const APPS_TO_EXCLUDE = [
+    'address-service'
```
SavelevMatthew, line 22: why?

bin/test-translations.js
```diff
@@ -0,0 +1,285 @@
...
+
+const STATE = {errors: 0, processed: 0, complete: 0}
+
+function findTranslationKeysInCode (filePath) {
+    const contents = fs.readFileSync(filePath, 'utf-8')
+    let ast
+    try {
+        const sourceType = filePath.endsWith('.js') ? 'script' : 'module'
+        ast = parser.parse(contents, {
+            allowImportExportEverywhere: true,
```
SavelevMatthew, line 119: Should not be hard-coded too, why exactly this config?

bin/test-translations.js
```diff
@@ -0,0 +1,285 @@
...
+
+                if (callee.type === 'Identifier' && callee.name === 'i18n' && path.node.arguments.length > 0) {
+                    const arg = path.node.arguments[0];
+
+                    if (arg.type === 'StringLiteral') {
+                        translationKeys.add(arg.value);
+                    }
+                }
+
+                // Check if the call is to `intl.formatMessage`
```
SavelevMatthew, line 176: What about <FormattedMessage> component

bin/test-translations.js
SavelevMatthew, line 1: What are you testing? I guess lint is more appropriate for this case

bin/test-translations.js
```diff
@@ -0,0 +1,285 @@
+const path = require('path')
+const fs = require('fs')
+const parser = require('@babel/parser')
```
SavelevMatthew, line 3: Do we have any faster alternatives? I've heard babel is slow

bin/test-translations.js
```diff
@@ -0,0 +1,285 @@
...
+const traverse = require('@babel/traverse').default
+
+/*
+    Inspects .ts, .js, .tsx, .jsx files in all apps inside folders "domain", "pages"
+    Searching for occurrences like { messageForUser: <key> } ; i18n(<key>) ; itntl.formatMessage({ id: <key> })
```
SavelevMatthew, line 8:
```suggestion
    Searching for occurrences like { messageForUser: <key> } ; i18n(<key>) ; intl.formatMessage({ id: <key> })
```

Alllex202:
Are you going to run this in CI?

We have a script `bin/lint-i18n-translations.js`, what will happen to it?

In fact, you do not check all the places where there may be translations. At least there is `<FormattedMessage id={<some-key>} .../>` (possibly something else)

What about translations in code that are calculated at runtime?

bin/test-translations.js
```diff
@@ -0,0 +1,285 @@
...
+function getDirectoriesNames (source) {
+    return fs.readdirSync(source, {withFileTypes: true})
+        .filter(dirent => dirent.isDirectory())
+        .map(dirent => dirent.name)
+}
+
+function getLanguageJSONPaths (source) {
+    const subFiles = fs.readdirSync(source, {withFileTypes: true})
+    let enJson = subFiles.find(v => v.name === 'en.json')
+    let ruJson = subFiles.find(v => v.name === 'ru.json')
```
pahaz, line 41: why its not for all translations?

pahaz:
Could you please wait before https://github.com/open-condo-software/condo/pull/5243
I also want to add a runtime GQLError guard inside constructor.

You solution is interesting. It's hard to maintain but may be it's not so important for linter.

PR #5319, opened 2024-10-04
feat(condo): INFRA-538 migrate condo domains to use new generate.server.utils

ekabardinsky:
Migrate models from acquiring, analytics, banking, billing domains to use new way to interact with DB models

### DIFF
--- a/apps/condo/bin/billing/discoverServiceConsumers.js
+++ b/apps/condo/bin/billing/discoverServiceConsumers.js
@@ -8,14 +8,13 @@
 const path = require('path')
 
 const commander = require('commander')
-const { map } = require('lodash')
+const { map, chunk } = require('lodash')
 
 const { getLogger } = require('@open-condo/keystone/logging')
 const { prepareKeystoneExpressApp } = require('@open-condo/keystone/prepareKeystoneApp')
+const { find } = require('@open-condo/keystone/schema')
 
-const { BillingAccount } = require('@condo/domains/billing/utils/serverSchema')
 const { UUID_REGEXP } = require('@condo/domains/common/constants/regexps')
-const { loadListByChunks } = require('@condo/domains/common/utils/serverSchema')
 const { discoverServiceConsumers } = require('@condo/domains/resident/utils/serverSchema')
 
 const logger = getLogger('discoverServiceConsumersScript')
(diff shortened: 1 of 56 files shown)


apps/condo/domains/common/utils/serverSchema/index.js
```diff
@@ -234,7 +236,16 @@ const loadListByChunks = async ({
             throw new Error('Operation timed out')
         }
 
-        newChunk = await list.getAll(context, where, { sortBy, first: chunkSize, skip: skip })
+        // TODO INFRA-538 remove this condition once migration to new server utils to be done
+        const resolvedFields = !isNil(fields) ? fields : 'id'
```
AleX83Xpert, line 240: Is it possible to make only one place where we say that `'id'` is default fragment?

apps/condo/domains/resident/schema/DiscoverServiceConsumersService.js
```diff
@@ -150,29 +151,21 @@ const DiscoverServiceConsumersService = new GQLCustomSchema('DiscoverServiceCons
...
-                                ...organizationsToAcquiringContextsMap,
-                                [organizationId]: [
-                                    ...(organizationsToAcquiringContextsMap[organizationId] || []),
-                                    acquiringContextId,
-                                ],
-                            }
-                        })
-                        return []
-                    },
+                const allAcquiringIntegrationContext = find('AcquiringIntegrationContext', {
```
AleX83Xpert, line 154:
```suggestion
                const allAcquiringIntegrationContexts = find('AcquiringIntegrationContext', {
```

apps/condo/domains/common/utils/serverSchema/index.js
ekabardinsky: I'll remove this code once migration to be done. This part of code - like a notification for developers: you should rewrite code where you calling loadListByChunks

apps/condo/bin/billing/discoverServiceConsumers.js
```diff
@@ -43,26 +42,23 @@ async function main () {
...
-            try {
-                const result = await discoverServiceConsumers(context, { ...DV_SENDER, billingAccountsIds })
-                logger.info({ msg: `chunk[${chunk.length}]`, billingAccountsIds, result })
-            } catch (err) {
-                logger.error({ msg: `chunk[${chunk.length}] error`, billingAccountsIds, err })
-            }
-
-            return []
-        },
+    const accounts = await find('BillingAccount', {
```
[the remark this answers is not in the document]
ekabardinsky: This is a bin script and going to be called from the command line. I guess it is better to use more memory instead of load PG server CPU.

nomerdvadcatpyat (approved):

PR #5320, opened 2024-10-04
feat(codegen): INFRA-538 support fields params for loadListByChunks util

ekabardinsky:


packages/codegen/generate.server.utils.js
```diff
@@ -641,6 +640,7 @@ function generateServerUtils (gqlOrSchemaName) {
         delete: delete_,
         softDelete,
         softDeleteMany,
+        hasFieldsParam: true, // TODO INFRA-538 remove this once migration to new server utils to be done
```
Alllex202, line 643: What is this for?

Alllex202: Ah, I see
Old utilities don't have this flag

Please write a comment about it

ekabardinsky: This a temporary flag, just to notify developers - we have not migrated loadListByChunks utility to the new codegen utils.

apps/condo/domains/common/utils/serverSchema/index.js
```diff
@@ -234,7 +236,16 @@ const loadListByChunks = async ({
             throw new Error('Operation timed out')
         }
 
-        newChunk = await list.getAll(context, where, { sortBy, first: chunkSize, skip: skip })
+        // TODO INFRA-538 remove this condition once migration to new server utils to be done
+        const resolvedFields = !isNil(fields) ? fields : 'id'
+        if (get(list, 'hasFieldsParam') && isNil(fields)) {
+            // the case when we migrate model to use new server utils, but loadListByChunks use it in old way
+            console.trace('Provide required fields parameter for loadListByChunks util')
+        }
```
Alllex202, line 244: Do you want the fields to be explicitly passed?
Isn't it enough that there is a default value ("id")?

ekabardinsky: Yes, 99.9% usages of loadListByChunks - get model data, not only ids. There are another case: model was migrated to the new codegen utils, but loadListByChunks are calling in old way. So I would like to notify developer about such case - to trigger him to rewrite code and provide required "fields" param.

AleX83Xpert (approved):

SavelevMatthew (approved):

abshnko (approved):

PR #5327, opened 2024-10-07
hotfix(condo): refactored useAllObjects hook

ekabardinsky:


Alllex202 (approved):

toplenboren (approved):

packages/codegen/generate.hooks.ts
```diff
@@ -380,64 +380,58 @@ export function generateReactHooks<
...
-                const { data: fetchedData } = await fetchMore({
+                await fetchMore({
                     variables: {
                         skip: skip,
                     },
+                    updateQuery (previousData, { fetchMoreResult }) {
+                        // @ts-ignore
+                        const updatedObjs = [...previousData.objs, ...fetchMoreResult.objs]
+                        // @ts-ignore
+                        return { ...previousData, objs: updatedObjs, count: fetchMoreResult.count }
```
toplenboren, line 409: why tsignore is here? Best to fix it

packages/codegen/generate.hooks.ts
```diff
@@ -380,64 +380,58 @@ export function generateReactHooks<
...
-        }, [count, data.length, error, fetchMoreError, firstPageLoaded, loadMore, skip])
+            if (objs.length >= count) {
+                return
+            }
+
+            loadMore(objs.length)
+        }, [loadMore, objs.length, count])
 
-        const allDataLoaded = data.length === 0 ? objs.length === count : data.length === count
+        const allDataLoaded = objs.length >= count && !loading
```
toplenboren, line 429: boolean stuff is best prefixed with is, but this is just a nitpick :-)

packages/codegen/generate.hooks.ts
Alllex202: The problem with types.
I'm not sure we'll do anything about it

packages/codegen/generate.hooks.ts
Alllex202: Yes, but this is an old name that is used everywhere, so we will not change it.

PR #5334, opened 2024-10-09
Feat(condo): DOMA-10367 add fields rawAddress and recipient in Payment

YEgorLu:
"rawAddres" is address as it was imported. When company exporting payments back they expect to match them with their data by addres, but we normalizing it instead. So expecting to use this field just for that

"recipient" is like recipient in billing receipt. We will need to have full requisites later

We need to extend Payment model to have all the fields that we are using to create PaymentRegistryFiles
Later we can give an API to organizations (getAllPayments) to automate their workflows


apps/condo/domains/acquiring/schema/Payment.js
```diff
@@ -271,17 +272,29 @@ const Payment = new GQLListSchema('Payment', {
...
         recipientBankAccount: {
-            schemaDoc: 'Bank account number of recipient organization, used for matching payments with receipts in case of multiple receipts per account + address',
+            schemaDoc: '@deprecated --use "recipient" Bank account number of recipient organization, used for matching payments with receipts in case of multiple receipts per account + address',
             type: 'Text',
             isRequired: true,
         },
 
+        recipient: {
+            ...RECIPIENT_FIELD,
+            schemaDoc: 'Recipient. Should contain all meta information to identify the organization',
```
dkoviazin, line 288: and bank account

apps/condo/domains/acquiring/schema/Payment.js
```diff
@@ -271,17 +272,29 @@ const Payment = new GQLListSchema('Payment', {
...
+            isRequired: false,
+        },
+
+        rawAddress: {
```
dkoviazin, line 292: Non-normalized address that was imported from the organization

apps/condo/domains/acquiring/schema/fields/Recipient.js
```diff
@@ -9,6 +9,7 @@ const RECIPIENT_FIELDS_DEFINITION = {
     offsettingAccount: 'String',
     tin: 'String!',
     iec: 'String',
+    // NOTE(YEgorLu): rename to routingNumber
```
dkoviazin, line 12: better to add TODO to make all recipients models and bankAccounts to use the same naming

apps/condo/migrations/20241009105402-0433_payment_rawaddress_payment_recipient_and_more.js
```diff
@@ -0,0 +1,58 @@
...
+--
+ALTER TABLE "PaymentHistoryRecord" DROP COLUMN "recipient" CASCADE;
+--
+-- Add field rawAddress to paymenthistoryrecord
+--
+ALTER TABLE "PaymentHistoryRecord" DROP COLUMN "rawAddress" CASCADE;
+--
+-- Add field recipient to payment
+--
+ALTER TABLE "Payment" DROP COLUMN "recipient" CASCADE;
```
dkoviazin, line 50: why CASCADE everywhere?

YEgorLu: It's default. Dangerous if fields were relations, but they are not

abshnko (approved):

PR #5340, opened 2024-10-10
refactor(registry): INFRA-591 update server utils in registry

nomerdvadcatpyat:


apps/registry
sitozzz, line 1: Strange commit point here, isn't it?

nomerdvadcatpyat: No, just we have a non main point on registry in condo

sitozzz (approved):

SavelevMatthew (approved):

abshnko (approved):

PR #5342, opened 2024-10-11
refactor(condo): DOMA-10456 trying to increase performance of `_inter…

AleX83Xpert:
…nalSyncContactsWithResidentsForOrganization`


apps/condo/domains/contact/schema/_internalSyncContactsWithResidentsForOrganizationService.js
```diff
@@ -47,55 +50,52 @@ const _internalSyncContactsWithResidentsForOrganizationService = new GQLCustomSc
...
-                        const existedContacts = await Contact.getAll(context, {
-                            organization: { id: organization.id },
+                    const user = await getById('User', resident.user)
+                    const { unitType, unitName, property } = resident
+                    const { phone, email, name } = user
+
+                    if (phone) {
+                        const existedContacts = await find('Contact', {
                             phone: phone,
+                            organization: { id: organization.id },
```
Alllex202, line 69: Should we filter by a property a resident belongs to?

Alllex202: And unitName/unitType?

Alllex202: Although the user can enter the unit incorrectly, so it is not worth it. But why do not filter by property? This means that if I am a resident with one phone number in two properties, then only one resident is synchronized for me, not two

AleX83Xpert: ¯\\\_(ツ)_/¯
The existing filtering is quite suitable for users. And there are no requests to change it.

AleX83Xpert: But yes, in case you described we will create only one contact instead of two.

AleX83Xpert: > in case you described we will create only one contact instead of two.

This looks ok according to tests

apps/condo/domains/contact/schema/_internalSyncContactsWithResidentsForOrganizationService.js
```diff
@@ -47,55 +50,52 @@ const _internalSyncContactsWithResidentsForOrganizationService = new GQLCustomSc
...
+
                 const createdContacts = []
+
                 for (const resident of residents) {
-                    const user = await getById('User', resident.user.id)
-                    const phone = user.phone
-                    if (phone){
-                        const existedContacts = await Contact.getAll(context, {
-                            organization: { id: organization.id },
+                    const user = await getById('User', resident.user)
```
sitozzz, line 62: Maybe check for user is not deleted here?

Alllex202 (approved):

apps/condo/domains/contact/schema/_internalSyncContactsWithResidentsForOrganizationService.js
sitozzz: Now it's more clear imo

sitozzz (approved):

PR #5346, opened 2024-10-11
fix(infra): DOMA-10462 update certificate (because of alert)

AleX83Xpert:
copy-paste cert from condo->pfx_extended to condo->pfx and to registryma->pfx


.helm
sitozzz, line 1: You're had multiple changes inside that update. Was that how it was planned?

sitozzz (approved):

.helm
AleX83Xpert: All changes were synced

dkoviazin (approved):

PR #5364, opened 2024-10-15
refactor(global): INFRA-592 update server utils in eps and condorb 

nomerdvadcatpyat:


packages/codegen/generate.server.utils.js
```diff
@@ -335,6 +335,11 @@ function _generateServerUtilsDeprecated (gql) {
     }
 }
 
+function wrapIfNeeded (fields) {
```
SavelevMatthew, line 338:
```suggestion
function _wrapFileldsIfNeeded (fields) {
```

SavelevMatthew (approved):
Good job! 🔥

packages/codegen/generate.server.utils.js
Alllex202, line 338: Please don't add `IfNeeded` to function name, let it be `IfMissing`

Maybe `wrapInBracesIfMissing`/`addBracesIfMissing` or `...CurlyBraces...` would be better?

Alllex202 (approved):

PR #5370, opened 2024-10-16
feat(condo): DOMA-10397 add act of paid works in ticket

tolmachev21:
Problem:
We need to generate a paid works act. Similar to how it happens with the act of completion work. This functionality will help our clients work with the marketplace with greater comfort.

Solution:
This document is available under the following conditions: Russian locale, paid application and availability of an invoice.

The generated document is attached.

[cm2bxzpsx005whcs686014jhn-paid_works_ticket_da79a441-f81a-4fe5-9e21-401c50a61bd1_16_10_2024.docx](https://github.com/user-attachments/files/17396956/cm2bxzpsx005whcs686014jhn-paid_works_ticket_da79a441-f81a-4fe5-9e21-401c50a61bd1_16_10_2024.docx)

Some data is loaded in runtime via FinanceInfoClient. Therefore, the case when the application is old, but some details have changed - is not processed


apps/condo/package.json
```diff
@@ -166,6 +166,7 @@
     "react-yandex-metrika": "^2.6.0",
     "recheck": "^4.4.5",
     "redlock": "^5.0.0-beta.2",
+    "rubles": "^0.2.0",
```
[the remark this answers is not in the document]
toplenboren: I think it is better to just inline the code

apps/condo/lang/en/en.json
```diff
@@ -917,6 +917,7 @@
   "pages.condo.ticket.exportBlank.HideAllComments.label": "Hide comments",
   "pages.condo.ticket.generateDocument.label": "Generate document",
   "pages.condo.ticket.generateDocument.completionWorks.label": "Work completion certificate",
+  "pages.condo.ticket.generateDocument.paidWorks.label": "Work paid certificate",
```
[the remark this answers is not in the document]
toplenboren: Better to name Work completion certificate (paid)

apps/condo/domains/ticket/hooks/useTicketDocumentGenerationTask.tsx
```diff
@@ -52,8 +53,13 @@ export const useTicketDocumentGenerationTask = ({ ticket, user, timeZone }) => {
                 key: 'generate-document-of-completion-works',
                 onClick: getHandleClick(TICKET_DOCUMENT_TYPE.COMPLETION_WORKS),
             },
+            ticket.isPaid && get(invoices, '0') && isSupportedDocumentTypeByLocale(TICKET_DOCUMENT_TYPE.PAID_WORKS, locale) && {
+                label: PaidWorksLabel,
+                key: 'generate-document-of-paid-works',
```
toplenboren, line 58: I'd have vent with

document-of-works-completion
document-of-paid-works-completion

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,159 @@
...
+const renderMoney = (amount, currencyCode, locale) => {
+    const options = { currency: currencyCode }
+    const numberFormat = new Intl.NumberFormat(locale, options)
+    const parts = numberFormat.formatToParts(amount)
+    return parts.map((part) => part.value).join('')
+}
+
+
+const generateTicketDocumentOfPaidWorks = async ({ task, baseAttrs, context, locale, ticket, organization }) => {
+    const dadata = new FinanceInfoClient()
```
toplenboren, line 51: Do you need to initialaze it every time the generation is triggered?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,159 @@
...
+    let psrn = null, organizationAddress = null, iec = null
+    try {
+        ({ iec, psrn, organizationAddress } = await dadata.getOrganization(organization.tin))
+    } catch (error) {
+        console.error(error)
```
toplenboren, line 56: Best to write what you expect to happen if finance info client fails to retrieve the data

toplenboren: Not sure that just sending stuff to console is okay way to go

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,159 @@
...
+    }
+    const { format, timeZone: timeZoneFromUser } = task
+
+    const timeZone = normalizeTimeZone(timeZoneFromUser) || DEFAULT_ORGANIZATION_TIMEZONE
+    const printDate = dayjs().tz(timeZone).locale(locale)
+
+    const contact = ticket.contact
+        ? await getById('Contact', ticket.contact)
```
toplenboren, line 64: what if contact is deleted?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,159 @@
...
+        : null
+    const employee = organization.id && ticket.executor
```
toplenboren, line 66: what if employee is deleted?

toplenboren: also, why do you need to check organization.id existence ?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,159 @@
...
+        ? await getByCondition('OrganizationEmployee', {
+            organization: { id: organization.id },
+            user: { id: ticket.executor },
+        })
+        : null
+
+    const invoices = await Invoice.getAll(context, {
+        ticket: { id: ticket.id },
+        deletedAt: null,
```
toplenboren, line 75: 👍

apps/condo/domains/ticket/hooks/useTicketDocumentGenerationTask.tsx
```diff
@@ -52,8 +53,13 @@ export const useTicketDocumentGenerationTask = ({ ticket, user, timeZone }) => {
                 key: 'generate-document-of-completion-works',
                 onClick: getHandleClick(TICKET_DOCUMENT_TYPE.COMPLETION_WORKS),
             },
+            ticket.isPaid && get(invoices, '0') && isSupportedDocumentTypeByLocale(TICKET_DOCUMENT_TYPE.PAID_WORKS, locale) && {
```
toplenboren, line 56: what if invoice is deleted?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,159 @@
...
+        sortBy: ['createdAt_ASC'],
+    })
+
+    const currencyCode = get(invoices, '0.currencyCode') || DEFAULT_INVOICE_CURRENCY_CODE
+    let totalSum = 0, totalVAT = 0, vat = get(invoices, '0.rows.0.vatPercent')
+
+    const listOfWorks = invoices.reduce((acc, invoice) => {
+        const rows = Array.isArray(invoice.rows) ? invoice.rows : []
+        acc.push(...rows.map((row, index) => {
+            const price = parseFloat(!row.isMin ? row.toPay : null)
```
toplenboren, line 87: you should not use float to store prices, use decimal (big.js)

https://docs.python.org/3/tutorial/floatingpoint.html

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,159 @@
...
+
+    totalVAT = totalSum * vat / 100
+    totalVAT.toFixed(2)
+
+    const documentData = {
+        header: {
+            generalDate: printDate.format('DD.MM.YYYY'),
+        },
+        company: {
+            name: get(organization, 'name') || '_______________________',
```
toplenboren, line 112: Maybe move blank to const?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,159 @@
...
+            psrn: psrn || '_______________________',
+            tin: get(organization, 'tin') || '_______________________', 
+            iec: iec || '_______________________',
+            address: organizationAddress || '_______________________',
+            phone: '_______________________', 
+
+        },
+        bankDetails: {
+            accountNumber: get(invoices, '0.accountNumber') || '_______________________',
```
toplenboren, line 121: lodash.get has a default value as third argument

accountnumber: get(invoices, '0.accountNumber', BLANK)

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,159 @@
...
+    const invoices = await Invoice.getAll(context, {
+        ticket: { id: ticket.id },
+        deletedAt: null,
+        status_not: INVOICE_STATUS_CANCELED,
+    }, {
+        sortBy: ['createdAt_ASC'],
+    })
+
+    const currencyCode = get(invoices, '0.currencyCode') || DEFAULT_INVOICE_CURRENCY_CODE
+    let totalSum = 0, totalVAT = 0, vat = get(invoices, '0.rows.0.vatPercent')
```
toplenboren, line 82: what if no vat?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,159 @@
...
+
+    const listOfWorks = invoices.reduce((acc, invoice) => {
+        const rows = Array.isArray(invoice.rows) ? invoice.rows : []
+        acc.push(...rows.map((row, index) => {
+            const price = parseFloat(!row.isMin ? row.toPay : null)
+            const sum = price * row.count
+            totalSum += sum
+
+            return {
+                number: index + 1 || '',
```
toplenboren, line 92: why || '' ?

toplenboren (changes requested):

apps/condo/package.json
Alllex202, line 169: This is a very old unsupported library, maybe there is something newer? 
If not, a suitable one, then you can leave only numbers in the document, without converting to words

And is there a package with support for different languages?

apps/condo/lang/en/en.json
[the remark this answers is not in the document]
tolmachev21: The tests are swearing

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
tolmachev21: Since the same employee can be connected to several organizations. Therefore, the condition must be both the user ID and organization ID in which he is an employee

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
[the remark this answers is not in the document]
tolmachev21: I don't think that the tax on different goods can be different. Initially, the organization chooses its own tax system. However, the structure is suitable for each product. But then there is a chance to make a mistake in the calculations. We need to clarify this case

apps/condo/package.json
tolmachev21: There is a package with support for different currencies, newer and with normal generation of kopecks. https://github.com/vicimpa/vicimpa-library

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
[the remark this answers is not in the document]
tolmachev21: I think that no one will fill out this form manually if something does not load, except for signatures. The document format is DOKH, so they will increase it if necessary. Well, and the land should be enough for the most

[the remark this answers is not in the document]
Alllex202: I would provide for different situations, it doesn't seem difficult

Up to you :)

apps/condo/domains/ticket/hooks/useTicketDocumentGenerationTask.tsx
toplenboren: generate-document-of-works-completion
generate-document-of-paid-works-completion

apps/condo/package.json
toplenboren: Why not simply inline the code?

apps/condo/domains/ticket/hooks/useTicketDocumentGenerationTask.tsx
```diff
@@ -45,15 +46,22 @@ export const useTicketDocumentGenerationTask = ({ ticket, user, timeZone }) => {
         handleRunTask({ taskAttrs: { documentType } })
     }, [handleRunTask, loading])
 
+    const isValidInvoice = invoices.find((invoice) => get(invoice, 'status') !== 'Canceled' &&  get(invoice, 'deletedAt') === null)
```
nomerdvadcatpyat, line 49: use const from `domains/marketplace/constants`:
```suggestion
    const isValidInvoice = invoices.find((invoice) => get(invoice, 'status') !== INVOICE_STATUS_CANCELED &&  get(invoice, 'deletedAt') === null)
```

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
nomerdvadcatpyat: I think it's better to use logger

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,153 @@
...
+        sortBy: ['createdAt_ASC'],
+    })
+
+    const currencyCode = get(invoices, '0.currencyCode') || DEFAULT_INVOICE_CURRENCY_CODE
+    let totalSum = Big(0), totalVAT = Big(0)
+
+    const listOfWorks = invoices.reduce((acc, invoice) => {
+        const rows = Array.isArray(invoice.rows) ? invoice.rows : []
+        acc.push(...rows.map((row, index) => {
+            const price = Big(!row.isMin ? row.toPay : 0)
```
nomerdvadcatpyat, line 79: We need only exact prices? Is there no need to write that the price is contract?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
Alllex202: I wouldn't use `lodash.get` at all :)

There is `optional chaining`, it are very good at coding for frontend with typescript.
The downside of `lodash.get` is that it ignores any type errors that typescript would highlight

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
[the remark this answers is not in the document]
Alllex202: Please write here about what you managed to find out

apps/condo/package.json
tolmachev21: The library is very cool. There are many useful functions. In general, yes, it can be rewritten. If you don't like the library, I'll rewrite it.

Alllex202: > Why not simply inline the code? It is one script, like 100 characters ...

No need to write your own library...
I doubt it will fit in 100 characters

Alllex202: Let's leave only numbers then, without transforming them into words. This option will also be fine

apps/condo/package.json
```diff
@@ -79,6 +79,7 @@
     "@tinymce/tinymce-react": "^4.1.0",
     "@types/lodash": "4.14.186",
     "@types/wavesurfer.js": "^6.0.6",
+    "@vicimpa/rubles": "^0.1.18",
```
toplenboren, line 82: adding another library for one locale specific task seems not a good idea. Why not simply inline the code? It was 100 lines or smth

Why inlining is better
- less attack surface for supply chain attack (see XZ case)
- less bloated node_modules
- less stuff to update

toplenboren (changes requested):

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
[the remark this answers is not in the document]
tolmachev21: The structure of the account allows us to check the tax for each service. I think that this was done for a reason, so we will display the tax percentage for each service separately.

Alllex202 (changes requested):

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,157 @@
...
+        },
+        client: {
+            name: get(contact, 'name') || '',
+        },
+        listOfWorks,
+        sum: {
+            totalSum: !Number.isNaN(totalSum) ? renderMoney(totalSum, currencyCode, locale) : '',
+            totalVAT: !Number.isNaN(totalVAT) ? renderMoney(totalVAT, currencyCode, locale) : '',
+            totalSumInWords: buildEmptyLineDifferentLength(true),
+            totalVATInWords: buildEmptyLineDifferentLength(true),
```
tolmachev21, line 130: It is difficult for us to translate numbers into words
- we don't want to drag some lib (for various reasons)
- we don't want to make our own lib (it is difficult/long/needs support)
- we don't want to make crutches either :)

apps/condo/package.json
toplenboren: I'm also in only numbers club, but it is expected in Ru so that you'd have this `words` near numbers :-( 

--

**How to approach this?**

I'm a bit against on introducing a new library just for the task, 

when introducing a new library you add one more supply chain attack source and bloat your node_modules (so you now have to download more stuff from npm every time you run `npm i`)

Solution i'd like will be to simply inline the library inside the document generator code, and maybe add some tests. This way you'd incapsulate this logic in this feature and wouldn't introduce a new dependency

-- 

However this solution is bad if:

- You need to do this `number-to-text` transform anywhere else, for example in billing code
- You need this code to work in non russian environment

In this scenario its wise to add one more dependency. 

- Dependency should support major locales
- Code should be OK (have tests, etc)
- Dependency should be maintained well – check issues and pull request. Does author reply to them?
- Dependency should be small. No need to drag 5mb of JS code into our codebase
- The less dependencies this dependency have – the better
- License should be permissive. Best one out there – MIT

The dependency named [ruble](https://www.npmjs.com/package/@vicimpa/rubles) will not be ok for use, becuase it doesnt support 

The package I kinda like is this: https://www.npmjs.com/package/to-words but it doesnt support RU-ru locale, so you might want to do a PR there first :-)

--

In a nutshell
- Consult legal team. Can you ship the feature without number-to-words?
- Consult billing team. Do they have the need for a separate package that does number-to-words?
- If billing team does not need the solution, i'd have inlined the code to the generator (it is RU-ru anyway) and would've added some tests, so that it is easier for us to support it
- If billing team needs the solution – you'd need to choose package and maybe create a package in open condo

apps/condo/domains/ticket/utils/index.js
```diff
@@ -69,6 +69,100 @@ const renderMoney = (amount, currencyCode, locale) => {
...
+const bigDischarges = [
+    ['…n-лион', 'ов', '', 'а'],
+    ['тысяч', '', 'а', 'и'],
+    ['миллион', 'ов', '', 'а'],
+    ['миллиард', 'ов', '', 'а'],
+]
+
+const smallDischarges = [
+    ['ноль'],
+    ['-', 'один', 'два', 'три', 'четыре', 'пять', 'шесть', 'семь', 'восемь', 'девять'],
```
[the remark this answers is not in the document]
toplenboren: - If it is to be in `ticket/utils/index` then it should support other locales and have tests

If you want to inline this code just for this feature – you should move it to act generator

apps/condo/domains/ticket/hooks/useTicketDocumentGenerationTask.tsx
```diff
@@ -45,15 +47,22 @@ export const useTicketDocumentGenerationTask = ({ ticket, user, timeZone }) => {
         handleRunTask({ taskAttrs: { documentType } })
     }, [handleRunTask, loading])
 
+    const isValidInvoice = invoices.find((invoice) => get(invoice, 'status') !== INVOICE_STATUS_CANCELED &&  get(invoice, 'deletedAt') === null)
```
toplenboren, line 50: it is hasValidInvoice

apps/condo/domains/ticket/utils/index.js
```diff
@@ -52,9 +53,28 @@ const buildFullClassifierName = (classifier) => {
...
+const buildEmptyLineDifferentLength = (long = false) => {
+    return long ? '______________________________________________' : '_______________________'
+}
+
+
+const formatDate = (date, timeZone, format = 'DD.MM.YYYY') => {
+    return dayjs(date).tz(timeZone).format(format)
+}
+
+const renderMoney = (amount, currencyCode, locale) => {
```
toplenboren, line 65: Why in ticket utils?

apps/condo/domains/ticket/utils/index.js
```diff
@@ -52,9 +53,28 @@ const buildFullClassifierName = (classifier) => {
     ].filter(Boolean).join(' » ')
 }
 
+const buildEmptyLineDifferentLength = (long = false) => {
```
toplenboren, line 56: No, this is not about ticket, this is about some generation utils. Not all documents are word docs with ___ instead of blanks

apps/condo/domains/ticket/utils/index.js
```diff
@@ -52,9 +53,28 @@ const buildFullClassifierName = (classifier) => {
...
+    return long ? '______________________________________________' : '_______________________'
+}
+
+
+const formatDate = (date, timeZone, format = 'DD.MM.YYYY') => {
```
toplenboren, line 61: formatDateToTimezone? formatDate is ambiguous

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,157 @@
+const Big = require('big.js')
```
toplenboren, line 1: 👍

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,157 @@
...
+                vat: !Number.isNaN(vatPercent) ? renderMoney(vatPercent, currencyCode, locale) : '',
+                sum: !Number.isNaN(sum) ? renderMoney(sum, currencyCode, locale) : '',
+            }
+        }))
+
+        return acc
+    }, [])
+
+    totalSum.toFixed(2)
+    totalVAT.toFixed(2)
```
toplenboren, line 101: This works only with rubles.

toplenboren: Best practice is to get the locale from somewhere else, add a todo at least

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,157 @@
...
+        header: {
+            generalDate: printDate.format('DD.MM.YYYY'),
+        },
+        company: {
+            name: get(organization, 'name') || buildEmptyLineDifferentLength(true),
+            psrn: psrn || buildEmptyLineDifferentLength(),
+            tin: get(organization, 'tin') || buildEmptyLineDifferentLength(),
+            iec: iec || buildEmptyLineDifferentLength(),
+            address: organizationAddress || buildEmptyLineDifferentLength(true),
+            phone: get(organization, 'phone') || buildEmptyLineDifferentLength(),
```
toplenboren, line 113: I'd have went the different way

1. companyData = { name, psrn, iec ... }
2. changeUndefinedDataToBlanks(companyData)

changeUndefinedDataToBlanks will just iterate over the object and if value is null or undefined it will simply use BLANK

you might want to store meta info about how long this is somewhere too

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,157 @@
...
+
+        },
+        bankDetails: {
+            accountNumber: get(invoices, '0.accountNumber') || buildEmptyLineDifferentLength(),
+            bankName: get(invoices, '0.recipient.bankName') || buildEmptyLineDifferentLength(),
+            bankAccount: get(invoices, '0.recipient.bankAccount') || buildEmptyLineDifferentLength(),
+            bic: get(invoices, '0.recipient.bic') || buildEmptyLineDifferentLength(),
+        },
+        client: {
+            name: get(contact, 'name') || '',
```
toplenboren, line 123: why not blank?

apps/condo/package.json
Alllex202: > but it is expected in Ru so that you'd have this `words` near numbers :-(

we were told a long time ago that it is not necessary to translate numbers into words :)
In my understanding, this moment only delays the release of the feature. If you really want, you can do it later. What do you think about this? :)

If you want to do it, do it well, so that your implementation is well tested and can support implementation for different locales (right now you can do only ru, but adding other languages should not be a problem).
And it is definitely not "100 characters", as you said earlier)))

And also if you copy code from another repository, then it would probably be nice to leave a link to the source :)

Alllex202: > The package I kinda like is this: https://www.npmjs.com/package/to-words but it doesnt support RU-ru locale, so you might want to do a PR there first :-)

Yeah, I saw it too and I liked the library, but there is no ru locale. And I looked at the pull request list - there are two requests for adding locales that are over a year old, so I don't think adding ru locale will be a quick thing :(

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,157 @@
...
+            fileUploadInput = buildUploadInputFrom(await buildExportWordFile({ task, documentData, locale, timeZone }))
+            break
+        }
+
+        default: {
+            throw new Error(`unexpected format "${format}" for a document generation`)
+        }
+    }
+
+    logger.info({ msg: 'finish genereate document of paid completion works ', ticketId: ticket.id })
```
nomerdvadcatpyat, line 150: log `taskId` also

nomerdvadcatpyat (approved):

apps/condo/domains/ticket/utils/index.js
tolmachev21: Because this value is in the table

apps/condo/domains/common/utils/moneyToWords/index.js
```diff
@@ -0,0 +1,146 @@
...
+}
+
+const moneyToWords = (input, options = {}) => {
+    options = Object.assign({}, defaultOptions, options)
+
+    if (options.locale) setWordsToConvert(options.locale, options.currencyCode)
+
+    if (!isValidNumber(input)) throw new Error(`Invalid Number '${input}'`)
+
+    let [base = '0', dop = '00'] = parseNumber(input)
```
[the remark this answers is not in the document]
toplenboren: If this is about decimal part of money, like 20.05 or smth like this, you should take thta info from currency

example: BTC, 0.02928294 is normal for crypto

apps/condo/domains/common/utils/moneyToWords/locales/ru.spec.js
```diff
@@ -0,0 +1,79 @@
...
+describe('Test Integers with optioins = { format: `$WholeString $wholeCurrency` }', () => {
+    test.concurrent.each(testIntegersWithCurrency)('moneyToWords %d => %s', (input, expected) => {
+        expect(moneyToWords(input, { locale, currencyCode, format: '$WholeString $wholeCurrency' })).toBe(expected)
+    })
+})
+
+const testFloatsWithCurrency = [
+    [0.0, 'Ноль рублей ноль копеек'],
+    [0.04, 'Ноль рублей четыре копейки'],
+    [0.0468, 'Ноль рублей четыре копейки'],
```
toplenboren, line 64: Shouldnt this be exception? You have chosen incorrect currency if this is the case. Not sure that I like the implicit rounding

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+const { Invoice } = require('@condo/domains/marketplace/utils/serverSchema')
+const { DEFAULT_ORGANIZATION_TIMEZONE } = require('@condo/domains/organization/constants/common')
+const { TICKET_DOCUMENT_GENERATION_TASK_FORMAT } = require('@condo/domains/ticket/constants/ticketDocument')
+const { formatDateToTimezone } = require('@condo/domains/ticket/utils')
+
+const logger = getLogger('generateDocumentOfPaidWorksCompletion')
+
+const financeInfoClient = new FinanceInfoClient()
+
+async function getLocaleToPullFinanceInfoClient (locale, organization) {
```
toplenboren, line 24: Function is called `getLocale ...` yet you return some organization data. Best to rename this function like:
`getFinanceInfoDataByLocale (organization, { locale })`

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+    let psrn = null, organizationAddress = null, iec = null
```
toplenboren, line 25: What is PSRN? Why is it not under `locale = ru` ?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+
+    if (locale === 'ru') {
+        try {
+            ({ iec, psrn, organizationAddress } = await financeInfoClient.getOrganization(organization.tin))
+        } catch (error) {
+            logger.info({ msg: 'fall financeInfoClient when get organization by tin', organizationId: organization.id, tin: organization.tin, error: error })
+        }
+    }
+
+    return { iec, psrn, organizationAddress }
```
toplenboren, line 35: what is psrn? It should be international if it is returned from this function for all locales :-)

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+    const chengedDocumentDataInLine = Object.fromEntries(Object.entries(documentDataInLine).map(([key, value]) => {
+        let changedValue = {}
+        for (let subValue in value) {
+            changedValue[subValue] = value[subValue] ? value[subValue] : metaInfoAboutLongWords.hasOwnProperty(subValue) ? '____________________________________' : '__________________'
+        }
+        return [key, changedValue]
+    }))
+    
+    return chengedDocumentDataInLine
+}
```
toplenboren, line 52: Why is this function called changeUndefinedData?

It starts with `change`, so I expect that you will do some in place modifications or something like this, yet you return a new object

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    return { iec, psrn, organizationAddress }
+}
+
+const changeUndefinedDataToBlanks = (documentDataInLine) => {
+    const metaInfoAboutLongWords = {
+        name: true,
+        address: true,
+    }
```
toplenboren, line 42: What is `long word` ?

If you define some consts, they should be defined outside of function. This object will be created every time on function invocation

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+    const chengedDocumentDataInLine = Object.fromEntries(Object.entries(documentDataInLine).map(([key, value]) => {
```
toplenboren, line 43: it is chAnged

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+            ({ iec, psrn, organizationAddress } = await financeInfoClient.getOrganization(organization.tin))
+        } catch (error) {
+            logger.info({ msg: 'fall financeInfoClient when get organization by tin', organizationId: organization.id, tin: organization.tin, error: error })
+        }
+    }
+
+    return { iec, psrn, organizationAddress }
+}
+
+const changeUndefinedDataToBlanks = (documentDataInLine) => {
```
toplenboren, line 38: What is documentDataInLine ? What type is this variable?

Why not simply documentData?

toplenboren, line 38: If you create a function and it is public (doesnt start with _ or is exported) then you shoud write a jsDoc

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
toplenboren: I'd have done smth like this

```js
replaceUndefinedDocumentDataWithBlanks(documentData) {}

// usage
replaceUndefinedDocumentDataWithBlanks(documentData) 
```

and then would've returne updated object. This way function would remind of native replaceAll

or more generic way

```js
replaceUndefinedValues ( object, replacer ) {}

// usage
replaceUndefinedValues(documentData, (k, v) => (k === 'address' || k ==='otherLongStuff') LONG_BLANK : SHORT_BLANK
```

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+        encoding: DOCX_FILE_META.encoding,
+        meta: {
+            listkey: 'TicketDocumentGenerationTask',
+            id,
+        },
+    }
+}
+
+const generateTicketDocumentOfPaidWorks = async ({ task, baseAttrs, context, locale, ticket, organization }) => {
+    const { iec, psrn, organizationAddress } = await getLocaleToPullFinanceInfoClient(locale, organization)
```
toplenboren, line 75: is psrn locale independent?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+
+    const { format, timeZone: timeZoneFromUser } = task
+
+    const timeZone = normalizeTimeZone(timeZoneFromUser) || DEFAULT_ORGANIZATION_TIMEZONE
+    const printDate = dayjs().tz(timeZone).locale(locale)
+
+    const contact = ticket.contact ? await getByCondition('Contact', {
+        id: ticket.contact,
+        deletedAt: null,
+    }) : null
```
toplenboren, line 85: Can we continue document generation if there are no contact?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+            user: { id: ticket.executor },
+            deletedAt: null,
+        })
+        : null
+
+    const invoices = await Invoice.getAll(context, {
+        ticket: { id: ticket.id },
+        deletedAt: null,
+        status_not: INVOICE_STATUS_CANCELED,
+    }, 'id currencyCode accountNumber rows { isMin toPay name count vatPercent } recipient { bankName bankAccount bic }', {
```
toplenboren, line 99: why specify fields like that? Is it how it should be done? Why not `graphql` file?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+                sum: !Number.isNaN(sum) ? renderMoney(sum, currencyCode, locale) : '',
+            }
+        }))
+
+        return acc
+    }, [])
+
+    // NOTE: There are two different types of data in the document template
+    // The first is data inserted into tables
+    // The second is data inserted directly into the document paragraph
```
toplenboren, line 130: If you say `data inserted in document paragraph` in docs, you should use this terminology consistently thoroughout the code

const documentDataInParagraph 

or in docs 

// Note: this doc needs two variables to be generated:
// 1. Values that will go into the table (worksData)
// 2. Values that will be embeded into the text of the document (textData)

const documentTextData = { ... }

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
toplenboren: Why || then?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+            bankAccount: get(invoices, '0.recipient.bankAccount'),
+            bic: get(invoices, '0.recipient.bic'),
+        },
+        totalInWords: {
+            totalSum: moneyToWords(totalSum.toFixed(2), { locale, currencyCode }),
+            totalVAT: moneyToWords(totalVAT.toFixed(2), { locale, currencyCode }),
+        },
+    }
+
+    const documentDataInTable = {
```
toplenboren, line 156: Check my comment above please

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+        return acc
+    }, [])
+
+    // NOTE: There are two different types of data in the document template
+    // The first is data inserted into tables
+    // The second is data inserted directly into the document paragraph
+
+    const documentDataInLine = {
+        header: {
+            generalDate: printDate.format('DD.MM.YYYY'),
```
toplenboren, line 134: Why date is in this format? If you want this doc to be localizable you should localize this data too, or at least put a todo

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+            phone: get(organization, 'phone'),
+        },
+        bankDetails: {
+            accountNumber: get(invoices, '0.accountNumber'),
+            bankName: get(invoices, '0.recipient.bankName'),
+            bankAccount: get(invoices, '0.recipient.bankAccount'),
+            bic: get(invoices, '0.recipient.bic'),
+        },
+        totalInWords: {
+            totalSum: moneyToWords(totalSum.toFixed(2), { locale, currencyCode }),
```
toplenboren, line 151: Check out my comment about `2` in money

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+            totalVAT: !Number.isNaN(totalVAT) ? renderMoney(totalVAT, currencyCode, locale) : '',
+        },
+        executor: {
+            name: get(employee, 'name') || '',
+        },
+    }
+
+    const documentData = {
+        ...changeUndefinedDataToBlanks(documentDataInLine),
+        ...documentDataInTable,
```
toplenboren, line 172: why documentDataInTable doesnt use this changeUndefined lalala

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,195 @@
...
+            fileUploadInput = buildUploadInputFrom(await buildExportWordFile({ task, documentData, locale, timeZone }))
+            break
+        }
+
+        default: {
+            throw new Error(`unexpected format "${format}" for a document generation`)
+        }
+    }
+
+    logger.info({ msg: 'finish genereate document of paid completion works ', taskId: task.id, ticketId: ticket.id })
```
toplenboren, line 188:
```suggestion
    logger.info({ msg: 'finished genereating a document of paid completion works ', taskId: task.id, ticketId: ticket.id })
```

apps/condo/domains/common/utils/moneyToWords/index.js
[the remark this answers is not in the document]
Alllex202: > example: BTC, 0.02928294 is normal for crypto

It seems that for regular currencies two decimal places should be enough. You won't be able to pay for an amount less than one kopeck, cent or any other regular currency

!Not for all currencies, there are exceptions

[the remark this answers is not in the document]
Alllex202: I don't think we should focus on cryptocurrencies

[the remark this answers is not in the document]
Alllex202: > If this is about decimal part of money, like 20.05 or smth like this, you should take thta info from currency

But if it can be customized, it will be good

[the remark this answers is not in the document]
Alllex202: Perhaps we should describe some limitations so as not to complicate the utility.

apps/condo/domains/common/utils/moneyToWords/index.js
```diff
@@ -0,0 +1,146 @@
+const LOCALES = require('./locales')
+
+const formatRegExp = /\$([a-z]+)/gi
+
+let currency, smallDischarges, bigDischarges
+
+const setWordsToConvert = (locale = 'ru', currencyCode = 'RUB') => {
```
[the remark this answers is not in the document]
tolmachev21: Yes, because the locale is used without region

apps/condo/domains/ticket/utils/index.js
```diff
@@ -52,9 +53,14 @@ const buildFullClassifierName = (classifier) => {
     ].filter(Boolean).join(' » ')
 }
 
+const formatDateToTimezone = (date, timeZone, format = 'DD.MM.YYYY') => {
+    return dayjs(date).tz(timeZone).format(format)
+}
```
toplenboren, line 58: Why default format is russian?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/completionWorks.js
```diff
@@ -28,7 +24,7 @@ const buildExportWordFile = async ({ task, documentData, locale, timeZone }) =>
 
     return {
         stream,
-        filename: `completion_works_ticket_${ticket.id}_${formatDate(undefined, timeZone, 'DD_MM_YYYY')}.docx`,
+        filename: `completion_works_ticket_${ticket.id}_${formatDateToTimezone(undefined, timeZone, 'DD_MM_YYYY')}.docx`,
```
toplenboren, line 27: Why do you need to create this function?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,220 @@
...
+        }
+    }
+
+    return { iec, psrn, organizationAddress }
+}
+
+const getDataFormatByLocale = ( locale ) => {
+    if (locale === 'ru') return 'DD.MM.YYYY'
+    return 'YYYY-MM-DD'
+}
```
toplenboren, line 44: Use Intl or dayjs internal formatting tools

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,220 @@
...
+
+const _changeToBlanks = (key, value) => {
+    if (!value) {
+        return (key === 'address' || key === 'name') ? LONG_BLANK : SHORT_BLANK
```
toplenboren, line 48: General rule of thumb, if you use something domain-specific (like address and name here) then name of the function should reflect that, so that you wouldnt be confused 

I suggest smth like `changePaidWorksDocumentUndefinedDataToBlanks`

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
toplenboren: ^^

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,220 @@
...
+    let totalSum = Big(0), totalVAT = Big(0)
+
+    const listOfWorks = invoices.reduce((acc, invoice) => {
+        const rows = Array.isArray(invoice.rows) ? invoice.rows : []
+        acc.push(...rows.map((row, index) => {
+            let price
+            if (row.isMin) {
+                price = Big(0)
+            } else {
+                price = !Number.isNaN(parseFloat(row.toPay)) ? Big(row.toPay) : Big(0)
```
toplenboren, line 130: Why do you use parseFloat instead of Big.js here?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,220 @@
...
+            totalSum: renderMoney(totalSum, currencyCode, locale),
+            totalVAT: renderMoney(totalVAT, currencyCode, locale),
+        },
+        executor: {
+            name: get(employee, 'name'),
+        },
+    }
+
+    const documentData = {
+        ...replaceAllUndefinedValues(documentTextData, _changeToBlanks),
```
toplenboren, line 196: If you use function one time, best to just inline it

toplenboren (changes requested):

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
Alllex202, line 44: Do you really need this function?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,220 @@
...
+
+    return { iec, psrn, organizationAddress }
+}
+
+const getDataFormatByLocale = ( locale ) => {
+    if (locale === 'ru') return 'DD.MM.YYYY'
+    return 'YYYY-MM-DD'
+}
+
+const _changeToBlanks = (key, value) => {
```
Alllex202, line 46: This name is confusing

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,220 @@
...
+    if (!value) {
+        return (key === 'address' || key === 'name') ? LONG_BLANK : SHORT_BLANK
+    }
+    return value
+}
+
+const _changeToEmptyCell = (key, value) => {
```
Alllex202, line 53: same^^

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,220 @@
...
+        },
+        executor: {
+            name: get(employee, 'name'),
+        },
+    }
+
+    const documentData = {
+        ...replaceAllUndefinedValues(documentTextData, _changeToBlanks),
+        ...replaceAllUndefinedValues(documentDataInTable, _changeToEmptyCell),
+    }
```
Alllex202, line 198: I really don't understand why these transformations are needed...
Usually just use inline default values, it's much simpler

Alllex202: Why exactly undefined? That the value will be null or empty line?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
Alllex202: > why specify fields like that? Is it how it should be done? Why not `graphql` file?

Because that's how server utilities work now :) Now you should request only the fields you need

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
tolmachev21: If you pass Big(" ") to big.js, there will be an error

Therefore, Number.isNaN(parseFloat(row.toPay) will check that the passed value is not an empty string or a string of spaces

apps/condo/domains/common/utils/moneyToWords/locales/ru.spec.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+    row[0] = -row[0]
+})
+
+const testIntegersWithCurrency = [
+    [0, 'Ноль рублей'],
+    [137, 'Сто тридцать семь рублей'],
+    [700, 'Семьсот рублей'],
+    [1100, 'Одна тысяча сто рублей'],
+    [4680, 'Четыре тысячи шестьсот восемьдесят рублей'],
```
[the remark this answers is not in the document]
tolmachev21: This is done to check the formatting of the line generation with a capital letter. But I will add this to the test title

apps/condo/domains/common/utils/moneyToWords/locales/en.spec.js
```diff
@@ -0,0 +1,86 @@
...
+    [37.06, 'Thirty seven six'],
+    [37.068, 'Thirty seven seven'],
+    [37.68, 'Thirty seven sixty eight'],
+    [37.683, 'Thirty seven sixty eight'],
+]
+
+const testFloatsWithCurrency = [
+    [0.0, 'Zero dollars zero cents'],
+    [0.04, 'Zero dollars four cents'],
+    [0.0468, 'Zero dollars five cents'],
```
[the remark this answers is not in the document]
tolmachev21: Yes, for now this function is only supported by locales with decimal rounding.
Otherwise, the function that generates the act must pass rounded values, and this in turn violates its locale independence.
Therefore, rounding will be in coinToWords until it becomes locale independent.

apps/condo/domains/common/utils/moneyToWords/locales/ru.spec.js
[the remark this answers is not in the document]
Alllex202: some implicit formatting?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/completionWorks.js
tolmachev21: This function was there before. I just changed the name from formatDate to formatDateToTimezone. I use this function in several places

apps/condo/domains/ticket/hooks/useTicketDocumentGenerationTask.tsx
```diff
@@ -45,15 +47,22 @@ export const useTicketDocumentGenerationTask = ({ ticket, user, timeZone }) => {
         handleRunTask({ taskAttrs: { documentType } })
     }, [handleRunTask, loading])
 
+    const hasValidInvoice = invoices.find((invoice) => get(invoice, 'status') !== INVOICE_STATUS_CANCELED &&  get(invoice, 'deletedAt') === null)
```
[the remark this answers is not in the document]
tolmachev21: No, but if something changes, we will be ready for it)

apps/condo/domains/common/utils/moneyToWords/locales/ru.spec.js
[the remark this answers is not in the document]
tolmachev21: We have a parameter format: "$WholeString", which makes the string capitalized. By the way, with the help of capitalize

apps/condo/domains/common/utils/moneyToWords/locales/en.spec.js
```diff
@@ -0,0 +1,86 @@
...
+    [0.999, 'One dollars zero cents'],
+    [37.06, 'Thirty seven dollars six cents'],
+    [37.068, 'Thirty seven dollars seven cents'],
+    [37.68, 'Thirty seven dollars sixty eight cents'],
+    [37.683, 'Thirty seven dollars sixty eight cents'],
+]
+
+
+describe('Test "en" locale moneyToWords', () => {
+    test.concurrent.each(testIntegers)('moneyToWords(testIntegers) %d => %s', (input, expected) => {
```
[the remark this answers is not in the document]
tolmachev21: I ran different versions of the tests. And locally my tests without concurrent were executed even faster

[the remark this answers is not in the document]
Alllex202: Why do you use `concurrent ` then, if tests with it take longer to run? :)

apps/condo/domains/common/utils/moneyToWords/locales/ru.spec.js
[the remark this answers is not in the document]
Alllex202: Ok, I get it. It's good that this can be changed

apps/condo/domains/ticket/hooks/useTicketDocumentGenerationTask.tsx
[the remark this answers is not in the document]
Alllex202: We should not query deleted data on the client. If an object has `deletedAt` set, then it should be considered that it is deleted and does not exist (even if in reality it is still present in the database)

So if you are requesting deleted data, there is something wrong with your logic.
(This is all assuming that the app doesn't work offline - and it doesn't work offline))

apps/condo/domains/common/utils/date.js
```diff
@@ -37,11 +37,28 @@ const getStartDates = (dateRaw) => {
  */
 const getCurrTimeStamp = () => Math.floor(Date.now() / 1000)
 
+/**
+ * Форматирует дату с учетом указанного часового пояса.
```
Alllex202, line 41: 請問你能用英文寫同樣的嗎？謝謝你））

apps/condo/domains/common/utils/moneyToWords/moneyToWords.spec.js
```diff
@@ -0,0 +1,80 @@
...
+    const testInvalidDecimialInputs = [
+        '124.124',
+        '124.000',
+        '124.29102480918209480129',
+        '0.000',
+        '0.1241',
+        '0.1',
+        '0.9',
+        '0.0',
+        '-0.0',
```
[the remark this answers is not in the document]
Alllex202: '0.1',
 '0.9',
 '0.0',
 '-0.0',

apps/condo/domains/common/utils/moneyToWords/moneyToWords.spec.js
```diff
@@ -0,0 +1,80 @@
...
+        '-0.00000',
+        '-0.513511',
+        0.40,
+        -0.40,
```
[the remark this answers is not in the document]
Alllex202: 0.40,
 -0.40,

apps/condo/domains/common/utils/moneyToWords/locales/en.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * @constant { Array } dischargesMoreThanThousand 
+ * dischargesMoreThanThousand[n][0] - word root;
+ * dischargesMoreThanThousand[n][1] - word root + ending for units from 5 to 9
+ * dischargesMoreThanThousand[n][2] - word root + ending for 1 unit
+ * dischargesMoreThanThousand[n][3] - word root + ending for units from 2 to 4
+ */
+
+const dischargesMoreThanThousand = [
```
toplenboren, line 9: why not one array?

apps/condo/domains/common/utils/moneyToWords/locales/en.js
```diff
@@ -0,0 +1,67 @@
...
+ * @constant { 
+*  currencyCode: string[][],
+* } currency 
+*/
+
+const supportedCurrencies = {
+    'USD': currencyUSD,
+}
+
+const mathematicalSymbol = {
```
toplenboren, line 58: Why do you implement your own Intl? :-) I'd have just handled minus in frontend code

apps/condo/domains/common/utils/moneyToWords/locales/en.spec.js
```diff
@@ -0,0 +1,77 @@
...
+    [4680, 'four thousand six hundred eighty'],
+    [63892, 'sixty three thousand eight hundred ninety two'],
+    [86100, 'eighty six thousand one hundred'],
+    [792581, 'seven hundred ninety two thousand five hundred eighty one'],
+    [2741034, 'two million seven hundred forty one thousand thirty four'],
+    [86429753, 'eighty six million four hundred twenty nine thousand seven hundred fifty three'],
+]
+
+const testNegativeIntegers = cloneDeep(testIntegers)
+testNegativeIntegers.map((row, i) => {
```
toplenboren, line 24: map already returns you a new object, is it not? 
image

toplenboren: Why do you need cloneDeep here?

apps/condo/domains/common/utils/moneyToWords/moneyToWords.js
```diff
@@ -0,0 +1,179 @@
...
+ * @exception This function can work with numbers up to >10000000000. But it can be easily extended to numbers of higher orders.
+ */
+
+function moneyToWords (input, options = {}) {
+    options = { ...DEFAULT_OPTIONS, ...options }
+
+    if (!(options.locale in LOCALES)) throw new Error(`Supported only ${Object.keys(LOCALES).join(', ')} locales`)
+    if (!(options.currencyCode in LOCALES[options.locale].supportedCurrencies)) throw new Error(`Supported only ${Object.keys(LOCALES[options.locale].supportedCurrencies).join(', d')} locales`)
+
+    if (!_isValidNumber(input)) throw new Error(`Invalid Number "${input}"`)
```
toplenboren, line 52: If you use function only one time and it is one liner, best to inline it. Who knows what a `validNumber` is to you?

apps/condo/domains/common/utils/moneyToWords/moneyToWords.js
```diff
@@ -0,0 +1,179 @@
...
+
+    const { currency, dischargesLessThanThousand, dischargesMoreThanThousand, mathematicalSymbol } = _setWordsToConvert(options.locale, options.currencyCode)
+
+    const isNegativeNumber = input < 0
+    if (isNegativeNumber) {
```
toplenboren, line 57: Why not inline this?

apps/condo/domains/ticket/hooks/useTicketDocumentGenerationTask.tsx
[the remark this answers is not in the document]
toplenboren: General rule of thumb

Use find / server utils -> explicitly check deletedAt

Use getAll like utils -> you can skip checking deletedAt

apps/condo/domains/ticket/utils/index.js
```diff
@@ -1,3 +1,4 @@
+const dayjs = require('dayjs')
```
toplenboren, line 1: Why do you add import to this file?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/completionWorks.js
toplenboren: But it is in diff

apps/condo/domains/common/utils/date.js
```diff
@@ -37,11 +37,28 @@ const getStartDates = (dateRaw) => {
...
+ * @param {Date|string} date - The original date in Date format or a string that can be converted to Date.
+ * @param {string} timeZone - Time zone name (e.g. 'Europe/Moscow').
+ * @param {string} [format=DATE_FORMAT] - Date output format. By default, a constant is used DATE_FORMAT.
+ * @returns {string} Formatted date in the specified time zone and format.
+ *
+ * @example
+ * const formattedDate = formatDateToTimezone(new Date(), 'America/New_York', 'YYYY-MM-DD HH:mm:ss');
+ */
+
+const formatDateToTimezone = (date, timeZone, format = DATE_FORMAT) => {
```
toplenboren, line 52: Why do you add this function?

apps/condo/domains/common/utils/money.js
```diff
@@ -0,0 +1,21 @@
...
+* @param {string} [locale='en-US'] The locale to format (defaults to 'en-US'). But you can also just 'ru' 
+* @returns {string} The formatted currency value.
+* @throws {TypeError} If `amount` is not a number.
+*/
+
+const renderMoney = (amount, currencyCode, locale) => {
+    if (Number.isNaN(parseFloat(amount))) return '' 
+    const options = { currency: currencyCode, style: 'currency' }
+    const numberFormat = new Intl.NumberFormat(locale, options)
+    const parts = numberFormat.formatToParts(parseFloat(amount))
```
toplenboren, line 15: parseFloat detected

toplenboren: What this adds to Intl?

apps/condo/domains/common/utils/moneyToWords/moneyToWords.js
```diff
@@ -0,0 +1,175 @@
...
+                    )
+                } else {
+                    row.push(dischargesLessThanThousand[1][units])
+                }
+            }
+
+            if (deep) {
+                row.push(_counterWord(dischargesMoreThanThousand[deep] ?? dischargesMoreThanThousand[0], current))
+            }
+        }
```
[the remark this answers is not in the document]
toplenboren: Lets put a todo that will address this issue and rewrite this together after merge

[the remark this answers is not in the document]
tolmachev21: dischargesLessThanThousand[n][units]
Supports only digits up to a thousand, i.e. n >= 6 is not possible. Thousands digits are supported by dischargesMoreThanThousand.

apps/condo/domains/common/utils/moneyToWords/moneyToWords.js
```diff
@@ -0,0 +1,176 @@
...
+    if (!(options.currencyCode in LOCALES[options.locale].supportedCurrencies)) throw new Error(`Supported only ${Object.keys(LOCALES[options.locale].supportedCurrencies).join(', d')} locales`)
+
+    if (!(!isNaN(parseFloat(input)) && isFinite(input))) throw new Error(`Invalid Number "${input}"`)
+
+    const { currency, dischargesLessThanThousand, dischargesMoreThanThousand, texts } = _setWordsToConvert(options.locale, options.currencyCode)
+
+    const isNegativeNumber = input < 0
+    if (isNegativeNumber) input = input.toString().slice(1)
+
+    let [whole = '0', decimal = '00'] = _parseNumber(input)
```
toplenboren, line 59: why not use toNumber from lodash?

apps/condo/domains/common/utils/moneyToWords/moneyToWords.js
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    return output.join(' ')
+}
+
+function _counterWord (input, counter) {
+    const units = counter % 10
+
+    if (counter > 10 && counter < 20)
+        return _joinWord(input, 0)
```
toplenboren, line 159: best to just use join from array

toplenboren (approved):

Alllex202 (approved):

apps/condo/domains/common/utils/moneyToWords/moneyToWords.js
```diff
@@ -0,0 +1,176 @@
+const { capitalize } = require('lodash')
+
+const LOCALES = require('./locales')
+
+const FORMAT_REG_EXP = /\$([a-z]+)/gi
+
+const DEFAULT_OPTIONS = {
+    locale: 'ru',
```
pahaz, line 8: YOU SHOULD NOT CONNECT LOCALE AND CURRENCY! It's a bad idea

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/completionWorks.js
```diff
@@ -28,7 +23,7 @@ const buildExportWordFile = async ({ task, documentData, locale, timeZone }) =>
 
     return {
         stream,
-        filename: `completion_works_ticket_${ticket.id}_${formatDate(undefined, timeZone, 'DD_MM_YYYY')}.docx`,
+        filename: `completion_works_ticket_${ticket.id}_${dayjs().tz(timeZone).format('DD_MM_YYYY')}.docx`,
```
pahaz, line 26: why not UTC format?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,202 @@
...
+    const { format, timeZone: timeZoneFromUser } = task
+
+    const timeZone = normalizeTimeZone(timeZoneFromUser) || DEFAULT_ORGANIZATION_TIMEZONE
+
+    const contact = ticket.contact ? await getByCondition('Contact', {
+        id: ticket.contact,
+        deletedAt: null,
+    }) : null
+
+    const employee = organization.id && ticket.executor
```
pahaz, line 84: why you don't use `OrganizationEmployee.getAll` here?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -0,0 +1,202 @@
...
+}
+
+const generateTicketDocumentOfPaidWorks = async ({ task, baseAttrs, context, locale, ticket, organization }) => {
+    const { iec, psrn, organizationAddress } = await getFinanceInfoDataByLocale(organization, { locale })
+
+    const { format, timeZone: timeZoneFromUser } = task
+
+    const timeZone = normalizeTimeZone(timeZoneFromUser) || DEFAULT_ORGANIZATION_TIMEZONE
+
+    const contact = ticket.contact ? await getByCondition('Contact', {
```
pahaz, line 79: why you don't use `Contact.getAll`?

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/completionWorks.js
tolmachev21: This function is used only in Russian locale. It is possible to add support for universal locale, but then it will break the document template with underscores

apps/condo/domains/common/utils/moneyToWords/moneyToWords.js
tolmachev21: Change this behavior in this PR - https://github.com/open-condo-software/condo/pull/5688

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
tolmachev21: find easier and faster

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
tolmachev21: find easier and faster

PR #5386, opened 2024-10-23
feat(keystone): DOMA-10930 added "GraphiQL" instead of the outdated "graphql-playground"

Alllex202:
- [x] use `nunjucks`
- [x] ability to change path to api
- [x] lock library versions
- [x] disable old playground
- [x] add static files 

## Preview

https://github.com/user-attachments/assets/7102f5e8-cb1f-4e8f-a25f-8c357d093382

### DIFF
--- a/packages/keystone/KSv5v6/v5/prepareKeystone.js
+++ b/packages/keystone/KSv5v6/v5/prepareKeystone.js
@@ -190,13 +190,17 @@ function prepareKeystone ({ onConnect, extendKeystoneConfig, extendExpressApp, s
             new IpBlackListMiddleware(),
             new KeystoneTracingApp(),
             ...((apps) ? apps() : []),
-            ...(IS_ENABLE_DANGEROUS_GRAPHQL_PLAYGROUND ? [new GraphiqlApp({ graphiqlPath: '/admin/graphiql' })] : []),
+            ...(IS_ENABLE_DANGEROUS_GRAPHQL_PLAYGROUND ? [
+                new GraphiqlApp({
+                    allowedQueryParams: ['deleted'],
+                }),
+            ] : []),
             new GraphQLApp({
                 apollo: {
                     formatError: safeApolloErrorFormatter,
                     debug: IS_ENABLE_APOLLO_DEBUG,
                     introspection: IS_ENABLE_DANGEROUS_GRAPHQL_PLAYGROUND,
-                    playground: IS_ENABLE_DANGEROUS_GRAPHQL_PLAYGROUND,
+                    playground: false,
                     plugins: apolloPlugins,
                 },
                 ...(graphql || {}),
--- a/packages/keystone/KSv5v6/v5/prepareKeystone.js
+++ b/packages/keystone/KSv5v6/v5/prepareKeystone.js
@@ -32,6 +32,9 @@ const { KeystoneTracingApp } = require('@open-condo/keystone/tracing')
 const { Keystone } = require('./keystone')
 const { validateHeaders } = require('./validateHeaders')
 
+const { GraphiqlApp } = require('../../graphiql')
+
+
 const IS_BUILD_PHASE = conf.PHASE === 'build'
 const IS_WORKER_PROCESS = conf.PHASE === 'worker'
 const IS_BUILD = conf['DATABASE_URL'] === 'undefined'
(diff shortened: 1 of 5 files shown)


packages/keystone/graphiql/graphiql.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Playground - __REPLACE_TO_YOUR_SERVER_URL__</title>
```
sitozzz, line 6:
```suggestion
    <title>Playground</title>
```

sitozzz: Or even condo playground)

packages/keystone/graphiql/graphiql.html
```diff
@@ -0,0 +1,32 @@
...
+    <link rel="stylesheet" href="https://unpkg.com/graphiql/graphiql.min.css" />
+</head>
+<body style="margin: 0;">
+<div id="graphiql" style="height: 100vh;"></div>
+<script crossorigin src="https://unpkg.com/react/umd/react.production.min.js"></script>
+<script crossorigin src="https://unpkg.com/react-dom/umd/react-dom.production.min.js"></script>
+<script crossorigin src="https://unpkg.com/graphiql/graphiql.min.js"></script>
+<script>
+    const apiPath = '__REPLACE_TO_YOUR_API_PATH__'
```
sitozzz, line 15: That's not looking good)

sitozzz: Oh yeah, I found the reason below

packages/keystone/graphiql/graphiql.html
Alllex202: The point here is to be able to differentiate between sandbox tabs of different applications.
This is exactly how it works for us now.
However, we can forget about this moment)

packages/keystone/graphiql/index.js
```diff
@@ -0,0 +1,40 @@
...
+        const html = fileBuffer.toString()
+        return html
+            .replace('__REPLACE_TO_YOUR_API_PATH__', apiPath)
+            .replace('__REPLACE_TO_YOUR_SERVER_URL__', SERVER_URL)
+    }
+
+    async prepareMiddleware () {
+        // nosemrep: javascript.express.security.audit.express-check-csurf-middleware-usage.express-check-csurf-middleware-usage
+        const app = express()
+        app.get(this._graphiqlPath, (req, res) => {
```
pahaz, line 31: may be we need to move it to another URL?

packages/keystone/graphiql/graphiql.html
```diff
@@ -0,0 +1,32 @@
...
+    const graphQLFetcher = async (graphQLParams) =>
+        fetch(apiPath, {
+            method: 'POST',
+            headers: {
+                'Content-Type': 'application/json',
+            },
+            credentials: 'omit',
```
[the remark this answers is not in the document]
Alllex202: The behavior matches what is in our current playground.

Authentication cookies are not taken from the platform. 
All requests are made with the auth token in the header

packages/keystone/graphiql/index.js
Alllex202: We can try to leave the old playground, but move it to another URL. 
And leave the new playground on the current URL, so as not to update the links in all the documentation.

The old playground is almost impossible to use due to overlapping popups when writing queries, but there you can change URL to API

Alllex202: moved the new playground to another URL

AleX83Xpert (approved):

packages/keystone/KSv5v6/v5/prepareKeystone.js
```diff
@@ -185,6 +188,7 @@ function prepareKeystone ({ onConnect, extendKeystoneConfig, extendExpressApp, s
             new IpBlackListMiddleware(),
             new KeystoneTracingApp(),
             ...((apps) ? apps() : []),
+            ...(IS_ENABLE_DANGEROUS_GRAPHQL_PLAYGROUND ? [new GraphiqlApp({ graphiqlPath: '/admin/graphiql' })] : []),
```
sitozzz, line 191: What about disable existing one via passing `IS_ENABLE_DANGEROUS_GRAPHQL_PLAYGROUND` to `playground` property a little lower in the code?

packages/keystone/graphiql/graphiql.html
sitozzz: Now looks good at this point

packages/keystone/KSv5v6/v5/prepareKeystone.js
Alllex202: for now I decided not to remove the old version - now there are links from different documentation and also the ability to change the path to the API

Alllex202: disabled old playground

packages/keystone/KSv5v6/v5/prepareKeystone.js
```diff
@@ -184,12 +187,13 @@ function prepareKeystone ({ onConnect, extendKeystoneConfig, extendExpressApp, s
...
             new KeystoneTracingApp(),
             ...((apps) ? apps() : []),
+            ...(IS_ENABLE_DANGEROUS_GRAPHQL_PLAYGROUND ? [new GraphiqlApp()] : []),
             new GraphQLApp({
                 apollo: {
                     formatError: safeApolloErrorFormatter,
                     debug: IS_ENABLE_APOLLO_DEBUG,
                     introspection: IS_ENABLE_DANGEROUS_GRAPHQL_PLAYGROUND,
-                    playground: IS_ENABLE_DANGEROUS_GRAPHQL_PLAYGROUND,
+                    playground: false,
```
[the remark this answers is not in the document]
Alllex202: I don't think it's critical :)

sitozzz (approved):
Let's go

PR #5388, opened 2024-10-23
chore(eps): DOMA-1893 check bicrypt build

vovaaxeapolla:


SavelevMatthew (changes requested):
You should have only 1 yarn and 1 yarn.lock located in the root of monorepo.

SavelevMatthew:
Can you use newly added workspace instead of `cd`?

Like:
```
yarn workspace lib-name build && ...
```

Instead of:
```
cd path/to/lib && yarn build && ...
```

SavelevMatthew (approved):
LGTM. Great job 🔥

dkoviazin (approved):

PR #5396, opened 2024-10-24
Idea: lets send only silent pushes  

toplenboren:


tolmachev21:
What is silent push? Push that we can customize?

Alllex202:
Explain what you did, what problem it solves and why? Nothing is clear...

And it looks like something is wrong with tests

pahaz (approved):

PR #5415, opened 2024-10-31
chore(condo): log string body

toplenboren:
This adds a body logger to the common fetch.js package. Case: i want to know what was sent to external services


sitozzz (changes requested):
That's might be really unsafe to do so. Why do you need this?

toplenboren:
> That's might be really unsafe to do so. Why do you need this?

If by unsafe you refer to logging some safe data, then
1. We already log headers, tokens and so on
2. We already log a lot of private stuff

I needed this PR to debug pushes to HCM (Huawei cloud), but I found another way to do this, so closing this PR

PR #5424, opened 2024-11-05
fix(condo): DOMA-10557 fix iteration over chunk uploading result

AleX83Xpert:


abshnko (approved):

apps/condo/domains/meter/tasks/importMeters.js
```diff
@@ -140,7 +140,7 @@ async function importMeters (taskId) {
             await MeterReadingsImportTask.update(context, taskId, {
                 ...dvAndSender,
                 status: COMPLETED,
-            })
+            }, 'id')
```
Alllex202, line 143: `id` is default value

Alllex202 (approved):

apps/condo/domains/meter/tasks/importMeters.js
AleX83Xpert: Yes, but function signature don't know about it :)

PR #5427, opened 2024-11-05
Hotfix/fix endless rerendering on meter page

abshnko:


apps/condo/pages/meter/index.tsx
```diff
@@ -126,41 +128,40 @@ const MetersPage: IMeterIndexPage = () => {
...
+        await updateQuery(router, { newParameters }, { resetOldParameters: true, routerAction: 'replace' } )
     }, [router])
 
 
     useEffect(() => {
-        if (!activeTab) return
-        if (!tab || tab !== activeTab) {
-            changeRouteToActiveTab(activeTab)
+        if (!activeType && !activeTab) return
+        if ((!type || type !== activeType) && (!tab || tab !== activeTab)) {
```
Alllex202, line 145:
```suggestion
        if ((!type || type !== activeType) || (!tab || tab !== activeTab)) {
```
It seems that if we now reset the parameters when changing the route, then we can leave only one such condition

Alllex202 (approved):

PR #5439, opened 2024-11-07
fix(condo): DOMA-10616 prepare should only create .env files

pahaz:
Use same logic as in others apps


packages/codegen/templates/app00/bin/prepare.js
```diff
@@ -13,10 +12,6 @@ const main = async ([B2BAppName = '{{name}}']) => {
     await updateAppEnvFile(appName, 'ACCEPT_LANGUAGE', 'accept-language')
     // 2) prepare B2BApp in condo
     await prepareCondoAppB2BAppConfig(appName, B2BAppName, true)
-    // 3) migrate initial data and maketypes
-    await runAppPackageJsonScript(appName, 'makemigrations')
```
nomerdvadcatpyat, line 17: You should also remove it from app01 template

nomerdvadcatpyat (approved):

PR #5441, opened 2024-11-07
feat(condo): DOMA-10511 authentication by token for service users

YEgorLu:


apps/condo/bin/prepare.js
```diff
@@ -8,11 +8,16 @@ async function updateAppEnvAddressSuggestionConfig (serviceName) {
     await updateAppEnvFile(serviceName, 'FAKE_FINANCE_INFO_CLIENT', 'true')
 }
 
+async function updateAppEnvLocalSecrets (serviceName) {
```
pahaz, line 11: Can you use the existing secret?

apps/condo/domains/common/utils/session.js
```diff
@@ -0,0 +1,70 @@
...
+ * @prop {string[]?} b2bPermissionKeys - active permissions
+ * @prop {Record<string, boolean>?} b2bPermissions
+ */
+
+/**
+ * @param {SessionDataArgs?} args
+ * @returns {{organizations: string[] | null, b2bPermissionKeys: string[] | null}}
+ */
+function makeSessionData (args = {}) {
+    let { organizations, b2bPermissionKeys, b2bPermissions } = args
```
pahaz, line 50: check the arguments

apps/condo/bin/prepare.js
```diff
@@ -8,11 +8,16 @@ async function updateAppEnvAddressSuggestionConfig (serviceName) {
...
+    await updateAppEnvFile(serviceName, 'ACCESS_TOKEN_SECRET_KEY', 'secret-key-very-looooooooooooong')
```
pahaz, line 12: if need to add a new variable it should be randomly generated! But really you can check during the code something like: if we didn't set this secret we should use the existing secret cookie secret.

And its better to use something like DATABASE_ENCRYPION_KEY

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,255 @@
...
+                    return get(existingItem, fieldPath)
+                },
+            },
+        },
+
+        token: {
+            schemaDoc: 'Token, ready to be put in Authorization header',
+            type: 'Virtual',
+            graphQLReturnType: 'String',
+            resolver: (parent, args, context, info, extra) => {
```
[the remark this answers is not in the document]
YEgorLu: Resolver returns token on creation and null on other operations
Item can be accessed / created only by admin or employee with role

access: { create: true, ...:false } adds more complexity for tests, and have pretty same result: token can be seen only once on creation. 

Looks like more work for nothing

YEgorLu:
I'm expecting to make field "sessionId" of type [SymmetricEncryptedText](https://github.com/open-condo-software/condo/pull/5450) (after merging PR with him) , so comments about prepare.js and secret will become irrelevant

apps/condo/migrations/20241126164304-0440_b2baccesstoken_b2baccesstokenhistoryrecord_and_more.js
```diff
@@ -0,0 +1,211 @@
...
+--
+ALTER TABLE "B2BAppAccessRightSetHistoryRecord" ADD COLUMN "name" text NULL;
+--
+-- Add field type to b2bappaccessrightsethistoryrecord
+--
+ALTER TABLE "B2BAppAccessRightSetHistoryRecord" ADD COLUMN "type" text NULL;
+--
+-- Add field canManageB2BAccessTokens to userrightsset
+--
+ALTER TABLE "UserRightsSet" ADD COLUMN "canManageB2BAccessTokens" boolean DEFAULT false NOT NULL;
```
dkoviazin, line 58: Do we really need this  in UserRightsSet

apps/condo/domains/user/utils/directAccess/config.js
```diff
@@ -32,6 +32,7 @@ const DIRECT_ACCESS_AVAILABLE_SCHEMAS = {
     // Note: This list should be alphabetical when possible
     lists: [
         // Miniapps domain
+        'B2BAccessToken',
```
dkoviazin, line 35: Direct access looks very dangerous

apps/condo/domains/miniapp/utils/clientSchema/index.ts
```diff
@@ -17,4 +17,5 @@ export * as MessageAppBlackList from './MessageAppBlackList'
 export * as B2BAppAccessRightSet from './B2BAppAccessRightSet'
 export * as B2BAppNewsSharingConfig from './B2BAppNewsSharingConfig'
 export * as B2CAppMessageSetting from './B2CAppMessageSetting'
+export * as B2BAccessToken from './B2BAccessToken'
```
dkoviazin, line 20: do we need it on client schema?

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -344,14 +345,107 @@ describe('B2BAppAccessRightSet', () => {
...
-            await createTestB2BAppAccessRightSet(admin, app)
+            const [rightSet] = await createTestB2BAppAccessRightSet(admin, app)
+            console.error('sssss', rightSet)
 
             await expectToThrowUniqueConstraintViolationError(async () => {
                 await createTestB2BAppAccessRightSet(admin, app)
             }, 'b2b_app_access_right_set_unique_app')
         })
+
+        test('Cannot be created 11 active sets of access rights with type: token for a single app', async () => {
```
dkoviazin, line 358: You need to use constant

apps/condo/domains/miniapp/schema/B2BAccessToken.test.js
```diff
@@ -0,0 +1,637 @@
...
+            const [accessToken] = await createTestB2BAccessToken(admin, b2bAppContext, scopedRightSet)
+
+            await expectToThrowAccessDeniedErrorToObj(async () => {
+                await updateTestB2BAccessToken(client, accessToken.id, updateInput)
+            })
+
+            const [updatedToken] = await updateTestB2BAccessToken(client, accessToken.id, pick(updateInput, 'deletedAt'))
+            expect(updatedToken).toBeDefined()
+        })
+        // const USER_READONLY_FIELDS = [
```
dkoviazin, line 346: if you do not use this code, better to remove it

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,260 @@
...
+            const b2bAppContext = await getByCondition('B2BAppContext', {
+                id: get(newItem, 'context'),
+                deletedAt: null,
+            })
+            const accessRightSet = await getByCondition('B2BAppAccessRightSet', {
+                id: get(newItem, 'rightSet'),
+                deletedAt: null,
+            })
+            
+            function throwErrorOrDelete (error) {
```
dkoviazin, line 186: please, try to find another solution (with no new function inside resolver)

dkoviazin (approved):

apps/condo/domains/common/utils/session.js
```diff
@@ -0,0 +1,75 @@
...
+]
+
+function _getSession (sessionOrWrapper) {
+    for (const sessionPath of POSSIBLE_SESSION_PATHS) {
+        if (has(sessionOrWrapper, sessionPath)) {
+            return get(sessionOrWrapper, sessionPath)
+        }
+    }
+
+    return sessionOrWrapper
```
SavelevMatthew, line 17: Why do you return `sessionOrWrapper ` here? 

It must be null, not passed argument

apps/condo/domains/common/utils/session.js
```diff
@@ -0,0 +1,75 @@
+const get = require('lodash/get')
+const has = require('lodash/has')
+const uniq = require('lodash/uniq')
+
+const POSSIBLE_SESSION_PATHS = [
+    'req.session',
+    'session',
+]
+
+function _getSession (sessionOrWrapper) {
```
SavelevMatthew, line 10: what is `sessionOrWrapper`?

SavelevMatthew: Name is not clear, I have no idea where should this functions be used?

apps/condo/domains/common/utils/session.js
```diff
@@ -0,0 +1,75 @@
...
+    for (const sessionPath of POSSIBLE_SESSION_PATHS) {
```
SavelevMatthew, line 11: Why do you need multiple paths to scan?

apps/condo/domains/common/utils/session.js
```diff
@@ -0,0 +1,75 @@
...
+ * @property {string[] | null } enabledB2BPermissions
+ */
+
+/**
+ * @param {SessionDataArgs?} args
+ * @returns {{allowedOrganizations: string[] | null, enabledB2BPermissions: string[] | null}}
+ */
+function makeSessionData (args = {}) {
+    Object.keys(args).forEach(key => {
+        args[key] = args[key] ? uniq(args[key].filter(Boolean)) : args[key]
```
SavelevMatthew, line 58: You export function, which does something strange....

What if args[key] not an Array? Why do even need this util?

apps/condo/domains/common/utils/session.js
```diff
@@ -0,0 +1,75 @@
...
+    })
+
+    const {
+        allowedOrganizations,
+        enabledB2BPermissions,
+    } = args
+
+    return {
+        allowedOrganizations: allowedOrganizations || null,
```
SavelevMatthew, line 67: A bit unclear...

I can read `allowedOrganizations: null` as "None organizations is allowed"

apps/condo/domains/miniapp/access/B2BAccessToken.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    return false
+}
+
+async function canManageB2BAccessTokens (args) {
+    const { authentication: { item: user } } = args
+    if (!user) return throwAuthenticationError()
+    if (user.deletedAt) return false
+    if (user.isAdmin) return true
```
SavelevMatthew, line 28: Why support is not included? Or directly-accessed to be more precise

apps/condo/domains/miniapp/constants.js
```diff
@@ -1,3 +1,5 @@
+const dayjs = require('dayjs')
```
SavelevMatthew, line 1: Unused import?

apps/condo/domains/miniapp/constants.js
```diff
@@ -76,7 +78,17 @@ const INVALID_PERMISSIONS_ERROR = 'INVALID_PERMISSIONS'
 const APP_BLACK_LIST_ERROR = 'APP_BLACK_LIST_ERROR'
 const PERMISSION_KEY_WRONG_FORMAT_ERROR = 'PERMISSION_KEY_WRONG_FORMAT'
 const ACCESS_RIGHT_SET_NOT_FOR_CONNECTED_B2B_APP = 'ACCESS_RIGHT_SET_NOT_FOR_CONNECTED_B2B_APP'
+const ACCESS_RIGHT_SET_TOO_MANY_OF_TYPE = 'ACCESS_RIGHT_SET_TOO_MANY_OF_TYPE'
 const PERMISSION_NAME_INVALID_LENGTH_ERROR = 'PERMISSION_NAME_INVALID_LENGTH'
+const NAME_REQUIRED = 'NAME_REQUIRED'
```
SavelevMatthew, line 83: Please be more specific

apps/condo/domains/miniapp/constants.js
```diff
@@ -99,6 +111,24 @@ const NEWS_SHARING_PUSH_NOTIFICATION_SETTINGS = {
...
+
+const ACCESS_RIGHT_SET_MAX_ITEMS_GLOBAL_TYPE = 1
+const ACCESS_RIGHT_SET_MAX_ITEMS_SCOPED_TYPE = 10
+
+const ACCESS_RIGHT_SET_MAX_ITEMS_FOR_TYPE = {
+    [ACCESS_RIGHT_SET_GLOBAL_TYPE]: ACCESS_RIGHT_SET_MAX_ITEMS_GLOBAL_TYPE,
+    [ACCESS_RIGHT_SET_SCOPED_TYPE]: ACCESS_RIGHT_SET_MAX_ITEMS_SCOPED_TYPE,
+}
+
+const ACCESS_TOKEN_SESSION_ID_PREFIX = 'b2bAccessToken:'
```
SavelevMatthew, line 130: Should this be hashed or something?

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -35,15 +76,112 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
             },
         },
 
+        name: {
+            schemaDoc: 'Name for right set to distinguish it',
+            type: 'Text',
+            isRequired: true,
+            defaultValue: 'default',
```
SavelevMatthew, line 83: `default` is not matching schemaDoc, if all sets became default = no way to distinguish them :)

apps/condo/domains/miniapp/schema/B2BAccessToken.js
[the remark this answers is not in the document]
SavelevMatthew: > Looks like more work for nothing

You should protect sensitive fields. Model access can be changed any time, this one is important

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,280 @@
...
+                },
+            },
+        },
+
+        token: {
+            schemaDoc: 'Token, ready to be put in Authorization header',
+            type: 'Virtual',
+            graphQLReturnType: 'String',
+            resolver: (parent, args, context, info, extra) => {
+                if (info.operation.name.value === 'createB2BAccessToken') {
```
SavelevMatthew, line 132: This will not work on aliases, use gqlContext instead

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,280 @@
...
+            const newItem = {
+                ...(existingItem || {}),
+                ...resolvedData,
+            }
+
+            if (newItem.deletedAt) {
+                return
+            }
+
+            const b2bAppContext = await getByCondition('B2BAppContext', {
```
SavelevMatthew, line 211: You should check, that all apps in fetched models are the same

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,280 @@
...
+ */
+
+const crypto = require('crypto')
+
+const cookieSignature = require('cookie-signature')
+const dayjs = require('dayjs')
+const get = require('lodash/get')
+const has = require('lodash/has')
+const isEmpty = require('lodash/isEmpty')
+const isNil = require('lodash/isNil')
```
SavelevMatthew, line 12: Are u really need utils like that?)

(Not the biggest fan of 50 imported mini-utils for simple tasks)

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,280 @@
...
+    ACCESS_TOKEN_SERVICE_USER_DOES_NOT_MATCH_CONTEXT,
+    ACCESS_TOKEN_NEGATIVE_TTL,
+    ACCESS_RIGHT_SET_SCOPED_TYPE,
+} = require('@condo/domains/miniapp/constants')
+const { SERVICE_USER_FIELD } = require('@condo/domains/miniapp/schema/fields/accessRight')
+const { getEnabledPermissions } = require('@condo/domains/miniapp/utils/permissions')
+
+// NOTE(YEgorLu): using deprecated function to get same cookieSecret, that app uses
+// otherwise I would need to copy code to get same result
+const { cookieSecret } = prepareDefaultKeystoneConfig(conf)
```
SavelevMatthew, line 39: Why not conf?

apps/condo/domains/miniapp/schema/B2BAppContext.js
```diff
@@ -135,6 +135,13 @@ const B2BAppContext = new GQLListSchema('B2BAppContext', {
 
             if (isSoftDeletedHappened) {
                 await deleteB2BAppRoles.delay(appId, organizationId)
+                const accessTokens = await find('B2BAccessToken', {
+                    context: { id: updatedItem.id },
+                    deletedAt: null,
+                })
+                for (const accessToken of accessTokens) {
+                    await B2BAccessToken.update(context, accessToken.id, { dv: accessToken.dv, sender, deletedAt: new Date().toISOString() }, 'id')
```
SavelevMatthew, line 143: If app failed here, should there be some cron task to kill tokens, it'll be more secure

apps/condo/domains/miniapp/utils/permissions.js
```diff
@@ -0,0 +1,41 @@
+const { B2B_APP_SERVICE_USER_ACCESS_AVAILABLE_SCHEMAS } = require('@condo/domains/miniapp/utils/b2bAppServiceUserAccess/config')
+const { generatePermissionFields } = require('@condo/domains/miniapp/utils/b2bAppServiceUserAccess/schema.utils')
+
+
+const PERMISSION_FIELDS_KEYS = Object.keys(generatePermissionFields({ config: B2B_APP_SERVICE_USER_ACCESS_AVAILABLE_SCHEMAS }))
```
SavelevMatthew, line 5: This should be taken from shema, shouldn't it?

apps/condo/domains/miniapp/schema/B2BAccessToken.js
YEgorLu: Already have. 
1. Get B2BAppContext
2. Get B2BAppRightsSet
3. Get B2BAppAccessRight with same app, as b2bAppContext (b2bAppContext.app === accessRightSet.app)
4. Check app equality of context and right set

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
YEgorLu: It does match. We had one B2BAppAccessRightSet on miniapp. Now we have multiple. And to make multiple sets we must have one of type GLOBAL first. Sets, which we already have, becomes GLOBAL aka 'default'

apps/condo/domains/common/utils/session.js
```diff
@@ -0,0 +1,78 @@
...
```
SavelevMatthew, line 70: `* | Array<string>`

apps/condo/domains/miniapp/schema/B2BAccessToken.js
SavelevMatthew: think about deleting random

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,282 @@
...
+    if (!b2bAppContext) {
+        return ERRORS.CONTEXT_NOT_EXISTS
+    }
+    if (!accessRightSet) {
+        return ERRORS.RIGHT_SET_NOT_EXISTS
+    }
+    if (get(accessRightSet, 'type') !== ACCESS_RIGHT_SET_SCOPED_TYPE) {
+        return ERRORS.WRONG_RIGHT_SET_TYPE
+    }
+    if (get(accessRightSet, 'app') && (get(accessRightSet, 'app') !== get(b2bAppContext, 'app'))) {
```
SavelevMatthew, line 55: accessRightSet.app

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,282 @@
...
+    ERRORS.NEGATIVE_TTL,
+])
+
+const B2BAccessToken = new GQLListSchema('B2BAccessToken', {
+    schemaDoc: 'Access to service user with additional restrictions by Organization and B2BAppAccessRightSet',
+    fields: {
+        sessionId: {
+            schemaDoc: 'Encrypted sessionId of session',
+            type: 'EncryptedText',
+            isRequired: true,
```
SavelevMatthew, line 109: pass manager directly

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,282 @@
...
+            access: access.updatableOnlyForAdmin,
+            hooks: {
+                resolveInput ({ operation, existingItem, fieldPath, originalInput }) {
+                    const inputSessionId = get(originalInput, fieldPath)
+                    // need to have option to manually set sessionId, so we can migrate already given tokens to this
+                    if (!isNil(inputSessionId) && !isEmpty(inputSessionId)) {
+                        return inputSessionId
+                    }
+                    if (operation === 'create') {
+                        return ACCESS_TOKEN_SESSION_ID_PREFIX + crypto.randomUUID()
```
SavelevMatthew, line 119: default behaviour?

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,282 @@
...
+        token: {
+            schemaDoc: 'Token, ready to be put in Authorization header. Shown only once after creation',
+            type: 'Virtual',
+            graphQLReturnType: 'String',
+            resolver: (item) => {
+                if (!isNil(item.token)) {
+                    return item.token
+                }
+                return null
+            },
```
SavelevMatthew, line 135: access?

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,282 @@
...
+
+        user: {
+            ...SERVICE_USER_FIELD,
+            isRequired: true,
+            access: access.updatableOnlyForAdmin,
+            hooks: {
+                ...SERVICE_USER_FIELD.hooks,
+                async resolveInput ({ operation, existingItem, resolvedData, fieldPath }) {
+                    if (operation === 'create' && !has(resolvedData, fieldPath)) {
+                        const contextId = get(resolvedData, 'context') || get(existingItem, 'context')
```
SavelevMatthew, line 156: const newItem = { ...existingItem, ...resolvedData }

const context = get(newItem, context)

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,282 @@
...
+                        throw new GQLError(ERRORS.NEGATIVE_TTL, context)
+                    }
+                },
+            },
+        },
+    },
+    hooks: {
+        async validateInput ({ operation, resolvedData, existingItem, context }) {
+            const newItem = {
+                ...(existingItem || {}),
```
SavelevMatthew, line 201:
```suggestion
                ...existingItem,
```

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,282 @@
...
+                ...resolvedData,
+            }
+
+            if (newItem.deletedAt) {
```
SavelevMatthew, line 205: update({ somethingWrong, deletedAt: '...' }) -> true
update({ deletedAt: null })

apps/condo/domains/miniapp/access/B2BAccessToken.js
YEgorLu: We afraid to give rights for this entity for supports. Adding direct access -> adding rule in UserRightSet -> supports can give rights to themselves with no validation

apps/condo/domains/miniapp/schema/B2BAppContext.js
YEgorLu: For start there will be only few tokens, so task can appear later

apps/condo/domains/common/utils/session.js
YEgorLu: I asked gpt, I don't want to do something like 
```
{
    [key]: { hasRestrictions: boolean, values: [...] }
}
```
because it looks gross and big.

Returning `true` should be okay to indicate, that everything is allowed, I think

apps/condo/domains/miniapp/schema/B2BAccessToken.js
YEgorLu: I found out that in case of virtual field, only 'read' access is called. And it does not contain information about operation (creating entity, 'create' access is not called, only 'read')

apps/condo/domains/miniapp/schema/B2BAccessToken.js
YEgorLu: I found out that in case of virtual field, only 'read' access is called. And it does not contain information about operation (creating entity, 'create' access is not called, only 'read').

So access for virtual field does nothing, aside of User checks, which defined by list access

apps/condo/domains/common/utils/session.js
```diff
@@ -0,0 +1,80 @@
...
+    } = Object.fromEntries(arrayArgsEntries)
+
+    return {
+        allowedOrganizations,
+        enabledB2BPermissions,
+    }
+}
+
+function _uniqNonNull (array) {
+    return uniq(array).filter(item => item !== null && item !== undefined)
```
SavelevMatthew, line 74: `import { nonNull } from '@open-condo/miniapp-utils/helpers/collections'`

apps/condo/domains/miniapp/access/B2BAccessToken.js
```diff
@@ -0,0 +1,60 @@
...
+ * Generated by `createschema miniapp.B2BAccessToken 'sessionId:Text; context:Relationship:B2BAppContext:CASCADE; rightSet:Relationship:B2BAppAccessRightSet:CASCADE; expiresAt:Text;'`
+ */
+
+const { userIsAdmin, isSpecificQuery } = require('@open-condo/keystone/access')
+const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
+
+const { canReadObjectsAsB2BAppServiceUser, canManageObjectsAsB2BAppServiceUser } = require('@condo/domains/miniapp/utils/b2bAppServiceUserAccess')
+const { SERVICE } = require('@condo/domains/user/constants/common')
+
+async function canReadB2BAccessTokens (args) {
```
SavelevMatthew, line 11: I guess you should protect filtering by any sensitive fields here, like we do in user...

So queries like `allB2BAccessTokens(where: { token_starts_with: })` are banned, because you can brute force them then

YEgorLu: Virtual fields are not queriable
Admin needs to be able to query by `sessionId` I'm sure
Service users anyway can see only their's tokens, so it is pointless
But if we restrict to them access for `sessionId`, which is surely not needed for them, it might have some cons, so made these 2 restrictions for them

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,258 @@
...
+        id: item.rightSet,
+        deletedAt: null,
+    })
+    if (!accessRightSet) {
+        return ERRORS.RIGHT_SET_NOT_EXISTS
+    }
+    if (accessRightSet.type !== ACCESS_RIGHT_SET_SCOPED_TYPE) {
+        return ERRORS.WRONG_RIGHT_SET_TYPE
+    }
+    if (accessRightSet.app && (accessRightSet.app !== b2bAppContext.app)) {
```
SavelevMatthew, line 56: logic is not exactly right, It potentially can be possible to have no accessRightSet.app (during set_null delete or something), so you should check that both context and rightSet has app field and they're matching

SavelevMatthew: Right now if accessRightSet.app === null, check will'be skipped

SavelevMatthew: Easiest way to fix it, add app condition in find where above, like 

```
await getByCondition('B2BAppAccessRightSet', {
        id: item.rightSet,
        deletedAt: null,
        app: { deletedAt: null }
        user: { deletedAt: null }
    })
```

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,258 @@
...
+            const newItem = {
+                ...existingItem,
+                ...resolvedData,
+            }
+            const validationError = await getValidationError(newItem)
+
+            if (operation === 'create' && validationError) {
+                throw new GQLError(validationError, context)
+            }
+            if (operation === 'update' && validationError && ERRORS_FOR_DELETION.has(validationError)) {
```
SavelevMatthew, line 214: What this branch is doing? Seems strange...

SavelevMatthew:
```
const { isSoftDelete } = require('@open-condo/keystone/access')

// NOTE: Allow setting deletedAt, but not changing other fields with it
if (isSoftDelete) {
    return
}

const validationError = await getValidationError(newItem)
if (validationError) {
   throw new ...
}
```

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,258 @@
...
+                resolvedData.deletedAt = resolvedData.deletedAt ? resolvedData.deletedAt : dayjs().toISOString()
+            }
+        },
+        async afterChange ({ context, operation, existingItem, updatedItem }) {
+            const softDeleted = operation === 'update' && !existingItem['deletedAt'] && updatedItem['deletedAt']
```
SavelevMatthew, line 219: Why not just use `updatedItem`? Other checks can isn't necessary. And even wrong

SavelevMatthew: Security issue!

SavelevMatthew: In request above isSoftDelete won't be triggered, so new session will be created and will exist in redis, but not in Postgres. So its a shadow session, which is dangerous

SavelevMatthew: Fix this and add test for that

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,258 @@
...
+            hooks: {
+                ...SERVICE_USER_FIELD.hooks,
+                async resolveInput ({ operation, existingItem, resolvedData, fieldPath }) {
+                    const newItem = {
+                        ...existingItem,
+                        ...resolvedData,
+                    }
+                    if (operation === 'create' && !newItem[fieldPath]) {
+                        const contextId = get(resolvedData, 'context') || get(existingItem, 'context')
+                        const context = await getById('B2BAppContext', contextId)
```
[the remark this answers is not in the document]
SavelevMatthew: Add test for that

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
```diff
@@ -239,6 +239,31 @@ describe('B2BAppAccessRight', () => {
                 })
             }, ERRORS.ACCESS_RIGHT_SET_NOT_FOR_CONNECTED_B2B_APP)
         })
+
+        test('Can not create or update "B2BAppAccessRight" if "accessRightSet"."type" != "GLOBAL"', async () => {
+            const admin = await makeLoggedInAdminClient()
+            const [serviceUser] = await registerNewServiceUserByTestClient(admin)
```
SavelevMatthew, line 245: Please make tests more real, create service user from support. 

Because its additional real-world protection, if you messed up with global logic this can be additional indicator, that existing business process isn't working

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
```diff
@@ -239,6 +239,31 @@ describe('B2BAppAccessRight', () => {
...
+
+            const [app] = await createTestB2BApp(admin)
```
SavelevMatthew, line 247: Same, at least replace it with support. Try to make tests closer to real world

apps/condo/domains/miniapp/schema/B2BAppAccessRight.test.js
```diff
@@ -239,6 +239,31 @@ describe('B2BAppAccessRight', () => {
...
+
+            const [accessRightSetGlobal] = await createTestB2BAppAccessRightSet(admin, app)
```
SavelevMatthew, line 249: same

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -35,15 +75,112 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
...
-        ...generatePermissionFields({ config: B2B_APP_SERVICE_USER_ACCESS_AVAILABLE_SCHEMAS }),
+        name: {
+            schemaDoc: 'Name for right set to distinguish it',
+            type: 'Text',
+            isRequired: true,
+            defaultValue: 'default',
+            hooks: {
+                validateInput ({ resolvedData, existingItem, context }) {
+                    const nextItem = {
+                        ...(existingItem || {}),
```
SavelevMatthew, line 86:
```suggestion
                        ...existingItem,
```

Spreading undefined is safe

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -35,15 +75,112 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
...
+                        ...resolvedData,
+                    }
+                    const name = get(nextItem, 'name')
+                    if (isEmpty(name)) {
```
SavelevMatthew, line 90: How about to trim then ?

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -35,15 +75,112 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
...
+            },
+        },
 
+        ...B2B_PERMISSION_FIELDS,
+
+    },
+    hooks: {
+        async validateInput ({ resolvedData, existingItem, context }) {
+            const nextItem = {
+                ...(existingItem || {}),
```
SavelevMatthew, line 113: same

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -35,15 +75,112 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
...
+                }
+            }
+
+            if (operation === 'update' && updatedItem.type === ACCESS_RIGHT_SET_SCOPED_TYPE) {
+                const accessTokens = await find('B2BAccessToken', {
+                    rightSet: { id: updatedItem.id },
+                    deletedAt: null,
+                })
+                for (const accessToken of accessTokens) {
+                    await B2BAccessToken.update(context, accessToken.id, {
```
SavelevMatthew, line 170: Please add a note, why do you use empty update. I've figured out to trigger some hooks in them, but it's not obvious in the file context, so someone can delete this lines

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -344,14 +345,106 @@ describe('B2BAppAccessRightSet', () => {
...
+        test('Cannot be created 2 active set of access rights with type: GLOBAL for a single app', async () => {
             const [app] = await createTestB2BApp(admin)
             await createTestB2BAppAccessRightSet(admin, app)
 
             await expectToThrowUniqueConstraintViolationError(async () => {
                 await createTestB2BAppAccessRightSet(admin, app)
             }, 'b2b_app_access_right_set_unique_app')
         })
+
+        test('Cannot be created 11 active sets of access rights with type: SCOPED for a single app', async () => {
```
SavelevMatthew, line 357: Use consts please, not hard-code 11, 12 or any other numbers

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -344,14 +345,106 @@ describe('B2BAppAccessRightSet', () => {
...
+            await expectToThrowGQLError(async () => {
+                await createTestB2BAppAccessRightSet(admin, app, { type: 'SCOPED' })
+            }, {
+                code: 'BAD_USER_INPUT',
+                type: 'ACCESS_RIGHT_SET_TOO_MANY_OF_TYPE',
+                message: 'Too many items of type "SCOPED". Maximum is "10"',
+            }, 'obj')
+        })
+
+        describe('name', () => {
```
SavelevMatthew, line 373: This is not part of constraints block, you should move it out

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -344,14 +345,106 @@ describe('B2BAppAccessRightSet', () => {
...
+            await expectToThrowGQLError(async () => {
+                await updateTestB2BAppAccessRightSet(admin, tokenRightSet.id, { canReadMeters: true })
+            }, {
+                code: 'BAD_USER_INPUT',
+                type: 'ACCESS_RIGHT_SET_TOO_MANY_PERMISSIONS',
+                message: 'You trying to give to right set more permissions, than "GLOBAL" right set has',
+            }, 'obj')
+        })
+
+        test('Removing permissions on "GLOBAL" B2BAppAccessRightSet leads to removing these permissions from entities with type "SCOPED"', async () => {
```
SavelevMatthew, line 433: deleting global should lead to deleting scoped

apps/condo/domains/miniapp/schema/B2BAppContext.js
```diff
@@ -135,6 +135,13 @@ const B2BAppContext = new GQLListSchema('B2BAppContext', {
...
+                }
```
SavelevMatthew, line 144: test?

apps/condo/domains/miniapp/schema/fields/b2bAccessRightSet.js
```diff
@@ -0,0 +1,10 @@
+const { B2B_APP_SERVICE_USER_ACCESS_AVAILABLE_SCHEMAS } = require('../../utils/b2bAppServiceUserAccess/config')
```
SavelevMatthew, line 1: Use absolute import.

`./something` is fine
`../something` is ok, if part of the same module / block
`../../something`, probably not ok

apps/condo/index.js
```diff
@@ -145,9 +147,27 @@ const extendExpressApp = (app) => {
     app.use(Sentry.Handlers.errorHandler())
 }
 
+const authStrategyOpts = {
+    hooks: {
+        // forbid unauth with B2BAccessToken
+        beforeUnauth ({ context }) {
```
SavelevMatthew, line 153: Why?

SavelevMatthew: Something like:
```
Sessions linked can be permanent tokens can only be destroyed by deleting linking model, such as B2BAppToken, ....
```

packages/keystone/KSv5v6/v5/prepareKeystone.js
```diff
@@ -151,6 +151,7 @@ function prepareKeystone ({ onConnect, extendKeystoneConfig, extendExpressApp, s
                 // NOTE: if key not found returns 0, else 1.
                 await redisClient.expire(sessKey, SERVICE_USER_SESSION_TTL_IN_SEC)
             },
+            ...get(authStrategyOpts, 'hooks', {}),
```
SavelevMatthew, line 154: Should you compose hooks, like in plugins?

apps/condo/domains/miniapp/schema/B2BAccessToken.js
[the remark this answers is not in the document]
YEgorLu: I have no idea how to write this kind of test without random fails.
From my understanding getById returning deleted entity should be super rare case, because it is being validated in `validateInput`

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -21,6 +55,12 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
         '\nNOTE: Some schemas have links with other models.' +
         ' Therefore, it is necessary to take this into account when specifying the necessary rights.' +
         '\nFor example, to create a contact, in addition to having access to managing properties, you also need to have access to read organization',
+
+    resolveLabel (item) {
```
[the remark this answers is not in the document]
YEgorLu: I forgot where i saw it...
Replaced with labelResolver

apps/condo/domains/miniapp/schema/B2BAppContext.js
YEgorLu: testing that in B2BAccessToken

packages/keystone/KSv5v6/v5/prepareKeystone.js
YEgorLu: Why not
Let's compose default hooks, written in prepareKeystone and exported from app hooks

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,276 @@
...
+                        const [accessRight] = await find('B2BAppAccessRight', {
```
SavelevMatthew, line 169: What if app is null? Please add getByCondition above, does not rely on logic in other places...

apps/condo/domains/miniapp/schema/B2BAccessToken.js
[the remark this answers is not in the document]
SavelevMatthew: I don't want a "super rare case" to even exist, I want logic to work 100% of time, so once you implement it good, it can be left for years

packages/keystone/KSv5v6/v5/prepareKeystone.js
```diff
@@ -115,7 +116,28 @@ function prepareKeystone ({ onConnect, extendKeystoneConfig, extendExpressApp, s
...
+            // Example: 12345678901234567890123456789012.asdhaksdjhajskdhajskdhjkas
+            // Session token can be build like so "sess:{prefixBeforeDot}"
+            const sessToken = token.split('.')[0]
+            const sessKey = `sess:${sessToken}`
+            const redisClient = getRedisClient()
+            // NOTE: if key not found returns 0, else 1.
+            await redisClient.expire(sessKey, SERVICE_USER_SESSION_TTL_IN_SEC)
+        },
+    }
+    const authStrategyConfig = get(authStrategyOpts, 'config', { hooks: { afterAuth: [defaultAuthStrategyConfigHooks.afterAuth] } })
```
SavelevMatthew, line 140: Why to use this get hack? Why not to just compose all hooks together?

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,276 @@
...
+
+            if (isSoftDelete(originalInput)) {
+                return
+            }
+
+            const validationError = await getValidationError(newItem)
+            if (operation === 'create' && validationError) {
+                throw new GQLError(validationError, context)
+            }
+            if (operation === 'update' && validationError) {
```
SavelevMatthew, line 224: Can you explain this branch?

What is ERROR_FOR_DELETION, what's happening with deletedAt? Seems like some not obvious hacks

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -35,15 +75,128 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
...
+                        ...resolvedData,
+                    }
+                    return get(nextItem, 'name', '').trim()
+                },
+                validateInput ({ resolvedData, existingItem, context }) {
+                    const nextItem = {
+                        ...existingItem,
+                        ...resolvedData,
+                    }
+                    const name = get(nextItem, 'name', '')
```
SavelevMatthew, line 96: trim? or regex check?

SavelevMatthew: Because right now I can pass something like `" "`...Is this valuable?

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,276 @@
...
+                if (!ERRORS_FOR_DELETION.has(validationError)) {
+                    throw new GQLError(validationError, context)
+                }
+                resolvedData.deletedAt = resolvedData.deletedAt ? resolvedData.deletedAt : dayjs().toISOString()
```
SavelevMatthew, line 228: Bad practice to mutate data inside

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,276 @@
...
+        },
+        async afterChange ({ context, operation, existingItem, updatedItem }) {
+            const softDeleted = Boolean(operation === 'update' && !existingItem['deletedAt'] && updatedItem['deletedAt'])
+            const sessionId = encryptionManager.decrypt(updatedItem.sessionId)
+
+            if (softDeleted) {
+                await destroySession(context.req.sessionStore, sessionId)
+            }
+
+            if (!softDeleted && !updatedItem['deletedAt']) {
```
SavelevMatthew, line 239:
```suggestion
            if (!updatedItem['deletedAt']) {
```

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,276 @@
...
+                    throw new GQLError(validationError, context)
+                }
+                resolvedData.deletedAt = resolvedData.deletedAt ? resolvedData.deletedAt : dayjs().toISOString()
+            }
+        },
+        async afterChange ({ context, operation, existingItem, updatedItem }) {
+            const softDeleted = Boolean(operation === 'update' && !existingItem['deletedAt'] && updatedItem['deletedAt'])
+            const sessionId = encryptionManager.decrypt(updatedItem.sessionId)
+
+            if (softDeleted) {
```
SavelevMatthew, line 235:
```suggestion
            if (updatedItem['deletedAt']) {
```

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,276 @@
...
+                // Strange case, if these entities were deleted in time between hooks
+                if (!rightSet) {
+                    throw new GQLError(ERRORS.RIGHT_SET_NOT_EXISTS, context)
+                }
+                if (!b2bAppContext) {
+                    throw new GQLError(ERRORS.CONTEXT_NOT_EXISTS, context)
+                }
+                const permissions = Object.keys(getEnabledPermissions(rightSet))
+                const additionalFields = makeSessionData({
+                    allowedOrganizations: [get(b2bAppContext, 'organization')],
```
SavelevMatthew, line 251: `[undefined]`

apps/condo/domains/common/utils/session.js
```diff
@@ -0,0 +1,86 @@
...
+ * @typedef SessionData
+ * @property {string[] | null} allowedOrganizations
+ * @property {string[] | null } enabledB2BPermissions
+ */
+
+/**
+ * @param {SessionDataArgs?} args
+ * @returns {{allowedOrganizations: string[] | null, enabledB2BPermissions: string[] | null}}
+ */
+function makeSessionData (args = {}) {
```
SavelevMatthew, line 48: payload, initialState

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -35,15 +75,128 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
...
+                        { type: ACCESS_RIGHT_SET_GLOBAL_TYPE },
+                        { type_not: ACCESS_RIGHT_SET_GLOBAL_TYPE, deletedAt: null },
+                    ],
+                })
+                
+                const globalRightSet = rightSetsForApp.find(set => set.type === ACCESS_RIGHT_SET_GLOBAL_TYPE)
+                if (!globalRightSet && !resolvedData['deletedAt']) {
+                    throw new GQLError(ERRORS.GLOBAL_RIGHT_SET_REQUIRED, context)
+                }
+                const sameTypeRightSets = rightSetsForApp.filter(set => set.type === nextItem.type)
```
SavelevMatthew, line 136: create -> delete -> create

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -35,15 +75,128 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
...
+            if (nextItem.type !== ACCESS_RIGHT_SET_GLOBAL_TYPE) {
+                const rightSetsForApp = await find('B2BAppAccessRightSet', {
+                    app: { id: nextItem.app },
+                    OR: [
+                        { type: ACCESS_RIGHT_SET_GLOBAL_TYPE },
+                        { type_not: ACCESS_RIGHT_SET_GLOBAL_TYPE, deletedAt: null },
+                    ],
+                })
+                
+                const globalRightSet = rightSetsForApp.find(set => set.type === ACCESS_RIGHT_SET_GLOBAL_TYPE)
```
SavelevMatthew, line 132: check deletedAt

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -35,15 +75,128 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
...
+                    deletedAt: null,
+                })
+                for (const rightSet of otherTypeRightSets) {
+                    const dvSender = { dv: rightSet.dv, sender: updatedItem.sender }
+                    const diffPermissions = getPermissionsDiff(updatedItem, rightSet)
+                    const permissionsToDisable = getEnabledPermissions(diffPermissions)
+                    const disabledPermissions = Object.fromEntries(Object.keys(permissionsToDisable).map((perm) => [perm, false]))
+                    const updateInput = { deletedAt: updatedItem.deletedAt ? new Date().toISOString() : undefined }
+                    if (Object.keys(disabledPermissions).length) {
+                        Object.assign(updateInput, disabledPermissions)
```
SavelevMatthew, line 167:
```suggestion
                        updateInput = Object.assign(updateInput, disabledPermissions)
```

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -35,15 +75,128 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
...
+            }
+
+            if (operation === 'update' && updatedItem.type === ACCESS_RIGHT_SET_SCOPED_TYPE) {
+                const accessTokens = await find('B2BAccessToken', {
+                    rightSet: { id: updatedItem.id },
+                    deletedAt: null,
+                })
+                for (const accessToken of accessTokens) {
+                    // NOTE(YEgorLu): Do not pass anything there, because B2BAccessToken knows how to manage himself
+                    await B2BAccessToken.update(context, accessToken.id, {
```
SavelevMatthew, line 186: chunk by 500 -> updateMany

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -344,14 +346,120 @@ describe('B2BAppAccessRightSet', () => {
...
+            const [app] = await createTestB2BApp(support)
+            const [globalRightsSet] = await createTestB2BAppAccessRightSet(support, app, { canReadOrganizations: true })
+
+            for (let i = 0; i < 3; i++) {
```
SavelevMatthew, line 413: <3

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -344,14 +346,120 @@ describe('B2BAppAccessRightSet', () => {
...
+            scopedRightSets.forEach(scopedRightSet => {
+                expect(scopedRightSet.canReadOrganizations).toEqual(false)
+            })
+        })
+
+        test('Deleting "GLOBAL" B2BAppAccessRightSet is leading for related right sets of other types deletion', async () => {
+            const [app] = await createTestB2BApp(support)
+            const [globalRightsSet] = await createTestB2BAppAccessRightSet(support, app, { canReadOrganizations: true })
+
+            for (let i = 0; i < 3; i++) {
```
SavelevMatthew, line 429: <3

apps/condo/domains/miniapp/schema/B2BAppContext.js
```diff
@@ -135,6 +135,13 @@ const B2BAppContext = new GQLListSchema('B2BAppContext', {
 
             if (isSoftDeletedHappened) {
                 await deleteB2BAppRoles.delay(appId, organizationId)
+                const accessTokens = await find('B2BAccessToken', {
+                    context: { id: updatedItem.id },
+                    deletedAt: null,
+                })
+                for (const accessToken of accessTokens) {
```
SavelevMatthew, line 142: updateMany preferable

apps/condo/domains/miniapp/utils/permissions.js
```diff
@@ -0,0 +1,38 @@
+const { B2B_PERMISSION_FIELDS } = require('@condo/domains/miniapp/schema/fields/b2bAccessRightSet')
+
+const PERMISSION_FIELDS_KEYS = Object.keys(B2B_PERMISSION_FIELDS)
+
+/** Get permission fields from b2bAppAccessRightSetRight, which values differs from b2bAppAccessRightSetLeft */
+function getPermissionsDiff (b2bAppAccessRightSetLeft, b2bAppAccessRightSetRight) {
```
SavelevMatthew, line 6:
```suggestion
function getPermissionsDiff (referenceRightSet, inspectedRightSet) {}
```

packages/keystone/session.js
```diff
@@ -0,0 +1,136 @@
...
+ *          "keystoneListKey": "User",
+ *          "keystoneItemId": "uuid"
+ *          "rights": ["canManageIntegrations]
+ *      }
+ *  *\/
+ *
+ *  await destroySession(context.req.sessionStore, sessionId)
+ *  // context.req.session = null
+ */
+function setSession (sessionStore, config) {
```
SavelevMatthew, line 89: Think about dynamic listKey as it KS does

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,276 @@
...
+                    enabledB2BPermissions: permissions,
+                })
+                const expiresAt = get(updatedItem, 'expiresAt')
+                await setSession(context.req.sessionStore, {
+                    sessionId,
+                    userId: get(updatedItem, 'user'),
```
SavelevMatthew, line 257: Check user existence?

apps/condo/domains/miniapp/schema/B2BAccessToken.test.js
```diff
@@ -0,0 +1,623 @@
...
+                    message: 'B2BAppContext and B2BRightSet connected to different B2BApps',
+                }, 'obj')
+            }
+
+        })
+    })
+
+    describe('Real-life cases', () => {
+
+        test('Can not update deleted access token to store it in redis', async () => {
```
SavelevMatthew, line 378: Maybe spec?

apps/condo/domains/miniapp/schema/B2BAccessToken.test.js
```diff
@@ -0,0 +1,623 @@
...
+const {
+    makeLoggedInAdminClient, makeClient, expectValuesOfCommonFields,
+    expectToThrowAuthenticationErrorToObj, expectToThrowAuthenticationErrorToObjects,
+    expectToThrowAccessDeniedErrorToObj, expectToThrowGQLError, expectToThrowAccessDeniedErrorToObjects,
+    makeLoggedInClient, expectToThrowAccessDeniedError,
+} = require('@open-condo/keystone/test.utils')
+
+const { encryptionManager } = require('@condo/domains/common/utils/encryption')
+const { B2BAccessToken, createTestB2BAccessToken, createTestB2BAppContext,
+    createTestB2BAppAccessRightSet, createTestB2BAppAccessRight, createTestB2BAccessTokenReadonly,
```
SavelevMatthew, line 21: const {
    A,
    A,
    A,
    A,
} = require('...')

apps/condo/domains/miniapp/schema/B2BAccessToken.test.js
```diff
@@ -0,0 +1,623 @@
...
+    let scopedRightSet
+    let globalRightSet
+    let serviceUser
+    beforeEach(async () => {
+        admin = await makeLoggedInAdminClient()
+        ;[b2bApp] = await createTestB2BApp(admin, {
+            name: faker.company.name().replace(/ /, '-').toUpperCase() + ' B2B APP',
+            developer: faker.company.name(),
+        })
+        ;[organization] = await createTestOrganization(admin)
```
SavelevMatthew, line 47: ;?)

apps/condo/domains/miniapp/schema/B2BAccessToken.test.js
```diff
@@ -0,0 +1,623 @@
...
+        ;[globalRightSet] = await createTestB2BAppAccessRightSet(admin, b2bApp)
+        ;[scopedRightSet] = await createTestB2BAppAccessRightSet(admin, b2bApp, {
+            name: faker.random.alphaNumeric(8),
+            type: 'SCOPED',
+        })
+        ;[serviceUser] = await registerNewServiceUserByTestClient(admin)
+        await createTestB2BAppAccessRight(admin, serviceUser, b2bApp, globalRightSet)
+    })
+
+    describe('CRUD tests', () => {
```
SavelevMatthew, line 58: image

apps/condo/domains/miniapp/schema/B2BAccessToken.test.js
```diff
@@ -0,0 +1,623 @@
...
+
+            test('user can\'t', async () => {
+                const client = await makeClientWithNewRegisteredAndLoggedInUser()
+                await expectToThrowAccessDeniedErrorToObj(async () => {
+                    await createTestB2BAccessTokenReadonly(client, b2bAppContext, scopedRightSet)
+                })
+            })
+
+            test('serviceUser can with rights', async () => {
+                const [b2bApp] = await createTestB2BApp(admin)
```
SavelevMatthew, line 86: why not support?

apps/condo/domains/miniapp/schema/B2BAccessToken.test.js
```diff
@@ -0,0 +1,623 @@
...
+                const [organization] = await createTestOrganization(admin)
+                const [b2bAppContext] = await createTestB2BAppContext(admin, b2bApp, organization, { status: 'Finished' })
+                const [globalRightSet] = await createTestB2BAppAccessRightSet(admin, b2bApp, { canManageB2BAccessTokens: true, canReadB2BAccessTokens: true })
+                const [scopedRightSet] = await createTestB2BAppAccessRightSet(admin, b2bApp, { type: 'SCOPED' })
+
+                const serviceUserClient = await makeClientWithServiceUser()
+                await createTestB2BAppAccessRight(admin, serviceUserClient.user, b2bApp, globalRightSet)
+                const [accessToken] = await createTestB2BAccessTokenReadonly(serviceUserClient, b2bAppContext, scopedRightSet)
+                const t = await B2BAccessTokenReadonly.getOne(serviceUserClient, { sessionId: accessToken.sessionId })
```
SavelevMatthew, line 95: token

apps/condo/domains/miniapp/schema/B2BAccessToken.test.js
```diff
@@ -0,0 +1,623 @@
...
+                name: faker.random.alphaNumeric(8),
+                type: 'SCOPED',
+            })
+            const serviceUserClient = await makeClientWithServiceUser()
+            await createTestB2BAppAccessRight(admin, serviceUserClient.user, b2bApp, globalRightSet)
+
+            const [accessToken] = await createTestB2BAccessToken(serviceUserClient, b2bAppContext, scopedRightSet)
+
+            await expectToThrowAccessDeniedErrorToObj(async () => {
+                await updateTestB2BAccessTokenReadonly(serviceUserClient, accessToken.id, updateInput)
```
SavelevMatthew, line 322: {
  a, b, c
}

-> test.each .. { a }, { b }, { c }

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
YEgorLu: It's actually not testing maximum count of entities, just constant to create some right sets for testing, not related to anything.
Made this number random between 1 and maximum scoped right sets count

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
YEgorLu: t's actually not testing maximum count of entities, just constant to create some right sets for testing, not related to anything.
Made this number random between 1 and maximum scoped right sets count

packages/keystone/KSv5v6/v5/prepareKeystone.js
YEgorLu: this line (140) is my mistake. Config should be {} as before
In prepare let's join defined here hooks and hooks, provided from app
Getting many hooks of same type from app seems illogical, better let app handle their composing or merge before passing here

packages/keystone/KSv5v6/v5/prepareKeystone.js
```diff
@@ -131,27 +153,7 @@ function prepareKeystone ({ onConnect, extendKeystoneConfig, extendExpressApp, s
...
-                // Example: 12345678901234567890123456789012.asdhaksdjhajskdhajskdhjkas
-                // Session token can be build like so "sess:{prefixBeforeDot}"
-                const sessToken = token.split('.')[0]
-                const sessKey = `sess:${sessToken}`
-                const redisClient = getRedisClient()
-                // NOTE: if key not found returns 0, else 1.
-                await redisClient.expire(sessKey, SERVICE_USER_SESSION_TTL_IN_SEC)
-            },
-        },
+        hooks: composeHooks(defaultAuthStrategyConfigHooks, get(authStrategyOpts, 'hooks', {})),
```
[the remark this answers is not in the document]
YEgorLu: compose functions have arguments (originalHook, newHook) => ...
seems logical to past defined here hooks as original hooks

packages/keystone/session.js
YEgorLu: keystone's auth strategy does not have dynamic list key. It's predefined constant. So let's just allow passing needed listKey

apps/condo/domains/common/utils/session.js
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * @param {SessionDataArgs?} payload
+ * @returns {{allowedOrganizations: string[] | null, enabledB2BPermissions: string[] | null}}
+ */
+function makeSessionData (payload = {}) {
+    const parsedPayloadEntries = Object.entries(PAYLOAD_CONFIG)
+        .map(([key, config]) => {
+            let value = payload[key]
+            if (value === undefined) {
```
SavelevMatthew, line 56: if (payload.hasOwnProperty())

apps/condo/domains/common/utils/session.js
```diff
@@ -0,0 +1,95 @@
+const { Session } = require('express-session')
+const get = require('lodash/get')
+const uniq = require('lodash/uniq')
+
+const { nonNull } = require('@open-condo/miniapp-utils')
+
+const PAYLOAD_CONFIG = {
```
SavelevMatthew, line 7:
```suggestion
const PAYLOAD_SCHEMA = {
```

apps/condo/domains/common/utils/session.spec.js
```diff
@@ -0,0 +1,52 @@
+
+const { faker } = require('@faker-js/faker')
+
+const { makeSessionData } = require('@condo/domains/common/utils/session')
+
+describe('session', () => {
+    
+    describe('makeSessionData', () => {
+        
+        describe('organizations', () => {
```
SavelevMatthew, line 10: describe.each(field1, field2)

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,289 @@
...
+                        ...resolvedData,
+                    }
+                    if (operation === 'create' && !newItem[fieldPath]) {
+                        const contextId = get(resolvedData, 'context')
+                        const context = await getByCondition('B2BAppContext', {
+                            id: contextId,
+                            app: { deletedAt: null },
+                            organization: { deletedAt: null },
+                            status: CONTEXT_FINISHED_STATUS,
+                            deletedAt: null })
```
SavelevMatthew, line 168: line-break

apps/condo/domains/miniapp/schema/B2BAccessToken.js
```diff
@@ -0,0 +1,289 @@
...
+                    throw new GQLError(ERRORS.CONTEXT_NOT_EXISTS, context)
+                }
+                const existingUser = await getByCondition('User', { id: updatedItem.user, deletedAt: null })
+                if (!existingUser) {
+                    throw new GQLError(ERRORS.SERVICE_USER_DOES_NOT_MATCH_CONTEXT, context)
+                }
+                const permissions = Object.keys(getEnabledPermissions(rightSet))
+                const additionalFields = makeSessionData({
+                    allowedOrganizations: [b2bAppContext.organization],
+                    enabledB2BPermissions: permissions,
```
SavelevMatthew, line 265:
```suggestion
                    enabledB2BAppPermissions: permissions,
```

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -35,15 +77,153 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
...
     },
+    hooks: {
+        async validateInput ({ resolvedData, existingItem, context }) {
+            const nextItem = {
+                ...existingItem,
+                ...resolvedData,
+            }
+            if (nextItem.type !== ACCESS_RIGHT_SET_GLOBAL_TYPE) {
+                const rightSetsForApp = await find('B2BAppAccessRightSet', {
+                    app: { id: nextItem.app },
```
SavelevMatthew, line 127: deletedAt: null

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.js
```diff
@@ -35,15 +77,153 @@ const B2BAppAccessRightSet = new GQLListSchema('B2BAppAccessRightSet', {
...
+            }
+
+            if (operation === 'update' && updatedItem.type === ACCESS_RIGHT_SET_SCOPED_TYPE && !updatedItem.deletedAt) {
+                const accessTokens = await find('B2BAccessToken', {
+                    rightSet: { id: updatedItem.id },
+                    deletedAt: null,
+                })
+                
+                const updateInputs = accessTokens.map(token => ({ id: token.id, data: { dv: 1, sender: { dv: 1, fingerprint: updatedItem.sender.fingerprint } } }))
+                // Update enabled permissions in sessions
```
SavelevMatthew, line 214:
```suggestion
                // NOTE: Update trigger is required to update sessions in B2BAccessToken afterChange hook
```

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -350,14 +353,120 @@ describe('B2BAppAccessRightSet', () => {
...
+            await expectToThrowGQLError(async () => {
+                await createTestB2BAppAccessRightSet(support, app, { type: 'SCOPED' })
+            }, {
+                code: 'BAD_USER_INPUT',
+                type: 'ACCESS_RIGHT_SET_TOO_MANY_OF_TYPE',
+                message: `Too many items of type "SCOPED". Maximum is "${ACCESS_RIGHT_SET_MAX_ITEMS_SCOPED_TYPE}"`,
+            }, 'obj')
+        })
+
+        test('Cannot create or update set with type "SCOPED" if type "GLOBAL" doesn\'t exists', async () => {
```
SavelevMatthew, line 381: deletedAt case?

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -350,14 +353,120 @@ describe('B2BAppAccessRightSet', () => {
...
+        test('Deleting "GLOBAL" B2BAppAccessRightSet is leading for related right sets of other types deletion', async () => {
+            const [app] = await createTestB2BApp(support)
+            const [globalRightsSet] = await createTestB2BAppAccessRightSet(support, app, { canReadOrganizations: true })
+
+            for (let i = 0; i < Math.round(Math.random() * ACCESS_RIGHT_SET_MAX_ITEMS_SCOPED_TYPE) + 1; i++) {
+                const [scopedRightSet] = await createTestB2BAppAccessRightSet(support, app, { type: 'SCOPED', canReadOrganizations: true })
+                expect(scopedRightSet.canReadOrganizations).toEqual(true)
+            }
+            await updateTestB2BAppAccessRightSet(support, globalRightsSet.id, { deletedAt: new Date().toISOString() })
+            const deletedScopedRightSets = await B2BAppAccessRightSet.getAll(support, { app: { id: app.id }, type: 'SCOPED' })
```
SavelevMatthew, line 441: length 0?

SavelevMatthew: expect(deletedScopedRightSets.length).toBeGreaterThan(0)

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -1360,4 +1468,73 @@ describe('B2BApp permissions for service user', () => {
 
         })
     })
+
+    describe('B2BAccessTokens domain', () => {
```
SavelevMatthew, line 1472:
```suggestion
    describe('B2BAccessTokens hooks test', () => {
```

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
```diff
@@ -1360,4 +1468,73 @@ describe('B2BApp permissions for service user', () => {
...
+            [context] = await createTestB2BAppContext(support, app, organization, { status: 'Finished' })
+            const [right] = await createTestB2BAppAccessRight(support, serviceUser.user, app);
+            [accessRightSet] = await createTestB2BAppAccessRightSet(support, app, {
+                canReadB2BAccessTokens: true,
+                canManageB2BAccessTokens: true,
+            });
+            [scopedRightSet] = await createTestB2BAppAccessRightSet(support, app, { type: 'SCOPED' })
+            await updateTestB2BAppAccessRight(support, right.id, { accessRightSet: { connect: { id: accessRightSet.id } } })
+        })
+
```
SavelevMatthew, line 1494: describe('Service user')

apps/condo/domains/miniapp/utils/permissions.spec.js
```diff
@@ -0,0 +1,179 @@
...
+describe('Permission Helper Functions', () => {
+    describe('getPermissionsDiff', () => {
+        it('skips fields, not defined in right set config', () => {
+            const rightSet = {
+                canManageContacts: false,
+                canReadMeters: true,
+                canManageMeters: false,
+            }
+            const referenceRightSet = {
+                canReadContacts2: true,
```
SavelevMatthew, line 18: sender: { dv: 1, fingerprint: "something" }
...
sender: { dv: 1, fingerprint: "other" }

apps/condo/domains/miniapp/schema/B2BAccessToken.test.js
```diff
@@ -0,0 +1,612 @@
...
+            test('anonymous can\'t', async () => {
+                const client = await makeClient()
+
+                await expectToThrowAuthenticationErrorToObj(async () => {
+                    await createTestB2BAccessToken(client, b2bAppContext, scopedRightSet)
+                })
+            })
+        })
+
+        describe('update', () => {
```
SavelevMatthew, line 110: service user?

apps/condo/domains/miniapp/schema/B2BAccessToken.test.js
```diff
@@ -0,0 +1,612 @@
...
+                const [objCreated] = await createTestB2BAccessToken(admin, b2bAppContext, scopedRightSet)
+
+                const client = await makeClient()
+                await expectToThrowAuthenticationErrorToObj(async () => {
+                    await updateTestB2BAccessTokenReadonly(client, objCreated.id)
+                })
+            })
+        })
+
+        describe('hard delete', () => {
```
SavelevMatthew, line 148: service user

apps/condo/domains/miniapp/schema/B2BAccessToken.test.js
```diff
@@ -0,0 +1,612 @@
...
+    })
+
+    describe('Validation tests', () => {
+        test('Should have correct dv field (=== 1)', async () => {
+            const [obj] = await createTestB2BAccessToken(admin, b2bAppContext, scopedRightSet)
+            expect(obj).toHaveProperty('dv', 1)
+        })
+
+        test('Accepts only context and rightSet for same app', async () => {
+            const [otherApp1] = await createTestB2BApp(admin)
```
SavelevMatthew, line 322: support?

apps/condo/domains/miniapp/schema/B2BAccessToken.test.js
```diff
@@ -0,0 +1,612 @@
...
+                    expect(res.data.obj).toEqual(expect.objectContaining({
+                        id: serviceUser.id,
+                        type: 'service',
+                    }))
+                })
+
+            })
+            
+            test('Can access only connected organization', async () => {
+                const [originalOrganization] = await registerNewOrganization(admin)
```
SavelevMatthew, line 494: support

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
YEgorLu: This test is testing that rights restrictions from B2BAppAccessRightSet are working, so these must be here

apps/condo/domains/miniapp/schema/B2BAppAccessRightSet.test.js
YEgorLu: Tests needed to test serviceUser restrictions
And in every operation with model, which permissions restricts user, are executed by service user. Why would I need describe block with him?

apps/condo/domains/common/utils/session.js
YEgorLu: it would say true for { a: undefined }.hasOwnProperty('a'). Seems opposite of what it needs to do

SavelevMatthew (approved):

PR #5442, opened 2024-11-07
feat(condo): DOMA-9789 added accountNumber field to MeterReading model

abshnko:


apps/condo/domains/meter/schema/MeterReading.js
```diff
@@ -145,6 +150,18 @@ const MeterReading = new GQLListSchema('MeterReading', {
                 update: access.canEditBillingStatusFields,
             },
         },
+        accountNumber: {
```
AleX83Xpert, line 153: Is it possible to make this field readonly?
```js
access: { read: true, create: true, update: false }
```

apps/condo/migrations/20241107225800-0436_meterreading_accountnumber_and_more.js
```diff
@@ -0,0 +1,50 @@
...
+    await knex.raw(`
+BEGIN;
+--
+-- [CUSTOM] Set Statement Timeout to some large amount - 25 min (25 * 60 => 1500 sec)
+--
+SET statement_timeout = '1500s';
+--
+-- Add field accountNumber to meterreading
+--
+ALTER TABLE "MeterReading" ADD COLUMN "accountNumber" text NULL;
```
AleX83Xpert, line 14: Do we need an index for this field?

AleX83Xpert (approved):

apps/condo/domains/meter/schema/MeterReading.js
abshnko: yes, fixed it, thank you! it's much more concise

vovaaxeapolla (approved):

PR #5451, opened 2024-11-10
feat(condo): DOMA-10613 parse YYYYMM and MMYYYY date formats in import

YEgorLu:
These formats violated previous excel date parsing, when it gives us number like 44012, we convert it to string and later, knowing that column should contain date, parsing that as number. So i accidentally found xlsx parser option, which parses dates in date type not as number, but as JS Date by itself...


apps/condo/domains/common/utils/import/date.spec.js
```diff
@@ -12,6 +12,10 @@ const validDateStrings = [
     '01-01-2024',
     '2024-01-01 00:00:00',
     '2024-01-01 00:00',
+    '202410',
```
pahaz, line 15: Why only this format? Is this locale specific?

PR #5452, opened 2024-11-10
refactor(resident-app): DOMA-10626 convert lib/gql queries into domain ones

SavelevMatthew:


yarn.lock
```diff
@@ -2002,6 +2002,7 @@ __metadata:
     "@babel/core": 7.20.12
     "@babel/preset-env": 7.20.2
     "@babel/preset-typescript": 7.18.6
+    "@graphql-codegen/add": ^5.0.3
```
pahaz, line 2005: Please add it to someone else. It may break the open source in a future. Can you add it to keystone package too?

SavelevMatthew: where? I use it in my app, I've added it in my app.

PR #5477, opened 2024-11-12
fix(condo): no more useless react warnings

pahaz:
🥳🥳🥳

image

### DIFF
--- a/apps/condo/pages/_document.tsx
+++ b/apps/condo/pages/_document.tsx
@@ -3,6 +3,17 @@ import { extractCritical } from 'emotion-server'
 import Document, { Html, Head, Main, NextScript } from 'next/document'
 import React from 'react'
 
+
+const CriticalStyles = ({ css, ids }) => {
+    const innerHtml = { __html: css }
+    return (
+        <style
+            data-emotion-css={ids.join(' ')}
+            dangerouslySetInnerHTML={innerHtml}
+        />
+    )
+}
+
 export default class MyDocument extends Document {
     static getInitialProps ({ renderPage }) {
         const page = renderPage()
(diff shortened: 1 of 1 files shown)


toplenboren (approved):

dkoviazin (approved):

apps/condo/pages/_document.tsx
```diff
@@ -17,7 +17,13 @@ export default class MyDocument extends Document {
                 <Head>
                     <link id='favicon' rel='shortcut icon' href='/favicon.ico' type='image/x-icon'/>
                 </Head>
-                <body>
+                {/*
+                    NOTE: Extensions similar to Grammarly, ColorZilla and LanguageTool are therefore the cause of this warning,
+                    so you have to find out which one is doing this and then disable/configure it
+                    to not run on the ports you usually use for development.
+                    https://stackoverflow.com/a/75339011/941020
+                */}
```
sitozzz, line 25: Do we really need note inside client-side part?

apps/condo/domains/common/components/containers/BaseLayout/components/styles.ts
```diff
@@ -88,7 +88,10 @@ interface IPageWrapper {
     isSmall: boolean
 }
 
-export const StyledPageWrapper = styled(Layout.Content)<IPageWrapper>`
+export const StyledPageWrapper = styled(Layout.Content, {
+    // NOTE: remove `isSmall` prop on a DOM element (avoid React warning)
+    shouldForwardProp: (prop) => prop !== 'isSmall',
+})<IPageWrapper>`
```
sitozzz, line 94: Looks like a hack)

pahaz:
> Why do you suppress all warning on the entire body?

In reality, we do not have and will never have any attributes in the body. However, this warning is quite common and will always appear, negatively impacting the developer experience (DX) for external contributions. I don't see a significant issue with adding this to the `body` and `html` since we won't be passing any attributes there. This approach would also help avoid unnecessary distractions from additional notifications. I would prioritize DX in this case, as attributes are often passed in the body. I don't perceive any substantial risks of accidental errors arising from this. Currently, I've had to address more than four warnings because everyone is accustomed to their presence. Therefore, I want to enhance the local development experience so that even warnings from third-party extensions do not divert attention.

apps/condo/domains/common/components/containers/BaseLayout/components/styles.ts
pahaz: Yes, the `styled` adds attributes to the HTML element by default. And this leads to the emergence of warning

SavelevMatthew:
> > Why do you suppress all warning on the entire body?
> 
> In reality, we do not have and will never have any attributes in the body. However, this warning is quite common and will always appear, negatively impacting the developer experience (DX) for external contributions and local development.
> 
> I don't see a significant issue with adding this to the `body` and `html` since we won't be passing any attributes there. This approach would also help avoid unnecessary distractions from additional warnings. I would prioritize DX in this case, as attributes are often passed in the body. I don't perceive any substantial risks of accidental errors arising from this. Currently, I've had to address more than four warnings because everyone is accustomed to their presence. Therefore, I want to enhance the local development experience so that even warnings from third-party extensions do not divert attention.

Does this only affects body? Because I'm worried about its behaviour, if hydration error occurred in children, will it be thrown?

SavelevMatthew:
According to react docs, it will suppress on children's too... 🤔 
image

SavelevMatthew (approved):

sitozzz (approved):

PR #5490, opened 2024-11-14
fix(condo): DOMA-10686 fix export drops

Alllex202:
Processing large templates may cause a "Maximum call stack size exceeded" error.
In this pr unnecessary empty lines were removed from some templates (a large number of unnecessary lines)


sitozzz (approved):
imo looks good

abshnko (approved):

toplenboren (approved):

PR #5491, opened 2024-11-15
fix(condorb): DOMA-9618 delete payments if can not create payment in sberSecurePay

YEgorLu:


apps/condorb
sitozzz, line 1: is this still relevant?

PR #5492, opened 2024-11-15
hotfix(global): adjust postgres config for local development

sitozzz:


SavelevMatthew (approved):

SavelevMatthew (commented):
Effective cache size x2 = +3Gb allocated memory?

sitozzz:
> Effective cache size x2 = +3Gb allocated memory?

That's only an upper limit, not a hard one

PR #5499, opened 2024-11-16
revert(eps): handle invalid failUrl

YEgorLu:


SavelevMatthew (changes requested):
In error branch you're redirecting to undefined, so it's basically broken logic here.

If url is not passed = you cannot finish logic handling.

So instead you should validate url at the top of handler, and in case of undefined = return response with 400 code (bad request)

By the way, the first thing you should do is figuring out why undefined is even occurred. Maybe its client-side error

abshnko (approved):

dkoviazin (approved):

PR #5500, opened 2024-11-16
fix(DOMA-10698): fixed some cases of unitName parsing

dkoviazin:
1. When unitName contains some keywords for unitType and they are not separated with spaces - this keyword will not be cleared
For example 
unitName "НП10" - before was cleared to warehouse 10, now it will be warehouse НП10
unitName "НП 10" - will still be: warehouse 10
2. Fixed bug when
unitName "клад. 10" - was converted to "ад. 10"


packages/clients/address-service-client/utils/parseAddressesFromString.js
```diff
@@ -2,11 +2,11 @@ const KEYWORDS = {
     parking: ['автоместо', 'парковка', 'паркинг', 'машиноместо', 'гараж', 'м/м', 'мм', 'место', 'м/место', 'а/м', 'бокс', 'парк'],
     apartment: ['аппарт', 'апарт', 'ап', 'к/п'],
     commercial: ['офис', 'оф'],
-    warehouse: ['помещение', 'подвал', 'помещ', 'пом', 'кл', 'кладовка', 'кладовая', 'нп'],
+    warehouse: ['помещение', 'подвал', 'помещ', 'пом', 'кл', 'кладовка', 'кладовая', 'нп', 'клад'],
     flat: ['квартира', 'кв', 'комн'],
 }
 
-const HOUSE_IDENTIFIERS = 'д|уч|дом|участок|двлд'
+const HOUSE_IDENTIFIERS = 'д|уч|дом|участок|двлд|домовладение'
```
sitozzz, line 9: Don't we think to make that configurable through environment?

PR #5507, opened 2024-11-19
feat(global): INFRA-650 restrict public access to User.name

SavelevMatthew:
## TODO List (in this PR)

- [x] Add necessary logic 
- [x] Make sure it works in User.test.js
- [x] Make sure it works in custom mutations
- [x] Write field-level condo tests
- [x] Pass existing CI

## TODO List (outside this PR)
- [ ] Clone logic outside of condo to restrict access in miniapps too

### DIFF
--- a/apps/condo/domains/user/schema/User.test.js
+++ b/apps/condo/domains/user/schema/User.test.js
@@ -26,6 +26,7 @@ const {
 } = require('@open-condo/keystone/test.utils')
 
 const { normalizeEmail } = require('@condo/domains/common/utils/mail')
+const { createTestB2BApp, B2BApp } = require('@condo/domains/miniapp/utils/testSchema')
 const {
     MIN_PASSWORD_LENGTH,
     MAX_PASSWORD_LENGTH,
@@ -50,7 +51,10 @@ const {
     createTestPhone,
     createTestEmail,
     makeClientWithResidentUser,
+    makeClientWithSupportUser,
+    registerNewUser,
 } = require('@condo/domains/user/utils/testSchema')
+
 
 describe('SIGNIN', () => {
     test('anonymous: SIGNIN_MUTATION', async () => {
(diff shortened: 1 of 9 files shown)


packages/keystone/KSv5v6/utils/resolvers.js
```diff
@@ -1,11 +1,6 @@
 const get = require('lodash/get')
 
-const { _internalGetAsyncLocalStorage } = require('@open-condo/keystone/executionContext')
-
-// NOTE(SavelevMatthew): separate context is required because I don't want to mix its logic executionContext one:
-// 1. it has different entry points (it is hard to manage the merge process + dependency on executionContext's shape appears)
-// 2. We may want to add call-stack later, merging states will become hard
-const graphqlCtx = _internalGetAsyncLocalStorage('graphqlCtx')
+const { graphqlCtx } = require('./graphqlCtx')
```
sitozzz, line 3: Why that util placed inside keystone compatibility module?

apps/condo/domains/user/schema/User.test.js
```diff
@@ -455,6 +459,89 @@ describe('User fields', () => {
...
+                    expect.objectContaining({ id: directAccessedClient.user.id, name: null }),
+                    expect.objectContaining({ id: defaultClient.user.id, name: defaultClient.user.name }),
+                ]))
+
+                const nameAccessDeniedErrors = errors.filter(error =>
+                    error && error.name &&
+                    error.path && error.name === 'AccessDeniedError' &&
+                    error.path.length === 3 &&
+                    error.path[0] === 'objs' && typeof error.path[1] === 'number' && error.path[2] === 'name'
+                )
```
sitozzz, line 516: That's really hard to understand)

packages/keystone/KSv5v6/utils/resolvers.js
SavelevMatthew: I was trying to keep it close to Keystone itself, since its core logic is inside Keystone resolvers

apps/condo/domains/user/schema/User.test.js
sitozzz: You're trying to collect name related error from all of the sources, right?

SavelevMatthew: Yep, not the best try-catch here, but we don't have utils for that :(

I need to check partial data (only self-name returned) and partial errors (only with path ['objs', idx, 'name']) since there's extra errors about right set

sitozzz (approved):
Seems okay

apps/condo/domains/user/schema/User.test.js
sitozzz: Yup, thats true)

packages/keystone/access.js
```diff
@@ -140,6 +142,20 @@ function isFilteringBy (where, fields) {
     return false
 }
 
+function isSpecificQuery (queryName) {
```
pahaz, line 145: This name could be misleading. Is it possible to choose something more intuitive?

PR #5523, opened 2024-11-21
fix(condo): Duplicated payments temporary fix

dkoviazin:


apps/condo/domains/billing/schema/AllResidentBillingReceiptsService.test.js
```diff
@@ -352,7 +411,8 @@ describe('AllResidentBillingReceiptsService', () => {
             expect(receiptAfterPayment.isPayable).toBeTruthy()
         })
 
-        test('paid field calculated when organization bank info changed', async () => {
+        // TODO(dkovyazin): Add additional field on balance change to calculate new payments and display on mobile app
+        test.skip('paid field calculated when organization bank info changed', async () => {
```
pahaz, line 415: !!

dkoviazin: because of this case - we changed the logic and reverted this commit

PR #5524, opened 2024-11-21
feat(condo): DOMA-10613 made parsing for similar date formats and support for YYYYMM MMYYYY formats

YEgorLu:
There is formats like YYYYMM and MMYYYY, which other systems gives for period.
In such formats we can surely know, which is year and which is month, because we live in 2024.
But dayjs by default accepts years like 0220 etc. For example '092009' is september of 2009, but dayjs thinks that is august of 920. 
In imported files surely will not contain such old dates, so in that case we can try other formats

### DIFF
--- a/apps/condo/domains/common/constants/import.js
+++ b/apps/condo/domains/common/constants/import.js
@@ -1,3 +1,7 @@
+const dayjs = require('dayjs')
+
+const VALID_YEAR_OFFSET_FROM_TODAY = 30
+
 const PROCESSING = 'processing'
 const COMPLETED = 'completed'
 const ERROR = 'error'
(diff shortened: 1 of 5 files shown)


apps/condo/domains/common/constants/import.js
```diff
@@ -35,17 +39,34 @@ const DATE_TIME_FORMATS = DATE_FORMATS
...
     'YYYY-MM-DDTHH:mm:ss.SSS',
     'YYYY-MM-DDTHH:mm:ssZZ',
     'YYYY-MM-DDTHH:mm:ssZ',
     'YYYY-MM-DDTHH:mm:ss',
-].sort((a, b) => b.length - a.length) // Order matters! see "Differences to moment" https://day.js.org/docs/en/parse/string-format
+].sort((a, b) => {
+    if (typeof a === 'string' && typeof b === 'string') {
+        return b.length - a.length
+    }
+    if (typeof a === 'string' || typeof b === 'string') {
```
SavelevMatthew, line 65: What's this?

apps/condo/domains/common/utils/import/date.js
```diff
@@ -107,16 +111,53 @@ function isDateStrValid (dateStr, formats = DEFAULT_DATE_PARSING_FORMATS) {
...
 }
 
+/**
+ * @param {(dateString: string, formats: string | string[], strict: boolean) => dayjs.Dayjs} dayjsFunc
+ * @param dateString
+ * @param {({format: string, validate: (date: dayjs.Dayjs) => boolean} | string)[]} formats
+ * @param strict
+ * @returns dayjs.Dayjs
+ * */
+function tryParseDateWithRestrictions (dayjsFunc, dateString, formats, strict) {
```
SavelevMatthew, line 127: Its usually a bad practice for functions to have more than 3 args, consider grouping something into options...

apps/condo/domains/common/constants/import.js
SavelevMatthew: Why not to compare a or a.format?

SavelevMatthew (changes requested):
What the `202021` date would look like?

apps/condo/domains/common/constants/import.js
YEgorLu: Need to sort formats from longest to shortest.
Formats are: strings, objects with "format" property

YEgorLu:
> What the `202021` date would look like?

Right now it will be invalid date. (isDateStringValid would say that it is invalid, toIsoString does not care and will try to parse to something)

PR #5530, opened 2024-11-22
chore(condo): DOMA-10737 improve lint-i18n-translations.js

pahaz:


bin/wait-apps-apis.js
```diff
@@ -3,55 +3,59 @@ const fetch = require('node-fetch')
 
 const { getAppServerUrl } = require('@open-condo/cli')
 
-program.option('-f, --filter <names...>', 'Filters apps by name')
-program.option('-s, --sleep <number>',  'Sleep interval between polling in ms', /\d{1,12}/, '3000')
-program.option('-t, --timeout <number>',  'Max polling time in ms', /\d{1,12}/, '120000')
```
pahaz, line 8: RegExp is deprecated

apps/condo/lang/ru/messages/RESET_PASSWORD/default.njk
Alllex202, line 1: It looks like we don't use the "RESET_PASSWORD" message type, so you should remove the translations for that type

pahaz: Yes, I want to add this check to the linter

bin/lint-i18n-translations.js
```diff
@@ -40,11 +54,146 @@ async function loadTranslations (langDir) {
...
+                    if (entry.isDirectory()) {
+                        try {
+                            const files = await fs.readdir(messageDirPath)
+                            const njkFiles = {}
+                            for (const file of files) {
+                                // nosemgrep: javascript.lang.security.audit.path-traversal.path-join-resolve-traversal.path-join-resolve-traversal
+                                const filePath = path.join(messageDirPath, file)
+                                if (file.endsWith('.njk')) {
+                                    try {
+                                        await fs.access(filePath)
```
SavelevMatthew, line 109: How about to nest code even more and wrap every single line in try-catch? Because right now the code reads too easily :)

SavelevMatthew: I guess we need to refactor this part heavily, right now its heavy 400-liner, no one can read and understand.
I hate this kind of files, because you cannot even tell somebody to fix something in it.

It's like `@open-condo/next`, very heavy and hard to understand files with unstructured code...

SavelevMatthew: You can partially fix it with naming (for example this utils scans for push / email templates, you can call it `loadNotificationTemplates` or something, remove nested try catches here, panic is enough for bin

PR #5536, opened 2024-11-25
fix(eps): DOMA-10688 added logger for static methods

vovaaxeapolla:


Alllex202 (commented):
Why isn't the existing logger suitable?

vovaaxeapolla:
> Why isn't the existing logger suitable?

Static method has no this value

Alllex202 (approved):

dkoviazin (approved):

PR #5549, opened 2024-11-27
fix(condo): DOMA-10568 Add sendAt date and check on success for updating date

Alexander-Turkin:
Add sendAt date and check on success for updating date

image


toplenboren (approved):

apps/condo/pages/news/[id]/index.tsx
```diff
@@ -256,9 +255,12 @@ const NewsItemCard: React.FC = () => {
         } else if (newsItemSharing.status === 'error') {
             status = ErrorMessage
         } else if (newsItemSharing.status === 'success') {
+            dateToShow = get(newsItemSharing, 'updatedAt', null)
```
Comment (nomerdvadcatpyat) [inline apps/condo/pages/news/[id]/index.tsx:258]: Is it possible to update newsItemSharing after submission? 
If so, then there may be an incorrect `updatedAt`

Comment (nomerdvadcatpyat) [reply apps/condo/pages/news/[id]/index.tsx:258]: Maybe we should prohibit updating NewsItemSharing with the `success` status via api, so that such situations definitely don't happen?

Comment (toplenboren) [reply apps/condo/pages/news/[id]/index.tsx:258]: OK for me, but we should leave the ability to update it to support users

apps/condo/domains/news/schema/NewsItemSharing.js
```diff
@@ -116,7 +123,11 @@ const NewsItemSharing = new GQLListSchema('NewsItemSharing', {
             if (operation === 'update') {
                 const oldStatus = existingItem.status
                 const newStatus = resolvedData.status
-                if (newStatus && !ALLOWED_TRANSITIONS[oldStatus].includes(newStatus)) {
+
+                if (!ALLOWED_TRANSITIONS[oldStatus]?.includes(newStatus)) {
+                    if (oldStatus === STATUSES.PUBLISHED) {
+                        throw new GQLError(ERRORS.CANT_CHANGE_PUBLISHED_NEWS, context)
+                    }
```
nomerdvadcatpyat, line 130: You need to move it outside of the `if` in line 127

apps/condo/domains/news/schema/NewsItemSharing.js
```diff
@@ -116,7 +123,11 @@ const NewsItemSharing = new GQLListSchema('NewsItemSharing', {
             if (operation === 'update') {
                 const oldStatus = existingItem.status
                 const newStatus = resolvedData.status
-                if (newStatus && !ALLOWED_TRANSITIONS[oldStatus].includes(newStatus)) {
+
+                if (!ALLOWED_TRANSITIONS[oldStatus]?.includes(newStatus)) {
```
nomerdvadcatpyat, line 127: Why you remove `newStatus &&` condition from `if`?
This block checks that `status` field is updated (`status` field existed in `resolvedData`) and it's forbidden status transition.

apps/condo/domains/news/schema/NewsItemSharing.test.js
```diff
@@ -325,7 +325,23 @@ describe('NewsItemSharing', () => {
             })
         })
 
-        describe('should not allow to make wrong status transitions', () => {
+        describe('should not allow to make wrong transitions', () => {
+            const extraAttrForCases = [
+                { statusMessage: 'test' },
```
nomerdvadcatpyat, line 330: You can use faker to generate random data. For example:

```
const { faker } = require('@faker-js/faker')

...

const statusMessage = faker.random.alphaNumeric(5)
```

apps/condo/domains/news/constants/errors.js
```diff
@@ -10,6 +10,7 @@ const NO_NEWS_ITEM_SCOPES = 'NO_NEWS_ITEM_SCOPES'
 
 const BAD_STATUS_TRANSITION = 'BAD_STATUS_TRANSITION'
 const NOT_NEWS_SHARING_CONTEXT = 'NOT_NEWS_SHARING_CONTEXT'
+const CANT_CHANGE_PUBLISHED_NEWS = 'CANT_CHANGE_PUBLISHED_NEWS'
```
toplenboren, line 13: CANT_CHANGE ... NEWS_ITEM_SHARINGS

nomerdvadcatpyat (approved):

PR #5562, opened 2024-12-02
chore(condo): DOMA-10785 add component with the text of the agreement + corrected ru translation

Alexander-Turkin:
Registration:

image

Entrance: 

image


toplenboren (approved):

apps/condo/domains/user/components/auth/InputPhoneForm.tsx
```diff
@@ -149,7 +143,7 @@ export const InputPhoneForm: React.FC<IInputPhoneFormProps> = ({ onFinish }) =>
                 >
                     <Row style={ROW_STYLES} gutter={[0, 28]}>
                         <ResponsiveCol span={24}>
-                            <Row>
+                            <Row gutter={[0, 28]}>
```
SavelevMatthew, line 146: consider moving static arrays outside of components, since they're reallocating on each serenader when passed inline like that

SavelevMatthew (approved):

PR #5564, opened 2024-12-03
feat(condo): DOMA-10780 add Message defaultContent field

nomerdvadcatpyat:
On the web client side, we need to display the notification text.
Since we won’t be creating a separate transport, we need to retrieve this text from somewhere.
Therefore, we decided to implement it using a virtual field `defaultContent` in Message that will compute the default notification content.


toplenboren (approved):

apps/condo/domains/notification/schema/Message.js
```diff
@@ -107,6 +110,24 @@ const Message = new GQLListSchema('Message', {
             },
         },
 
+        defaultContent: {
```
pahaz, line 113: why not `content` (maybe `body`)

pahaz (approved):

apps/condo/domains/notification/schema/Message.js
nomerdvadcatpyat: I think `defaultContent` it's better name because this field returns message content from `default` template

apps/condo/domains/notification/templates.js
```diff
@@ -105,6 +105,31 @@ function getEmailTemplate (locale, messageType) {
...
+function renderDefaultTemplate (message, locale) {
+    const env = {
+        serverUrl: SERVER_URL,
+    }
+    const { type } = message
+    const { templatePath } = getDefaultTemplate(locale, type)
+
+    if (templatePath) {
+        const messageTranslated = substituteTranslations(message, locale)
+        return unescape(nunjucks.render(templatePath, { message: messageTranslated, env }))
```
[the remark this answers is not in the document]
nomerdvadcatpyat: Copied this logic from other `render*Template`. Looks like it's not needed here, removed

PR #5572, opened 2024-12-05
fix(condo): DOMA-10792 fix getPollTicketComments query

nomerdvadcatpyat:
Query only last 100 comments (how it was before we rewrote the queries) and wrap `refetch` call in try/finally, to define `syncTicketCommentsAt` (date key in localStorage) even if the request fails

### DIFF
--- a/apps/condo/domains/ticket/hooks/usePollTicketComments.tsx
+++ b/apps/condo/domains/ticket/hooks/usePollTicketComments.tsx
@@ -34,21 +34,26 @@ export function usePollTicketComments ({
 
         const now = new Date().toISOString()
         const lastSyncAt = localStorage.getItem(LOCAL_STORAGE_SYNC_KEY)
+        let newSyncedAt
 
-        const result = await refetchSyncComments({
-            where: {
-                ...pollCommentsQuery,
-                updatedAt_gt: lastSyncAt || now,
-            },
-        })
-        const ticketComments = result?.data?.ticketComments?.filter(Boolean) || []
+        try {
+            const result = await refetchSyncComments({
+                where: {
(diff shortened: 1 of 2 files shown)


toplenboren (approved):

apps/condo/domains/ticket/hooks/usePollTicketComments.tsx
```diff
@@ -34,21 +34,26 @@ export function usePollTicketComments ({
...
-        })
-        const ticketComments = result?.data?.ticketComments?.filter(Boolean) || []
+        try {
+            const result = await refetchSyncComments({
+                where: {
+                    ...pollCommentsQuery,
+                    updatedAt_gt: lastSyncAt || now,
+                },
+            })
+            const ticketComments = result?.data?.ticketComments?.filter(Boolean) || []
```
toplenboren, line 46: thought we used get in these cases

toplenboren (approved):

apps/condo/domains/ticket/hooks/usePollTicketComments.tsx
sitozzz: I agree with this. 
it doesn't look quite good, since below in another block a code structure is used using the get function

sitozzz (approved):

apps/condo/domains/ticket/hooks/usePollTicketComments.tsx
nomerdvadcatpyat: lodash get erases types of return values, i think it's better to use ts where it's possible.
Rewrote block below to `?.` for more consistent

PR #5592, opened 2024-12-09
chore(condo): INFRA-674 codegen speed boost 🚀

Alllex202:
Executing `eslint --fix` takes a very loooooong time.
Removed it from codegen


.eslintrc.js
```diff
@@ -35,6 +35,7 @@ module.exports = {
     ignorePatterns: [
         'dist',
         '*schema.ts',
+        ...['condo', 'callcenter'].map(app => `apps/${app}/gql/*.ts`),
```
sitozzz, line 38: Why only 2 of apps, not all of them?

Alllex202: You are right, it is better to add it everywhere at once

Alllex202:
> Is that an expected behavior? I can't find any changes at the linter side but there are a lot of trailing commas everywhere

linter was turn off for codegen because run very long time

sitozzz (approved):
Okay, but you should resolve conflicts first)

nomerdvadcatpyat (approved):

PR #5603, opened 2024-12-11
feat(condo): DOMA-10801 ResetUserLimitAction

nomerdvadcatpyat:
We need a tool that will allow QA to reset SMS limit counters by himself


toplenboren (approved):

toplenboren:
Best to also check that IP address is not reserved one (so that I cannot use 127.0.0.1 as arg)

AleX83Xpert (approved):

apps/condo/domains/user/access/_internalResetSMSDayLimitCountersService.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Generated by `createservice user._internalResetSMSDayLimitCountersService`
+ */
+const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
+
+async function can_internalResetSMSDayLimitCounters ({ authentication: { item: user } }) {
+    if (!user) return throwAuthenticationError()
+    if (user.deletedAt) return false
+
+    if (user.isAdmin || user.isSupport) return true
```
SavelevMatthew, line 10: There's a lot of people in support right now, maybe we should add directAccessRight here instead?

apps/condo/domains/user/utils/serverSchema/guards.js
```diff
@@ -77,6 +77,21 @@ class RedisGuard {
...
+        const key = `${this.counterPrefix}${variable}`
+        return await this.redis.get(key)
+    }
+
+    async checkCounterExistence (variable) {
+        const key = `${this.counterPrefix}${variable}`
+        return await this.redis.exists(key)
+    }
+
+    async resetCounterToOne (variable) {
```
SavelevMatthew, line 90: I suggest to delete it, why 1?

SavelevMatthew (commented):
I don't really like using the default prefix, `guard_lock:` looks too generic. Maybe we should change it to something like `guard_lock:auth: `?

And resetting to 1 is strange, seems like you're copying someones local wrong-implemented script :)

apps/condo/domains/user/access/_internalResetSMSDayLimitCountersService.js
AleX83Xpert: Who will be the manager of the new UserRightSet model (connect it to the User models and disconnect sometimes)?

SavelevMatthew: The head of support I guess

apps/condo/domains/user/utils/clientSchema/ResetUserLimitAction.ts
SavelevMatthew, line 1: Remove this file :)

Can you also remove it from template as well?

apps/condo/domains/user/utils/clientSchema/index.ts
```diff
@@ -8,4 +8,5 @@ export * as ConfirmPhoneAction from './ConfirmPhoneAction'
 export * as OidcClient from './OidcClient'
 export * as ExternalTokenAccessRight from './ExternalTokenAccessRight'
 export * as UserRightsSet from './UserRightsSet'
+export * as ResetUserLimitAction from './ResetUserLimitAction'
```
SavelevMatthew, line 11: remove

apps/condo/domains/user/utils/directAccess/config.js
```diff
@@ -59,6 +59,7 @@ const DIRECT_ACCESS_AVAILABLE_SCHEMAS = {
         'OidcClient',
         { schemaName: 'User', readonly: true },
         'UserRightsSet',
+        'ResetUserLimitAction',
```
SavelevMatthew, line 62: Keep alphabetical sort inside domain please

SavelevMatthew (approved):

PR #5610, opened 2024-12-11
fix(condo): add missing 'es' locale to statusTranslations

paulo-rossy:


apps/condo/domains/ticket/constants/statusTransitions.js
```diff
@@ -24,6 +24,7 @@ const DEFAULT_STATUS_TRANSITIONS = {
 const COUNTRY_RELATED_STATUS_TRANSITIONS = {
     [RU_LOCALE]: DEFAULT_STATUS_TRANSITIONS,
     [EN_LOCALE]: DEFAULT_STATUS_TRANSITIONS,
+    [ES_LOCALE]: DEFAULT_STATUS_TRANSITIONS,
```
SavelevMatthew, line 27: How about creating it more dynamically?

Like so:
```
const { LOCALES } = require('@condo/domains/common/constants/locale')
// ...
const COUNTRY_RELATED_STATUS_TRANSITIONS = Object.from(Object.keys(LOCALES).map(locale => [locale, DEFAULT_STATUS_TRANSITIONS]))
```

SavelevMatthew (approved):

toplenboren (approved):

PR #5615, opened 2024-12-12
feat(registry): DOMA-10823 add task for receipt import

YEgorLu:


dkoviazin (commented):
1. getAll will return only 100 receipts
2. Better to use single task to sync every service as one task will bee to slow
3. If you want to sync every service - you need to add bankAccount and category to where input
4. createOrUpdateSppAccrual - better to check if you need an update
5. BILLING_RECEIPT_CHUNK_SIZE will fail if you request relations

vovaaxeapolla (approved):

abshnko (approved):

PR #5622, opened 2024-12-16
fix(condo): DOMA-10516 add test and update callcenter

tolmachev21:
We want to make sure that push notifications do not come to the technician's mobile app when mass reassigning requests. However, the logic requires a custom query name, so the test in this PR checks this logic.

And here are added the current changes for the callcenter, so that the feature works, and the project is assembled

### DIFF
--- a/apps/condo/domains/ticket/schema/Ticket.js
+++ b/apps/condo/domains/ticket/schema/Ticket.js
@@ -44,7 +44,6 @@ const {
     OMIT_TICKET_CHANGE_TRACKABLE_FIELDS,
     REVIEW_VALUES,
     DEFERRED_STATUS_TYPE,
-    DISABLE_PUSH_NOTIFICATION_FOR_OPERATIONS,
 } = require('@condo/domains/ticket/constants')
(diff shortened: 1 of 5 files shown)


toplenboren (approved):

nomerdvadcatpyat (approved):

apps/condo/domains/ticket/schema/Ticket.js
```diff
@@ -965,7 +964,7 @@ const Ticket = new GQLListSchema('Ticket', {
             )(...args)
 
             /* NOTE: this sends different kinds of notifications on ticket create/update except bulk update operation */
-            if (!DISABLE_PUSH_NOTIFICATION_FOR_OPERATIONS.includes(get(context, ['req', 'body', 'operationName'], null))) {
+            if (!Array.isArray(get(context, ['req', 'body', 'variables', 'data'], null))) {
```
Alllex202, line 967:
```suggestion
            const isBulkOperation = Array.isArray(get(context, ['req', 'body', 'variables', 'data'], null))
            if (!isBulkOperation) {
```

Alllex202 (approved):
It would be great to add a couple of lines to the schema documentation about these changes, so that third-party developers are also aware of this nuance.

PR #5632, opened 2024-12-18
fix(condo): DOMA-10516 fix redirect user

tolmachev21:
Screenshot 2024-12-18 at 15 30 29
After deleting an employee, we remain on this page

We want to redirect the user to the page '/employee/'


apps/condo/pages/employee/[id]/index.tsx
```diff
@@ -339,7 +339,9 @@ export const EmployeeInfoPage: PageComponentType = () => {
     const employeeWithSpecializations = { ...employee, specializations: organizationEmployeeSpecializations.map(scope => scope.specialization) }
 
     const updateEmployeeAction = OrganizationEmployee.useUpdate({}, () => refetch())
-    const softDeleteAction = OrganizationEmployee.useSoftDelete(() => refetch())
+    const softDeleteAction = OrganizationEmployee.useSoftDelete(() => {
+        if (window.location.pathname === `/employee/${employeeId}`) push('/employee/')
```
Comment (toplenboren) [inline apps/condo/pages/employee/[id]/index.tsx:343]: You should double check that window is available, in server side rendering context you do not have window

Best to double check it, so 

if window && window.location ...

toplenboren (approved):

nomerdvadcatpyat (approved):

Alllex202 (approved):

PR #5644, opened 2024-12-19
fix(condo): DOMA-10241 fixes on qa review

toplenboren:


apps/condo/domains/marketplace/utils/clientSchema/Invoice.ts
```diff
@@ -259,6 +258,9 @@ export function processRowsFromInvoiceTicketForm (rawRows, intl) {
         }
 
         const resultRow = { ...row, toPay: resultToPay }
+        if (!row.measure) {
+            resultRow.measure = undefined
+        }
```
AleX83Xpert, line 263: You will set `resultRow.measure` to `undefined` even if `row.measure` is 0, false, null, empty string. Is it OK?

apps/condo/domains/marketplace/components/Invoice/BaseInvoiceForm.tsx
```diff
@@ -502,7 +502,7 @@ const ServicesList = ({ organizationId, propertyId, form, currencySymbol, disabl
             const priceObj = get(pricesArray, '0')
             const price = get(priceObj, 'price')
             const isMin = get(priceObj, 'isMin')
-            const measure = get(priceObj, 'measure')
+            const measure = get(priceObj, 'measure', undefined)
```
AleX83Xpert, line 505: Just FYI
if `priceObj.measure` is `null` `get` will return `null`
`undefined` is default value for 3rd parameter, so nothing changed here
Did you meant:
```suggestion
            const measure = get(priceObj, 'measure') || undefined
```
?

toplenboren: Thanks

AleX83Xpert (approved):

apps/condo/domains/marketplace/utils/clientSchema/Invoice.ts
AleX83Xpert: Anyway, strange 3-line logic :)

PR #5651, opened 2024-12-20
feat(condo): INFRA-764 reset user rate-limit

SavelevMatthew:
## TODO:

- [x] Generate schemaDocs from resetters
- [x] Write good tests to test rate-limit (Ok, good tests are not possible, since we'll probably disable rate-limit in CI to avoid unauthed errors, so I'll just set variable and reset it like @nomerdvadcatpyat did with auth)


apps/condo/domains/user/utils/limits.js
```diff
@@ -0,0 +1,70 @@
...
+        const identifierType = getIdentifierType(identifier)
+        if (!this.supportedIdentifiers.includes(identifierType)) {
+            return false
+        }
+
+        return await this.guard.deleteCounter(`${AUTH_COUNTER_LIMIT_TYPE}:${identifier}`)
+    }
+}
+
+class RateLimitResetter {
```
nomerdvadcatpyat, line 35: I think it's better to move Resetters to separate files

apps/condo/domains/user/utils/limits.js
```diff
@@ -0,0 +1,70 @@
...
+
+    constructor () {
+        this.guard = new RedisGuard()
+    }
+
+    async checkExistence (identifier) {
+        const identifierType = getIdentifierType(identifier)
+        if (!this.supportedIdentifiers.includes(identifierType)) {
+            return false
+        }
```
nomerdvadcatpyat, line 20: This logic can be a separate method, for example, `isIdentifierSupported`.
And reused here, in reset method and in ResetUserLimitAction.afterChange

apps/condo/domains/user/utils/identifiers.spec.js
```diff
@@ -0,0 +1,39 @@
+const { faker } = require('@faker-js/faker')
+
+const { createTestPhone } = require('@condo/domains/user/utils/testSchema')
+
+const { getIdentifierType, IPv4_TYPE, UUID_TYPE, PHONE_TYPE } = require('./identifiers')
+
+describe('User identifiers utils', () => {
+    describe('getIdentifierType', () => {
+        const SAMPLES_COUNT = 1000
```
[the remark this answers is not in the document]
SavelevMatthew: Yep, all tests are done under a second locally

apps/condo/domains/user/utils/limits.js
SavelevMatthew: Separated logic into `utils/limits/guards`

apps/condo/domains/user/utils/limits.js
SavelevMatthew: Moved this logic into `AbstractResetter`

nomerdvadcatpyat (approved):

PR #5654, opened 2024-12-21
feat(condo): DOMA-10910 a touch of winter magic ❄️✨

Alllex202:
https://github.com/user-attachments/assets/2c7376a2-780c-40d0-adc5-54a692584931


SavelevMatthew (approved):

apps/condo/next.config.js
```diff
@@ -106,6 +107,7 @@ let nextConfig = withTM(withLess(withCSS({
         termsOfUseUrl,
         privacyPolicyUrl,
         dataProcessingConsentUrl,
+        isSnowfallDisabled,
```
SavelevMatthew, line 110: Can we turn it on in runtime without restarts?

Alllex202: You can do this by setting the number of snowflakes to zero, via the feature flag (default is zero).

A flag from environment variables allows you to completely disable the snowfall logic from the application

tolmachev21 (approved):

toplenboren (approved):

nomerdvadcatpyat (approved):

PR #5659, opened 2024-12-23
feat(keystone): DOMA-10804 FileStorageText field is added in keystone

vovaaxeapolla:
### **Add FileStorageText field for Keystone**
This PR introduces a new FileStorageText field for Keystone, designed to store and retrieve large text data seamlessly. The field is optimized for storing data in external cloud storage while behaving like a standard text string field in the Keystone schema.

**Use Cases:**

- Storing extensive logs, parsed data, or large text that exceed typical database text field limits.
- Applications requiring scalable, cost-effective storage for large textual data.


toplenboren:
I do not really like the naming. Maybe lets name it FileContent ? or CloudTextField?

LargeTextField has me under an impression that data is stored in db directly

toplenboren:
CloudStorageDataField (?)

packages/keystone/fields/CloudStorageText/adapters/index.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    async _readFileFromAdapter (fileMetadata) {
+        const fileStream = await getObjectStream(fileMetadata, this.fileAdapter)
+        const fileContent = await readFileFromStream(fileStream)
+        return fileContent.toString()
+    }
+
+    _generateFilename () {
+        return `${new Date().toISOString()}`
```
YEgorLu, line 68: This can eventually generate same names for strings, when many clients are creating model at the same time?

If so, i heard we had problems before with same names of files, newest replaces olds one or something like that.

So maybe add here more randomness? Like
`${new Date().toISOString()}-${uuid()}`

packages/keystone/fields/CloudStorageText/adapters/index.js
```diff
@@ -0,0 +1,84 @@
+const { Text } = require('@keystonejs/fields')
+const cuid = require('cuid')
+const isNil = require('lodash/isNil')
```
dkoviazin, line 3: const { isNil } = require('lodash') on server

PR #5663, opened 2024-12-24
feat(global): INFRA-776 add ability to increase quotas per IP / user in rate-limit

SavelevMatthew:


sitozzz (approved):

packages/keystone/rateLimiting/config.utils.js
```diff
@@ -0,0 +1,32 @@
...
+
+const ajv = new Ajv()
+
+const PLUGIN_OPTIONS_SCHEMA = {
+    type: 'object',
+    properties: {
+        customQuotas: { type: 'object', additionalProperties: { type: 'number' } },
+        queryWeight: { type: 'number' },
+        mutationWeight: { type: 'number' },
+        window: { type: 'string' },
```
sitozzz, line 11: What is the goal of this field?

SavelevMatthew: Sliding window of quotas (TTL)

toplenboren (approved):

packages/keystone/rateLimiting/config.utils.js
Alllex202: Maybe then we should call the field `quotaWindow`?
As your const `DEFAULT_QUOTA_WINDOW`

Alllex202 (approved):

PR #5670, opened 2024-12-25
fix(condo): DOMA-10927 added skip in global providers

nomerdvadcatpyat:
Added `skip` for requests in TicketVisibilityContextProvider and GlobalAppsContainer, so that they are not executed until the user is authorized


sitozzz (approved):
Well done!

SavelevMatthew (approved):

PR #5682, opened 2025-01-10
fix(condo): DOMA-10863 use organization's tin to boost searching by address from QR

AleX83Xpert:


apps/condo/domains/billing/schema/ValidateQRCodeService.spec.js
```diff
@@ -0,0 +1,70 @@
...
+        await expectToThrowGQLError(
+            async () => {
+                await validateQRCodeByTestClient(adminClient, { qrCode: stringifyQrCode(qrObj) })
+            },
+            {
+                mutation: 'validateQRCode',
+                code: 'INTERNAL_ERROR',
+                type: 'NOT_FOUND',
+                message: 'Organization with provided TIN and having provided address is not registered',
+            },
```
sitozzz, line 62: You can use constant from schema called ```ERRORS.NO_ORGANIZATION```. In case if you need to change message or something else you don't need to change that twice

sitozzz: This is just a suggestion, not necessarily a correction

apps/condo/domains/billing/utils/receiptQRCodeUtils.js
```diff
@@ -191,12 +191,17 @@ function formatPeriodFromQRCode (period) {
  */
 async function findAuxiliaryData (qrCodeFields, errors) {
     const addressServiceClient = createInstance()
-    const normalizedAddress = await addressServiceClient.search(getQRCodeField(qrCodeFields, 'PayerAddress'), { extractUnit: true })
+    const searchParams = { extractUnit: true }
+    const tin = getQRCodeField(qrCodeFields, 'PayeeINN')
+    if (tin) {
+        searchParams.helpers = JSON.stringify({ tin })
```
YEgorLu, line 197: Will search for properties below fail to find existing property if tin is not present?

YEgorLu (approved):

apps/condo/domains/billing/schema/ValidateQRCodeService.spec.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function stringifyQrCode (qrCodeObj) {
+    return Buffer.from(Object.keys(qrCodeObj).reduce((qrStr, field) => {
+        return `${qrStr}|${field}=${qrCodeObj[field]}`
+    }, 'ST00012')).toString('base64')
+}
+
+describe('ValidateQRCodeService spec', () => {
+    setFakeClientMode(index)
```
sitozzz, line 43: R we agreed with fakeclientmode usage? I've heard discussions that this is not the best option for writing tests

apps/condo/domains/billing/schema/ValidateQRCodeService.spec.js
AleX83Xpert: Here we try to find an organization related to the property. I want to show that the pair organization-property does not exist.

apps/condo/domains/billing/utils/receiptQRCodeUtils.js
AleX83Xpert: The `PayeeINN` (tin) is the required field. It is checked before calling this function.

apps/condo/domains/billing/schema/ValidateQRCodeService.spec.js
AleX83Xpert: For spec too?

sitozzz (approved):

PR #5688, opened 2025-01-13
refactor(condo): DOMA-10610 remove moneyToWords and renderMoney

tolmachev21:
Small changes, which removed suppord currecy display for different locales. And add lib, which conversation number to words. 

 How now looks document, u can see here:
- [CompWorks](https://github.com/user-attachments/files/18397757/cm5v44g570060mlvzabps0kah-completion_works_ticket_f9f00fee-e6f4-497f-a79d-241b50034c2d_13_01_2025.docx)

- [PaidWorks1](https://github.com/user-attachments/files/18397750/cm5v44b2x005wmlvz5i47fbc4-paid_works_ticket_f9f00fee-e6f4-497f-a79d-241b50034c2d_13_01_2025.docx)

- [PaidWorks2](https://github.com/user-attachments/files/18397763/cm5v3wdjj004imlvz6slt66pf-paid_works_ticket_a0c25bd5-57bc-44d6-b41b-9b06e974962f_13_01_2025.docx)


apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/completionWorks.js
```diff
@@ -72,8 +71,16 @@ const generateTicketDocumentOfCompletionWorks = async ({ task, baseAttrs, contex
...
+                price: !Number.isNaN(parseFloat(price)) ? new Intl.NumberFormat(locale, {
+                    style: 'currency',
+                    currency: currencyCode,
+                    currencyDisplay: 'code',
+                }).format(price).replace(/[A-Z]{3}/, '').trim() : '',
+                sum: !Number.isNaN(parseFloat(price * row.count)) ? new Intl.NumberFormat(locale, {
+                    style: 'currency',
+                    currency: currencyCode,
+                    currencyDisplay: 'code',
+                }).format(price * row.count).replace(/[A-Z]{3}/, '').trim() : '',
```
SavelevMatthew, line 83: `''` is bad default value to store...

SavelevMatthew (changes requested):
I'm a bit missing this PR.

1. Is this code used on server or client?

If on client, why its in serverSchema
And if on server, why do you default values `''`. Empty string is not good value to store in DB, consider null / undefined or throw an exception.

2. Why do you need this strange import? Especially globally defined...Also seems strange to me. Please explain, what exactly is happening

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/completionWorks.js
tolmachev21: This code is executed on the server because this task. An empty string is the value that will be inserted into the document (for us, an empty string is an empty space in the document that can be filled in manually). This value is not stored in the DB

tolmachev21:
> 1. Is this code used on server or client?
> 
> If on client, why its in serverSchema And if on server, why do you default values `''`. Empty string is not good value to store in DB, consider null / undefined or throw an exception.
> 

1. Is this code used on server or client? - An empty string is a value that will be inserted into the document. It is not saved in the database.

> 2. Why do you need this strange import? Especially globally defined...Also seems strange to me. Please explain, what exactly is happening

2. Why do you need this strange import? - This import is used here because the n2words library does not support СommunJS modules. Since we are currently on 16 node version, we also do not have support for ESM modules.

The solution to this situation is the dynamic import function - import(), which is supported since 12 version node and makes it possible to use ESM modules in files with CommunJS import

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -154,23 +162,36 @@ const generateTicketDocumentOfPaidWorks = async ({ task, baseAttrs, context, loc
...
+                .map(p => p.type === 'fraction' ? p.value : '')
+                .join(''), { lang: locale, feminine: true }),
+            wholePartTotalVAT: n2words(numberFormatByLocale
+                .formatToParts(totalVAT)
+                .map(p => p.type === 'integer' ? p.value : '')
+                .join(''), { lang: locale }),
+            decimalPartTotalVAT: n2words(numberFormatByLocale
+                .formatToParts(totalVAT)
+                .map(p => p.type === 'fraction' ? p.value : '')
+                .join(''), { lang: locale, feminine: true }),
```
Alllex202, line 180: It seems that the need to use the `feminine` parameter depends on the locale.

Right now it is not very important (since our document is only in Russian), but it is worth keeping in mind

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -4,21 +4,22 @@ const { get } = require('lodash')
...
+const { Contact } = require('@condo/domains/contact/utils/serverSchema')
 const { DEFAULT_INVOICE_CURRENCY_CODE, INVOICE_STATUS_CANCELED } = require('@condo/domains/marketplace/constants')
 const { Invoice } = require('@condo/domains/marketplace/utils/serverSchema')
 const { DEFAULT_ORGANIZATION_TIMEZONE } = require('@condo/domains/organization/constants/common')
+const { OrganizationEmployee } = require('@condo/domains/organization/utils/serverSchema')
 const { TICKET_DOCUMENT_GENERATION_TASK_FORMAT } = require('@condo/domains/ticket/constants/ticketDocument')
 
 const logger = getLogger('generateDocumentOfPaidWorksCompletion')
 
+const n2wordsImport = import('n2words').then(module => module.default)
```
toplenboren, line 21: add a comment describing why this syntax is used

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
```diff
@@ -154,23 +162,36 @@ const generateTicketDocumentOfPaidWorks = async ({ task, baseAttrs, context, loc
...
-            totalVAT: moneyToWords(totalVAT.toFixed(2), { locale, currencyCode }),
+            wholePartTotalSum: n2words(numberFormatByLocale
+                .formatToParts(totalSum)
+                .map(p => p.type === 'integer' ? p.value : '')
+                .join(''), { lang: locale }),
+            decimalPartTotalSum:  n2words(numberFormatByLocale
+                .formatToParts(totalSum)
+                .map(p => p.type === 'fraction' ? p.value : '')
+                .join(''), { lang: locale, feminine: true }),
+            wholePartTotalVAT: n2words(numberFormatByLocale
```
toplenboren, line 173: naming-wise its best to specify entity first and then specify the type

TotalWholePart
TotalDecimalPart

VATWholePart
VATDecimalPart

apps/condo/domains/ticket/utils/serverSchema/TicketDocumentGeneration/paidWorks.js
toplenboren: add a if ru -> use feminine

Alllex202 (approved):

SavelevMatthew (approved):

PR #5689, opened 2025-01-14
ci: INFRA-783 Build open-source condo on multiple OS runners 🏃

SavelevMatthew:


packages/ui/package.json
```diff
@@ -29,8 +29,8 @@
   ],
   "scripts": {
     "build:deps": "turbo build --filter=@open-condo/ui^...",
-    "build:styles": "find src/tokens -mindepth 1 -not -name 'tokens.json' -delete && node bin/prepare-tokens.js && node bin/buildStyleVars/runner.js --web less css",
```
sitozzz, line 32: Find not working at the windows environment, right?)

packages/ui/bin/buildStyleVars/runner.js
```diff
@@ -1,3 +1,4 @@
+const fs = require('fs')
```
sitozzz, line 1: I don't think it's used
```suggestion
const fs = require('fs')
```

packages/ui/package.json
SavelevMatthew: They have different args and flags there, yep

.github/workflows/nodejs.condo.ci.yml
```diff
@@ -206,6 +212,68 @@ jobs:
...
+          ref: ${{ env.REF }}
+      - name: Setup native Postgres (Windows Runner only)
+        if: matrix.os == 'windows-latest'
+        uses: ikalnytskyi/action-setup-postgres@v7
+        with:
+          postgres-version: 16
+      - name: Setup native Redis (Windows Runner only)
+        if: matrix.os == 'windows-latest'
+        run: |
+          choco install redis --version 6.2.17 -y
```
sitozzz, line 243: Maybe add matrix values for win based libs?

sitozzz (approved):
In general looks awesome

packages/ui/package.json
sitozzz: ok then that's acceptable

PR #5700, opened 2025-01-16
fix(condo): fix cannot read property 'address' of undefined in success toast

crowncode-development:
When creating Address Injections manually and then adding a house to this address, an error occurs: `Cannot read properties of undefined (reading 'address') ` when displaying success toast. To solve it, you can output the address property of the property object, otherwise address from the props.


SavelevMatthew (approved):
🏅

pahaz (approved):

PR #5712, opened 2025-01-23
fix(condo): DOMA-11007 clear ROOT_QUERY when page is update

tolmachev21:
Bug:
If you keep the ticket page (which contains comments) open for one minute and then start editing it (even without changing it), then after saving the data the comments disappear.

Before update:
Screenshot 2025-01-23 at 13 41 14

After update without fix:

Screenshot 2025-01-23 at 13 44 27

Solution:
It's a cache issue. Apollo cache loses the TicketComment:xyz cache after a minute of going to the ticket editing page, but the link to the cache in KShChChE_YGUKN remains. And after returning to the ticket/[id] page, the data is taken from the cache (i.e. the request is not made, since there is a link to this cache), but there is nothing at this link. Therefore, if you add fetchPolicy: 'cache-and-network', the data will first be taken from the cache, and then the current data will be loaded in the background and the comments will appear. The fix should work on both v1 and cc.

After update with fix:
Screenshot 2025-01-23 at 13 42 37

### DIFF
--- a/apps/condo/domains/ticket/components/BaseTicketForm/index.tsx
+++ b/apps/condo/domains/ticket/components/BaseTicketForm/index.tsx
@@ -804,6 +804,7 @@ export const BaseTicketForm: React.FC<ITicketFormProps> = (props) => {
         client.cache.evict({ id: 'ROOT_QUERY', fieldName: 'allTicketChanges' })
         client.cache.evict({ id: 'ROOT_QUERY', fieldName: 'allContacts' })
         client.cache.evict({ id: 'ROOT_QUERY', fieldName: 'allInvoices' })
+        // TODO: DOMA-11038 delete this evict, then cache in ROOT_QUERY works correctly
         client.cache.evict({ id: 'ROOT_QUERY', fieldName: 'allTicketComments' })
         client.cache.gc()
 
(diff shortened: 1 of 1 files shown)


apps/condo/pages/ticket/[id]/index.tsx
```diff
@@ -581,6 +581,7 @@ export const TicketPageContent = ({ ticket, pollCommentsQuery, refetchTicket, or
             ticketId: id,
         },
         skip: !persistor,
+        fetchPolicy: 'cache-and-network',
```
Comment (SavelevMatthew) [inline apps/condo/pages/ticket/[id]/index.tsx:584]: why?

Comment (SavelevMatthew) [reply apps/condo/pages/ticket/[id]/index.tsx:584]: @nomerdvadcatpyat You've already implemented comment polling properly, why do you need to use network again here?

toplenboren (approved):

tolmachev21:
The current solution is temporary to address a pain point for the customer center. Opened an issue with @nerdwallet/apollo-cache-policies to clarify their understanding of cleaning broken links

Alllex202 (approved):
Write note that this is a temporary solution.

nomerdvadcatpyat (approved):

PR #5715, opened 2025-01-23
hotfix: fix new receipts push task fallings

vovaaxeapolla:


apps/condo/domains/resident/tasks/sendBillingReceiptsAddedNotificationForOrganizationContextTask.js
```diff
@@ -73,7 +74,7 @@ const prepareAndSendNotification = async (keystone, context, receipt, resident,
         meta: { dv: 1, data },
         sender: { dv: 1, fingerprint: 'send-billing-receipts-added-notifications' },
         uniqKey: notificationKey,
-        organization: { id: resident.organization },
+        organization: organizationId && { id: organizationId },
```
AleX83Xpert, line 77: What if no organizationId?
Maybe you meant this?
```suggestion
        organization: organizationId && { id: organizationId },
        ...(organizationId ? { id: organizationId } : {}),
```

vovaaxeapolla: This returns undefined/null or { id: organizationId }, this is not required field and can be undefined/null  or { id: organizationId } but not { id: undefined/null }

SavelevMatthew: Its 3 separate behaviours:

1. If some filter is `undefined`, keystone will try to ignore it. So it basically the "allow all" filter
2. If some filter is `organization: null` this will throw error
3. And if id is passed, then it resolves correctly

SavelevMatthew: In code above you're relying on "Keystone magic that live under the hood", please specify correct behaviour explicitly

apps/condo/domains/resident/tasks/sendBillingReceiptsAddedNotificationForOrganizationContextTask.js
```diff
@@ -73,7 +78,7 @@ const prepareAndSendNotification = async (keystone, context, receipt, resident,
         meta: { dv: 1, data },
         sender: { dv: 1, fingerprint: 'send-billing-receipts-added-notifications' },
         uniqKey: notificationKey,
-        organization: { id: resident.organization },
+        organization: organizationId,
```
YEgorLu, line 81: So organizationId can be null in your code. It means that keystone will throw error `Cannot convert undefined or null to object`

Maybe keep organizationId as undefined initially: `let organizationId`
Or add it later:
```
const messageData = {
        lang: locale,
        to: { user: { id: resident.user } },
        type: messageType,
        meta: { dv: 1, data },
        sender: { dv: 1, fingerprint: 'send-billing-receipts-added-notifications' },
        uniqKey: notificationKey,
} // there is no 'organization' key
if (resident.organization) {
    messageData.organization = { id: resident.organization }
}
````

File: apps/condo/domains/resident/tasks/sendBillingReceiptsAddedNotificationForOrganizationContextTask.js (hunk 1 above)
Comment (YEgorLu) [reply apps/condo/domains/resident/tasks/sendBillingReceiptsAddedNotificationForOrganizationContextTask.js]: This is sent to sendMessageService, here 'organization' property is not required, and it does 
```
if (organization) { attributes['organization'] = { connect: organization } }
```
So undefined will not allow all, null will not throw

YEgorLu (approved):

SavelevMatthew (changes requested):
As @AleX83Xpert said, `organizationId` must contain id, not filter. Clear variable naming is the key to code sustainability...

And also, why is it  taken from resident? Aren't we notifying about service providers as well?

PR #5719, opened 2025-01-24
feat(global)!: INFRA-791 KV storage adapter with keyPrefix

sitozzz:
- [x] Add migration docs for community
- [x] Double check test cases, maybe add more business specific cases


bin/rename-redis-keys.js
```diff
@@ -0,0 +1,57 @@
...
+        console.log('keys count -> ', keys.length)
+        cursor = newCursor
+
+        if (keys.length) {
+            const multi = client.multi()
+
+            const filterQuery = direction === 'forward'
+                ? (key) => !key.startsWith(REDIS_KEY_PREFIX)
+                : (key) => key.startsWith(REDIS_KEY_PREFIX)
+            keys.filter(filterQuery).forEach((key) => {
```
SavelevMatthew, line 35: Isn't renamenx safer for our case?

https://redis.github.io/ioredis/classes/Redis.html#renamenx

packages/keystone/redis.js
```diff
@@ -1,12 +1,51 @@
...
+if (get(conf, 'REDIS_FALLBACK_ENABLED', 'false') === 'true') {
+    const originalSendCommand = IORedis.prototype.sendCommand
+    IORedis.prototype.sendCommand = function (...args) {
+        const [command] = args
+
+        if (!READONLY_COMMANDS.includes(command.name)) {
+            return originalSendCommand.apply(this, args)
+        }
+
+        const prefixedKey = command.args[0]
```
SavelevMatthew, line 25: Is this always in first place for all operations?

packages/keystone/redis.js
```diff
@@ -1,12 +1,51 @@
...
+
+        return new Promise((resolve, reject) => {
+            originalSendCommand.apply(this, args).then((value) => {
+                if (value !== null || !prefixedKey.startsWith(PREFIX)) {
+                    resolve(value)
+                } else {
+                    const nonPrefixedKey = prefixedKey.substring(PREFIX.length)
+
+                    const otherCommand = new IORedis.Command(command.name, [nonPrefixedKey], 'utf-8')
```
SavelevMatthew, line 34: Are all commands have only 1 argument?

packages/keystone/redis.spec.js
```diff
@@ -0,0 +1,79 @@
...
+
+
+describe('Redis adapter', () => {
+    let redisClient
+    let nonPrefixedClient
+
+    beforeAll(() => {
+        jest.resetModules()
+        process.env.REDIS_URL = conf['REDIS_URL'] || 'redis://127.0.0.1:6379'
+        process.env.REDIS_FALLBACK_ENABLED = conf['REDIS_FALLBACK_ENABLED'] || 'true'
```
SavelevMatthew, line 13: Should this be resetted in afterAll or not?

SavelevMatthew (commented):
I guess it will be helpful to add migration speed (in `keys / second`?), so other developers can predict their migration time.

IMHO it will be helpful metric, since Redis is single threaded, so we shouldn't expect this metric to vary a lot I guess, should we?

packages/keystone/redis.js
```diff
@@ -1,12 +1,51 @@
...
+const { get } = require('lodash')
 
 const conf = require('@open-condo/config')
 
 const { getLogger } = require('./logging')
 
+const READONLY_COMMANDS = ['bitcount', 'bitfield_ro', 'bitpos', 'dbsize', 'dump', 'eval_ro', 'evalsha_ro', 'exists', 'expiretime', 'fcall_ro', 'geodist', 'geohash', 'geopos', 'georadius_ro', 'georadiusbymember_ro', 'geosearch', 'get', 'getbit', 'getrange', 'hexists', 'hget', 'hgetall', 'hkeys', 'hlen', 'hmget', 'hrandfield', 'hscan', 'hstrlen', 'hvals', 'keys', 'lcs', 'lindex', 'llen', 'lolwut', 'lpos', 'lrange', 'mget', 'pexpiretime', 'pfcount', 'pttl', 'randomkey', 'scan', 'scard', 'sdiff', 'sinter', 'sintercard', 'sismember', 'smembers', 'smismember', 'sort_ro', 'srandmember', 'sscan', 'strlen', 'substr', 'sunion', 'touch', 'ttl', 'type', 'xlen', 'xpending', 'xrange', 'xread', 'xrevrange', 'zcard', 'zcount', 'zdiff', 'zinter', 'zintercard', 'zlexcount', 'zmscore', 'zrandmember', 'zrange', 'zrangebylex', 'zrangebyscore', 'zrank', 'zrevrange', 'zrevrangebylex', 'zrevrangebyscore', 'zrevrank', 'zscan', 'zscore', 'zunion']
+
+
+const PREFIX = process.cwd().split('/').pop() + ':'
```
pahaz, line 11: We have a similar logic inside the conf package. May be it's better to move this like inside conf?

apps/condo/migrations/20250124162345-0446_auto_20250124_1123.js
```diff
@@ -0,0 +1,22 @@
+// auto generated by kmigrator
+// KMIGRATOR:0446_auto_20250124_1123:IyBHZW5lcmF0ZWQgYnkgRGphbmdvIDQuMCBvbiAyMDI1LTAxLTI0IDExOjIzCgpmcm9tIGRqYW5nby5kYiBpbXBvcnQgbWlncmF0aW9ucwoKCmNsYXNzIE1pZ3JhdGlvbihtaWdyYXRpb25zLk1pZ3JhdGlvbik6CgogICAgZGVwZW5kZW5jaWVzID0gWwogICAgICAgICgnX2RqYW5nb19zY2hlbWEnLCAnMDQ0NV9vcmdhbml6YXRpb25lbXBsb3llZXJlcXVlc3RoaXN0b3J5cmVjb3JkX2FuZF9tb3JlJyksCiAgICBdCgogICAgb3BlcmF0aW9ucyA9IFsKICAgIF0K
+
+exports.up = async () => {
+    const { renameRedisKeys } = require('../../../bin/rename-redis-keys')
```
pahaz, line 5: You need to ensure that this file remains functional in the future. It's unclear to future developers whether their refactoring of this binary file will disrupt the migration process.

bin/rename-redis-keys.js
```diff
@@ -0,0 +1,57 @@
...
+                }
+            })
+
+            const result = await multi.exec()
+
+            console.log('rename for batch result -> ', result.length)
+        }
+
+        if (newCursor === 0 || newCursor === '0') {
+            console.log('CURSOR=0 -> stopping execution')
```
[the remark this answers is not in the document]
sitozzz: Removed non used logs & make existing more friendly

packages/keystone/redis.js
sitozzz: If I understand correctly, the problem is fixed

packages/keystone/redis.js
sitozzz: If I understand correctly, the problem is fixed

packages/keystone/redis.spec.js
sitozzz: it's not necessary, but it won't hurt

sitozzz:
> I guess it will be helpful to add migration speed (in `keys / second`?), so other developers can predict their migration time.
> 
> IMHO it will be helpful metric, since Redis is single threaded, so we shouldn't expect this metric to vary a lot I guess, should we?

I'll try to implement something similar. But it seems that this will not be so useful, because the migration logs are issued after its completion

apps/condo/migrations/20250124162345-0446_auto_20250124_1123.js
sitozzz: Yes, that's a good point

packages/keystone/redis.spec.js
```diff
@@ -0,0 +1,82 @@
...
+
+    afterAll(async () => {
+        await nonPrefixedClient.del('test')
+        await nonPrefixedClient.del('keystone:test1')
+        await nonPrefixedClient.del('keystone:incrTest')
+        await nonPrefixedClient.del('keystone:testList')
+        await nonPrefixedClient.disconnect()
+        await redisClient.disconnect()
+
+        process.env = { ...OLD_ENV }
```
SavelevMatthew, line 29: This will not work:
image

Since `process.env` is object, you're only creating new reference in OLD_ENV in line 7. So now you have 2 pointers, but same object

So then you're mutating object by adding new property (`process.env.something = 'something` is equal to `OLD_ENV.something = 'something'`, since pointers are same)

And after all you're spreading here not old copy, but mutated one

sitozzz: Ok, fixed that with deep copy

packages/keystone/redis.js
sitozzz: I didn't quite understand

bin/prepare.js
SavelevMatthew, line 1: unnecessary

bin/rename-redis-keys.js
SavelevMatthew, line 1: outdated

docker-compose.yml
SavelevMatthew, line 1: remove

packages/keystone/package.json
```diff
@@ -72,7 +72,7 @@
     "graphql-upload": "^15.0.2",
     "hot-shots": "^10.0.0",
     "inflection": "^1.13.1",
-    "ioredis": "^4.28.5",
+    "iovalkey": "^0.3.1",
```
SavelevMatthew, line 75: Rollback?

packages/keystone/session.spec.js
SavelevMatthew, line 1: rollback?

bin/prepare.js
```diff
@@ -112,7 +111,7 @@ async function prepare () {
             logWithIndent('Writing assigned urls / ports / dbs to app\'s .env', 2)
             const env = {
                 DATABASE_URL: `${LOCAL_PG_DB_PREFIX}/${app.pgName}`,
-                REDIS_URL: `${LOCAL_REDIS_DB_PREFIX}/${app.redisIndex}`,
```
SavelevMatthew, line 115: Are you sure you want to change that in this PR?

packages/keystone/kv.js
```diff
@@ -7,15 +11,48 @@ const { getLogger } = require('./logging')
...
+            const stats = fs.statSync(packageJsonPath, { throwIfNoEntry: false })
+            if (stats?.isFile()) {
+                break
+            }
+        } catch (e) { console.error(e) }
+
+        directory = path.dirname(directory)
+    }
+
+    return require(packageJsonPath).name.split('/')
```
SavelevMatthew, line 37: toLower?

packages/keystone/kv.spec.js
```diff
@@ -0,0 +1,74 @@
...
+    let nonPrefixedClient
+    let moduleName
+
+    beforeEach(() => {
+        nonPrefixedClient = new Redis(process.env.VALKEY_URL)
+        moduleName = require(process.cwd() + '/package.json').name.split('/').pop() + ':'
+
+        jest.resetModules()
+
+        const { getKVClient } = require('./kv')
```
SavelevMatthew, line 17: 🤔 
https://github.com/open-condo-software/condo/blob/main/packages/migrator/src/utils/packages.spec.ts#L142

packages/keystone/kv.js
```diff
@@ -7,15 +11,48 @@ const { getLogger } = require('./logging')
 const KV_CLIENTS = {}
 const logger = getLogger('kv')
 
+const getKVPrefix = () => {
+    const toPath = urlOrPath => urlOrPath instanceof URL ? fileURLToPath(urlOrPath) : urlOrPath
```
SavelevMatthew, line 15: What is URL used for?

packages/keystone/kv.js
```diff
@@ -7,15 +11,48 @@ const { getLogger } = require('./logging')
...
+                break
+            }
+        } catch (e) { console.error(e) }
+
+        directory = path.dirname(directory)
+    }
+
+    return require(packageJsonPath).name.split('/')
+        .pop()
+        .replace(/:/g, '')
```
SavelevMatthew, line 39: toLower?

packages/keystone/kv.spec.js
```diff
@@ -0,0 +1,74 @@
...
+const { getKVPrefix } = require('./kv')
+
+describe('Key value adapter', () => {
+    const OLD_ENV = JSON.parse(JSON.stringify(process.env))
+    let client
+    let nonPrefixedClient
+    let moduleName
+
+    beforeEach(() => {
+        nonPrefixedClient = new Redis(process.env.VALKEY_URL)
```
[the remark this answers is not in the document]
sitozzz: That rebase was kinda ****

packages/keystone/kv.js
sitozzz: Why not)

bin/prepare.js
sitozzz: I'm not completely sure

SavelevMatthew (approved):
👍

PR #5730, opened 2025-01-27
fix(condo): DOMA-11011 remove unnecessary queries

tolmachev21:
Problem:
After exiting the platform, there are extra requests that shouldn't be there.

Screenshot 2025-01-27 at 17 46 10

This happens because of extra re-renders of components, which after signOut() mutations change the cache state. And since the global user changes, all components make requests to the server and receive an error that the user is unauthorized.

Solution:
I slightly adjusted the logic of the component so that when re-rendering child components, you can rely on the fact that either the user will be empty (user: null) or isLoading will be true (isLoading: true). Then you can rely on this state in child components and not make extra requests

https://github.com/user-attachments/assets/19875d0c-a36a-4e8d-bb37-68bd12b1e5db

### DIFF
--- a/apps/condo/pages/_app.tsx
+++ b/apps/condo/pages/_app.tsx
@@ -458,10 +458,14 @@ const TasksProvider = ({ children }) => {
 }
 
 const MyApp = ({ Component, pageProps }) => {
+    console.log('Render MyApp')
     const intl = useIntl()
     useHotCodeReload()
     dayjs.locale(intl.locale)
     const router = useRouter()
+    const auth = useAuth()
+    console.log('MyApp user', auth.user)
+    console.log('MyApp isLoading', auth.isLoading)
     const { publicRuntimeConfig: { yandexMetrikaID, popupSmartConfig, UseDeskWidgetId, isSnowfallDisabled } } = getConfig()
 
     const LayoutComponent = Component.container || BaseLayout
@@ -619,6 +623,7 @@ if (!isDisabledSsr || !isSSR()) {
 
 const withCookies = () => (PageComponent: NextPage): NextPage => {
     const WithCookies = (props) => {
+        console.log('Render WithCookies')
         const ssrCookies = useVitalCookies(props?.pageProps)
 
         return (
--- a/apps/condo/pages/_app.tsx
+++ b/apps/condo/pages/_app.tsx
@@ -388,7 +388,7 @@ const MenuItems: React.FC = () => {
 }
 
 const TasksProvider = ({ children }) => {
-    const { user } = useAuth()
+    const { user, isLoading } = useAuth()
     // Use UI interfaces for all tasks, that are supposed to be tracked
     const { TicketDocumentGenerationTask: TicketDocumentGenerationTaskUIInterface } = useTicketDocumentGenerationTaskUIInterface()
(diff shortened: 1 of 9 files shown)


packages/next/auth.tsx
```diff
@@ -263,7 +263,9 @@ const _withAuthLegacy: WithAuthLegacyType = ({ ssr = false, ...opts } = {}) => P
 const AuthProvider: React.FC = ({ children }) => {
     const apolloClient = useApolloClient()
 
-    const { data, loading: userLoading, refetch } = useQuery(USER_QUERY)
+    const { data, loading: userLoading, refetch } = useQuery(USER_QUERY, {
+        notifyOnNetworkStatusChange: true,
```
Alllex202, line 267: What is this for? It looks like you are not using network status

tolmachev21: This is necessary to more clearly track the loading status of the request

nomerdvadcatpyat (approved):

packages/next/organization.tsx
```diff
@@ -384,23 +384,23 @@ const OrganizationProvider: React.FC<OrganizationProviderProps> = ({
...
             value={{
                 selectLink: handleSelectLink,
                 selectEmployee: handleSelectEmployee,
-                isLoading: (!auth.user || !activeEmployeeId) ? false : isLoading,
+                isLoading: (!user || !activeEmployeeId) ? false : isLoading,
                 link: (activeEmployee && activeEmployee.id) ? activeEmployee : null,
-                organization: (activeEmployee && activeEmployee.organization) ? activeEmployee.organization : null,
+                organization: (user && activeEmployee && activeEmployee.organization) ? activeEmployee.organization : null,
                 employee: (activeEmployee && activeEmployee.id) ? activeEmployee : null,
                 role: (activeEmployee && activeEmployee.role) ? activeEmployee.role : null,
```
Alllex202, line 405: if you changed the logic of the `organization` field, then you should change the logic of other fields too. Otherwise, it turns out that the organization may not exist, but the role, employee, etc. exist.

apps/condo/pages/_app.tsx
```diff
@@ -405,7 +405,7 @@ const TasksProvider = ({ children }) => {
     // Load all tasks with 'processing' status
     const { data, loading: isProcessingTasksLoading } = useGetProcessingTasksQuery({
         variables: { userId: user?.id || null, createdAtGte: dayjs().startOf('day').toISOString() },
-        skip: !user?.id,
+        skip: !user?.id || isLoading,
```
Alllex202, line 408: Do we now need to check organization and user load in all requests?

tolmachev21: Yes, because the user switches to a false not immediately, but after several renderings

packages/next/organization.tsx
tolmachev21: Okay, good point, thanks

Alllex202 (approved):

apps/condo/pages/_app.tsx
Alllex202: I don't think anyone used isUserLoading on purpose in requests)

tolmachev21: I think that there is a little bit of uneven logic there. Because somewhere the organization loading state is used, and somewhere the user loading state. Therefore, we need one depending on the context

Issue #5697, opened 2025-01-16
A ticket not from resident redirects to a non-existent page when clicking on the mouse wheel


crowncode-development:
When creating a ticket from a non-resident, when specifying his phone number and name in the ticket form, clicking on the client's name through the mouse wheel in the ticket card redirects to a non-existent `page/ticket/[id]#`. Screenshot below

![Image](https://github.com/user-attachments/assets/14289f27-adb2-4578-92e1-6a662c5b0a87)

sitozzz:
Hi, thank you for helping us improve the platform. 
This issue will be fixed in one of the upcoming releases. 
I will attach this post to the pull request with the fix.

PR #5734, opened 2025-01-30
fix(condo): INFRA-837 hide link from resident ticket[id] page if there is no resident

sitozzz:
That's fix for issue - https://github.com/open-condo-software/condo/issues/5697

### DIFF
--- a/apps/condo/domains/ticket/components/TicketId/TicketUserInfoField.tsx
+++ b/apps/condo/domains/ticket/components/TicketId/TicketUserInfoField.tsx
@@ -1,5 +1,4 @@
-import { Ticket, User } from '@app/condo/schema'
-import { get } from 'lodash'
+import { User } from '@app/condo/schema'
 import Link from 'next/link'
 import React, { useMemo } from 'react'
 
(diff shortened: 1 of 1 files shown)


apps/condo/domains/ticket/components/TicketId/TicketUserInfoField.tsx
```diff
@@ -34,24 +34,34 @@ interface ITicketUserInfoFieldProps {
 export const TicketUserInfoField: React.FC<ITicketUserInfoFieldProps> = (props) => {
     const id = useMemo(() => get(props, ['user', 'id']), [props])
     const name = useMemo(() => get(props, ['user', 'name']), [props])
-    const nameLink = useMemo(() => get(props, ['nameLink'], '#'), [props])
+    const nameLink = useMemo(() => get(props, ['nameLink']), [props])
     const phone = useMemo(() => get(props, ['user', 'phone']), [props])
     const email = useMemo(() => get(props, ['user', 'email']), [props])
     const userInfo = useMemo(() => [], [])
```
Alllex202, line 40: Could you please get rid of the `get` function call in this component at the same time? :)

Alllex202 (approved):

apps/condo/domains/ticket/components/TicketId/TicketUserInfoField.tsx
sitozzz: Do you propose to implement it in the following way?
```js
const nameLink = useMemo(() => props?.nameLink, [props])
```

Alllex202: Yeah, optional chains

sitozzz: Done that

nomerdvadcatpyat (approved):

PR #5740, opened 2025-01-31
chore(condo): DOMA-11041 pass category to registerMultiPaymentForVirt…

AleX83Xpert:
…ualReceipt

Also, throw an error if no receipt found after qr-code scanned

### DIFF
--- a/apps/condo/domains/acquiring/schema/CreatePaymentByLinkService.js
+++ b/apps/condo/domains/acquiring/schema/CreatePaymentByLinkService.js
@@ -5,7 +5,7 @@
 const Big = require('big.js')
 const { get } = require('lodash')
 
-const { GQLError, GQLErrorCode: { BAD_USER_INPUT } } = require('@open-condo/keystone/errors')
+const { GQLError, GQLErrorCode: { BAD_USER_INPUT, INTERNAL_ERROR  } } = require('@open-condo/keystone/errors')
 const { GQLCustomSchema, find } = require('@open-condo/keystone/schema')
 
 const access = require('@condo/domains/acquiring/access/CreatePaymentByLinkService')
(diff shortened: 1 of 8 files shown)


apps/condo/domains/acquiring/schema/CreatePaymentByLinkService.js
```diff
@@ -39,6 +39,12 @@ const ERRORS = {
         message: 'Provided receipt already paid',
         messageForUser: 'api.billing.billingReceipt.RECEIPT_ALREADY_PAID_ERROR',
     },
+    NO_PREV_RECEIPT: {
+        mutation: 'createPaymentByLink',
+        code: INTERNAL_ERROR,
```
pahaz, line 44: BAD USER INPUT

apps/condo/domains/acquiring/schema/CreatePaymentByLinkService.js
```diff
@@ -120,20 +127,25 @@ const CreatePaymentByLinkService = new GQLCustomSchema('CreatePaymentByLinkServi
                 }
 
                 const payForQR = async (lastBillingReceipt) => {
+                    if (!lastBillingReceipt) {
```
pahaz, line 130: please don't create a new function for every request. Could you please extract it as a normal function?

pahaz (approved):

apps/condo/domains/acquiring/schema/CreatePaymentByLinkService.js
AleX83Xpert: Fixed

apps/condo/domains/acquiring/schema/CreatePaymentByLinkService.js
AleX83Xpert: I need to pass this function as a resolver to another asynchronous function. I also need to have rw-access to context and other variables outside. The closure is good for that.

abshnko (approved):

PR #5767, opened 2025-02-05
doma-9987 synchronization template

Alexander-Turkin:
add emoji, add symbols of transfer, add receiving template


apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -1462,8 +1466,9 @@ export const BaseNewsForm: React.FC<BaseNewsFormProps> = ({
                                     (getStepTypeByStep(currentStep) === 'sharingApp') && (
                                         // TODO (DOMA-9328) Move onSkip to BaseNewsForm component, since steps are handled here!
                                         <NewsItemSharingForm
+                                            template={{ id: templateId, ...templates[templateId] }}
                                             ctxId={getStepDataByStep(currentStep).sharingAppData.id}
-                                            onSkip={() => handleStepSkip({ skip: true, step: currentStep })}
+                                            onSkip={()  => handleStepSkip({ skip: true, step: currentStep })}
```
toplenboren, line 1471:
```suggestion
                                            onSkip={() => handleStepSkip({ skip: true, step: currentStep })}
```

apps/condo/domains/news/components/NewsForm/CreateNewsForm.tsx
```diff
@@ -191,8 +191,8 @@ export const CreateNewsForm: React.FC = () => {
     const templates = isNewsItemTemplatesFetching || !newsItemTemplates?.length ? null : newsItemTemplates
         .reduce((acc, template) => {
             acc[template.id] = {
-                title: template.title,
-                body: template.body,
+                title: template.title.replace(/<\/br>/g, '\n'),
```
toplenboren, line 194: Why do you need this logic?

apps/condo/domains/news/components/TemplatesSelect.tsx
```diff
@@ -69,7 +69,7 @@ export const TemplatesSelect: React.FC<INewsFormProps> = ({ items, onChange, has
             const result = []
             for (const category of (Object.keys(templatesByCategory).reverse())) {
                 result.push({
-                    label: intl.formatMessage( { id: `news.template.category.${CATEGORIES[category]}` }),
+                    label: category ? intl.formatMessage( { id: `news.template.category.${CATEGORIES[category]}` }) : '',
```
toplenboren, line 72: Is this for default category?

Alexander-Turkin: There is an empty template that does not have a category => there is no translation and if you remove this logic, the user will see news.template.category.undefined

apps/condo/domains/news/components/NewsForm/CreateNewsForm.tsx
Alexander-Turkin: I replaced the /n on </br> templates because when sending the transfer symbol is shielded, well, here you need to return it

apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -1462,6 +1466,7 @@ export const BaseNewsForm: React.FC<BaseNewsFormProps> = ({
                                     (getStepTypeByStep(currentStep) === 'sharingApp') && (
                                         // TODO (DOMA-9328) Move onSkip to BaseNewsForm component, since steps are handled here!
                                         <NewsItemSharingForm
+                                            template={{ id: templateId, ...templates[templateId] }}
```
toplenboren, line 1469: why not just ...templates[templateId] ?

apps/condo/domains/news/components/NewsForm/CreateNewsForm.tsx
toplenboren: why not just use url encoding?

apps/condo/domains/news/components/TemplatesSelect.tsx
toplenboren: But shouldnt we return some default category instead of nothing?

apps/condo/domains/news/components/NewsForm/NewsItemSharingForm.tsx
```diff
@@ -1,4 +1,4 @@
-import { B2BApp } from '@app/condo/schema'
+import { B2BApp, NewsItemTemplateTypeType } from '@app/condo/schema'
```
toplenboren, line 1: why TypeType?

apps/condo/domains/news/schema/NewsItemTemplate.test.js
```diff
@@ -92,6 +92,16 @@ describe('NewsItemTemplate', () => {
                 expect(obj.body).toEqual(attrs.body)
                 expect(obj.type).toEqual(attrs.type)
             })
+            
+            test('create with special symbols', async () => {
```
toplenboren, line 96:
```suggestion
            test('can be created with special symbols', async () => {
```

apps/condo/domains/news/components/NewsForm/BaseNewsForm.tsx
```diff
@@ -621,7 +621,11 @@ export const BaseNewsForm: React.FC<BaseNewsFormProps> = ({
         Body.setTextLength(initialBody.length)
     }, [])
 
+    const [templateId, setTemplateId] = useState('')
```
toplenboren, line 624: I think when dealing with Ids its more clean to make values null || string

apps/condo/domains/news/schema/NewsItemTemplate.js
```diff
@@ -45,12 +46,22 @@ const NewsItemTemplate = new GQLListSchema('NewsItemTemplate', {
             schemaDoc: 'The title of a future news item',
             type: 'Text',
             isRequired: true,
+            hooks: {
+                resolveInput: async ({ resolvedData }) => {
+                    return normalizeText(resolvedData['title'])
+                },
+            },
```
Alllex202, line 53: Do we need to do same for the `name` field?

Alexander-Turkin: no, field is used in the selector

Alllex202: I see this type for the `name` field: `Text`, not `Select`

PR #5768, opened 2025-02-05
refactor(global): INFRA-840 a lot of renamings everywhere

sitozzz:


werf.yaml
```diff
@@ -451,7 +452,7 @@ docker:
   ENV:
     LANG: C.UTF-8
 ---
-image: classifier
+image: classifiers-api
```
SavelevMatthew, line 455: _

SavelevMatthew (approved):

PR #5779, opened 2025-02-07
fix(condo): DOMA-11023 prevent update the `client` field if no reside…

AleX83Xpert:
…nt found

### DIFF
--- a/apps/condo/domains/marketplace/schema/Invoice.js
+++ b/apps/condo/domains/marketplace/schema/Invoice.js
@@ -584,7 +584,9 @@ const Invoice = new GQLListSchema('Invoice', {
                             deletedAt: null,
                         })
 
-                        set(resolvedData, 'client', get(resident, 'user', null))
+                        if (resident) {
+                            set(resolvedData, 'client', get(resident, 'user'))
+                        }
                     }
                 }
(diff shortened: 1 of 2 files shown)


sitozzz (approved):

nomerdvadcatpyat (approved):

apps/condo/domains/marketplace/schema/Invoice.js
```diff
@@ -584,7 +584,9 @@ const Invoice = new GQLListSchema('Invoice', {
                             deletedAt: null,
                         })
 
-                        set(resolvedData, 'client', get(resident, 'user', null))
+                        if (resident) {
+                            set(resolvedData, 'client', get(resident, 'user'))
```
nomerdvadcatpyat, line 588: If we change the contact (client data) and no Resident is found for the new contact, the old client will remain here. Is that okay?

AleX83Xpert: I thought about this. There are two options:
1. Allow to change the `client` field for published invoices
2. Keep the `client` field as is in the case of no resident found

I've chosen the 2nd to keep the invoice visible for the same user.

PR #5780, opened 2025-02-07
chore(condo): DOMA-10985 update graphql

tolmachev21:
This PR updates graphql.
However, the major version changes from 15 to 16, so they stop supporting positional arguments and instead you need to pass an object.
Since keystone v5 was dropped, we had to make two patches with a small adjustment to the execute function

### DIFF
--- a/.yarn/patches/@keystonejs-keystone-npm-19.3.4-2a53e53061.patch
+++ b/.yarn/patches/@keystonejs-keystone-npm-19.3.4-2a53e53061.patch
@@ -1,5 +1,5 @@
 diff --git a/lib/Keystone/index.js b/lib/Keystone/index.js
-index 5f76cf925886570bc5416d50d7335923d3992ae7..d1eefcdcc6505b69b7391b15af1bba49ae97b042 100644
+index 5f76cf925886570bc5416d50d7335923d3992ae7..7da7b0d18242660b5aa5f458674429c8f3b50a41 100644
 --- a/lib/Keystone/index.js
 +++ b/lib/Keystone/index.js
 @@ -4,7 +4,7 @@ const memoize = require('micro-memoize');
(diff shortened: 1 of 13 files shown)


.yarn/patches/apollo-server-core-npm-2.26.0-6396d881cb.patch
tolmachev21, line 1: You only need to leave the update in the dist folder

tolmachev21 (commented):
Update all submodules

.yarn/patches/@keystonejs-keystone-npm-19.3.4-2a53e53061.patch
```diff
@@ -1,13 +1,13 @@
...
+index 5f76cf925886570bc5416d50d7335923d3992ae7..7da7b0d18242660b5aa5f458674429c8f3b50a41 100644
 --- a/lib/Keystone/index.js
 +++ b/lib/Keystone/index.js
 @@ -4,7 +4,7 @@ const memoize = require('micro-memoize');
  const falsey = require('falsey');
  const createCorsMiddleware = require('cors');
  const { execute, print } = require('graphql');
 -const { GraphQLUpload } = require('graphql-upload');
-+const GraphQLUpload= require('graphql-upload/GraphQLUpload.js');
++const GraphQLUpload = require('graphql-upload/GraphQLUpload.js');
```
Alllex202, line 10: Why not remove the file extension?
```suggestion
+const GraphQLUpload = require('graphql-upload/GraphQLUpload');
```

apps/callcenter
Alllex202, line 1: Do you really need to update submodules in this pr?

.yarn/patches/@keystonejs-keystone-npm-19.3.4-2a53e53061.patch
tolmachev21: These changes are already in the patch.

Since I was generating a new patch, I needed to collect all the changes from the existing patch

PR #5782, opened 2025-02-10
feat(condo): DOMA-11043 added restrictionEndDay field to MeterReportingPeriod

abshnko:
1. added isStrict and restrictionEndDay fields to MeterReportingPeriod
2. added validations on creating MeterReadings depending on whether the period allows for passing readings after the period's end
3. minor frontend fixes
image


apps/condo/domains/meter/schema/MeterReading.js
```diff
@@ -65,6 +74,36 @@ const MeterReading = new GQLListSchema('MeterReading', {
                                 messageInterpolation: { givenDate: date },
                             }, context)
                         }
+
+                        // NOTE: check against MeterReportingPeriod only readings passed from mobile app
+                        if (source === MOBILE_APP_SOURCE_ID) {
+                            const meter = await Meter.getOne(context, {
+                                id: get(resolvedData, 'meter', null),
```
YEgorLu, line 81: If "meter" is required, better just write "resolvedData.meter". get() says that meter field is nullable, and if you pass null here, function will throw if I remember correctly.

Add "deletedAt: null"?

apps/condo/domains/meter/schema/MeterReading.test.js
```diff
@@ -1984,4 +1986,273 @@ describe('MeterReading', () => {
             )
         })
     })
+
+    describe('Manipulating readings taking into account MeterReportingPeriod for corresponding property/organization', () => {
+        let organization, source, meter, property, reportingPeriod
```
AleX83Xpert, line 1991: Each test operates the same variable `reportingPeriod`. This variable may have a random value if tests run in parallel.

apps/condo/domains/meter/schema/MeterReading.test.js
```diff
@@ -1984,4 +1986,273 @@ describe('MeterReading', () => {
...
+  
+            const readingDate = dayjs().subtract(2, 'month').date(28).format('YYYY-MM-DD')
+            const [meterReading] = await createTestMeterReading(admin, meter, source, { date: readingDate })
+  
+            expect(meterReading.id).toMatch(UUID_RE)
+            expect(meterReading.date).toMatch(DATETIME_RE)
+        })
+  
+        test('skips MeterReportingPeriod validation if source is not mobile resident app', async () => {
+            await updateTestMeterReportingPeriod(admin, reportingPeriod.id, {
```
YEgorLu, line 2056: Better make requests as resident and not admin

apps/condo/domains/meter/utils/serverSchema/resolveHelpers.js
```diff
@@ -22,7 +23,54 @@ async function connectContactToMeterReading (context, resolvedData, existingItem
...
+ * If restrictionEnd is bigger than the number of days in the month, then return the last day of month, else return restrictionEnd
+ * @param restrictionEndDay
+ * @param lastDayOfMonth
+ * @returns {number}
+ */
+function getAdjustedRestrictionEnd (restrictionEndDay, lastDayOfMonth) {
+    return Math.min(restrictionEndDay, lastDayOfMonth)
+}
+
+function isReadingDateAllowed (date, meterReportingPeriod) {
```
AleX83Xpert, line 36: What do you think about a couple of specs for this function?

apps/condo/domains/meter/schema/MeterReading.js
```diff
@@ -65,6 +74,36 @@ const MeterReading = new GQLListSchema('MeterReading', {
                                 messageInterpolation: { givenDate: date },
                             }, context)
                         }
+
+                        // NOTE: check against MeterReportingPeriod only readings passed from mobile app
+                        if (source === MOBILE_APP_SOURCE_ID) {
+                            const meter = await Meter.getOne(context, {
```
AleX83Xpert, line 80: You use only IDs, so it's possible to use `getByCondition()`

apps/condo/domains/meter/utils/serverSchema/resolveHelpers.js
```diff
@@ -22,7 +23,54 @@ async function connectContactToMeterReading (context, resolvedData, existingItem
...
+    if (isRestrictionWithinAllowedPeriod) return true
+
+    // NOTE: restricted period can span two months, so we need to check both current and previous month
+    const monthStart = readingDate.startOf('month')
+    const prevMonthStart = monthStart.subtract(1, 'month')
+
+    const adjustedRestrictionEnd = getAdjustedRestrictionEnd(restrictionEndDay, readingDate.endOf('month').date())
+    const adjustedPrevRestrictionEnd = getAdjustedRestrictionEnd(restrictionEndDay, prevMonthStart.endOf('month').date())
+
+    const getPeriodEnd = (monthStart, notifyEnd) => monthStart.date(notifyEnd)
```
YEgorLu, line 57: It is said that function declaration is better, than function restriction. So maybe get these () => {} out from that function is separate function ...() {}?

apps/condo/domains/meter/utils/serverSchema/resolveHelpers.js
```diff
@@ -22,7 +23,54 @@ async function connectContactToMeterReading (context, resolvedData, existingItem
...
+        restrictionEndDay < notifyEnd
+            ? getPeriodEnd(monthStart, notifyEnd).add(1, 'month').date(restrictionEnd)
+            : monthStart.date(restrictionEnd)
+
+    const periodEnd = getPeriodEnd(monthStart, notifyEndDay)
+    const prevPeriodEnd = getPeriodEnd(prevMonthStart, notifyEndDay)
+    const restrictionEnd = getRestrictionEnd(monthStart, notifyEndDay, adjustedRestrictionEnd)
+    const prevRestrictionEnd = getRestrictionEnd(prevMonthStart, notifyEndDay, adjustedPrevRestrictionEnd)
+
+    return !(readingDate.isAfter(periodEnd, 'day') && !readingDate.isAfter(restrictionEnd, 'day') ||
```
YEgorLu, line 68: I guess it is hard expression to understand. Maybe split it by variables like

`
const a = readingDate.isAfter(periodEnd, 'day')
const b = !readingDate.isAfter(restrictionEnd, 'day')
const c = readingDate.isAfter(prevPeriodEnd, 'day')
const d = !readingDate.isAfter(prevRestrictionEnd, 'day')
const e = a && b || c && d
const f = !e
`

Or something like that. Can't give variables proper names, because i still don't understand fully what is happening here. I guess you are checking, if readingDate is not in proper date-range and then reversing result.

I saw tests for that logic, but it hard to understand, what will happen, if you set restrictionEndDay = 30, but current month have 28 days, maybe write test for that?

YEgorLu: And what if readingDate is after restrictionEndDay? Seems like last expression will give you "true", which means that resident can give meter readings for next period? should it be like that?

abshnko: the last expression allows readingDates that are after restrictionEndDay, but before the notifyEndDay (the end of allowed period). by default restrictionEndDay is 31, so starting with new month all readings will go into its month period

as for the date - it will be the date of creating meter reading, so they can only pass readings for the period that is right now (for example restrictionEndDay = 2, we turn the readings in on the 3rd: this will be allowed and the readings will go into this month's period)

YEgorLu: I think i understood logic behind restrictionEndDay.
But "date" is a field, potentially anyone can pass anything here, can't they?

abshnko: yes, you're right, perhaps we need to compare createdAt of MeterReading with its date and if these dates are not in the same date - throw validation error (so that there are no late turn-ins)

apps/condo/domains/meter/schema/MeterReading.test.js
abshnko: you're right, this turned out to be a bad practice. now i create separate entities for every test

apps/condo/domains/meter/utils/serverSchema/resolveHelpers.js
abshnko: decided to leave as is, they can be arrow functions

AleX83Xpert (approved):

dkoviazin (approved):

PR #5789, opened 2025-02-12
feat(ui): DOMA-11097 added switch component

nomerdvadcatpyat:
https://github.com/user-attachments/assets/0005279c-6995-467e-9ac3-539149c63500

[layouts](https://www.figma.com/design/Nbkug4IRM33sEyDuN0yO1C/UI-Kit-(web)?node-id=3356-6457&m=dev)

### DIFF
--- a/packages/ui/src/components/Switch/switch.tsx
+++ b/packages/ui/src/components/Switch/switch.tsx
@@ -1,0 +1,50 @@
+import {
+    Switch as DefaultSwitch,
+    SwitchProps as DefaultSwitchProps,
+} from 'antd'
+import React, { useCallback } from 'react'
+
+import { sendAnalyticsChangeEvent } from '../_utils/analytics'
+
+import './style.less'
+
+
+const SWITCH_CLASS_PREFIX = 'condo-switch'
+
+export type SwitchProps = Pick<DefaultSwitchProps,
+'className' |
+'id' |
+'autoFocus' |
+'checked' |
+'defaultChecked' |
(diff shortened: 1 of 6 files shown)


Alllex202:
Why doesn't the cursor change to a pointer when hovering over an interactive element?

nomerdvadcatpyat:
> Why doesn't the cursor change to a pointer when hovering over an interactive element?

Because there was a screen recording. I don’t know why this is so, but without screen recording the cursor becomes pointer

Alllex202 (approved):

packages/ui/src/components/Switch/switch.tsx
```diff
@@ -0,0 +1,50 @@
...
+const SWITCH_CLASS_PREFIX = 'condo-switch'
+
+export type SwitchProps = Pick<DefaultSwitchProps,
+'className' |
+'id' |
+'autoFocus' |
+'checked' |
+'defaultChecked' |
+'disabled' |
+'size' |
```
SavelevMatthew, line 21: I guess we should map our design system sizes (small / ~~medium~~ / large) to antd (small / default), shouldn't we?

nomerdvadcatpyat:
> The switch turns completely white for a split second when it turns off. Can we fix this?
> 
> The default `antd` doesn't have this problem
> 2025-02-13.15.04.58.mov

Fixed

https://github.com/user-attachments/assets/207716ef-0be0-443c-95bc-54fc971acdc6

PR #5799, opened 2025-02-14
feat(migrator): introduce @open-condo/migrator for backward incompatible changes 🏃🏻‍♂️🦖

SavelevMatthew:
## TODO

- [x] Package readme
- [x] E2E tests (cli.spec.ts)
- [x] Data version logic
- [ ] 3.0 Migration docs (outside this PR)
- [ ] Check redis version in migrations? (outside this PR)
- [x] Switch PR trigger back

### DIFF
--- a/packages/migrator/package.json
+++ b/packages/migrator/package.json
@@ -17,7 +17,8 @@
   ],
   "license": "MIT",
   "scripts": {
-    "build": "tsup"
+    "build": "tsup",
+    "test": "jest"
   },
   "bin": {
     "condo-migrator": "dist/cli.js",
@@ -39,11 +40,17 @@
     "access": "public"
   },
   "devDependencies": {
+    "@faker-js/faker": "^7.6.0",
     "@open-condo/tsconfig": "workspace:^",
     "@types/ioredis": "^4.28.5",
+    "@types/jest": "^29.5.14",
     "@types/node": "^16",
     "@types/prompts": "^2.4.9",
+    "@types/tmp": "^0.2.6",
     "commander": "^11.0.0",
+    "jest": "^29.7.0",
+    "tmp": "^0.2.3",
+    "ts-jest": "^29.2.5",
     "tsup": "^8.3.6",
     "typescript": "^5.5.4"
   },
--- a/packages/migrator/package.json
+++ b/packages/migrator/package.json
@@ -40,6 +40,7 @@
   },
   "devDependencies": {
     "@open-condo/tsconfig": "workspace:^",
+    "commander": "^11.0.0",
     "tsup": "^8.3.6"
   }
 }
--- a/packages/migrator/package.json
+++ b/packages/migrator/package.json
@@ -46,6 +46,8 @@
     "typescript": "^5.5.4"
   },
   "dependencies": {
+    "cli-table3": "^0.6.5",
+    "dotenv": "^16.4.7",
     "glob": "^11.0.1",
     "zod": "^3.24.2"
   }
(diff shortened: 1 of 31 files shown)


packages/migrator/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "@open-condo/migrator",
+  "version": "0.0.0-development",
```
sitozzz, line 3: Maybe set that to something different?

packages/migrator/src/cli.ts
```diff
@@ -0,0 +1,38 @@
...
+    .description([
+        'Adds application prefixes to KV database keys. ',
+        'You can provide connection string by REDIS_URL env when migration single app, or by .env files when migrating multiple apps locally.',
+        '',
+        'For example, the app keys “@app/example-keystone” would look like this:',
+        '"some_key" -> "example_keystone:some_key"',
+        '"bull:queue_name:something" -> "{example_keystone:bull:queue_name}:something"',
+        '',
+        'Example #1 (remote server migration):',
+        'REDIS_URL=redis://127.0.0.1:6379/6 npx @open-condo/migrator add-kv-prefixes',
```
sitozzz, line 28: It seems it would be better to give an example of a line with authorization
```suggestion
        'REDIS_URL=redis://remote_user:*******@127.0.0.1:6379/6 npx @open-condo/migrator add-kv-prefixes',
```

SavelevMatthew: fixed

packages/migrator/src/cli.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+import { createTestMonoRepo } from '@/utils/tests/repo'
+
+function expectSuccessfulExecution (result: ExecaReturnValue) {
+    const { stdout, stderr, exitCode } = result
+    expect(exitCode).toEqual(0)
+    expect(stderr).toBe('')
+    expect(stdout).not.toBe('')
+}
+
+let currentDB = -1
```
sitozzz, line 23: You can use `INFO keyspace` command to discover active databases.
That command will give you something like that:
```
db0:keys=10,expires=0
db1:keys=1,expires=0
db3:keys=1,expires=0
```

Btw your approach will work too)

packages/migrator/src/utils/kvdb/addAppPrefix.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+function getRedisVirtualDBConnection (host: string, port: number): string {
+    const dbIndex = ++currentDB
+    return `redis://${host}:${port}/${dbIndex}`
+}
+
+const TASKS_QUEUES = ['tasks', 'low', 'high']
+
+function randInt (min: number, max: number): number {
+    return Math.floor(Math.random() * (max - min)) + min
+}
```
sitozzz, line 33: Why not use a ready-made one if it is already installed?

```faker.datatype.number({min: 10, max: 50})```

SavelevMatthew: Had no idea this existed :)

So decided to write my own one-liner

packages/migrator/src/cli.spec.ts
SavelevMatthew: I guess your solution is more ultimate, but It has some cons: 
To extract connectionString via `info keyspace` you need connected client... And to connect client you need connection string 😅 

I can possibly connect to 0-indexed one, receive free index and re-connect, but I guess its overhead for now

But thanks for suggestion, will keep it in mind in case of future refactoring

PR #5800, opened 2025-02-14
feat(condo): DOMA-10862 added dv, sender, captcha, userType for authenticateUserWithPhoneAndPassword mutation

Alllex202:


apps/condo/domains/user/components/auth/SignInForm.tsx
```diff
@@ -68,9 +71,18 @@ export const SignInForm = (): React.ReactElement => {
     const onFormSubmit = useCallback((values) => {
         setIsLoading(true)
 
+        const sender = getClientSideSenderInfo()
         return runMutation({
```
tolmachev21, line 75: Just a question: or maybe we should immediately move to new hooks?

tolmachev21 (approved):

apps/condo/domains/user/components/auth/SignInForm.tsx
Alllex202: No, it will be in another pr

nomerdvadcatpyat (approved):

toplenboren (approved):

PR #5801, opened 2025-02-17
fix(spp): sbbol api now do not return userGuid

dkoviazin:
a combination of the "sub" (Subject) Claim and the "iss" (Issuer) Claim is always globally unique
we are using type sbbol instead of the iss, so we can use sub as a external user id

### DIFF
--- a/apps/condo/domains/organization/integrations/sbbol/utils/getSbbolUserInfoErrors.spec.js
+++ b/apps/condo/domains/organization/integrations/sbbol/utils/getSbbolUserInfoErrors.spec.js
@@ -21,9 +21,36 @@ const DEFAULT_SEED = {
 }
 
 describe('getSbbolUserInfoErrors', () => {
-    test('no userGuid', () => {
-        const errors = getSbbolUserInfoErrors(DEFAULT_SEED)
-        expect(errors).toEqual(['must have required property \'userGuid\''])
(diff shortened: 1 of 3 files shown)


apps/condo/domains/organization/integrations/sbbol/utils/getSbbolUserInfoErrors.spec.js
```diff
@@ -21,9 +21,36 @@ const DEFAULT_SEED = {
 }
 
 describe('getSbbolUserInfoErrors', () => {
-    test('no userGuid', () => {
-        const errors = getSbbolUserInfoErrors(DEFAULT_SEED)
-        expect(errors).toEqual(['must have required property \'userGuid\''])
+    test('no userGuid and no sub', () => {
+        const { sub, userGuid, ...info } = {
+            ...DEFAULT_SEED,
+            userGuid: '5a99ca6e22f321af3a',
```
Alllex202, line 27: Why do we set `userGuid` if we don't use it later?

Alllex202 (approved):

apps/condo/domains/organization/integrations/sbbol/utils/getSbbolUserInfoErrors.spec.js
dkoviazin: fixed

PR #5802, opened 2025-02-17
fix(condo): DOMA-11123 add 429 error

tolmachev21:
Added page when receiving 429 error code

Screenshot 2025-02-17 at 19 21 56


apps/condo/pages/_app.tsx
```diff
@@ -544,6 +545,7 @@ type NextAppContext = (AppContext & NextPageContext) & {
 if (!isDisabledSsr || !isSSR()) {
     MyApp.getInitialProps = async (appContext: NextAppContext): Promise<{ pageProps: Record<string, any> }> => {
         try {
+            // throw new Error('Произошла внутренняя ошибка сервера')
```
nomerdvadcatpyat, line 548: remove comment

apps/condo/pages/_app.tsx
```diff
@@ -612,7 +614,23 @@ if (!isDisabledSsr || !isSSR()) {
 
             return { pageProps }
         } catch (error) {
-            console.error('Error while running `MyApp.getInitialProps', error)
+            const tooManyRequests = error?.cause?.result?.errors
+                ?.some((error) => error?.message?.toLowerCase().includes('many requests recently'))
```
nomerdvadcatpyat, line 618: Can we use something other than a `message` to determine that this is a rate limit? Like code

apps/condo/pages/_app.tsx
Alllex202, line 548: ?

packages/keystone/rateLimiting/constants.js
```diff
@@ -1,7 +1,7 @@
 const DEFAULT_MUTATION_WEIGHT = 5
 const DEFAULT_QUERY_WEIGHT = 1
 const DEFAULT_MAX_TOTAL_RESULTS = 1000
-const DEFAULT_AUTHED_QUOTA = 50_000
+const DEFAULT_AUTHED_QUOTA = 50_0
```
Alllex202, line 4: ?

apps/condo/pages/_app.tsx
```diff
@@ -612,7 +614,23 @@ if (!isDisabledSsr || !isSSR()) {
...
+
+            if (tooManyRequests) {
+                let timestamp = 0
+                error?.cause?.result?.errors?.forEach((error) => {
+                    if (error?.extensions?.reset && timestamp < error?.extensions?.reset) {
+                        timestamp = error?.extensions?.reset
+                    }
+                })
+
+                return { pageProps: { statusCode: 429, resetTime: new Date(timestamp * 1000).toLocaleTimeString(dayjs.locale()) } }
```
nomerdvadcatpyat, line 631: what is locale used here?

packages/keystone/rateLimiting/plugin.js
```diff
@@ -253,6 +253,7 @@ class ApolloRateLimitingPlugin {
                         type: ERROR_TYPE,
                         message: 'You\'ve made too many requests recently, try again later',
                         messageForUser: `api.global.rateLimit.${ERROR_TYPE}`,
+                        reset: resetTimestampInSec,
```
Alllex202, line 256: better to use `messageInterpolation` field. Check in other errors

packages/keystone/rateLimiting/constants.js
toplenboren, line 4: ???

nomerdvadcatpyat (approved):

apps/condo/lang/ru/ru.json
```diff
@@ -1767,8 +1767,11 @@
   "pages.condo.employeeRole.update.title": "Редактировать роль",
   "pages.condo.error.Description": "Проводим технические работы. Зайдите, пожалуйста, немного позже.",
   "pages.condo.error.NotFoundDescription": "Страница удалена или её никогда не существовало",
-  "pages.condo.error.NotFoundTitle": "Страница не найдена",
+  "pages.condo.error.NotFoundTitle": "Страница не найдена",
   "pages.condo.error.PageTitle": "Улучшаем сервис",
+  "pages.condo.error.TooManyRequestsTitle": "Скоро доступ восстановится",
+  "pages.condo.error.TooManyRequestsDescription": "С вашего IP-адреса поступает слишком много запросов, поэтому мы ограничили доступ. Он автоматически восстановится в {resetTime}. Пожалуйста, подождите.",
```
Alllex202, line 1773: In fact, the restriction may not be by IP.
We can leave it like this, the main thing is that this text does not mislead us

apps/condo/pages/_app.tsx
```diff
@@ -612,7 +614,23 @@ if (!isDisabledSsr || !isSSR()) {
...
+            if (tooManyRequests) {
+                const timestamp = errors.reduce((max, err) => {
+                    const resetTime =  err?.extensions?.messageInterpolation?.resetTime
+                    if (resetTime > max) return resetTime
+                    return max
+                }, 0)
+
+                const date = new Date(timestamp * 1000)
+
+                return { pageProps: { statusCode: 429, resetTime: `${date.getHours()}:${date.getMinutes()}` } }
```
Alllex202, line 631: Will this work well if the server timezone is different from the client timezone?

apps/condo/lang/ru/ru.json
tolmachev21: Saw this, however
1) This message will be clearer for the user
2) This screen will not be shown on the authorization page. Therefore, the screen will only be shown when the limit has ended for the user

apps/condo/pages/_app.tsx
tolmachev21: The date object is created on the client side, so the timezone must be defined by the client.

Alllex202, line 631: How about doing these date conversions directly on the page? That way the correct locale should be taken from `useIntl` (If I'm not mistaken), and `dayjs` should have the correct locale

tolmachev21: Hmm, okay, everything will be ready now)

tolmachev21 (commented):
Only need update pointer in cc

Alllex202 (approved):

PR #5809, opened 2025-02-18
hotfix(condo): fixed phone normalization

Alllex202:


toplenboren (approved):

Alexander-Turkin (approved):

apps/condo/domains/common/utils/phone.js
```diff
@@ -1,8 +1,12 @@
 const { phone } = require('phone')
 
+
+const STRICT_PHONE_REGEX = /^\+[0-9\s()-]+$/
```
SavelevMatthew, line 4: What about spaces?

Alllex202: what do you mean? Spaces are used for formatting, so we allow them, there are tests for cases with spaces

PR #5810, opened 2025-02-18
fix(condo): DOMA-11038 add read func for cache

tolmachev21:
This fix should solve the problem of broken links in the cache

In some cases this caused a problem. For example, the story of comments disappearing after changing the page (not after updating):

If you are interested in how the bug is reproduced now:
https://github.com/user-attachments/assets/c2b650c4-40d4-4c11-98d9-e76c6b10fee3

### DIFF
--- a/packages/apollo/src/utils/lists.ts
+++ b/packages/apollo/src/utils/lists.ts
@@ -86,18 +86,18 @@ export class ListHelper {
             return undefined
         }
 
-        const currentList = existing.slice(skip, skip + first)
+        const currentList = existing.slice(skip, skip + first).filter(Boolean)
 
         const cache = options?.cache.extract()
 
-        const hasBrokenLinks = currentList.filter(Boolean).some((link) => {
+        const hasBrokenLinks = currentList.some((link) => {
             const ref = (link as Reference)?.__ref
             return ref && cache && !(ref in cache)
         })
 
         if (hasBrokenLinks) return undefined
 
-        return currentList.filter(Boolean)
+        return currentList
     }
 
     /**
--- a/packages/apollo/src/utils/lists.ts
+++ b/packages/apollo/src/utils/lists.ts
@@ -1,5 +1,7 @@
 import bindAll from 'lodash/bindAll'
 
+import { nonNull } from '@open-condo/miniapp-utils'
+
 import type { InitCacheOptions } from './cache'
 import type { FieldReadFunction, FieldFunctionOptions, Reference } from '@apollo/client'
 import type { NormalizedCacheObject } from '@apollo/client/core'
@@ -25,15 +27,15 @@ export type ClientPaginationBehaviour = 'paginate' | 'showAll'
  * Checks if the provided value is an Apollo Client Reference object
  * Used to determine if a value in the cache is a reference to another cached object
  * 
- * @param link - Value to check for Reference type
+ * @param ref - Value to check for Reference type
  * @returns True if the value is an Apollo Reference object, false otherwise
  */
-function hasRef (link: unknown): link is Reference {
+function hasRef (ref: unknown): ref is Reference {
     return (
-        typeof link === 'object' &&
-        link !== null &&
-        '__ref' in link &&
-        typeof link.__ref === 'string'
+        typeof ref === 'object' &&
+        ref !== null &&
+        '__ref' in ref &&
+        typeof ref.__ref === 'string'
     )
 
 }
@@ -45,16 +47,16 @@ function hasRef (link: unknown): link is Reference {
  * 
  * @param list - Array of potentially cached items to check
  * @param cache - Apollo normalized cache object
- * @returns True if the list contains any broken references or non-reference items, false if all references are valid
+ * @returns True if the list contains any broken references, false if all references are valid or if list contain non-reference items
  */
-function hasBrokenLinks (
+function hasBrokenRefs (
     list: ReadonlyArray<unknown>,
     cache: NormalizedCacheObject,
 ) {
-    return list.some((link) => {
-        if (hasRef(link)) {
-            const ref: string = link.__ref
-            return ref && !(ref in cache)
+    return list.some((ref) => {
+        if (hasRef(ref)) {
+            const cacheIdentifier = ref.__ref
+            return cacheIdentifier && !(cacheIdentifier in cache)
         }
         return false
     })
@@ -106,7 +108,7 @@ export class ListHelper {
             this.skipCacheOnRead = options.cacheOptions.skipOnRead
         }
 
-        bindAll(this, '_readPage', '_readList', '_networkOnlyRead', 'getReadFunction', 'mergeLists')
+        bindAll(this, '_readPage', '_readAll', '_networkOnlyRead', 'getReadFunction', 'mergeLists')
     }
 
     /**
@@ -128,11 +130,11 @@ export class ListHelper {
             return undefined
         }
 
-        const filteredPage = existing.slice(skip, skip + first).filter(Boolean)
+        const filteredPage = existing.slice(skip, skip + first).filter(nonNull)
 
-        const cache: NormalizedCacheObject = options.cache.extract()
+        const cache = options.cache.extract()
 
-        if (!cache || hasBrokenLinks(filteredPage, cache)) return undefined
+        if (hasBrokenRefs(filteredPage, cache)) return undefined
 
         return filteredPage
     }
@@ -150,10 +152,10 @@ export class ListHelper {
             return undefined
         }
 
-        const cache: NormalizedCacheObject = options.cache.extract()
+        const cache = options.cache.extract()
 
-        const filteredList = existing.filter(Boolean)
-        if (!cache || hasBrokenLinks(filteredList, cache)) {
+        const filteredList = existing.filter(nonNull)
+        if (hasBrokenRefs(filteredList, cache)) {
             return undefined
         }
 
@@ -170,7 +172,7 @@ export class ListHelper {
     /**
      * Selects the appropriate read function depending on cache options and desired pagination behaviour
      */
-    getReadFunction (clientPagination: ClientPaginationBehaviour): FieldReadFunction | undefined {
+    getReadFunction (clientPagination: ClientPaginationBehaviour): FieldReadFunction {
         if (this.skipCacheOnRead) {
             return this._networkOnlyRead
         }
@@ -179,11 +181,7 @@ export class ListHelper {
             return this._readPage
         }
 
-        if (clientPagination === 'showAll') {
-            return this._readAll
-        }
-
-        return undefined
+        return this._readAll
     }
 
     /**
--- a/packages/apollo/src/utils/lists.ts
+++ b/packages/apollo/src/utils/lists.ts
@@ -32,9 +32,10 @@ function hasRef (link: unknown): link is Reference {
     return (
         typeof link === 'object' &&
         link !== null &&
-        Object.prototype.hasOwnProperty.call(link, '__ref') &&
-        typeof (link as any).__ref === 'string'
+        '__ref' in link &&
+        typeof link.__ref === 'string'
     )
+
 }
 
 /**
@@ -47,13 +48,13 @@ function hasRef (link: unknown): link is Reference {
  * @returns True if the list contains any broken references or non-reference items, false if all references are valid
  */
 function hasBrokenLinks (
-    list: unknown[],
+    list: ReadonlyArray<unknown>,
     cache: NormalizedCacheObject,
 ) {
     return list.some((link) => {
-        if (hasRef(link) && typeof cache === 'object' && cache !== null) {
+        if (hasRef(link)) {
             const ref: string = link.__ref
-            return ref !== '' && !(ref in cache)
+            return ref && !(ref in cache)
         }
         return false
     })
@@ -105,7 +106,7 @@ export class ListHelper {
             this.skipCacheOnRead = options.cacheOptions.skipOnRead
         }
 
-        bindAll(this, '_readSlicePage', '_readAllList', '_networkOnlyRead', 'getReadFunction', 'mergeLists')
+        bindAll(this, '_readPage', '_readList', '_networkOnlyRead', 'getReadFunction', 'mergeLists')
     }
 
     /**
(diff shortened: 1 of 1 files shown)


packages/apollo/src/utils/lists.ts
```diff
@@ -86,7 +86,19 @@ export class ListHelper {
             return undefined
         }
 
-        return existing.slice(skip, skip + first)
+        const currentList = existing.slice(skip, skip + first)
+
+        const cache = options?.cache.extract()
+
+        const hasBrokenLinks = currentList.filter(Boolean).map((link) => {
+            const ref = get(link, ['__ref'], null)
```
Alllex202, line 94: don't use `get`

packages/apollo/src/utils/lists.ts
```diff
@@ -86,7 +86,19 @@ export class ListHelper {
...
+            if (cache && ref && !cache[ref]) return null
```
Alllex202, line 95: should we check for the presence of the key or just the value?
```suggestion
            if (cache && ref && !(ref in cache)) return null
```

packages/apollo/src/utils/lists.ts
```diff
@@ -86,7 +86,19 @@ export class ListHelper {
             return undefined
         }
 
-        return existing.slice(skip, skip + first)
+        const currentList = existing.slice(skip, skip + first)
+
+        const cache = options?.cache.extract()
+
+        const hasBrokenLinks = currentList.filter(Boolean).map((link) => {
```
Alllex202, line 93: should we check broken links only for the specified page (skip, first) or for full list?

packages/apollo/src/utils/lists.ts
```diff
@@ -86,7 +86,18 @@ export class ListHelper {
...
+        const cache = options?.cache.extract()
+
+        const hasBrokenLinks = currentList.filter(Boolean).some((link) => {
+            const ref = (link as Reference)?.__ref
+            return ref && cache && !(ref in cache)
+        })
+
+        if (hasBrokenLinks) return undefined
+
+        return currentList.filter(Boolean)
```
Alllex202, line 100: You changed the return values ​​of the function, is that ok? Before everything was returned, but now only what is non null

tolmachev21: Everything should be fine, because we usually filter out zero values ​​on the frontend anyway

Like this
Screenshot 2025-02-19 at 19 02 49

packages/apollo/src/utils/lists.ts
```diff
@@ -86,7 +86,18 @@ export class ListHelper {
             return undefined
         }
 
-        return existing.slice(skip, skip + first)
+        const currentList = existing.slice(skip, skip + first).filter(Boolean)
+
+        const cache = options?.cache.extract()
+
+        const hasBrokenLinks = currentList.some((link) => {
```
toplenboren, line 93: You need to add a big explanation to this code, as it is a bit uncliear from the get go

toplenboren (approved):
This looks okay-ish, but better to consult with somewone who has the cache context in head :-)

packages/apollo/src/utils/lists.ts
```diff
@@ -86,7 +86,18 @@ export class ListHelper {
...
+            const ref = (link as Reference)?.__ref
```
SavelevMatthew, line 94: It's better to avoid type casting here if it has `unknown` type...

SavelevMatthew: I would prefer something like

if (typeof item === 'object' && item !== null && item.hasOwnProperty('__ref') and typeof item.__ref === 'string')

packages/apollo/src/utils/lists.ts
```diff
@@ -21,6 +21,44 @@ export type ListHelperOptions = {
...
+ * @param list - Array of potentially cached items to check
+ * @param cache - Apollo normalized cache object
+ * @returns True if the list contains any broken references or non-reference items, false if all references are valid
+ */
+function hasBrokenLinks (
+    list: unknown[],
+    cache: NormalizedCacheObject,
+) {
+    return list.some((link) => {
+        if (hasRef(link) && typeof cache === 'object') {
```
Alllex202, line 54: if `cache = null` then `typeof cache` also return `object` :)

Alllex202: Or are we always sure that it is not empty?

packages/apollo/src/utils/lists.ts
```diff
@@ -1,8 +1,8 @@
 import bindAll from 'lodash/bindAll'
-import get from 'lodash/get'
```
Alllex202, line 2: 🏅

packages/apollo/src/utils/lists.ts
```diff
@@ -67,26 +105,58 @@ export class ListHelper {
...
+        existing: ReadonlyArray<TData> | undefined,
+        options: TOptions
+    ): Array<TData> | undefined {
+        if (!existing) {
+            return undefined
+        }
+
+        const cache: NormalizedCacheObject | undefined = options?.cache?.extract?.()
+
+        const list = existing?.filter(Boolean)
```
Alllex202, line 154: You have already checked above that the list exists

```suggestion
        const list = existing.filter(Boolean)
```

Alllex202 (approved):

packages/apollo/src/utils/lists.ts
tolmachev21: Yes, there is a check before calling the function that the object is not empty. However, it is worth adding a check so that such questions do not arise

packages/apollo/src/utils/lists.ts
```diff
@@ -21,6 +21,44 @@ export type ListHelperOptions = {
...
+ * Used to determine if a value in the cache is a reference to another cached object
+ * 
+ * @param link - Value to check for Reference type
+ * @returns True if the value is an Apollo Reference object, false otherwise
+ */
+function hasRef (link: unknown): link is Reference {
+    return (
+        typeof link === 'object' &&
+        link !== null &&
+        Object.prototype.hasOwnProperty.call(link, '__ref') &&
```
SavelevMatthew, line 35: this can be replaced by `'__ref' in link && typeof link.__ref === 'string'`

packages/apollo/src/utils/lists.ts
```diff
@@ -21,6 +21,44 @@ export type ListHelperOptions = {
...
+        typeof (link as any).__ref === 'string'
```
SavelevMatthew, line 36:
```suggestion
        '__ref' in link &&
        typeof link.__ref === 'string'
```

packages/apollo/src/utils/lists.ts
```diff
@@ -21,6 +21,44 @@ export type ListHelperOptions = {
...
+ * Validates if a list of cached items contains any broken references
+ * A broken reference occurs when a Reference object points to a cache entry that doesn't exist
+ * This is used to ensure data integrity when reading from the Apollo cache
+ * 
+ * @param list - Array of potentially cached items to check
+ * @param cache - Apollo normalized cache object
+ * @returns True if the list contains any broken references or non-reference items, false if all references are valid
+ */
+function hasBrokenLinks (
+    list: unknown[],
```
SavelevMatthew, line 50: We prefer `Array` syntax...

SavelevMatthew:
```suggestion
    list: ReadonlyArray<unknown>,
```

packages/apollo/src/utils/lists.ts
```diff
@@ -67,26 +105,58 @@ export class ListHelper {
             this.skipCacheOnRead = options.cacheOptions.skipOnRead
         }
 
-        bindAll(this, '_readPage', '_networkOnlyRead', 'getReadFunction', 'mergeLists')
+        bindAll(this, '_readSlicePage', '_readAllList', '_networkOnlyRead', 'getReadFunction', 'mergeLists')
```
SavelevMatthew, line 108: Why do you need to rename?

`_readPage` and `_readAll` is simple and good enough I guess

SavelevMatthew: page and slice are basically the same word. So it's like `_readPagePage` or `_readSliceSlice`...

I prefer page imho

packages/apollo/src/utils/lists.ts
```diff
@@ -67,26 +105,58 @@ export class ListHelper {
...
         // NOTE: partial data (skip only) is not enough for case
         // where you fetch (page + 1) item to show "Next page" pagination event
         if (!existing || existing.length < skip + first) {
             return undefined
         }
 
-        return existing.slice(skip, skip + first)
+        const sliceList = existing.slice(skip, skip + first).filter(Boolean)
+
+        const cache: NormalizedCacheObject | undefined = options?.cache?.extract?.()
```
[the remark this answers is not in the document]
tolmachev21: Well, there are some types, I think. But I wanted to be on the safe side)

packages/apollo/src/utils/lists.ts
```diff
@@ -67,26 +105,58 @@ export class ListHelper {
...
+        existing: ReadonlyArray<TData> | undefined,
+        options: TOptions
+    ): Array<TData> | undefined {
+        if (!existing) {
+            return undefined
+        }
+
+        const cache: NormalizedCacheObject | undefined = options?.cache?.extract?.()
+
+        const list = existing.filter(Boolean)
```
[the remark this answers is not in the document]
tolmachev21: Maybe 
- filteredPage for _readPage
- filteredList from _readAll ? 

"Page" is a part of a list, right?

packages/apollo/src/utils/lists.ts
```diff
@@ -21,6 +21,45 @@ export type ListHelperOptions = {
...
+ * @param cache - Apollo normalized cache object
+ * @returns True if the list contains any broken references or non-reference items, false if all references are valid
+ */
+function hasBrokenLinks (
+    list: ReadonlyArray<unknown>,
+    cache: NormalizedCacheObject,
+) {
+    return list.some((link) => {
+        if (hasRef(link)) {
+            const ref: string = link.__ref
```
SavelevMatthew, line 56: This string assertion can be deleted I guess, since you assert it be hasRef 
image

packages/apollo/src/utils/lists.ts
```diff
@@ -21,6 +21,45 @@ export type ListHelperOptions = {
...
+}
+
+/**
+ * Validates if a list of cached items contains any broken references
+ * A broken reference occurs when a Reference object points to a cache entry that doesn't exist
+ * This is used to ensure data integrity when reading from the Apollo cache
+ * 
+ * @param list - Array of potentially cached items to check
+ * @param cache - Apollo normalized cache object
+ * @returns True if the list contains any broken references or non-reference items, false if all references are valid
```
SavelevMatthew, line 48: Why this is default behaviour?

SavelevMatthew: If I have link `and` its not in cache (expired) => we need to refetch
If I have no link (`null` or some other values), then its is a valid list, isn't it?

packages/apollo/src/utils/lists.ts
```diff
@@ -67,26 +106,58 @@ export class ListHelper {
             this.skipCacheOnRead = options.cacheOptions.skipOnRead
         }
 
-        bindAll(this, '_readPage', '_networkOnlyRead', 'getReadFunction', 'mergeLists')
+        bindAll(this, '_readPage', '_readList', '_networkOnlyRead', 'getReadFunction', 'mergeLists')
```
SavelevMatthew, line 109: function below is called `_readAll`... Here's some leftover I guess

packages/apollo/src/utils/lists.ts
```diff
@@ -21,6 +21,45 @@ export type ListHelperOptions = {
...
+/**
+ * Validates if a list of cached items contains any broken references
+ * A broken reference occurs when a Reference object points to a cache entry that doesn't exist
+ * This is used to ensure data integrity when reading from the Apollo cache
+ * 
+ * @param list - Array of potentially cached items to check
+ * @param cache - Apollo normalized cache object
+ * @returns True if the list contains any broken references or non-reference items, false if all references are valid
+ */
+function hasBrokenLinks (
```
SavelevMatthew, line 50: I guess we should can it `refs`, not `links`, since this is commonly accepted  apollo term

packages/apollo/src/utils/lists.ts
```diff
@@ -67,26 +106,58 @@ export class ListHelper {
...
+        const first: number = options?.args?.[this.firstArgName] || 0
 
         // NOTE: partial data (skip only) is not enough for case
         // where you fetch (page + 1) item to show "Next page" pagination event
         if (!existing || existing.length < skip + first) {
             return undefined
         }
 
-        return existing.slice(skip, skip + first)
+        const filteredPage = existing.slice(skip, skip + first).filter(Boolean)
```
SavelevMatthew, line 131: We should filter only `null` / `undefined`, so if your return type is [Number], you can filter out zeros...

SavelevMatthew: It's better be more specific and use `nonNull` here:
```
import { nonNull } from '@open-condo/miniapp-utils/helpers/collections'
```

packages/apollo/src/utils/lists.ts
```diff
@@ -67,26 +106,58 @@ export class ListHelper {
...
+
+        const cache: NormalizedCacheObject = options.cache.extract()
+
+        if (!cache || hasBrokenLinks(filteredPage, cache)) return undefined
```
SavelevMatthew, line 135: It's cache guaranteed here?

image

packages/apollo/src/utils/lists.ts
```diff
@@ -67,26 +106,58 @@ export class ListHelper {
...
+        existing: ReadonlyArray<TData> | undefined,
+        options: TOptions
+    ): Array<TData> | undefined {
+        if (!existing) {
+            return undefined
+        }
+
+        const cache: NormalizedCacheObject = options.cache.extract()
+
+        const filteredList = existing.filter(Boolean)
```
SavelevMatthew, line 155: Same about `notNull` and `!cache`

packages/apollo/src/utils/lists.ts
```diff
@@ -108,6 +179,10 @@ export class ListHelper {
             return this._readPage
         }
 
+        if (clientPagination === 'showAll') {
```
SavelevMatthew, line 182: Remove `undefined` from return-type and return _readAll as last option (without if)

packages/apollo/src/utils/lists.ts
tolmachev21: Oops, you're right

packages/apollo/src/utils/lists.ts
tolmachev21: Method "extract" - guaranteed
Property "cache" - guaranteed
Anything else?

SavelevMatthew (approved):
🏅

PR #5812, opened 2025-02-19
feat(condo): INFRA-793 valkey cluster

sitozzz:


.github/workflows/_nodejs.condo.core.tests.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         env:
           REGISTRY: ${{ secrets.DOCKER_REGISTRY }}/doma/utils/
         run: |
-          docker-compose up -d redis postgresdb
+          docker-compose up -d valkey-cluster postgresdb
```
SavelevMatthew, line 46: How about adding some aliases? Like `docker compose up -d dbs`?

SavelevMatthew: Btw, dash syntax is deprecated....Use `docker compose` instead of `docker-compose`

bin/rename-redis-keys.js
SavelevMatthew, line 1: this one is outdated I guess, since migrator is already exists

packages/keystone/kv.js
```diff
@@ -63,7 +63,12 @@ function getKVClient (name = 'default', purpose = 'regular', opts = { kvOptions:
     if (!KV_CLIENTS[clientKey]) {
         const kvEnvName = `${name.toLowerCase()}_KV_URL`
         const redisEnvName = `${name.toUpperCase()}_REDIS_URL`
-        const kvUrl = conf[kvEnvName] || conf[redisEnvName] || conf.KV_URL || conf.REDIS_URL
+        let kvUrl = conf[kvEnvName] || conf[redisEnvName] || conf.KV_URL || conf.REDIS_URL
+        try {
+            kvUrl = JSON.parse(kvUrl)
+        } catch (err) {
+            // skip that error
```
SavelevMatthew, line 70: Add more detailed description, its a bit hard to understand whats going on without context of PR

valkey/valkey.conf
```diff
@@ -0,0 +1,4 @@
+cluster-enabled yes
+cluster-config-file nodes.conf
```
SavelevMatthew, line 2: What is nodes.conf? Should it be in repo?

SavelevMatthew: Or maybe we can inline 1 line into compose config?

bin/prepare.js
SavelevMatthew, line 1: We'll need to discuss format of prepare....It's becoming more and more complicated 🤔

sitozzz: I agree with this

valkey/valkey.conf
sitozzz: Here's a way to make it all work

sitozzz: Actually valkey.conf provides minimal configuration to make cluster up and running

bin/rename-redis-keys.js
sitozzz: you are right, it has already been fixed

SavelevMatthew (approved):

PR #5814, opened 2025-02-19
fix(condo): DOMA-11008 set reading date on server if no passed

abshnko:


apps/condo/domains/meter/schema/MeterReading.js
```diff
@@ -54,8 +54,16 @@ const MeterReading = new GQLListSchema('MeterReading', {
...
             hooks: {
+                resolveInput: async ({ resolvedData, fieldPath }) => {
+                    const date = resolvedData[fieldPath]
+                    if (!date) {
+                        return dayjs().toISOString()
+                    }
+
+                    return date
+                },
                 validateInput: async ({ context, operation, existingItem, resolvedData, fieldPath }) => {
```
YEgorLu, line 65: Is there any reason "date" was not required before? 
In RegisterMetersReading you must pass "date" to create reading, which sounds right logically. So if no "date" is provided let's think that "date" is now. But there is no "isRequired", do you know why is that so?

YEgorLu (approved):

apps/condo/domains/meter/schema/MeterReading.js
abshnko: that seemed weird to me, too. logically we should always require date and when the model was first created date field was required. i'll look into it

abshnko: update: the date is set in common resolveInput, it is not required, but will always be present, so
i removed the additional resolveInput hook

vovaaxeapolla (approved):

AleX83Xpert (approved):

PR #5857, opened 2025-02-27
fix(condo): DOMA-10800 fix resident to property connections when there are more than 100 residents at the address

nomerdvadcatpyat:
`getAll` only retrieves the first 100 items, so I rewrote it using `find` and added a test.

### DIFF
--- a/apps/condo/domains/resident/schema/RegisterResidentService.test.js
+++ b/apps/condo/domains/resident/schema/RegisterResidentService.test.js
@@ -10,16 +10,6 @@ const {
     expectToThrowAuthenticationErrorToResult, expectToThrowAccessDeniedErrorToResult,
 } = require('@open-condo/keystone/test.utils')
 
-const { CONTEXT_FINISHED_STATUS } = require('@condo/domains/acquiring/constants/context')
-const {
-    createTestAcquiringIntegration,
-    createTestAcquiringIntegrationContext,
-} = require('@condo/domains/acquiring/utils/testSchema')
-const {
-    createTestBillingAccount,
-    makeContextWithOrganizationAndIntegrationAsAdmin,
-    createTestBillingProperty,
-} = require('@condo/domains/billing/utils/testSchema')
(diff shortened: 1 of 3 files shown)


apps/condo/domains/resident/schema/RegisterResidentService.test.js
```diff
@@ -219,6 +209,55 @@ describe('manageResidentToPropertyAndOrganizationConnections worker task tests',
             expect(anotherResident.property.id).toEqual(olderProperty.id)
         })
     })
+
+    test('Must relink many residents after approve new properties', async () => {
```
nomerdvadcatpyat, line 213: Do we really need this test? Locally, it takes about a minute to run, sometimes even longer, due to the creation of 101 residents

toplenboren: Yea... I agree

toplenboren (approved):

apps/condo/domains/resident/schema/RegisterResidentService.test.js
```diff
@@ -219,6 +209,55 @@ describe('manageResidentToPropertyAndOrganizationConnections worker task tests',
...
+                return resident.id
+            })
+        )
+
+        // NOTE: Should connect to older property
+        await waitFor(async () => {
+            const residents = await Resident.getAll(admin, { id_in: residentIds }, { first: residentsInProperty })
+
+            for (const resident of residents) {
+                expect(resident.organization?.id).toEqual(organization.id)
```
tolmachev21, line 242: Just a question: why is ?. - required in these places - that Resident.getAll is returned? I thought that subqueries are made there, so there should be id inside, no?

tolmachev21 (approved):

apps/condo/domains/resident/schema/RegisterResidentService.test.js
nomerdvadcatpyat: Because organization may be undefined here

PR #5858, opened 2025-02-27
ci(global): INFRA-902 add test collector to condo / KS

SavelevMatthew:


.github/workflows/nodejs.condo.ci.yml
```diff
@@ -6,7 +6,8 @@ on:
       - main
 
   # NOTE: change pull_request_target to pull_request if you want to change the CI in your branch
-  pull_request_target:
+#  pull_request_target:
```
toplenboren, line 9: is it just so that tests are collected?

toplenboren (approved):

.github/workflows/nodejs.condo.ci.yml
SavelevMatthew: To test collection logic, since dispatching worklflows are also inherit target behaviour...And I need to change it

I'll rollback this as soon as I test it

PR #5868, opened 2025-03-02
feat(resident-app): INFRA-904 abstract analytics with groups and providers 📊

SavelevMatthew:


packages/miniapp-utils/src/helpers/analytics/types.ts
```diff
@@ -0,0 +1,24 @@
+import type { AnalyticsInstance, AnalyticsPlugin } from 'analytics'
+
+export type AnalyticsConfig = {
+    app?: string
+    version?: string | number
```
Alllex202, line 5: what is version used for? is it the version of the app?

packages/ui/src/events.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Checks if analytics parameters are valid
+ */
+export function isValidAnalyticsParams (params: unknown): params is AnalyticsParams {
+    const { success, data } = analyticsEventSchema.safeParse(params)
+    if (!success) {
+        return false
+    }
+
+    // NOTE: prevent overloaded messages
+    return Object.keys(data).length < 10
```
Alllex202, line 34: let's write something to the console if the parameters are invalid or there are too many messages

Alllex202 (approved):

nomerdvadcatpyat (approved):

packages/miniapp-utils/src/helpers/analytics/types.ts
SavelevMatthew: Yep, that's apps version:
https://getanalytics.io/api/#configuration

packages/ui/src/events.ts
SavelevMatthew: Oke

PR #5873, opened 2025-03-04
fix(condo): DOMA-11189 added rows count limit exceeded error in meter readings import

nomerdvadcatpyat:
Screenshot 2025-03-04 at 20 11 44

### DIFF
--- a/apps/condo/domains/meter/tasks/importMeters.js
+++ b/apps/condo/domains/meter/tasks/importMeters.js
@@ -4,7 +4,7 @@ const Upload = require('graphql-upload/Upload.js')
 const { get, isEmpty } = require('lodash')
 
 const { ConvertFileToTable, getObjectStream, readFileFromStream } = require('@open-condo/keystone/file')
-const { ROWS_COUNT_LIMIT, ROWS_COUNT_LIMIT_EXCEEDED } = require('@open-condo/keystone/file/utils')
+const { ROWS_COUNT_LIMIT, ROWS_COUNT_LIMIT_EXCEEDED } = require('@open-condo/keystone/file/constants')
 const FileAdapter = require('@open-condo/keystone/fileAdapter/fileAdapter')
 const { getSchemaCtx } = require('@open-condo/keystone/schema')
 const { i18n } = require('@open-condo/locales/loader')
--- a/apps/condo/domains/meter/tasks/importMeters.js
+++ b/apps/condo/domains/meter/tasks/importMeters.js
@@ -4,8 +4,10 @@ const Upload = require('graphql-upload/Upload.js')
 const { get, isEmpty } = require('lodash')
 
 const { ConvertFileToTable, getObjectStream, readFileFromStream } = require('@open-condo/keystone/file')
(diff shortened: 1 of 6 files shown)


packages/keystone/file/utils.js
```diff
@@ -11,6 +11,8 @@ const FORCE_ENCODING_CHANGE = {
     'KOI8-R': 'WINDOWS-1251', // TODO: Find why KOI8 detected on WINDOWS-1251
     'ISO-8859-2': 'WINDOWS-1251',
 }
+const ROWS_COUNT_LIMIT = 100_000
+const ROWS_COUNT_LIMIT_EXCEEDED = 'Rows count limit exceeded'
```
AleX83Xpert, line 15: Please move these constants to a separate file. Otherwise, we have to load all packages from utils.js when we need only two constants.

packages/keystone/file/file-types/excel.js
```diff
@@ -45,6 +48,10 @@ class ExcelParser {
         workbook.SheetNames.forEach((sheetName) => {
             const worksheet = workbook.Sheets[sheetName]
             const range = excel.utils.decode_range(worksheet['!ref'])
+            if (range.e.r > this.maxRowsCount) {
```
AleX83Xpert, line 51: What if `maxRowsCount` is undefined?

apps/condo/domains/meter/tasks/importMeters.js
```diff
@@ -146,8 +148,18 @@ async function importMeters (taskId) {
             await failWithErrorFile(context, taskId, errorFileContent, format)
         }
     } catch (err) {
-        const errorMessage = get(err, 'errors[0].extensions.messageForUser', get(err, 'message'))
-            || 'not recognized error'
+        let errorMessage = get(err, 'errors[0].extensions.messageForUser', get(err, 'message'))
+
+        if (errorMessage === ROWS_COUNT_LIMIT_EXCEEDED) {
```
AleX83Xpert, line 153: Could you please add a test for this error case?

packages/keystone/file/file-types/excel.js
nomerdvadcatpyat: I think this means there shouldn't be a limit. I'll add a check for the existence of `this.maxRowsCount`

apps/condo/domains/meter/tasks/importMeters.js
nomerdvadcatpyat: Added

toplenboren (approved):

packages/keystone/file/ConvertFileToTable.js
```diff
@@ -1,7 +1,7 @@
 const jschardet = require('jschardet')
 
+const { clearString, encode, detectEncoding, convertEncoding, ROWS_COUNT_LIMIT_EXCEEDED } = require('./constants')
```
AleX83Xpert, line 3: But functions still should be imported from utils. Isn't it?

AleX83Xpert (approved):

packages/keystone/file/ConvertFileToTable.js
nomerdvadcatpyat: oops :)
sure

PR #5876, opened 2025-03-05
fix(condo): DOMA-11188 display files attached message in ticket comment created notification

nomerdvadcatpyat:
We decided to display "(files attached)" message in notifications about ticket comment created. This means that the comment was created without content but with attached files (yes, through the API, we can create a comment without content or files, but we agreed on this solution).
Also updated the templates of TICKET_COMMENT_CREATED messages for cases when the notification data does not contain `authorName` (these are old comment created notifications).

### DIFF
--- a/apps/condo/domains/ticket/tasks/sendTicketCommentCreatedNotifications.js
+++ b/apps/condo/domains/ticket/tasks/sendTicketCommentCreatedNotifications.js
@@ -96,6 +96,7 @@ const sendTicketCommentCreatedNotifications = async (commentId, ticketId) => {
         const OpenTicketMessage = i18n(`notification.messages.${TICKET_COMMENT_CREATED_TYPE}.telegram.openTicket`, { locale: lang })
         const TicketStatusName = i18n(`ticket.status.${ticketStatus.type}.name`, { locale: lang })
         const TicketUnitType = i18n(`field.UnitType.prefix.${ticket.unitType}`, { locale: lang }).toLowerCase()
+        const FilesAttachedMessage = i18n(`notification.messages.${TICKET_COMMENT_CREATED_TYPE}.filesAttached`, { locale: lang })
         const CommentTypeMessage = getCommentTypeMessage(commentType, commentAuthorType, lang)
         const authorType = getCommentAuthorRoleMessage(commentAuthor, lang)
         const ticketClassifier = buildFullClassifierName(classifier)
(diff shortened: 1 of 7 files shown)


apps/condo/domains/ticket/tasks/sendTicketCommentCreatedNotifications.js
```diff
@@ -120,7 +121,7 @@ const sendTicketCommentCreatedNotifications = async (commentId, ticketId) => {
                         organizationId: organization.id,
                         organizationName: organization.name,
                         commentId,
-                        commentContent: createdComment.content || EMPTY_CONTENT,
+                        commentContent: createdComment.content || FilesAttachedMessage,
```
tolmachev21, line 124: Is it correct that there can't be an empty line or lines filled with spaces when sending only an attached file?

apps/condo/lang/en/messages/TICKET_COMMENT_CREATED/default.njk
```diff
@@ -1 +1,5 @@
-{{ message.meta.data.authorName }} ({{ message.meta.data.authorType }}): {{ message.meta.data.commentContent }}
\ No newline at end of file
+{%- if message.meta.data.authorName -%}
+{{ message.meta.data.authorName }} ({{ message.meta.data.authorType }}): {{ message.meta.data.commentContent }}
+{%- else -%}
```
tolmachev21, line 3: And if we have an attached file, the name of the author who attached it will not exist?

tolmachev21 (approved):

apps/condo/lang/en/messages/TICKET_COMMENT_CREATED/default.njk
tolmachev21: Uuups, Im read for this needed

apps/condo/domains/ticket/tasks/sendTicketCommentCreatedNotifications.js
tolmachev21: Simple:
In the first case, there may be a comment without attached files, but with content in the message that they are attached - and this is ok for us
In the second case, we will have a comment with an attached file, but with content in the message that '       ' - and it seems that it is not ok

Although the chance is small (considering the checks at the front)

nomerdvadcatpyat: If we send a comment with an empty content (or only with spaces), then this will be false and `FilesAttachedMessage` will be written in `commentContent`

Alllex202 (approved):

PR #5902, opened 2025-03-11
refactor(condo): INFRA-917 remove tracking context (Part 2)

SavelevMatthew:


sitozzz (approved):
looks okay

apps/condo/domains/common/components/Table/Filters.tsx
```diff
@@ -191,17 +186,6 @@ export const getOptionFilterDropdown: GetOptionFilterDropdownType = ({ container
...
-                logEvent({
-                    eventName,
-                    eventProperties: {
-                        component: {
-                            id: checkboxGroupProps.id,
-                            value: selectedValue,
-                        },
-                    },
-                })
-            }
```
sitozzz, line 204: Is something planned instead?

apps/condo/domains/common/components/Tabs.tsx
```diff
@@ -70,27 +70,35 @@ const TabWrapper = styled.div`
...
     showIcon = true,
     icon = <StarIcon/>,
-    eventName: propEventName,
     onClick,
 }) => {
-    const { getTrackingWrappedCallback, getEventName } = useTracking()
+    const onClickWrapped: React.MouseEventHandler<HTMLDivElement> = useCallback((e) => {
+        analytics.track('change', {
+            component: 'Tabs',
+            location: window.location.href,
```
sitozzz, line 91: Is it possible to transfer the page address automatically?

sitozzz: It's kinda basic info, that could be attached to the event. So for me it's strange that dev needs to pass location every time via constant

SavelevMatthew: It's legacy version of components here. 

All ui-kit components pass location automatically, so I add this to be compatible with UI-kit ones. The goal is to remove this component entirely

apps/condo/domains/common/components/Table/Filters.tsx
SavelevMatthew: For this specific filter - I don't think so... 

I've removed some old logs because I my head they was not used / valuable enough to build conversion / other product dashboards

SavelevMatthew: And "events just for events" for me is a bit useless. 

I would prefer combination of "basic events" (UI-kit) + "business valuable events" (on which you can compute DAU or something like that)

PR #5910, opened 2025-03-13
chore(condo): refactor PropertyReportCard.tsx

toplenboren:
This fixes the markup of Property Report Card component when Property report is active


sitozzz (approved):
lgtm

PR #5911, opened 2025-03-13
chore(condo): draft: language customization

toplenboren:


apps/condo/pages/settings/index.tsx
```diff
@@ -139,6 +141,7 @@ const SettingsPage: PageComponentType = () => {
             </Head>
             <PageWrapper>
                 <OrganizationRequired>
+                    {TestCustomTranslation}
```
sitozzz, line 144: Almost perfect, but I don't think that's necessary to provide an example inside code. IMO it'll be better to add that to docs

toplenboren: Yep, that is correct :-) I'll remove that once we have discussed this proposal internally :-)

PR #5914, opened 2025-03-14
feat(keystone): INFRA-931 add ApolloQueryBlockingPlugin

SavelevMatthew:


packages/keystone/apolloServerPlugins/queryBlocking/plugin.js
```diff
@@ -0,0 +1,59 @@
...
+                const blockedQueries = queries
+                    .map(query => query.name)
+                    .filter(name => this.#blockedQueries.has(name))
+
+                if (blockedMutations.length || blockedQueries.length) {
+                    throw new GQLError({
+                        code: FORBIDDEN,
+                        type: ERROR_TYPE,
+                        message: 'Request is rejected because it contains blocked queries / mutations: {blockedOperations}',
+                        messageForUser: `api.global.queryBlocking.${ERROR_TYPE}`,
```
pahaz, line 46: It looks like it should be hardcoded here. You can't chang it just by changing value number.

packages/keystone/apolloServerPlugins/queryBlocking/constants.js
```diff
@@ -0,0 +1,5 @@
+const ERROR_TYPE = 'FORBIDDEN_REQUEST'
```
pahaz, line 1: It's better to move it inside `packages/keystone/apolloServerPlugins/queryBlocking/plugin.js` it used only there ...

pahaz: its also better to use `FORBIDDEN_REQUEST` const name for better readability

pahaz (approved):

packages/keystone/apolloServerPlugins/queryBlocking/constants.js
SavelevMatthew: inlined variable inside plugin.js

sitozzz (approved):

PR #5915, opened 2025-03-14
fix(condo): DOMA-11246 split overcomplicated query for payments sum

YEgorLu:
And add env flag to disable this logic if something bad happens.

Problem with old query was that postgres decided to filter all Payments by sequence scan first, and only then use indexes and used 2 workers for this (usually it's just one worker).

I imitated production database Payments locally, with similar distribution of non nullable receipts, unique organizations, periods and accountNumbers. Also imitated case when all Payments will loose connection to receipt and tested query split and some indexes.

Looks like that easiest solution is to just split queries. In that case postgres uses indexes before sequence scan and works faster. 
Explain analyze of resulting sql on production gave me these results:
Organization with 800 payments - 16ms
Organization with 4.500 payments - 30ms
Organization with 25.000 payments - 80ms

In contrast old query gave around 3-4s on any parameters
And query without searching for payments with "receipt is null" and "invoice is null" gave 1-22ms. 

Better solution might be to add index "Payment (organization, accountNumber) where deletedAt is null", based on test results (around 2.1ms on any parameters). It would work with Payments with no relation for receipt, but I think that complete move from relations to invoice and receipt requires more sufficient index and maybe field, so let's think about it later

### DIFF
--- a/apps/condo/domains/billing/utils/serverSchema/index.js
+++ b/apps/condo/domains/billing/utils/serverSchema/index.js
@@ -8,7 +8,6 @@ const Big = require('big.js')
 
 const { generateServerUtils } = require('@open-condo/codegen/generate.server.utils')
 const { execGqlWithoutAccess } = require('@open-condo/codegen/generate.server.utils')
-const conf = require('@open-condo/config')
 const { find, getById } = require('@open-condo/keystone/schema')
 
 const { PAYMENT_DONE_STATUS, PAYMENT_WITHDRAWN_STATUS } = require('@condo/domains/acquiring/constants/payment')
@@ -30,8 +29,6 @@ const BillingReceipt = generateServerUtils('BillingReceipt')
 const BillingRecipient = generateServerUtils('BillingRecipient')
 const BillingCategory = generateServerUtils('BillingCategory')
 const BillingReceiptFile = generateServerUtils('BillingReceiptFile')
-
-const DISABLE_QR_NEW_PAYMENTS_SUM = String(conf.DISABLE_QR_NEW_PAYMENTS_SUM).toLowerCase() === 'true'
 
 async function registerBillingReceipts (context, data) {
     if (!context) throw new Error('no context')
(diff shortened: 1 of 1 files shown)


apps/condo/domains/billing/utils/serverSchema/index.js
```diff
@@ -30,6 +31,8 @@ const BillingRecipient = generateServerUtils('BillingRecipient')
 const BillingCategory = generateServerUtils('BillingCategory')
 const BillingReceiptFile = generateServerUtils('BillingReceiptFile')
 
+const DISABLE_QR_NEW_PAYMENTS_SUM = String(conf.DISABLE_QR_NEW_PAYMENTS_SUM).toLowerCase() === 'true'
```
toplenboren, line 34: This is really strange :-/ please check out the generic mechanism from @SavelevMatthew #5914 will it be enough for your case?

I believe we should either optimize the query or just remove it from the code. 

Just adding a toggle without any documentation will increase the complexity of the codebase :-(

toplenboren (dismissed):

apps/condo/domains/billing/utils/serverSchema/index.js
YEgorLu: Removed  flag.
Query should be fine right now after this split.

apps/condo/domains/billing/utils/serverSchema/index.js
```diff
@@ -86,20 +84,27 @@ const getNewPaymentsSum = async (receiptId) => {
...
+            { receipt_is_null: true },
+            { invoice_is_null: true },
+            { organization: { id: billingContext.organization } },
+            { period: receipt.period },
+            { accountNumber: account.number },
+            ...defaultConditions,
+        ]
+        const qrPayments = await find('Payment', {
+            AND: conditionsWithNoReceipt,
+        })
```
toplenboren, line 105: You should double check that query – you have removed `OR` condition from the `where`, this can cause issues with returning wrong data

toplenboren (approved):

apps/condo/domains/billing/utils/serverSchema/index.js
YEgorLu: checked.

It gives select from "Payment" left outer join "Organization" | "BillingReceipt" and true and (... [and (false or transferDate)]). Looks right

PR #5916, opened 2025-03-14
fix(condo): DOMA-11249 update cache config and add first arg to queries

tolmachev21:
After task DOMA-11056 we need update cache config and add to queries first arg, that query complexity calculation was correct


apps/condo/domains/ticket/queries/Ticket.graphql
```diff
@@ -238,8 +246,10 @@ query getTicketById ($id: ID!) {
 
 query getTicketByCreatedBy ($userId: ID!) {
```
toplenboren, line 247: This looks really strange, you might have more then 1 ticket here :-)

apps/condo/domains/ticket/queries/Ticket.graphql
```diff
@@ -80,7 +85,10 @@ query getTickets ($where: TicketWhereInput!, $sortBy: [SortTicketsBy!], $first:
 }
 
 query getTicketExistence ($where: TicketWhereInput!) {
```
toplenboren, line 87: checkTicketExistence

apps/condo/domains/ticket/queries/Ticket.graphql
```diff
@@ -261,7 +271,8 @@ query getTicketLastCommentsTime ($id: ID!) {
 query getContactTickets ($contactId: ID!) {
```
toplenboren, line 271: getTicketsByContact

toplenboren (approved):

apps/condo/domains/ticket/queries/Ticket.graphql
tolmachev21: It seems that this query is used as a heuristic that the application was created by the current user. This is needed to calculate the onboarding progress. However, it is not clear from the name of the query. But I will not change the name, because I only corrected the style code

SavelevMatthew (approved):

PR #5922, opened 2025-03-17
fix(condo): added the missing "canReadByResident" field when updating and adding a document

paulo-rossy:


toplenboren:
We can add this fields in condo UI, wyt? So they are available for all users

toplenboren (approved):

SavelevMatthew (approved):
lgtm

toplenboren (approved):

PR #5927, opened 2025-03-17
fix(condo): DOMA-11271 detect when isAutomatic is changed to false

abshnko:


apps/condo/domains/meter/utils/meter.utils.js
```diff
@@ -66,12 +66,12 @@ function shouldUpdateMeter (meter, changedFields, isPropertyMeters = false) {
...
-        if (result) {
-            return result
-        }
-        return !!get(changedFields, field) && get(changedFields, field) !== get(meter, field)
-    }, false)
+    return fieldsToUpdate.some(field => {
+        const newValue = get(changedFields, field)
+
+        // NOTE: isAutomatic is a true/false field, so we need to check if it was changed
+        return (newValue === false || Boolean(newValue)) && newValue !== get(meter, field)
```
dkoviazin, line 73: Please add a separate check for isAutomatic

dkoviazin (approved):

YEgorLu (approved):

apps/condo/domains/meter/utils/meter.utils.js
AleX83Xpert, line 73:
```suggestion
        return newValue !== undefined && newValue !== get(meter, field)
```

AleX83Xpert (commented):
What about a test?

apps/condo/domains/meter/utils/meter.utils.js
abshnko: yes, i initially did it like this, but we need to also check for !== '' because 'place' can be an empty string

PR #5931, opened 2025-03-18
feat/address-service/doma-11130/ability to update address

AleX83Xpert:


apps/address-service/domains/address/access/ActualizeAddressesService.js
```diff
@@ -0,0 +1,19 @@
...
+ * Generated by `createservice address.ActualizeAddressesService --type mutations 'addresses:[AddressWhereUniqueInput!]!'`
+ */
+const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
+
+async function canActualizeAddresses ({ authentication: { item: user } }) {
+    if (!user) return throwAuthenticationError()
+    if (user.deletedAt) return false
+
+    return !!(user.isAdmin || user.isSupport)
+}
```
sitozzz, line 11: how should it work? only in manual mode?

AleX83Xpert: For now - yes. Next I'll move logic to use it here and in cron task.

AleX83Xpert: Also, it may possible to add button within admin interface to refresh particular address.

apps/address-service/domains/address/schema/Address.js
```diff
@@ -70,6 +69,16 @@ const Address = new GQLListSchema('Address', {
...
+    kmigratorOptions: {
+        constraints: [
+            {
+                type: 'models.UniqueConstraint',
+                fields: ['key'],
+                condition: 'Q(deletedAt__isnull=True)',
+                name: 'address_key_unique_key',
+            },
+        ],
+    },
```
sitozzz, line 81: How it's related to this PR?

apps/address-service/domains/address/schema/AddressSource.js
```diff
@@ -67,6 +66,16 @@ const AddressSource = new GQLListSchema('AddressSource', {
...
+    kmigratorOptions: {
+        constraints: [
+            {
+                type: 'models.UniqueConstraint',
+                fields: ['source'],
+                condition: 'Q(deletedAt__isnull=True)',
+                name: 'addresssource_source_unique_key',
+            },
+        ],
+    },
```
sitozzz, line 78: it's the same here

apps/address-service/domains/address/schema/ActualizeAddressesService.js
```diff
@@ -0,0 +1,149 @@
...
+
+const access = require('@address-service/domains/address/access/ActualizeAddressesService')
+const { Address, AddressSource } = require('@address-service/domains/address/utils/serverSchema')
+const { DADATA_PROVIDER, GOOGLE_PROVIDER } = require('@address-service/domains/common/constants/providers')
+const { generateAddressKey } = require('@address-service/domains/common/utils/addressKeyUtils')
+const { DadataSuggestionProvider } = require('@address-service/domains/common/utils/services/suggest/providers')
+
+const logger = getLogger('ActualizeAddressesService')
+
+const ActualizeAddressesService = new GQLCustomSchema('ActualizeAddressesService', {
```
sitozzz, line 18: It's just a question - why not update?) IMO "Actualize" sounds a bit strange

apps/address-service/domains/address/schema/Address.js
AleX83Xpert: When the address was changed, it's possible that a new Address model was created. I need to remove it and update another Address model, which was used to create property in a condo. These two models have the same `key` field.
There was a restriction `isUnique: true`, but I don't want to restrict deleted models.
The mutation deletes one model and passes its data into another model created earlier.

apps/address-service/domains/address/schema/AddressSource.js
AleX83Xpert: The same. I want to restrict only not deleted models.

apps/address-service/domains/address/schema/ActualizeAddressesService.js
AleX83Xpert: The update sounds like we should pass some data on to update the model. But here we actualize the model with data loaded in runtime.

apps/address-service/domains/address/access/ActualizeAddressesService.js
AleX83Xpert: Just addes a button to admin interface
image

nomerdvadcatpyat (approved):

PR #6036, opened 2025-04-09
fix (condo): DOMA-10994 add suffix type from antd

Alexander-Turkin:


SavelevMatthew (changes requested):
Why do you need this?

PR #6060, opened 2025-04-14
feat(condo): DOMA-11464 add read access to Payment for B2BApp service users

YEgorLu:
- Adding "canReadPayments" flag for B2BAppAccessRightSet
- Adding check for b2bApp service user accesses in Payment read access

We need it for b2b app for payment receipts and for payment downloading with B2BAccessToken

### DIFF
--- a/apps/condo/domains/acquiring/access/Payment.js
+++ b/apps/condo/domains/acquiring/access/Payment.js
@@ -7,13 +7,15 @@ const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFo
 const { find } = require('@open-condo/keystone/schema')
 
 const { checkAcquiringIntegrationAccessRights } = require('@condo/domains/acquiring/utils/accessSchema')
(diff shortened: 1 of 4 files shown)


apps/condo/domains/acquiring/access/Payment.js
```diff
@@ -26,29 +28,40 @@ async function canReadPayments ({ authentication: { item: user }, listKey }) {
...
-            {
-                AND: [
-                    {
-                        invoice_is_null: false,
-                        organization: { employees_some: { user: { id: user.id }, role: { canReadPaymentsWithInvoices: true }, deletedAt: null, isBlocked: false } },
-                    },
-                ],
-            },
-        ],
+        OR: filters,
```
abshnko, line 64: i think it's better to separate these filters for different user types, so that the query is not so complex.  like so:
if (user.type === SERVICE) {
      // here logic only for service users
 }

if (user.type === STAFF) {
    // logic for STAFF users
}

return false

abshnko (approved):

dkoviazin (approved):

PR #6114, opened 2025-04-22
feat(condo): DOMA-11513 add "id" in SuggestServiceProviderOutput, clarify that output is not nullable

YEgorLu:
Query is not being used by anyone right now, but we need an Id to know what to bind for.

## Summary by CodeRabbit

- **New Features**
	- The service provider suggestion now includes ID and type fields, providing more detailed information for each suggestion.
- **Improvements**
	- The service provider suggestions query always returns a non-null list with no null entries, ensuring consistent and reliable results.

### DIFF
--- a/apps/condo/domains/resident/schema/SuggestServiceProviderService.js
+++ b/apps/condo/domains/resident/schema/SuggestServiceProviderService.js
@@ -59,14 +59,14 @@ const SuggestServiceProviderService = new GQLCustomSchema('SuggestServiceProvide
         },
         {
             access: true,
-            type: 'type SuggestServiceProviderOutput { tin: String!, name: String! }',
+            type: 'type SuggestServiceProviderOutput { id: String!, tin: String!, name: String!, type: String! }',
         },
     ],
     
     queries: [
         {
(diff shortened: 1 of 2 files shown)


apps/condo/domains/resident/schema/SuggestServiceProviderService.js
```diff
@@ -59,14 +59,14 @@ const SuggestServiceProviderService = new GQLCustomSchema('SuggestServiceProvide
         },
         {
             access: true,
-            type: 'type SuggestServiceProviderOutput { tin: String!, name: String! }',
+            type: 'type SuggestServiceProviderOutput { id: String! tin: String!, name: String! }',
```
pahaz, line 62: I think we don't have it in all cases. If it's suggest from third party provider we don't have any id

YEgorLu: ServiceProvider is Organization of ours and we search it here. That's why we have id

abshnko (approved):

vovaaxeapolla (approved):

PR #6118, opened 2025-04-22
fix(condo): DOMA-11376 remove the automatically enabled checkbox

tolmachev21:
## Summary by CodeRabbit

- **Refactor**
  - Updated the Checkbox component in floor and unit forms to use a new UI library for improved consistency.
  - Simplified form layouts by removing unused styles and unnecessary props from Checkbox elements.

### DIFF
--- a/apps/condo/domains/property/components/panels/Builder/forms/UnitForm.tsx
+++ b/apps/condo/domains/property/components/panels/Builder/forms/UnitForm.tsx
@@ -6,8 +6,8 @@ import { Row, Col, Space, Typography } from 'antd'
 import React, { useState, useRef, useEffect, useCallback, useMemo } from 'react'
 
 import { useIntl } from '@open-condo/next/intl'
+import { Checkbox } from '@open-condo/ui'
 
-import Checkbox from '@condo/domains/common/components/antd/Checkbox'
 import Input from '@condo/domains/common/components/antd/Input'
 import Select from '@condo/domains/common/components/antd/Select'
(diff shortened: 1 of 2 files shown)


toplenboren (approved):

nomerdvadcatpyat (approved):

apps/condo/domains/property/components/panels/Builder/forms/UnitForm.tsx
```diff
@@ -238,15 +237,13 @@ const UnitForm: React.FC<IPropertyMapModalForm> = ({ builder, refresh, setDuplic
...
-                            onChange={toggleRenameNextUnits}
-                            style={MODAL_FORM_CHECKBOX_STYLE}
-                            data-cy='property-map__unit-form__rename-units-checkbox'
-                        >
-                            {RenameNextUnitsLabel}
-                        </Checkbox>
                     </Space>
+                    <Checkbox
+                        onChange={toggleRenameNextUnits}
+                        data-cy='property-map__unit-form__rename-units-checkbox'
```
pahaz, line 243: Why is this name?

tolmachev21: Good question - you should ask whoever wrote this code. However, I assume the following:

- property-map - a common name for a group of inputs
- unit-form - a form type (same as "add-section" etc.). Important point, most likely there is a common name unit-form, which applies to both editing and adding a room
- rename-units-checkbox - a form element type (same as unit-type-select or section-select)

PR #6124, opened 2025-04-24
fix(condo): DOMA-11524 add validation for message.meta.data

tolmachev21:


apps/condo/domains/notification/schema/SendMessageService.js
```diff
@@ -78,7 +81,24 @@ async function checkSendMessageMeta (type, meta, context) {
         if (required && !value) {
             throw new GQLError({ ...ERRORS.MISSING_VALUE_FOR_REQUIRED_META_ATTRIBUTE, messageInterpolation: { attr } }, context)
         }
+
+        if (attr === 'data' && typeof value === 'object' && schema.data) {
+            const dataSchema = schema.data
+            for (const dataAttr of Object.keys(dataSchema)) {
+                const { required: dataRequired } = dataSchema[dataAttr]
```
Alllex202, line 88: If an attribute is described by an object, it does not mean that the data must necessarily be an object.

For example, we have a message type, whose meta field is described as follows:
```
    [DEVELOPER_IMPORTANT_NOTE_TYPE]: {
        dv: { defaultValue: '', required: true },
        type: { defaultValue: 'UNKNOWN', required: true },
        data: { defaultValue: null, required: true },
    },
```
Here the data can be absolutely anything `data: { defaultValue: null, required: true }` , but your check will expect an object with such keys, and in general it may fail.

apps/condo/domains/notification/schema/SendMessageService.js
```diff
@@ -78,7 +81,24 @@ async function checkSendMessageMeta (type, meta, context) {
...
+                if (dataRequired && !value[dataAttr]) {
+                    logger.info({ msg: 'Missing value for required "meta"', dataAttr })
```
Alllex202, line 90: How do you know from this log what type of message is being used?

Alllex202 (changes requested):

apps/condo/domains/notification/schema/SendMessageService.js
tolmachev21: If we have meta processing with a date without a field and the date is not transmitted, then we will send an error to this case above.

Therefore, we will not check anything inside in this case, since there is nothing to check. I fixed the destructuring error

apps/condo/domains/notification/schema/SendMessageService.js
```diff
@@ -78,7 +81,27 @@ async function checkSendMessageMeta (type, meta, context) {
...
         }
+
+        if (attr === 'data' && typeof value === 'object' && schema.data) {
+            const dataSchema = schema.data
+            for (const dataAttr of Object.keys(dataSchema)) {
+                if (dataAttr === 'required' || dataAttr === 'defaultValue') {
+                    continue
+                }
+                const { required: dataRequired } = dataSchema[dataAttr]
+                if (dataRequired && !value[dataAttr]) {
```
Alllex202, line 92: We should check only `undefined` and `null`?
I think that `0`, `false`, `<empty line>`(?) are valid values

apps/condo/domains/notification/schema/SendMessageService.js
```diff
@@ -78,7 +81,27 @@ async function checkSendMessageMeta (type, meta, context) {
...
+                    logger.info({ msg: 'Missing value for required "meta"', dataAttr, type })
+                }
+            }
+
+            for (const dataAttr of Object.keys(value)) {
+                if (!dataSchema[dataAttr]) {
+                    logger.info({ msg: 'Unknown attribute provided to "meta" variable', dataAttr, type })
+                }
+            }
```
Alllex202, line 101: Do we really need this check?

apps/condo/domains/notification/schema/SendMessageService.js
```diff
@@ -75,10 +78,30 @@ async function checkSendMessageMeta (type, meta, context) {
...
+        if (required && (value === undefined || value === null)) {
             throw new GQLError({ ...ERRORS.MISSING_VALUE_FOR_REQUIRED_META_ATTRIBUTE, messageInterpolation: { attr } }, context)
         }
+
+        if (attr === 'data' && typeof value === 'object' && schema.data) {
+            const dataSchema = schema.data
+            for (const dataAttr of Object.keys(dataSchema)) {
+                if (dataAttr === 'required' || dataAttr === 'defaultValue') {
+                    continue
+                }
```
Alllex202, line 90: Leave a comment on why this case should be skipped.

nomerdvadcatpyat (approved):

PR #6138, opened 2025-04-29
fix(condo): DOMA-11552 fix and refactor cache ttl config

tolmachev21:


apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.test.js
```diff
@@ -330,6 +330,24 @@ describe('SendB2CAppPushMessageService', () => {
                 message: 'You have to wait {minutesRemaining} min. to be able to send request again',
             })
         })
+
+        test('Check working CACHE_TTL config', async () => {
```
toplenboren, line 334: add a negative test?

apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.js
```diff
@@ -30,9 +30,9 @@ const { RedisGuard } = require('@condo/domains/user/utils/serverSchema/guards')
 
 const CACHE_TTL = {
     DEFAULT: DEFAULT_NOTIFICATION_WINDOW_DURATION_IN_SECONDS,
-    VOIP_INCOMING_CALL_MESSAGE: 2,
-    CANCELED_CALL_MESSAGE_PUSH_TYPE: 2,
-    B2C_APP_MESSAGE_PUSH: 3600,
+    [VOIP_INCOMING_CALL_MESSAGE_TYPE]: 2,
```
toplenboren, line 33: why?

toplenboren (approved):

nomerdvadcatpyat (approved):

apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.js
tolmachev21: I think this solution is the most extensible. Since before that there was a hardcode, but now the values are stored in variables.

apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.test.js
tolmachev21: Negative tests for other types of push messages were previously added there. I don't think that need to push anything else for this type of push, but I would like to check that it is the time indication in the configuration that works, and not just the presence of an error.

PR #6159, opened 2025-05-05
fix (condo): DOMA-11564 add compare with B2BAppNewsSharingConfig

Alexander-Turkin:
## Summary by CodeRabbit

- **New Features**
  - Improved the display logic for sharing news with other apps by using the total count of eligible apps, ensuring the "Other Apps" card appears only when applicable.
- **Enhancements**
  - Added support for retrieving the count of B2B apps with news sharing configurations, leading to more accurate sharing options.

### DIFF
--- a/apps/condo/domains/miniapp/queries/B2BAppNewsSharingConfig.graphql
+++ b/apps/condo/domains/miniapp/queries/B2BAppNewsSharingConfig.graphql
@@ -1,0 +1,5 @@
+query getB2BAppNewsSharingConfig {
+    B2BAppNewsSharingConfigs: allB2BAppNewsSharingConfigs {
+        id
+    }
+}
--- a/apps/condo/domains/miniapp/queries/B2BAppNewsSharingConfig.graphql
+++ b/apps/condo/domains/miniapp/queries/B2BAppNewsSharingConfig.graphql
@@ -1,5 +1,0 @@
-query getB2BAppNewsSharingConfig {
-    B2BAppNewsSharingConfigs: allB2BAppNewsSharingConfigs {
-        id
-    }
-}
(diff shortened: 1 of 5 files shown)


apps/condo/domains/miniapp/queries/B2BAppNewsSharingConfig.graphql
```diff
@@ -0,0 +1,5 @@
+query getB2BAppNewsSharingConfig {
```
toplenboren, line 1: why only Id?

Seems like you really just need to use count query here

And you also need to only get B2BAppNewsSharingConfigs that are active and are public

Consider this case:

I, as a big construction company create a miniapp integraion with my internal systems. This miniapp can have news sharing function. This miniapp is private and cannot be used by other managing companies

And judging by this query and by your code, u will show every user from every organization a prompt to go and check other services

toplenboren (changes requested):

apps/condo/domains/news/components/NewsForm/SelectSharingAppControl.tsx
```diff
@@ -97,6 +99,9 @@ const SelectSharingAppControl: React.FC<ISelectSharingAppControl> = ({ sharingAp
         })
     }, [sharingAppContexts, getNewsSharingRecipients, onError])
 
+    const { data: getCountB2BAppsWithNewsSharingConfigQuery } =  useGetCountB2BAppsWithNewsSharingConfigQuery()
```
nomerdvadcatpyat, line 102:
```suggestion
    const { data: countB2BAppsWithNewsSharingConfigResult } = useGetCountB2BAppsWithNewsSharingConfigQuery()
```

apps/condo/domains/news/components/NewsForm/SelectSharingAppControl.tsx
```diff
@@ -97,6 +99,9 @@ const SelectSharingAppControl: React.FC<ISelectSharingAppControl> = ({ sharingAp
...
+    const B2BAppWithNewsSharingConfigsCount = getCountB2BAppsWithNewsSharingConfigQuery?._allB2BAppsMeta?.count
```
nomerdvadcatpyat, line 103:
```suggestion
    const b2BAppWithNewsSharingConfigsCount = getCountB2BAppsWithNewsSharingConfigQuery?._allB2BAppsMeta?.count
```

nomerdvadcatpyat (approved):

PR #6169, opened 2025-05-07
fix(condo): DOMA-11581 add autoselect in contact card

Alexander-Turkin:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved tab selection logic to ensure the correct tab is activated when no specific tab is provided in the URL.
  - Enhanced consistency in displaying tab content by using fallback values when certain query parameters are missing.
  - Fixed an issue with tab index tracking for more reliable navigation between tabs.

### DIFF
--- a/apps/condo/pages/phone/[number]/index.tsx
+++ b/apps/condo/pages/phone/[number]/index.tsx
@@ -764,7 +764,7 @@ const ClientCardPageContent = ({
             tabDataWithProperty = tabsData[0]
         }
         else {
-            tabDataWithProperty = tabsData.find(tab => (
+            tabDataWithProperty = tabsData?.find(tab => (
                 tab.property.id === propertyId &&
                 tab.unitName === unitName &&
                 tab.unitType === unitType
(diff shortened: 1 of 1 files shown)


toplenboren (approved):

apps/condo/pages/phone/[number]/index.tsx
```diff
@@ -757,20 +757,43 @@ const ClientCardPageContent = ({
 
     useEffect(() => {
         const { type, property: propertyId, unitName, unitType, sectionType, sectionName } = parseCardDataFromQuery(tab)
-        const tabDataWithProperty = tabsData.find(({ property, unitName: tabUnitName, unitType: tabUnitType }) =>
-            property.id === propertyId && tabUnitName === unitName && tabUnitType === unitType)
+
+        const shouldReturnFirstTab = tabsData?.length > 0 && !unitName && !unitType
```
Comment (nomerdvadcatpyat) [inline apps/condo/pages/phone/[number]/index.tsx:761]: I would invert this logic — I think it would make things clearer that way

```suggestion
        const isTabSelected = tabsData?.length > 0 && unitName && unitType // check other variables?
        
        if (isTabSelected) {
         // find and select tab
        } else {
          // select first tab
        }
```

nomerdvadcatpyat (approved):

PR #6171, opened 2025-05-08
chore(condo): add some id and dataCy for elements in ticket and incident [id] page

oleg-murmur:
In automated tests, we need to check the change history and other data on the incidents and tickets page
### Changes:

1. Different "date-cy" for the "Change log of the ticket"(ticket) and "Changes hostory"(incident) blocks are set. Previously there was one common one with the name `ticket__history-change`. Now for ticket this attribute is `ticket__history-change`, for incident is `incident__history-change`
2. `id={change.id}` for every change item in history change block
3. `id={field}` for every message item in change item

изображение

4. `id='ticket_resident-features'` for resident features block (example: if contact have resident, we can check this in test)
изображение

5. `id='incident__create-info'` and `id='ticket__create-info'` for info block. This we need to check name user (createdBy)
изображение
изображение

6. `id='ticket__field-status'` for check statuses like: `isEmergency`, `isPayble`
изображение

For this case i use this: default ant component, like `Col`, use `id='xyz'`. For ui kit use `dataCy='xyz'`,

? Can we talk about naming `dataCy` and `data-cy` ? I made up the naming that was before, but now Cypress is not used, let's choose another one. I think wiil be good attribute is `dataTest` and  `data-test`. Or we can use `id` every time

## Summary by CodeRabbit

- **New Features**
    - Added unique HTML `id` attributes to components on incident and ticket pages, change history, and resident features sections for improved element identification and interaction in testing and automation.
- **Chores**
    - Extended component properties to support custom `id` attributes without impacting existing functionality or layout.


SavelevMatthew (approved):
lgtm

PR #6182, opened 2025-05-15
fix(condo): DOMA-11613 fix logic open filter modal

tolmachev21:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved filter template comparison by ignoring empty and nil fields, ensuring more accurate matching.
  - Enhanced state handling for filter templates to consistently use null when a template is not found.
  - Refined initial form values logic for filters, providing better handling when templates or fields are missing.
  - Fixed filter reset behavior to ensure proper asynchronous handling and consistent filter clearing.

### DIFF
--- a/apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
+++ b/apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
@@ -338,7 +338,7 @@ const ResetFiltersModalButton: React.FC<ResetFiltersModalButtonProps> = ({
     const { setSelectedFiltersTemplate } = useMultipleFilterContext()
 
     const handleReset = useCallback(async () => {
-        router.replace({ query: omit(router.query, ['filters', 'sort', 'offset']) }, undefined, { shallow: true })
+        await router.replace({ query: omit(router.query, ['filters', 'sort', 'offset']) }, undefined, { shallow: true })
         setSelectedFiltersTemplate(null)
 
         if (isFunction(handleResetFromProps)) {
--- a/apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
+++ b/apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
@@ -335,16 +335,14 @@ const ResetFiltersModalButton: React.FC<ResetFiltersModalButtonProps> = ({
     const intl = useIntl()
     const ClearAllFiltersMessage = intl.formatMessage({ id: 'ClearAllFilters' })
     const router = useRouter()
-    const { setSelectedFiltersTemplate } = useMultipleFilterContext()
 
     const handleReset = useCallback(async () => {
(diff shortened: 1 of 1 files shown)


apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
```diff
@@ -686,7 +687,7 @@ const Modal: React.FC<MultipleFiltersModalProps> = ({
         </TabPane>
     )), [ExistingFiltersTemplateNameInput, TemplateMessage, filtersTemplates])
 
-    const initialFormValues = useMemo(() => openedFiltersTemplate?.fields, [openedFiltersTemplate])
+    const initialFormValues = useMemo(() => openedFiltersTemplate?.fields ? openedFiltersTemplate?.fields : (selectedFiltersTemplate ? {} : filters), [filters, selectedFiltersTemplate, openedFiltersTemplate])
```
nomerdvadcatpyat, line 690: It's better to use if-else than a double ternary operator. It will be more readable.
And how is selectedFiltersTemplate different from openedFiltersTemplate?

apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
```diff
@@ -386,6 +386,7 @@ const isEqualSelectedFiltersTemplateAndFilters = (selectedFiltersTemplate, filte
     const templateFilters = selectedFiltersTemplate?.fields ?? null
     if (!templateFilters) return false
     if (has(templateFilters, '__typename')) delete templateFilters['__typename']
+    if (has(templateFilters, 'attributes')) delete templateFilters['attributes']
```
nomerdvadcatpyat, line 389: Why we need to remove `attributes` filter?

apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
tolmachev21: It seems that the open filter template is the one whose tab is currently open in the modal. The selected filter template is the one whose filters are applied.

Logic: if we have an open template, then insert its fields into the form. If there is no open template (that is, we are on the "New Filter" tab), then see if there is a selected template. If there is, then we want to reset the form ({}), and if not, then we want to see the filled form fields that correspond to the current filters (filters)

apps/condo/domains/common/hooks/useMultipleFiltersModal.tsx
tolmachev21: In some templates we have attributes for filtering. They are set in the useBooleanAttributesSearch component. If the filter template has an attribute, it will not be highlighted as an active window, since the check in the isEqualSelectedFiltersTemplateAndFilters function will not pass. Therefore, this attribute needs to be cleaned up

tolmachev21:
https://github.com/user-attachments/assets/5f07cec8-712b-4671-a4f7-eafff11846fa

nomerdvadcatpyat (approved):

toplenboren (approved):

PR #6184, opened 2025-05-16
feat(condo): DOMA-11617 add runtime stats for condo app

AleX83Xpert:
## Summary by CodeRabbit

- **New Features**
  - Introduced real-time tracking of active HTTP requests, categorized by type.
  - Added a new endpoint at "/.well-known/runtime-stats" to display current runtime request statistics in JSON format.
- **Chores**
  - Exported the default health check URL constant for easier configuration.

### DIFF
--- a/apps/condo/index.js
+++ b/apps/condo/index.js
@@ -210,12 +210,18 @@ const extendExpressApp = (app) => {
     })
 
     app.get('/.well-known/runtime-stats', (req, res) => {
-        res.json({
-            activeRequestsCount: runtimeStats.activeRequestsIds.size,
-            activeRequestsCountByType: runtimeStats.activeRequestsCountByType,
-            totalRequestsCount: runtimeStats.totalRequestsCount,
-            totalRequestsByTarget: runtimeStats.totalRequestsByTarget,
-        })
+        const token = req.query['token']
+        const accessToken = conf['RUNTIME_STATS_ACCESS_TOKEN']
+        if (!!accessToken && accessToken === token) {
+            res.json({
+                activeRequestsCount: runtimeStats.activeRequestsIds.size,
+                activeRequestsCountByType: runtimeStats.activeRequestsCountByType,
+                totalRequestsCount: runtimeStats.totalRequestsCount,
+                totalRequestsByTarget: runtimeStats.totalRequestsByTarget,
+            })
+        } else {
+            res.status(403).send()
+        }
     })
 
     app.get('/.well-known/change-password', function (req, res) {
--- a/apps/condo/index.js
+++ b/apps/condo/index.js
@@ -156,12 +156,15 @@ const extendExpressApp = (app) => {
             healthCheck: 0,
             other: 0,
         },
+        totalRequestsCount: 0,
+        totalRequestsByTarget: {},
     }
 
     app.use(function runtimeStatsMiddleware (req, res, next) {
         try {
             const url = new URL(`${conf['SERVER_URL']}${req.url}`).pathname
             const method = (req.method || 'get').toLowerCase()
+            const xTargetHeader = req.headers['x-target'] || 'other'
 
             const isPost = method === 'post'
 
@@ -189,6 +192,8 @@ const extendExpressApp = (app) => {
             if (requestType) {
                 runtimeStats.activeRequestsIds.add(req.id)
                 runtimeStats.activeRequestsCountByType[requestType] = (runtimeStats.activeRequestsCountByType[requestType] || 0) + 1
+                runtimeStats.totalRequestsCount = (runtimeStats.totalRequestsCount || 0) + 1
+                runtimeStats.totalRequestsByTarget[xTargetHeader] = (runtimeStats.totalRequestsByTarget[xTargetHeader] || 0) + 1
             }
 
             res.on('close', () => {
@@ -208,6 +213,8 @@ const extendExpressApp = (app) => {
         res.json({
             activeRequestsCount: runtimeStats.activeRequestsIds.size,
             activeRequestsCountByType: runtimeStats.activeRequestsCountByType,
+            totalRequestsCount: runtimeStats.totalRequestsCount,
+            totalRequestsByTarget: runtimeStats.totalRequestsByTarget,
         })
     })
 
--- a/apps/condo/index.js
+++ b/apps/condo/index.js
@@ -19,8 +19,8 @@ const {
     getPfxCertificateHealthCheck,
 } = require('@open-condo/keystone/healthCheck')
 const { prepareKeystone } = require('@open-condo/keystone/KSv5v6/v5/prepareKeystone')
+const { getLogger } = require('@open-condo/keystone/logging')
 const { RequestCache } = require('@open-condo/keystone/requestCache')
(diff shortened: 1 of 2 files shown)


apps/condo/index.js
```diff
@@ -143,6 +143,53 @@ const apps = () => {
...
+        },
+    }
+
+    app.use(function runtimeStatsMiddleware (req, res, next) {
+        const url = req.url
+        const method = (req.method || 'get').toLowerCase()
+
+        const isPost = method === 'post'
+
+        const isAdminApi = url.startsWith('/admin/api')
```
SavelevMatthew, line 161:
```suggestion
        const isGraphql = url.startsWith('/admin/api')
```

SavelevMatthew: Should we also parse exact path here? Like:
```
const url = new URL(req.url)

 const isGraphql = url.pathname === '/admin/api' 
```

apps/condo/index.js
```diff
@@ -143,6 +143,53 @@ const apps = () => {
...
+                runtimeStats.activeRequestsCountByType[requestType] = Math.max(0, runtimeStats.activeRequestsCountByType[requestType] - 1)
+            }
+        })
+
+        next()
+    })
+
+    app.get('/.well-known/runtime-stats', (req, res) => {
+        res.json({
+            activeRequestsIds: Array.from(runtimeStats.activeRequestsIds.keys()),
```
SavelevMatthew, line 187: Why do you need all ids here?

apps/condo/index.js
```diff
@@ -143,6 +143,53 @@ const apps = () => {
...
+            activeRequestsCount: runtimeStats.activeRequestsIds.size,
```
SavelevMatthew, line 188: is req.id always defined? Otherwise it can lead to incorrect counts

apps/condo/index.js
AleX83Xpert: Why not. Looks useful for investigations.

SavelevMatthew (commented):
Why not just simple shape like this? Why all IDs needed?
```
{
  "oidc": 2,
  "graphql": 10,
  "other": 3,
  "total": 15
}
```

apps/condo/index.js
SavelevMatthew: Who should use it? I have no real scenario on my mind, since you're not logging it anywhere. And you will not investigate anything  manually, since its expiring like each 200 ms?

apps/condo/index.js
```diff
@@ -143,6 +143,53 @@ const apps = () => {
...
+        const isPost = method === 'post'
+
+        const isAdminApi = url.startsWith('/admin/api')
+        const isOidc = url.startsWith('/oidc')
+
+        /** @type {'graphql' | 'oidc' | 'other'} */
+        let requestType = 'other'
+
+        if (isPost && isAdminApi) requestType = 'graphql'
+        if (isOidc) requestType = 'oidc'
```
pahaz, line 168: Add “well-known” group and “health-check” group

pahaz: And also: https://github.com/open-condo-software/condo/blob/main/apps/condo/index.js#L134

/api/version /api/files /api/auth … may be something others

It’s ok to extract “files” and “others-api” group by /api prefix.

apps/condo/index.js
AleX83Xpert: I think no. What is the profit?

apps/condo/index.js
AleX83Xpert: Looks like [yes](https://github.com/open-condo-software/condo/blob/a42d776f1a4e5885b22d1f247dbe604f3e8dd340/packages/keystone/KSv5v6/v5/prepareKeystone.js#L264).

AleX83Xpert:
> Why not just simple shape like this? Why all IDs needed?
> 
> ```
> {
>   "oidc": 2,
>   "graphql": 10,
>   "other": 3,
>   "total": 15
> }
> ```

I do not want to mix separated values with summarized values.

pahaz (approved):

apps/condo/index.js
SavelevMatthew: Because some weird request can affect your stat

Like request to `/admin/api/something/resolving/404`

SavelevMatthew: And since your target is to separately monitor amount of GQL requests, you want to separate specific endpoint, so other routes will count not as gql, but as others instead

apps/condo/index.js
AleX83Xpert: A new group `api` was added

AleX83Xpert: `wellKnown` and `healthCheck` groups too

apps/condo/index.js
AleX83Xpert: > 

`isAdminApi` is about the URL itself.

apps/condo/index.js
```diff
@@ -143,6 +144,64 @@ const apps = () => {
...
+        },
+    }
+
+    app.use(function runtimeStatsMiddleware (req, res, next) {
+        const url = req.url
+        const method = (req.method || 'get').toLowerCase()
+
+        const isPost = method === 'post'
+
+        const isAdminApiUrl = url === '/admin/api'
```
SavelevMatthew, line 162: This one will not work in case like `/admin/api?deleted=true` as far as I remember, since req.url contains query params
image


Consider using (new URL + pathname) option as I mentioned before

apps/condo/index.js
```diff
@@ -143,6 +146,84 @@ const apps = () => {
...
+        activeRequestsCountByType: {
+            graphql: 0,
+            api: 0,
+            oidc: 0,
+            wellKnown: 0,
+            healthCheck: 0,
+            other: 0,
+        },
+        totalRequestsCount: 0,
+        totalRequestsByTarget: {},
```
SavelevMatthew, line 160: Should we compute total by type? I thought we agreed to count it as well?

apps/condo/index.js
```diff
@@ -143,6 +146,84 @@ const apps = () => {
...
+    }
+
+    app.use(function runtimeStatsMiddleware (req, res, next) {
+        try {
+            const url = new URL(`${conf['SERVER_URL']}${req.url}`).pathname
+            const method = (req.method || 'get').toLowerCase()
+            const xTargetHeader = req.headers['x-target'] || 'other'
```
SavelevMatthew, line 167: You should sanitise this!!!

Otherwise I can brute-force values and memory leak

AleX83Xpert: Done

apps/condo/index.js
```diff
@@ -143,6 +146,97 @@ const apps = () => {
...
+        totalRequestsByTarget: {},
+    }
+
+    app.use(function runtimeStatsMiddleware (req, res, next) {
+        try {
+            const url = new URL(`${conf['SERVER_URL']}${req.url}`).pathname
+            const method = (req.method || 'get').toLowerCase()
+
+            // According to ADR35
+            const possibleXTargetValues = [
```
SavelevMatthew, line 169: take it from env, dont hard-code it

AleX83Xpert: moved to envs

SavelevMatthew (approved):

dlastskii (approved):

PR #6191, opened 2025-05-17
feat(keystone): DOMA-11626 add /api/debug for runtime analysis

pahaz:
use `if (isDebuggingEnabled) addDebugTools(app)`

### DIFF
--- a/packages/keystone/debugging.js
+++ b/packages/keystone/debugging.js
@@ -1,0 +1,162 @@
+const { Session } = require('inspector')
+const os = require('os')
+const {
+    monitorEventLoopDelay,
+    performance,
+    PerformanceObserver,
+} = require('perf_hooks')
+const { promisify } = require('util')
+const v8 = require('v8')
+
+const conf = require('@open-condo/config')
+
+const SAMPLE_MS = 20 // sample period
+const HISTORY = 10 // how many samples
+const TOKEN = conf['RUNTIME_STATS_ACCESS_TOKEN']
+
+const HOSTNAME = os.hostname()
+const CPU_CORES = os.cpus().length
+
+// keeps a rolling HDR-histogram of event-loop latency (how long the loop is blocked); docs: https://nodejs.org/api/perf_hooks.html#perf_hooksmonitoreventloopdelayoptions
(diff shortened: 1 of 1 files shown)


packages/keystone/debugging.js
```diff
@@ -0,0 +1,162 @@
...
+
+    LOOP_MONITOR.enable()
+    setInterval(collect, SAMPLE_MS).unref()
+
+    // TODO(pahaz): think about standards for `diagnostics_channel` to collect some business events timing
+
+    app.get('/api/debug/heap', (req, res) => {
+        if (req.header('x-token') !== TOKEN) return res.sendStatus(401)
+        const file = v8.writeHeapSnapshot()
+        res.json({ ok: true, host: HOSTNAME, file })
```
pahaz, line 142: should return file stream!

packages/keystone/debugging.js
```diff
@@ -0,0 +1,162 @@
...
+    })
+
+    app.get('/api/debug/cpu', async (req, res) => {
+        if (req.header('x-token') !== TOKEN) return res.sendStatus(401)
+        try {
+            const file = await captureCpu(10_000)  // 10s ~ 50–100 MB
+            res.json({ ok: true, host: HOSTNAME, file })
```
pahaz, line 149: should return file stream!

pahaz:
We also have HeapProfiling, something like a CpuProfiling, and we can add it too!

```
async function writeSamplingProfile(ms = 30_000) {
  const s = new inspector.Session(); s.connect();
  const post = util.promisify(s.post).bind(s);
  await post('HeapProfiler.enable');
  await post('HeapProfiler.startSampling', { samplingInterval: 256*1024 }); // 256 KB
  await new Promise(r => setTimeout(r, ms));
  const { profile } = await post('HeapProfiler.stopSampling');
  const file = path.join(os.tmpdir(), `heap-${Date.now()}.heapprofile`);
  await fs.promises.writeFile(file, JSON.stringify(profile));
  return file;
}
```

pahaz:
We also have old legacy `heap-profiler` here: https://www.npmjs.com/package/heap-profile?activeTab=readme

Issue #6181, opened 2025-05-15
Display public ticket info on `/share` page for unauthenticated users


paulo-rossy:
Hello team!

**Problem**
Some contractors cannot or do not want to register on the platform, yet we still need to share tickets with them. Currently, when an unauthenticated user opens a link like `/share/<ticketId>`, they are redirected to the login page and cannot see any information about the ticket.

**Proposed solution**
Update **`apps/condo/pages/share.tsx`**:

1. Detect whether the user is authenticated.
2. If **not authenticated**:

   * show only non-private ticket data — its number and description text;
   * display a message such as “To view the full content, please log in”;
   * include a “Log in / Sign up” button that leads to the auth page.
3. If the user is authenticated, keep the existing behavior (full access to the ticket).

This lets contractors immediately understand the issue without an account, while still encouraging them to log in for complete details (files, comments, etc.).

I will open a PR with the described changes in a few days.

Thank you!

toplenboren:
Hey, @paulo-rossy, thank you for the issue and 🚀 super-thanks 🚀  for the PR!

Just merged it into `master` branch.. Closing the issue as completed :-)

PR #6203, opened 2025-05-20
feat(condo): added public preview for shared tickets

paulo-rossy:
According to the issue #6181

**Background**
Before the change the /share page did only one thing – immediately redirected the visitor to /ticket/{id}.
As a result, contractors that do not have an account could not see anything – they landed on the login screen right away.

**How it works**
```
/share?q=<packed>&redirectToTicketPage=true   → always redirect to ticket page
/share?q=<packed>&redirectToTicketPage=false  → always show public preview ticket
/share?q=<packed>                            → fallback to config flag
```

Set
```env
SHOW_TICKET_INFO_ON_SHARE=true
```

image_2025-05-20_17-55-07

## Summary by CodeRabbit

- **New Features**
	- Added support for displaying public ticket information on the ticket sharing page, including ticket number, date, and problem description, with an option to view details.
	- Introduced a button to redirect users to the ticket detail page from the sharing view.
	- Enhanced social sharing with improved metadata.
	- Added configuration to control ticket info display behavior on the sharing page.

- **Localization**
	- Updated English, Spanish, and Russian translations for ticket sharing, providing separate messages for sharing with or without a date, and a new label for the "View details" button.

- **Refactor**
	- Improved the ticket sharing page layout and logic to support configurable display modes and clearer user experience.


Alexander-Turkin:
Hi! Good idea! But I see two potential issues:

Access – Unregistered users won’t be able to fetch ticket data.

Rate limits – Unauthenticated users have very strict rate limits, which could lead to frequent bans.

apps/condo/pages/share.tsx
```diff
@@ -59,39 +134,63 @@ const Share: PageComponentType<ShareProps> = ({ date, number, details, id }) =>
...
+    if (redirectToTicketPage === 'true') {
+        return (
+            <>
+                <HeadWithMeta />
+                <RedirectToTicket ticketId={id} />
+            </>
+        )
+    }
+    
+    if (redirectToTicketPage === 'false') {
```
pahaz:
```suggestion
    } else if (redirectToTicketPage === 'false') {
```

apps/condo/pages/share.tsx
```diff
@@ -59,39 +134,63 @@ const Share: PageComponentType<ShareProps> = ({ date, number, details, id }) =>
...
-        className='top-menu-only-layout'
-    >
-        <PageWrapper>
-            {children}
-        </PageWrapper>
-    </BaseLayout>
+    return {
+        props: {
+            ...shareParams,
+            redirectToTicketPage,
```
pahaz: You can obtain this value on the front end without handling it on the server side, if you prefer to avoid a server-side redirect.

It looks like you need to leve this function without any changes

apps/condo/pages/share.tsx
```diff
@@ -59,39 +134,63 @@ const Share: PageComponentType<ShareProps> = ({ date, number, details, id }) =>
         } = getConfig())
     }
 
+    const HeadWithMeta = () => <ShareMeta
+        title={ShareTitleMessage}
+        desc={ShareDetailsMessage}
+        image={`${origin}/logoSnippet.png`}
+    />
+
+    if (redirectToTicketPage === 'true') {
```
pahaz: You need to define a variable `shouldRedirect` like so:

```
const shouldRedirect = (redirectToTicketPage === 'false' || redirectToTicketPage === 'true') ? redirectToTicketPage === 'true' : !displayTicketInfoOnShare
```

paulo-rossy: done

apps/condo/pages/share.tsx
paulo-rossy: done

apps/condo/pages/share.tsx
```diff
@@ -29,20 +32,95 @@ function RedirectToTicket ({ ticketId }) {
     return <h1>{RedirectingMessage}</h1>
 }
 
+const ShareMeta = ({ title, desc, image }) => (
+    <Head>
+        <meta property='og:site_name' content={title} />
+        <meta property='og:title' content={title} />
+        <meta property='og:description' content={desc} />
+        <meta property='og:updated_time' content='14400000' />
```
toplenboren, line 40: Can you please explain why this number is chosen?

apps/condo/pages/share.tsx
```diff
@@ -29,20 +32,95 @@ function RedirectToTicket ({ ticketId }) {
...
+    </Head>
+)
+
+const TicketPublicInfo = ({ date, number, details, ticketId }) => {
+    const [redirect, setRedirect] = useState(false)
+
+    const intl = useIntl()
+    const locale = get(LOCALES, intl.locale)
+    const localizedDate = locale ? dayjs(date || 0).locale(locale) : dayjs(date || 0)
+    const dateFormatted = localizedDate.format('D MMMM YYYY')
```
toplenboren, line 51: Hmm, this date format is different for other countries.. I think locale should be taken in arguments of this function and date should be formatted accordingly

apps/condo/pages/share.tsx
```diff
@@ -59,39 +137,42 @@ const Share: PageComponentType<ShareProps> = ({ date, number, details, id }) =>
         } = getConfig())
     }
 
+    const HeadWithMeta = () => <ShareMeta
+        title={ShareTitleMessage}
+        desc={ShareDetailsMessage}
+        image={`${origin}/logoSnippet.png`}
```
toplenboren, line 143: Best practice would be to convert the image to webp or other lossy compression format (avif?)

toplenboren (approved):

apps/condo/pages/share.tsx
paulo-rossy: @toplenboren, hello, thanks for the review

The `14400000` value wasn’t introduced in this PR — it was copied verbatim from the original `apps/condo/pages/share.tsx` when I started the implementation for #6181 https://github.com/open-condo-software/condo/blob/a58281577705b448c7093ecb7e88a4acfa00099a/apps/condo/pages/share.tsx#L68
My current scope is limited to the unauth / auth split described in the ticket, so I deliberately left existing OG-tags as-is.

Agree that a hard-coded timestamp isn’t correct; happy to either
 1. remove the tag entirely, or
 2. replace it with a real `Math.floor(Date.now()/1000)` value in this PR or a follow-up one – whichever the team prefers.

 Let me know which option you’d like and I’ll update accordingly.

apps/condo/pages/share.tsx
paulo-rossy: @toplenboren thanks for catching that 

Just like with the `og:updated_time` value, this date-format block is **unchanged from the original `apps/condo/pages/share.tsx`** 
https://github.com/open-condo-software/condo/blob/a58281577705b448c7093ecb7e88a4acfa00099a/apps/condo/pages/share.tsx#L43

Because the goal of this PR is limited to showing partial ticket data to unauthenticated users, I left existing formatting as-is to avoid accidental regressions. If the team wants to make the format fully locale-aware (e.g. pass `locale` or `format` as an argument and use Day.js locale files), I can:

1. add a quick fix here, or
2. open a follow-up issue/PR so we can tackle it across the codebase consistently.

Let me know which path you prefer — happy to adjust!

apps/condo/pages/share.tsx
paulo-rossy: @toplenboren thanks 
The reference to `logoSnippet.png` is also legacy code — I copied it unchanged from the original **apps/condo/pages/share.tsx** 

I agree that switching to a WebP/AVIF (or adding automatic conversion in the build) is a good improvement. I can:
1. replace the asset and update the pipeline right in this PR, **or**
2. open a follow-up issue/PR so we handle image formats consistently across the repo.

Just let me know which way you’d like to proceed

PR #6244, opened 2025-05-29
fix(condo): DOMA-11656 fix meter create and update validations

abshnko:
1. added validation of accountNumber when either unitName or unitType were changed
2. added check for meter with same accountNumber, but in a different unitType (like we use on server side)

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying unit type when creating or updating meter records.
  - Forms now automatically validate the account number field when unit name or unit type changes.

- **Improvements**
  - Enhanced validation logic to prevent duplicate account numbers across different unit types.
  - Reorganized meter forms for better structure and improved dependency tracking.

- **Removals**
  - Removed modal forms and hooks related to meter creation and updating, simplifying the meter management interface.


apps/condo/domains/meter/components/Meters/BaseMetersFormFields.tsx
```diff
@@ -126,17 +127,17 @@ export const BaseMetersFormFields: React.FC<BaseMetersFormFieldsProps> = ({
         earlierThanFirstVerificationDateValidator,
         meterWithSameAccountNumberInOtherUnitValidation,
         meterResourceOwnerValidation,
-    } = useMeterValidations(isPropertyMeter, installationDate, verificationDate, propertyId, initialUnitName, organizationId, initialMeterNumber, addressKey)
+    } = useMeterValidations(isPropertyMeter, installationDate, verificationDate, propertyId, unitName, organizationId, initialMeterNumber, addressKey, unitType)
```
YEgorLu, line 130: useMeterValidations expects non nullable unitName and unitType and adds it in "where" query.

Maybe change types to be nullable or add "?? '' " as coderabbit comment suggests

YEgorLu (approved):

PR #6281, opened 2025-06-10
feat(condo): INFRA-1009 custom file field

sitozzz:


apps/condo/bin/prepare.js
```diff
@@ -8,11 +10,18 @@ async function updateAppEnvAddressSuggestionConfig (serviceName) {
     await updateAppEnvFile(serviceName, 'FAKE_FINANCE_INFO_CLIENT', 'true')
 }
 
+async function updateAppEnvFileClients (appName) {
+    await updateAppEnvFile(appName, 'FILE_APP_CLIENTS', JSON.stringify({
+        appName: { name: `${appName}-app`, secret: uuid() },
```
SavelevMatthew:
```suggestion
        [appName]: { secret: uuid() },
```

apps/condo/domains/common/schema/file.test.js
```diff
@@ -0,0 +1,23 @@
+const path = require('path')
+
+const conf = require('@open-condo/config')
+const { FileMiddlewareUnitTests } = require('@open-condo/files/fileMiddleware.test')
+const { FileMiddlewareTests } = require('@open-condo/files/schema/models/index')
+const { FileMiddlewareUtilsTests } = require('@open-condo/files/utils.test')
```
SavelevMatthew:
```suggestion
const { FileMiddlewareRoutesTests } = require('@open-condo/files/fileMiddleware.test')
const { FileMiddlewareE2ETests } = require('@open-condo/files/schema/models/index')
const { FileMiddlewareUtilsTests } = require('@open-condo/files/utils.test')
```

apps/condo/index.js
```diff
@@ -7,6 +7,8 @@ const utc = require('dayjs/plugin/utc')
 
 const conf = require('@open-condo/config')
 const { FeaturesMiddleware } = require('@open-condo/featureflags/FeaturesMiddleware')
+const { FileMiddleware } = require('@open-condo/files/fileMiddleware')
+const { getFileModels } = require('@open-condo/files/schema/models/index')
```
SavelevMatthew:
```suggestion
const { getFileModels } = require('@open-condo/files/schema/models')
```

packages/files/schema/models/CondoFile.js
```diff
@@ -0,0 +1,43 @@
+const { historical, versioned, uuided, tracked, softDeleted, dvAndSender } = require('@open-condo/keystone/plugins')
+const { GQLListSchema } = require('@open-condo/keystone/schema')
+
+const CondoFile = new GQLListSchema('CondoFile', {
```
SavelevMatthew:
```suggestion
const File = new GQLListSchema('File', {
```

packages/files/schema/models/CondoFile.js
```diff
@@ -0,0 +1,43 @@
...
+    schemaDoc: 'File uploaded to platform and meta with ownership data',
+    fields: {
+        file: {
```
SavelevMatthew:
```suggestion
        fileMeta: {
```

packages/files/schema/models/CondoFile.js
```diff
@@ -0,0 +1,43 @@
...
+            type: 'Json',
```
SavelevMatthew: Should this be typed?

packages/files/schema/models/CondoFile.js
```diff
@@ -0,0 +1,43 @@
...
+            isRequired: true,
+            schemaDoc: 'File uploaded to platform',
```
SavelevMatthew:
```suggestion
            schemaDoc: 'Information about file including its encoding, mime type, filename and ...',
```

packages/files/schema/models/CondoFile.js
```diff
@@ -0,0 +1,43 @@
...
+        },
+        signature: {
+            type: 'Text',
+            isRequired: true,
+            schemaDoc: 'File signature created from user and storage meta and application secret',
+        },
```
SavelevMatthew: redundant?

packages/files/schema/models/CondoFile.js
```diff
@@ -0,0 +1,43 @@
...
+        user: {
+            type: 'Relationship',
+            ref: 'User',
+            schemaDoc: 'User who uploaded the file',
+            isRequired: true,
+            knexOptions: { isNotNullable: true },
+            kmigratorOptions: { null: false, on_delete: 'models.PROTECT' },
+        },
+        attach: {
```
SavelevMatthew:
```suggestion
        isAttached: {
```

SavelevMatthew: Why do we use it?

packages/files/schema/models/CondoFile.js
```diff
@@ -0,0 +1,43 @@
...
+            knexOptions: { isNotNullable: false },
+        },
+    },
+    plugins: [uuided(), versioned(), tracked(), softDeleted(), dvAndSender(), historical()],
+    access: {
+        read: false,
+        create: false,
+        update: false,
+        delete: false,
+        auth: true,
```
SavelevMatthew:
```suggestion
        auth: false,
```

packages/files/schema/models/index.js
```diff
@@ -0,0 +1,11 @@
+const { CondoFile } = require('./CondoFile')
+const { FileMiddlewareTests } = require('./file.test')
+
+const getFileModels = () => {
+    return { CondoFile }
+}
+
+module.exports = {
+    getFileModels,
+    FileMiddlewareTests,
```
SavelevMatthew: R u sure its good to export tests from models folder

packages/files/schema/gql.js
SavelevMatthew, line 1: redundant file?

packages/files/fileMiddleware.js
```diff
@@ -0,0 +1,107 @@
...
+    validateAndParseAppClients,
+} = require('./utils')
+
+const DEFAULT_USER_QUOTA = 100
+const DEFAULT_IP_QUOTA = 100
+
+
+class FileMiddleware {
+    constructor ({
+        apiUrl = '/api/files',
```
SavelevMatthew:
```suggestion
        apiPrefix = '/api/files',
```

packages/files/fileMiddleware.js
```diff
@@ -0,0 +1,107 @@
...
+        maxFieldSize = 200 * 1024 * 1024,
+        maxFileSize = 200 * 1024 * 1024,
```
SavelevMatthew:
```suggestion
        maxFileSizeMb = 200 * 1024 * 1024,
```

SavelevMatthew: also our body limit currently is 100mb, why 200 is here?

packages/files/fileMiddleware.js
```diff
@@ -0,0 +1,107 @@
...
+} = require('./utils')
+
+const DEFAULT_USER_QUOTA = 100
+const DEFAULT_IP_QUOTA = 100
+
+
+class FileMiddleware {
+    constructor ({
+        apiUrl = '/api/files',
+        maxFieldSize = 200 * 1024 * 1024,
```
SavelevMatthew: R us sure json can be 200 mb?

packages/files/fileMiddleware.js
```diff
@@ -0,0 +1,107 @@
...
+        this.processRequestOptions = { maxFieldSize, maxFileSize, maxFiles }
+
+        this.quota = this.loadQuota()
+        this.appClients = this.loadAppClients()
+    }
+
+    loadQuota () {
+        let quota
+        try {
+            const raw = conf['FILE_QUOTA']
```
SavelevMatthew: FILE_UPLOAD_CONFIG = { quota, clients }

packages/files/package.json
```diff
@@ -0,0 +1,28 @@
...
+  "name": "@open-condo/files",
+  "version": "1.0.0",
+  "author": {
+    "name": "Ivan Sitnikov",
+    "email": "isitnikov@doma.ai",
+    "url": "https://github.com/sitozzz"
+  },
+  "license": "MIT",
+  "exports": {
+    "./*": "./*.js"
```
SavelevMatthew: It's better to explicitly specify exports

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+const { CondoFile } = require('@open-condo/files/schema/utils/serverSchema')
+const FileAdapter = require('@open-condo/keystone/fileAdapter/fileAdapter')
+const { getKVClient } = require('@open-condo/keystone/kv')
+const { getLogger } = require('@open-condo/keystone/logging')
+
+const logger = getLogger('file-middleware')
+
+const AppClientSchema = z.object({
+    name: z.string().min(3),
+    modelNames: z.array(z.string()).min(1).optional(),
```
SavelevMatthew: unused?

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+/** Validate and parse FILE_APP_CLIENTS. Returns the validated object. */
+function validateAndParseAppClients (data) {
+    return AppClientsSchema.parse(data)
+}
+
+// ---------------------- utilities ----------------------
+
+const FINGERPRINT_RE = /^[a-zA-Z0-9!#$%()*+-;=,:[\]/.?@^_`{|}~]{5,42}$/
+
+function sendError (res, status, message) {
```
SavelevMatthew: all you need is to call next with GQLError

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+    logger.error({ msg: message })
+    res.status(status).json({ error: message })
+}
+
+function extractRequestIp (req) {
+    return String(req.ip || '').split(':').pop()
```
SavelevMatthew:
```suggestion
    return req.ip
```

SavelevMatthew: `::1` -> `1`

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+}
+
+// ---------------------- guards ----------------------
+
+class RedisGuard {
+    constructor () {
+        this.counterPrefix = 'guard_counter:'
```
SavelevMatthew, line 61: this.#kv = getKVClient('guards')

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+    res.status(status).json({ error: message })
+}
+
+function extractRequestIp (req) {
+    return String(req.ip || '').split(':').pop()
+}
+
+// ---------------------- guards ----------------------
+
+class RedisGuard {
```
SavelevMatthew, line 59: can we reuse single guard?

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+        if (!this._redis) this._redis = getKVClient('guards')
+        return this._redis
+    }
+
+    async incrementHourCounter (variable) {
+        return this.incrementCustomCounter(variable, 3600)
+    }
+
+    async incrementCustomCounter (variable, ttl) {
+        let afterIncrement = await this.redis.incr(`${this.counterPrefix}${variable}`)
```
SavelevMatthew: incr return number by default

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+        if (afterIncrement === 1) {
+            await this.redis.expire(`${this.counterPrefix}${variable}`, ttl)
+        }
+        return afterIncrement
+    }
+}
+
+// ---------------------- validation ----------------------
+
+function parseAndValidateMeta (raw, req, res, onError) {
```
SavelevMatthew:
```suggestion
function parseAndValidateFileMeta (raw, req, res, onError) {
```

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+
+    if (meta.authedItem !== req.user.id) {
+        return onError(() => sendError(res, 403, 'Wrong authedItem. Unable to upload file for another user'))
+    }
+
+    if (!meta.appId) {
+        return onError(() => sendError(res, 400, 'Missing appId field for meta object'))
+    }
+
+    if (!meta.modelNames) {
```
SavelevMatthew: zod schema?

SavelevMatthew: Array.isArray?

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+
+const authHandler = () => async (req, res, next) => {
+    if (!req.user || req.user.deletedAt !== null) {
+        logger.error({ msg: 'auth handler error' })
+        return res.status(403).json({ error: 'Authorization is required' })
+    }
+    next()
+}
+
+const rateLimitHandler = ({ quota, guard }) => async (req, res, next) => {
```
SavelevMatthew:
```suggestion
const rateLimitHandler = ({ quota, guard }) => async function fileRateLimit (req, res, next) {
```

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+        return res.status(429).json({ error: 'You have reached the request limit, try again later.' })
+    }
+
+    next()
+}
+
+const parserHandler = ({ processRequestOptions }) => {
+    return async function (req, res, next) {
+        if (!req.is('multipart/form-data')) {
+            return res.status(405).json({ error: 'Wrong request method type. Only "multipart/form-data" is allowed' })
```
SavelevMatthew: why method not allowed here?

SavelevMatthew: 400?

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+
+    const ipCounter = await guard.incrementHourCounter(`file:${requestIp}`)
+    if (ipCounter > quota.ip) {
+        return res.status(429).json({ error: 'You have reached the request limit, try again later.' })
+    }
+
+    next()
+}
+
+const parserHandler = ({ processRequestOptions }) => {
```
SavelevMatthew: Please add notes to SRC

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+    return async function (req, res, next) {
```
SavelevMatthew, line 229: named function

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+
+        const condoFiles = await CondoFile.createMany(
+            context,
+            savedFiles.map((data, index) => {
+                const fileData = {
+                    ...data,
+                    originalFilename: files[index].filename,
+                    mimetype: files[index].mimetype,
+                    encoding: files[index].encoding,
+                    meta,
```
SavelevMatthew: can we store info about destination, like s3 provider / bucket?

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+        })
+
+        res.once('close', () => { released = true })
+        req.pipe(parser)
+    }
+}
+
+const fileStorageHandler = ({ keystone, appClients }) => {
+    return async function (req, res) {
+        const { meta, files } = req
```
SavelevMatthew: Maybe more scecific field on req instead of meta

SavelevMatthew: const key = Symbol('fileMeta')
req[key]

[[['fileMeta']]]

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+                        sender: meta.sender,
+                        signature,
+                        user: { connect: { id: req.user.id } },
+                    },
+                }
+            }),
+            'signature file { id }'
+        )
+
+        res.json(condoFiles)
```
SavelevMatthew: How about be more GQL-like, 
```json
{
  data: {
     files: []
   }
}
```

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+    return async function (req, res) {
+        const data = req.body
+        if (!data.signature) {
+            return res.status(403).json({ error: 'Missing id or signature in request body' })
+        }
+
+        const context = await keystone.createContext({ skipAccessControl: true })
+        const file = await CondoFile.getOne(
+            context,
+            { signature: data.signature },
```
SavelevMatthew: is it guaranteed to be unique

packages/files/utils.js
```diff
@@ -0,0 +1,388 @@
...
+                }
+            }),
+            'signature file { id }'
+        )
+
+        res.json(condoFiles)
+    }
+}
+
+const attachHandler = ({ keystone }) => {
```
SavelevMatthew: We need to think about this. 

Our clients might not use this endpoint, and we does not notify them in current implementation, what file can be deleted

packages/keystone/KSv5v6/v5/keystone.js
```diff
@@ -1,13 +1,39 @@
 const { Keystone: DefaultKeystone } = require('@open-keystone/keystone')
+const { GraphQLScalarType, GraphQLError } = require('graphql')
+const Upload = require('graphql-upload/Upload.js')
 
 const { _patchResolverWithGQLContext } = require('../utils/resolvers')
 
+
+const CustomFileScalar = new GraphQLScalarType({
```
SavelevMatthew: move to original KS?

packages/keystone/fields/CustomFile/Implementation.js
```diff
@@ -0,0 +1,184 @@
...
+                    type: 'ACCESS_DENIED',
+                    variable: [this.path],
+                    message: 'Owner of this file restrict connection to this model',
+                })
+            }
+
+            let condoFile
+
+            if (this.fileServiceUrl) {
+                const response = await fetch(this.fileServiceUrl + '/api/files/attach', {
```
SavelevMatthew: We need to think about it, maybe task or remove?

sitozzz: I think at the moment it should be removed, because it is not obvious who will use it except for its own developers

packages/files/utils.js
sitozzz: Yep, i'll remove it

packages/files/utils.js
sitozzz: Good point, I'll rewrite this to internal id

packages/files/schema/gql.js
sitozzz: Yep, removed

packages/files/schema/models/CondoFile.js
sitozzz: I can't rename it to just File due to conflict with same type. So I decided to move it to FileRecord

packages/files/schema/models/CondoFile.js
sitozzz: Done

packages/files/schema/models/CondoFile.js
sitozzz: Yep, removed

packages/files/schema/models/CondoFile.js
sitozzz: Actually not use it anymore)

packages/files/schema/models/index.js
sitozzz: Moved to the host dir

packages/files/fileMiddleware.js
sitozzz: Good point

packages/files/fileMiddleware.js
sitozzz: Fixed it

packages/files/fileMiddleware.js
sitozzz: Ofc not, I already changed this line

packages/files/package.json
sitozzz: Done

packages/files/fileMiddleware.js
sitozzz: Done

packages/files/utils.js
sitozzz: Do a massive refactor for that. Also change tests to check error format, but I need to add expressErrorHandler at the end of the my middleware, because it connected slightly different way

packages/files/utils.js
sitozzz: Don't really get it, but I'm completely remove this helper

packages/files/utils.js
sitozzz: Okay

packages/files/utils.js
sitozzz: I think so, but not in this PR

packages/files/utils.js
sitozzz: Added provider type to the metadata of file

sitozzz:
@coderabbitai pause

apps/condo/bin/prepare.js
```diff
@@ -1,3 +1,5 @@
+const { v4: uuid } = require('uuid')
```
SavelevMatthew: use the one from miniapp-utils instead

apps/condo/bin/prepare.js
```diff
@@ -8,11 +10,21 @@ async function updateAppEnvAddressSuggestionConfig (serviceName) {
     await updateAppEnvFile(serviceName, 'FAKE_FINANCE_INFO_CLIENT', 'true')
 }
 
+async function updateAppEnvFileClients (appName) {
+    await updateAppEnvFile(appName, 'FILE_UPLOAD_CONFIG', JSON.stringify({
+        clients: {
+            condo: { secret: uuid() },
+            miniapp: { secret: uuid() },
```
SavelevMatthew: can individual apps do this instead? like preparing themself by adding something in condo?

apps/condo/index.js
```diff
@@ -64,6 +66,7 @@ const schemas = () => [
     require('@condo/domains/document/schema'),
     require('@condo/domains/ai/schema'),
     getWebhookModels('@app/condo/schema.graphql'),
+    getFileModels(),
```
SavelevMatthew: Why do you need to call getter here? why not just require it like a schemas?

packages/files/schema/models/FileRecord.js
```diff
@@ -0,0 +1,57 @@
...
+    schemaDoc: 'Stores uploaded file meta data and owner',
+    fields: {
+        fileMeta: {
+            type: 'Json',
+            isRequired: true,
+            schemaDoc: 'Information about file including its encoding, mime type, filename and user related metadata',
+            graphQLReturnType: 'FileRecordMeta',
+            extendGraphQLTypes: [
+                'type FileSender { dv: Int!, fingerprint: String! }',
+                'type FileRecordUserMeta { dv: Int!, sender: FileSender!, authedItem: String!, appId: String!, modelNames: [String!]!, fileAdapter: String, sourceAppId: String }',
```
SavelevMatthew: strange mixing of `somethingId` and `something`
```suggestion
                'type FileRecordUserMeta { dv: Int!, sender: FileSender!, authedItemId: String!, appId: String!, modelNames: [String!]!, fileAdapter: String, sourceAppId: String }',
```

SavelevMatthew: What's the sourceAppId used for?

packages/files/schema/models/FileRecord.js
```diff
@@ -0,0 +1,57 @@
...
+        },
+        user: {
+            type: 'Relationship',
+            ref: 'User',
+            schemaDoc: 'User who uploaded the file',
+            isRequired: true,
+            knexOptions: { isNotNullable: true },
+            kmigratorOptions: { null: false, on_delete: 'models.PROTECT' },
+        },
+        fileKey: {
```
SavelevMatthew: If this is separate field, why adapter itself is not?

packages/files/utils.js
```diff
@@ -0,0 +1,457 @@
...
+    }).optional().default({
+        user: DEFAULT_USER_HOUR_QUOTA,
+        ip: DEFAULT_IP_HOUR_QUOTA,
+    }),
+})
+
+function validateAndParseFileConfig (config) {
+    const { data, success, error } = AppConfigSchema.safeParse(config)
+    if (!success) {
+        console.error(error.issues[0]?.message || 'Invalid file upload config')
```
SavelevMatthew: Its better to use z.prettifyError(error) instead of this

packages/files/schema/models/FileRecord.js
sitozzz: sourceAppId is a marker to original application id, from which it was shared

packages/files/schema/models/FileRecord.js
sitozzz: I move this one to a separate field for ease of searching in the database. So search by jsonb filed works not so good

sitozzz: I don't think there is a need to save adapter as separate field so

packages/files/utils.js
sitozzz: Good point, I'll refactor this one

apps/condo/index.js
sitozzz: Firstly I make it look like webhookmodels

packages/files/schema/models/FileRecord.js
```diff
@@ -0,0 +1,57 @@
...
+    fields: {
+        fileMeta: {
+            type: 'Json',
+            isRequired: true,
+            schemaDoc: 'Information about file including its encoding, mime type, filename and user related metadata',
+            graphQLReturnType: 'FileRecordMeta',
+            extendGraphQLTypes: [
+                'type FileSender { dv: Int!, fingerprint: String! }',
+                'type FileRecordUserMeta { dv: Int!, sender: FileSender!, authedItemId: ID!, appId: String!, modelNames: [String!]!, fileAdapter: String, sourceAppId: String }',
+                'type FileRecordMeta { id: ID, shareId: String, path: String, filename: String!, originalFilename: String, mimetype: String!, encoding: String!, meta: FileRecordUserMeta! }',
```
SavelevMatthew:
```suggestion
                'type FileRecordMeta { id: ID, recordId: String, path: String, filename: String!, originalFilename: String, mimetype: String!, encoding: String!, meta: FileRecordUserMeta! }',
```

packages/files/schema/models/FileRecord.js
```diff
@@ -0,0 +1,57 @@
...
+    schemaDoc: 'Stores uploaded file meta data and owner',
+    fields: {
+        fileMeta: {
+            type: 'Json',
+            isRequired: true,
+            schemaDoc: 'Information about file including its encoding, mime type, filename and user related metadata',
+            graphQLReturnType: 'FileRecordMeta',
+            extendGraphQLTypes: [
+                'type FileSender { dv: Int!, fingerprint: String! }',
+                'type FileRecordUserMeta { dv: Int!, sender: FileSender!, authedItemId: ID!, appId: String!, modelNames: [String!]!, fileAdapter: String, sourceAppId: String }',
```
SavelevMatthew: fileAdapter is not related to user

packages/files/fileMiddleware.test.js
```diff
@@ -0,0 +1,553 @@
...
+        // Clean rate limits
+        const kv = getKVClient('guards')
+
+        const client = await makeClient()
+        serverUrl = client.serverUrl + '/api/files/upload'
+        serverShareUrl = client.serverUrl + '/api/files/share'
+        admin = await makeLoggedInAdminClient()
+        filestream = fs.readFileSync(testFile)
+
+        console.log(admin.user.id)
```
SavelevMatthew: ?

packages/files/fileMiddleware.test.js
```diff
@@ -0,0 +1,553 @@
...
+                    method: 'POST',
+                    body: form,
+                })
+                const json = await result.json()
+
+                expect(result.status).toEqual(401)
+                expect(json).toEqual({
+                    errors: [
+                        expect.objectContaining({
+                            name: 'GQLError',
```
SavelevMatthew: It's better to check error type too

packages/files/fileMiddleware.test.js
```diff
@@ -0,0 +1,553 @@
...
+                const json = await result.json()
+                expect(result.status).toEqual(401)
+                expect(json).toEqual({
+                    errors: [
+                        expect.objectContaining({
+                            name: 'GQLError',
+                            message: 'Authorization is required',
+                        }),
+                    ],
+                })
```
SavelevMatthew: expect code and type as well

packages/files/utils.js
```diff
@@ -0,0 +1,457 @@
...
+        const fileRecords = await FileRecord.updateMany(context,
+            createdFiles.map(e => ({
+                id: e.id, data: { fileMeta: { ...e.fileMeta, shareId: e.id }, dv: meta.dv, sender: meta.sender },
+            })), `id fileKey fileMeta ${FILE_RECORD_META_FIELDS}`)
+
+        res.json({
+            data: {
+                files: fileRecords.map(file => ({
+                    ...file,
+                    signature: jwt.sign(
```
SavelevMatthew, line 429: explicitly specify singing algorithm

packages/files/fileMiddleware.test.js
```diff
@@ -0,0 +1,553 @@
...
+
+                const json = await result.json()
+
+                expect(result.status).toEqual(200)
+                expect(json.data.files).toHaveLength(1)
+                const file = json.data.files[0]
+                expect(file).toHaveProperty('signature')
+                expect(file).toHaveProperty('id')
+                expect(file).toHaveProperty('fileMeta')
+                expect(file.fileMeta).toHaveProperty(['meta', 'fileAdapter'], conf['FILE_FIELD_ADAPTER'])
```
SavelevMatthew: we should hide it from user imho

SavelevMatthew: Infra details should not be shown to user

packages/files/fileMiddleware.test.js
```diff
@@ -0,0 +1,553 @@
...
+                    body: form,
+                    headers: { Cookie: user.getCookie() },
+                })
+
+                const json = await result.json()
+
+                expect(result.status).toEqual(200)
+                expect(json.data.files).toHaveLength(1)
+                const file = json.data.files[0]
+                expect(file).toHaveProperty('signature')
```
SavelevMatthew: more strict expect

SavelevMatthew: expect(response).toEqual({})

SavelevMatthew: to check response shape

packages/files/fileMiddleware.test.js
```diff
@@ -0,0 +1,553 @@
...
+                        modelNames: ['AnotherModel'],
+                        ...DV_AND_SENDER,
+                    }),
+                })
+
+                expect(shareResult.status).toEqual(200)
+                const shareResultJson = await shareResult.json()
+                expect(shareResultJson).toHaveProperty(['data', 'file', 'id'])
+                const sharedFile = shareResultJson.data.file
+                expect(sharedFile).toHaveProperty('signature')
```
SavelevMatthew: more sctirct response shape check please

packages/files/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@open-condo/files",
+  "version": "1.0.0",
```
SavelevMatthew:
```suggestion
  "version": "0.0.0-development",
```

packages/files/utils.js
```diff
@@ -0,0 +1,457 @@
...
+
+const { ERRORS } = require('./errors')
+
+const DEFAULT_USER_HOUR_QUOTA = 100
+const DEFAULT_IP_HOUR_QUOTA = 100
+const UUID_REGEX = /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$/
+
+const AppClientSchema = z.object({
+    name: z.string().min(3).optional(),
+    modelNames: z.array(z.string()).min(1).optional(),
```
SavelevMatthew: redundant?

packages/files/utils.js
```diff
@@ -0,0 +1,457 @@
...
+        } catch {
+            return onError(() => sendError(req, next, { ...ERRORS.INVALID_META, message:'Invalid type for the "meta" multipart field' }))
+        }
+    } else if (typeof raw !== 'object' || raw === null) {
+        return onError(() => sendError(req, next, { ...ERRORS.INVALID_META, message:'Invalid type for the "meta" multipart field' }))
+    }
+
+    const result = MetaSchema.safeParse(candidate)
+    if (!result.success) {
+        const message = result.error.issues[0]?.message
```
SavelevMatthew:
```suggestion
        const error = new Error(z.prettifyError(result.error))
        next(new GQLError(ERRORS_INVALID_META, { req }, [error])
```

packages/files/utils.js
```diff
@@ -0,0 +1,457 @@
...
+            return next(error)
+        }
+
+        next()
+    }
+}
+
+function parserHandler ({ processRequestOptions }) {
+    return async function (req, res, next) {
+        if (!req.is('multipart/form-data')) {
```
SavelevMatthew, line 250: SRC comment here

packages/files/utils.js
```diff
@@ -0,0 +1,457 @@
...
+        })
+
+        res.once('close', () => { released = true })
+        req.pipe(parser)
+    }
+}
+
+function fileStorageHandler ({ keystone, appClients }) {
+    return async function (req, res, next) {
+        const { meta, files } = req
```
SavelevMatthew: maybe make meta less obvious

packages/keystone/fields/CustomFile/Implementation.js
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    async resolveInput ({ resolvedData, existingItem, context, listKey }) {
+        const uploadData = resolvedData[this.path]
+        // New way to 'upload' - connect file
+        if (get(uploadData, 'signature')) {
+
+            let file
+            try {
+                file = jwt.verify(uploadData['signature'], this._fileSecret)
```
SavelevMatthew: Should we check shape here?

packages/keystone/fileAdapter/fileAdapter.js
SavelevMatthew, line 1: What will be source in case of chained sharing?
`file <- copy1 <- copy2` or `file <- copy1` + `file <- copy2`

packages/keystone/KSv5v6/v5/keystone.js
```diff
@@ -1,12 +1,32 @@
...
+    },
+})
+
+
 class Keystone extends DefaultKeystone {
     getResolvers ({ schemaName }) {
         const originalResolvers = super.getResolvers({ schemaName })
 
-        return _patchResolverWithGQLContext(originalResolvers)
+        return _patchResolverWithGQLContext({ ...originalResolvers, FileUpload: CustomFileScalar })
```
SavelevMatthew: why not registering it officially?

packages/files/schema/models/FileRecord.js
```diff
@@ -0,0 +1,74 @@
...
+    fields: {
+        fileMeta: {
+            type: 'Json',
+            isRequired: true,
+            schemaDoc: 'Information about file including its encoding, mime type, filename and user related metadata',
+            graphQLReturnType: 'FileRecordMeta',
+            extendGraphQLTypes: [
+                'type FileSender { dv: Int!, fingerprint: String! }',
+                'type FileRecordUserMeta { dv: Int!, sender: FileSender!, authedItemId: ID!, appId: String!, modelNames: [String!]!, sourceAppId: String }',
+                'type FileRecordMeta { id: ID, fileAdapter: String, recordId: ID, path: String, filename: String!, originalFilename: String, mimetype: String!, encoding: String!, meta: FileRecordUserMeta! }',
```
SavelevMatthew: why id, fileAdapter and recordId is not mandatory? Every file should have ones, no?

packages/files/utils.js
```diff
@@ -0,0 +1,486 @@
...
+        }
+    } else if (typeof raw !== 'object' || raw === null) {
+
+        return onError(() => next(new GQLError(ERRORS.INVALID_META, { req })))
+    }
+
+    const result = MetaSchema.safeParse(candidate)
+    if (!result.success) {
+
+        return onError(() => next(new GQLError(ERRORS.INVALID_META, { req }, [result.error])))
```
SavelevMatthew:
```suggestion
        return onError(() => next(new GQLError(ERRORS.INVALID_META, { req }, [new Error(z.prettifyError(result.error))])))
```

SavelevMatthew: Because original error contains too much fields (avoid giga-logs)

sitozzz: Done

packages/files/schema/models/FileRecord.js
sitozzz: Yep, that not necessary so I make it required here

packages/keystone/KSv5v6/v5/keystone.js
sitozzz: Because there is no way to do this without patch a k lib)

packages/files/schema/models/CondoFile.js
sitozzz: Done that btw

packages/files/utils.js
sitozzz: added

packages/files/utils.js
sitozzz: This is done

packages/keystone/KSv5v6/v5/keystone.js
sitozzz: Maybe later

apps/condo/bin/prepare.js
sitozzz: Good take

packages/files/fileMiddleware.js
```diff
@@ -0,0 +1,90 @@
...
+        this.processRequestOptions = { maxFieldSize, maxFileSize, maxFiles }
+        const { clients, quota } = this.loadConfig()
+        this.quota = quota
+        this.appClients = clients
+    }
+
+    loadConfig () {
+        const config = conf['FILE_UPLOAD_CONFIG']
+        if (!config) {
+            console.warn('File upload config not set, disabling v2 upload')
```
pahaz, line 35: logging?

packages/files/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "./fileMiddleware.test": "./fileMiddleware.test.js",
+    "./utils": "./utils.js",
+    "./utils.test": "./utils.test.js"
+  },
+  "dependencies": {
+    "@open-condo/config": "workspace:^",
+    "@open-condo/miniapp-utils": "workspace:^",
+    "busboy": "^1.6.0",
+    "express": "^4.21.2",
+    "fs-capacitor": "^6.2.0",
```
pahaz: its just 30+ stars and 200 LOC: https://github.com/mike-marcacci/fs-capacitor

packages/files/schema/models/FileRecord.js
```diff
@@ -0,0 +1,74 @@
...
+        },
+        user: {
+            type: 'Relationship',
+            ref: 'User',
+            schemaDoc: 'User who uploaded the file',
+            isRequired: true,
+            knexOptions: { isNotNullable: true },
+            kmigratorOptions: { null: false, on_delete: 'models.PROTECT' },
+        },
+        sourceId: {
```
pahaz: May be `sourceFile`? or sourceFileRecord?

pahaz: We don't use `Id` suffix for relations

packages/keystone/fileAdapter/sberCloudFileAdapter.js
```diff
@@ -153,26 +156,42 @@ class SberCloudFileAdapter {
...
         // 1. we want file urls to point to our server
         // 2. if we give access token to file here - then the user can face such scenario:
         //    user opens ticket page
         //    then after 5 minutes, he decides to download the file and click on the URL
         //    the token is expired - user needs to reload the page to generate a new access token
+        let folder = this.folder
+        let sign
+        if ('meta' in props && props['meta']['appId']) {
+            folder = props['meta']['appId']
+            sign = jwt.sign({ id: props.id, filename, appId: props.meta.appId, user }, conf['FILE_SECRET'], { expiresIn: '1m', algorithm: 'HS256' })
```
pahaz, line 185: What if the FILE_SECRET is empty or undefined?

packages/keystone/fileAdapter/fileAdapter.js
```diff
@@ -31,20 +32,86 @@ class NoFileAdapter {
...
+    publicUrl ({ filename, ...props }, user) {
+        if ('meta' in props && props['meta']['appId']) {
+            const meta = props['meta']
+
+            const appId = meta.sourceAppId || meta.appId
+            let sign
+            if (user !== null) {
+                sign = jwt.sign(
+                    { id: props.id, user, appId },
+                    conf['FILE_SECRET'],
```
pahaz, line 54: What if the FILE_SECRET is empty or undefined?

packages/keystone/fileAdapter/awsFileAdapter.js
```diff
@@ -118,18 +120,35 @@ class AwsFileAdapter {
         return `${id}${path.extname(originalFilename).replace(forbiddenCharacters, '')}`
     }
 
-    publicUrl ({ filename, originalFilename }) {
+    publicUrl ({ filename, originalFilename, ...props }, user) {
+        let folder = this.folder
+        let sign
+        if ('meta' in props && props['meta']['appId']) {
+            folder = props['meta']['appId']
+            sign = jwt.sign({ id: props.id, filename, appId: props.meta.appId, user }, conf['FILE_SECRET'], { expiresIn: '1m', algorithm: 'HS256' })
```
pahaz, line 131: What if the FILE_SECRET is empty or undefined?

packages/keystone/fileAdapter/sberCloudFileAdapter.js
```diff
@@ -190,12 +209,13 @@ class SberCloudFileAdapter {
...
 
     const obsConfig = SBERCLOUD_OBS_CONFIG ? JSON.parse(SBERCLOUD_OBS_CONFIG) : {}
+    const appClients = conf['FILE_APP_CLIENTS'] ? JSON.parse(conf['FILE_APP_CLIENTS']) : {}
     const Acl = new SberCloudObsAcl(obsConfig)
 
     return async function (req, res, next) {
         try {
             if (!req.user) {
-            // TODO(zuch): Ask where error pages are located in keystone - 403 is probably missing
+                // TODO(zuch): Ask where error pages are located in keystone - 403 is probably missing
```
pahaz: you can remove this comment. looks outdated

packages/keystone/fileAdapter/sberCloudFileAdapter.js
```diff
@@ -205,95 +225,134 @@ const obsRouterHandler = ({ keystone }) => {
...
+                    hasAccessToReadFile = items.length > 0
+                }
+
+                if (itemId && !hasAccessToReadFile) {
+                    let returnFields = 'id'
 
-            if (itemId && !hasAccessToReadFile) {
-                let returnFields = 'id'
+                    // for checking property we have to include property name in the return fields list
+                    if (isString(propertyQuery)) {
```
pahaz, line 280: Looks like propertyQuery injection. Need to validate it

packages/keystone/KSv5v6/v5/keystone.js
```diff
@@ -1,12 +1,34 @@
 const { Keystone: DefaultKeystone } = require('@open-keystone/keystone')
+const { GraphQLScalarType, GraphQLError } = require('graphql')
+const Upload = require('graphql-upload/Upload.js')
 
 const { _patchResolverWithGQLContext } = require('../utils/resolvers')
 
+
+const FileMetaScalar = new GraphQLScalarType({
```
pahaz, line 8: It's not clear. One of the main reasons to move the file upload logic to one URL is to restrict GQL style upload. Probably you are trying to add backward compatibility for existing uploads. But it's not clear for me.

pahaz: Can we add an environment variable to disable the legacy backward‑compatibility logic? Like FILE_COMPATIBILITY_MODE=1

sitozzz: > It's not clear. One of the main reasons to move the file upload logic to one URL is to restrict GQL style upload. Probably you are trying to add backward compatibility for existing uploads. But it's not clear for me.

To be honest I don't understand the question. You just answered it yourself.
I don't see any other options for solving this problem - using Scalar looks neat and not bulky

sitozzz: > Can we add an environment variable to disable the legacy backward‑compatibility logic? Like FILE_COMPATIBILITY_MODE=1

I can do it, but it seems like we'll shoot ourselves in the foot during the final migration

sitozzz:
> Could you please add some `docs` related to file upload logic. It's too complex to easy understand it from the sources

Added to README at the package dir

packages/keystone/fileAdapter/sberCloudFileAdapter.js
sitozzz: Added error thrown on each case. Btw I don't think it's really changed something

packages/files/schema/models/FileRecord.js
sitozzz: Done that

packages/files/package.json
sitozzz: That's also backward compatibility. I just took same lib to not engage file processing and parsing stuff

packages/keystone/fileAdapter/sberCloudFileAdapter.js
sitozzz: Removed

packages/keystone/fileAdapter/sberCloudFileAdapter.js
sitozzz: It's original code of the adapter, so I don't think it's related to this PR

packages/files/README.md
```diff
@@ -0,0 +1,193 @@
+### Files API
+
+This middleware has endpoints for uploading binaries and sharing previously uploaded files between apps/users.
+
+- **Base path**: `/api/files`
+- **Endpoints**:
```
pahaz: `GET /api/files/<file-id>`

packages/files/README.md
```diff
@@ -0,0 +1,193 @@
...
+- Only authenticated, non-deleted users can call the endpoints. Auth is taken from the app session cookie (same as other Keystone-secured endpoints).
+- Unauthorized or deleted users receive `UNAUTHENTICATED/AUTHORIZATION_REQUIRED`.
+
+### Configuration
+
+Set `FILE_UPLOAD_CONFIG` env var to a JSON string:
+
+```json
+{
+  "clients": {
```
pahaz, line 27: Need to add more details about clients, it's not clear

packages/files/README.md
```diff
@@ -0,0 +1,193 @@
...
+### Uploading files
+
+- Method: `POST`
+- Content-Type: `multipart/form-data`
+- Fields:
+  - `file`: one or multiple file binaries
+  - `meta`: JSON string with:
+    - `dv`: `1`
+    - `sender`: `{ dv: 1, fingerprint: string }`
+    - `authedItemId`: UUID of the current user; must equal the authenticated user id
```
pahaz: `userId` ? or `user: { id }`

packages/files/README.md
```diff
@@ -0,0 +1,193 @@
...
+    - `appId`: one of keys from `clients`
```
pahaz: What about `organization`?

packages/files/README.md
pahaz: We use some GraphQL-style objects for foreign keys in kustom mutations. Probably we should use the same pattern here

packages/files/README.md
```diff
@@ -0,0 +1,193 @@
...
+- `signature` is an HS256 JWT (expires in 5 minutes) signed with `clients[appId].secret`. Payload contains file meta, including `recordId`, original filename, mimetype, encoding, and `meta` with app/user info.
+
+### FileRecord model and signature payload
+
+The signed payload corresponds to the public meta of `FileRecord`:
+
+```json
+{
+  "id": "<storage-id>",
+  "recordId": "<FileRecord.id>",
```
pahaz, line 145: Can we use FileRecord.id as the ID? One ID to rule them all

packages/files/README.md
```diff
@@ -0,0 +1,193 @@
...
+  "path": "<string|null>",
+  "filename": "<generated-name>",
+  "originalFilename": "<original name>",
+  "mimetype": "<mime>",
+  "encoding": "<encoding>",
+  "meta": {
+    "dv": 1,
+    "sender": { "dv": 1, "fingerprint": "<string>" },
+    "authedItemId": "<uuid>",
+    "appId": "<string>",
```
pahaz: I'm expecting something like: `user: { id: "<uuid>" }`, `app: { id: "<uuid>" }` and it's unclear which app ID should I use B2BApp OIDCApp B2CApp ... looks like we need to use some new word for that

packages/files/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+- Requirements:
+  - The current user must be the owner indicated by `authedItemId`.
+  - The model’s list key must be present in `meta.modelNames`.
+  - Use the signature before it expires.
+
+Create example:
+
+```graphql
+mutation CreateSomeModel($data: SomeModelCreateInput!) {
```
pahaz, line 183: need to provide a `data` example

packages/files/README.md
```diff
@@ -0,0 +1,193 @@
...
+- Method: `POST`
+- Content-Type: `multipart/form-data`
+- Fields:
+  - `file`: one or multiple file binaries
+  - `meta`: JSON string with:
+    - `dv`: `1`
+    - `sender`: `{ dv: 1, fingerprint: string }`
+    - `authedItemId`: UUID of the current user; must equal the authenticated user id
+    - `appId`: one of keys from `clients`
+    - `modelNames`: non-empty array of Models to which this file should be connected
```
pahaz: Is it optional? If so, you need to remove it from the minimal example

packages/files/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+Set `FILE_UPLOAD_CONFIG` env var to a JSON string:
+
+```json
+{
+  "clients": {
+    "some-app-internal-id": { "name": "Some app", "secret": "<HS256-signing-secret>" }
+  },
+  "quota": {
+    "user": 100,
```
pahaz: 100 Gb?

packages/files/README.md
```diff
@@ -0,0 +1,193 @@
...
+    - `dv`: `1`
+    - `sender`: `{ dv: 1, fingerprint: string }`
+    - `authedItemId`: UUID of the current user; must equal the authenticated user id
+    - `appId`: one of keys from `clients`
+    - `modelNames`: non-empty array of Models to which this file should be connected
+
+Example (curl):
+
+```bash
+curl -X POST "$SERVER/api/files/upload" \
```
pahaz, line 75: What about `objectId` + `schemaName`?

packages/files/fileMiddleware.test.js
```diff
@@ -0,0 +1,597 @@
...
+        filestream = fs.readFileSync(testFile)
+
+        // Clear rate limits
+        await kv.del(`guard_counter:file:${admin.user.id}`)
+        await kv.del('guard_counter:file:::ffff:127.0.0.1')
+    })
+
+    describe('file middleware', () => {
+        describe('security', () => {
+            test('Unauthorized request should fail with 403 error', async () => {
```
pahaz, line 60: You need to hardcode 403 inside the test or change the test name

packages/files/schema/models/FileRecord.js
```diff
@@ -0,0 +1,74 @@
...
+
+const FILE_RECORD_USER_META = '{ dv sender { dv fingerprint } authedItemId appId modelNames sourceAppId }'
+const FILE_RECORD_META_FIELDS = `{ id fileAdapter recordId path filename originalFilename mimetype encoding meta ${FILE_RECORD_USER_META} }`
+const FILE_RECORD_PUBLIC_META_FIELDS = `{ id recordId path filename originalFilename mimetype encoding meta ${FILE_RECORD_USER_META} }`
+
+
+const FileRecord = new GQLListSchema('FileRecord', {
+    schemaDoc: 'Stores uploaded file meta data and owner',
+    fields: {
+        fileMeta: {
```
pahaz, line 23: Why it's json? it's better to use plain fields filename, mimetype, originalFilename, path, encoding

packages/keystone/fields/CustomFile/Implementation.js
```diff
@@ -0,0 +1,158 @@
...
+
+                return {
+                    publicUrl: this.fileAdapter.publicUrl(itemValues, user),
+                    ...itemValues,
+                }
+            },
+        }
+    }
+
+    async validateInput (props) {
```
pahaz, line 87: I'm expecting to call `fetch` to connect the file to the model and ID. Somewhere near here

packages/files/README.md
sitozzz: That's is not related to exact middleware. It's on adapter side

packages/files/README.md
sitozzz: No, it's required. Here is a test for that and schema validation as well

packages/files/README.md
sitozzz: 100 files for user/ip per hour

packages/files/README.md
sitozzz: What do u mean? What is `schemaName`? It's already `modelNames` array for that reason if I get it right. 
But `objectId` is kinda tricky to understand for me as well

packages/files/README.md
sitozzz: I don't like to turn REST into something transitional. Also there is no really an app, to which u like connect to. It's a pair of public-secret

packages/files/schema/models/FileRecord.js
sitozzz: Because it's allows u to store extra metadata of the file and don't really relate with adapter specific fields. Also there is backward compatibility step, because originally it's a json based field

packages/keystone/fields/CustomFile/Implementation.js
sitozzz: ??

packages/files/README.md
sitozzz: No, because of storage id comes before we even can save an object. Also adapter does not guarantee uniqueness of that field

packages/files/README.md
sitozzz: U may provide as much fields as u can. But there is only required ones

packages/files/README.md
sitozzz: authedItem it's a broader concept because theoretically you can be authorized for any entity

packages/files/README.md
pahaz: I think we need to put all the logic relaten to files in one place to recude the complexity. It's hard to maintain if you need to know about some other file-related logic ... Is it possible to collect it in one place?

packages/files/README.md
pahaz: It looks like extra security checks. May be move it to optional

packages/files/README.md
pahaz: It looks like internal parts. You don't need to shou extra IDs.

packages/files/README.md
pahaz: No at the moment. But why do you assume we won’t add it as a model in the near future? And should we change it in a future?

pahaz: It's better to use the same IDs patter for all endpoints. REST, GQL, etc

packages/files/README.md
```diff
@@ -0,0 +1,250 @@
...
+  "originalFilename": "<original name>",
+  "mimetype": "<mime>",
+  "encoding": "<encoding>",
+  "meta": {
+    "dv": 1,
+    "sender": { "dv": 1, "fingerprint": "<string>" },
+    "authedItemId": "<uuid>",
+    "appId": "<string>",
+    "modelNames": ["<string>", "..."],
+    "sourceAppId": "<string|null>"
```
pahaz: If i relly know the model can I attach it with `modelName` + `itemId` fields here?

packages/files/README.md
```diff
@@ -0,0 +1,250 @@
...
+  "recordId": "<FileRecord.id>",
+  "path": "<string|null>",
+  "filename": "<generated-name>",
+  "originalFilename": "<original name>",
+  "mimetype": "<mime>",
+  "encoding": "<encoding>",
+  "meta": {
+    "dv": 1,
+    "sender": { "dv": 1, "fingerprint": "<string>" },
+    "authedItemId": "<uuid>",
```
pahaz: just `user`

packages/files/README.md
```diff
@@ -0,0 +1,250 @@
...
+
+Creates a new file record that points to the original binary; useful to “reassign” an uploaded file to another app/user/model.
+
+- Method: `POST`
+- Content-Type: `application/json`
+- Body:
+  - `dv`: `1`
+  - `sender`: `{ dv: 1, fingerprint: string }`
+  - `id`: UUID of existing file record (must belong to the authenticated user)
+  - `authedItemId`: UUID of the target user (new owner)
```
pahaz: What about organization ID?

packages/files/README.md
```diff
@@ -0,0 +1,250 @@
...
+  - `appId`: target app id (must exist in `clients`)
+  - `modelNames` (optional): array of strings
+
+Example:
+
+```bash
+curl -X POST "$SERVER/api/files/share" \
```
pahaz: Can we merge `attach` and `share`?

packages/files/schema/models/FileRecord.js
```diff
@@ -0,0 +1,86 @@
+const { historical, versioned, uuided, tracked, softDeleted, dvAndSender } = require('@open-condo/keystone/plugins')
```
pahaz, line 1: We use `schema` folder name for GQLLists. Why `model`?

packages/files/README.md
sitozzz: That's important part of a new API. U should strictly define where u want to connect ur file. Without that u just be able to upload file into the storage and don't use it anymore and anywhere

packages/files/README.md
```diff
@@ -0,0 +1,251 @@
...
+
+Clients is a object of fileClientId as a key and value is an object with payload of secret to encrypt and decrypt files 
+```json
+{
+  "clients": {
+    "some-app-internal-id": { "secret": "<HS256-signing-secret>" }
+  },
+  "quota": {
+    "user": 100,
+    "ip": 100
```
nomerdvadcatpyat: Why we have limit by `ip` if only authorized users can upload files?

Btw, we have an integration (call records), which creates files with same ip (integration ip) and different user ids. Can we add it to whitelist?

packages/files/README.md
```diff
@@ -0,0 +1,251 @@
...
+- Content-Type: `multipart/form-data`
+- Fields:
+  - `file`: one or multiple file binaries
+  - `meta`: JSON string with:
+    - `dv`: `1`
+    - `sender`: `{ dv: 1, fingerprint: string }`
+    - `user`: `{ id: "<USER_ID>" }` Object with id key, that should contain UUID of the current user; must equal the authenticated user id
+    - `fileClientId`: one of keys from `clients`
+    - `modelNames`: non-empty array of Models to which this file should be connected
+    - `organization`: `{id: "<ORGANIZATION_ID"}` Optional - object with id key, that should contain id of user organization
```
nomerdvadcatpyat:
```suggestion
    - `organization`: `{id: "<ORGANIZATION_ID"}` Optional - object with id key, that should contain id of organization to which the file belongs
```
?

packages/files/README.md
```diff
@@ -0,0 +1,251 @@
...
+{
+  "data": {
+    "file": { "id": "<uuid>", "signature": "<jwt>" }
+  }
+}
+```
+
+- The returned `signature` is signed with the target `fileClientId` secret and includes `meta.sourcefileClientId` (original app) and `recordId` of the new file record.
+
+### Attaching a file to a model
```
nomerdvadcatpyat: It's like a "Connect a file to a GraphQL model" above but rest way?
Maybe it's worth combining them in one part?

sitozzz: Yes, it's about adding a file to a gql model. 

Okay, I'll try to combine this two methods into one

packages/files/README.md
sitozzz: What the real question?)

packages/files/README.md
sitozzz: U're correct here. It was done before I make it "authed only"

packages/files/README.md
nomerdvadcatpyat: I think it's more correct to write here that this is the organization to which the file belongs, rather than the user's organization. A user can have many organizations

packages/keystone/fields/CustomFile/Implementation.js
```diff
@@ -0,0 +1,223 @@
...
+                    variable: [this.path],
+                    message: 'Signature is wrong or expired',
+                })
+            }
+
+            const { data, success, error } = validateFileUploadSignature(fileData)
+            if (!success) {
+                throw new GQLError({
+                    code: 'BAD_USER_INPUT',
+                    type: 'WRONG_SIGNATURE',
```
nomerdvadcatpyat: Can you add message like in other errors?

packages/keystone/fields/CustomFile/Implementation.js
```diff
@@ -0,0 +1,223 @@
...
+            } catch (err) {
+                throw new GQLError({
+                    code: 'BAD_USER_INPUT',
+                    type: 'WRONG_SIGNATURE',
+                    variable: [this.path],
+                    message: 'Signature is wrong or expired',
+                })
+            }
+
+            validateFileUploadSignature(fileMeta)
```
nomerdvadcatpyat: You need to handle validation error

packages/keystone/fields/CustomFile/Implementation.js
```diff
@@ -0,0 +1,223 @@
...
+            })
+        }
+
+        // === Legacy flow (Upload stream): delegate to base ===
+        return super.resolveInput({ resolvedData, existingItem })
+    }
+
+    async beforeChange ({ resolvedData, context, listKey }) {
+        const key = `${listKey}.${this.path}`
+        const marker = context._fileNewFlow && context._fileNewFlow[key]
```
nomerdvadcatpyat: I think `marker` is not obvious

```suggestion
        const hasFileInRequest = context._fileNewFlow && context._fileNewFlow[key]
```

packages/keystone/fields/CustomFile/Implementation.js
```diff
@@ -0,0 +1,223 @@
...
+        if (!marker) return
+
+        const payload = {
+            id: resolvedData.id,
+            modelName: listKey,
+            signature: context._fileNewFlow[key].signature,
+            fileClientId: this._fileClientId,
+            dv: 1, sender: { dv: 1, fingerprint: 'file-attach-handler' },
```
nomerdvadcatpyat: We don't have user sender here?

packages/keystone/fields/CustomFile/Implementation.js
```diff
@@ -0,0 +1,223 @@
...
+
+        try {
+            const res = await fetch(this._fileServiceUrl, {
+                method: 'POST',
+                headers: { 'Content-Type': 'application/json' },
+                body: JSON.stringify(payload),
+            })
+
+            if (!res.ok) {
+                return
```
nomerdvadcatpyat: Why just return here (and below in afterChange)? Can we throw an error to user?

packages/keystone/fields/CustomFile/Implementation.js
sitozzz: That's fixed

packages/keystone/fields/CustomFile/Implementation.js
sitozzz: Yes we have! I'll fix that

packages/files/utils.js
```diff
@@ -0,0 +1,632 @@
...
+                })
+            )
+        )
+
+        // clean tmp files
+        files.map(file => {
+            if (file.filepath) {
+                fs.promises.unlink(file.filepath).catch(() => {})
+            }
+        })
```
nomerdvadcatpyat, line 355: Could it be that files that are created on the server (temporary files created in parserHandler) will not be deleted here?
Will they then be deleted by the operating system as temporary?

nomerdvadcatpyat (approved):

packages/keystone/fields/CustomFile/Implementation.js
sitozzz: That's because of tricky api miscommunications. So from that part we had an graphql like syntax, but new API is REST-like. 
Anyway now I catch all of the errors, convert it back to GQLError factory and finally throw to user. 
Also if there is something unexpected - I throw that too. 

Maybe also I want to add some logs to the unexpected part, so developers may understand what's going bad here

packages/keystone/fields/CustomFile/Implementation.js
sitozzz: Done that, plz check

packages/files/utils.js
sitozzz: Sooo, it really depends on how do u deploy that application. I

f it application container is created by k8s, docker, or something without relying on persistent storage - answer is yes. 

But if u running that locally, it mostly depends from ur OS.

 For example, debian like systems will clean tmp dirs on boot, but some distros might do this by cron job)

PR #6286, opened 2025-06-11
feat(ui): DOMA-11779 added expandable description prop to Alert compnent

abshnko:
Added props `maxLines` to `Alert` component. If it is passed then the description is shortened to the number of lines that was passed and a new toggle appears to show less/show more.

It is also possible to pass custom text to the Show less/Show more toggle with the use of `showMoreText` and `showLessText` props

Collapsed:
image

Expanded:
image

### DIFF
--- a/packages/ui/src/components/Alert/alert.tsx
+++ b/packages/ui/src/components/Alert/alert.tsx
@@ -1,19 +1,82 @@
 import { Alert as DefaultAlert, AlertProps as DefaultAlertProps } from 'antd'
-import React from 'react'
+import classNames from 'classnames'
+import React, { useState } from 'react'
+
+import { ChevronDown, ChevronUp } from '@open-condo/icons'
+
+import { Typography } from '../Typography'
 
 type CondoAlertProps = {
     type: 'success' | 'info' | 'warning' | 'error'
+    expandable?: boolean
+    maxLines?: number
+    showMoreText?: string
+    showLessText?: string
 }
 
 const ALERT_CLASS_PREFIX = 'condo-alert'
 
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/Alert/alert.tsx
```diff
@@ -1,19 +1,82 @@
...
+}) => {
+    const [isExpanded, setIsExpanded] = useState(false)
+    const toggleDescription = () => setIsExpanded(!isExpanded)
+
+    const renderDescription = () => {
+        if (!description) return null
+
+        return (
+            <>
+                <div style={{
```
SavelevMatthew, line 40: Why inline style is here?

packages/ui/src/components/Alert/alert.tsx
```diff
@@ -1,19 +1,82 @@
...
 
-const Alert: React.FC<AlertProps> = (props) => {
+const ALERT_DESCRIPTION_MAX_HEIGHT = '300px'
+const ALERT_DEFAULT_SHOW_MORE_TEXT = 'Show more'
+const ALERT_DEFAULT_SHOW_LESS_TEXT = 'Show less'
+
+const Alert: React.FC<AlertProps> = ({
+    description,
+    maxLines,
+    showMoreText,
```
SavelevMatthew, line 28: What's this used for?

SavelevMatthew (changes requested):
All this will not work properly, because message and description is react nodes...

I also think, that more / less must be solved inside Typography instead:
https://4x.ant.design/components/typography/#components-typography-demo-ellipsis

PR #6293, opened 2025-06-16
chore(condo): Add an ability to configure sending limits for debug application

toplenboren:
## Why

Developers want to test sending push messages w/o creating a B2CApp first. 

For this they use hardcoded constant `kDomaLocalTestingApp' as id

```
{
    "data": {
      "dv": 1,
      "resident": {"id": "af787995-xxxx-xxxx-xxxx-4a34897c2c7f"},
      "sender": { "dv": 1, "fingerprint": "test-sender" },
      "user": { "id": "8bebb8b1-xxxx-xxxx-xxxx-157de137b81e" },
      "app": { "id": "kDomaLocalMiniappId" },
      "type": "B2C_APP_MESSAGE_PUSH",
      "data": {
        "body": "A debug notification!",
        "title": "Open the application",
        "B2CAppContext": "{\"screen1\": \"payment1\", \"paymentId\": \"123456\"}"
      }
  }
}
```

Because of the limits developers needs to wait 1 hour between each push message.

Previously this was fixed as hardcoded logic:

```
if DEVELOPER_PROVIDED_A_DEBUG_APP_ID -> skip all of the limits
```

Code above was executed in all environments (so by knowing app ID you were theoretically able to implement DoS attack on push sending infrastructure) 

After discovering a possibility for a DoS attack the code mentioned above was removed from the codebase and the problem with 1 hour delay reappeared

## Solution
 
1. Move `DEBUG_APP_ID` constant from the code to env
2. Add ability to configure delay between push messages thorough env

## FAQ

Q: Why not simply change limits for the whole environment?
A: In other words, lets have a lesser delay between sending push messages in a development environment, then in a production one. I decided agains that solution as it would create more differences between environments

## Summary by CodeRabbit

- **New Features**
  - Introduced a debug mode for sending B2C app push messages without requiring a registered app, enabling easier testing and development.

- **Chores**
  - Removed the debug app ID constant from shared constants and updated related tests accordingly.


apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.js
```diff
@@ -152,8 +161,10 @@ const SendB2CAppPushMessageService = new GQLCustomSchema('SendB2CAppPushMessageS
                 let B2CAppName = 'Debug app'
                 let appSettings = {}
 
+                const x = conf
```
SavelevMatthew: leftover?

apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.js
```diff
@@ -164,6 +175,10 @@ const SendB2CAppPushMessageService = new GQLCustomSchema('SendB2CAppPushMessageS
                     B2CAppName = appExisted.name
                 }
 
+                if (DEBUG_APP_ENABLED && b2cAppId === DEBUG_APP_ID) {
```
SavelevMatthew: why not just...else?

toplenboren: yea, fixed this

apps/condo/domains/miniapp/schema/SendB2CAppPushMessageService.js
toplenboren: fixed this

PR #6295, opened 2025-06-17
feat(condo): DOMA-11788 bulk meters delete

nomerdvadcatpyat:
1) Update Meter and PropertyMeter accesses, allowed bulk operations. Wrote test.
2) Added action bars with selected items to meter and property meter table pages

изображение


apps/condo/domains/meter/access/Meter.js
```diff
@@ -61,44 +62,96 @@ async function canManageMeters (args) {
         return await canManageObjectsAsB2BAppServiceUser(args)
     }
 
+    const isBulkRequest = Array.isArray(originalInput)
```
abshnko, line 65: i think it is possible to use this and get rid of else branch
```suggestion
    const inputs = Array.isArray(originalInput) ? originalInput : [originalInput]
```

apps/condo/domains/meter/access/PropertyMeter.js
```diff
@@ -27,48 +28,101 @@ async function canReadPropertyMeters ({ authentication: { item: user }, context
     }
 }
 
-async function canManagePropertyMeters ({ authentication: { item: user }, originalInput, operation, itemId, context }) {
+async function canManagePropertyMeters ({ authentication: { item: user }, originalInput, operation, itemId, itemIds, context }) {
     if (!user) return throwAuthenticationError()
     if (user.deletedAt) return false
     if (user.isAdmin) return true
 
+    const isBulkRequest = Array.isArray(originalInput)
```
abshnko, line 36: and here
```suggestion
    const inputs = Array.isArray(originalInput) ? originalInput : [originalInput]
```

Alexander-Turkin (approved):

PR #6312, opened 2025-06-24
fix(condo): DOMA-11841 remove skipUserPrefetch from initial page

nomerdvadcatpyat:
Because of this, the switching of the organizationLink cookie in the webview wasn't working


tolmachev21:
How this work? I'd like to know about it. Could you tell me about it?

PR #6361, opened 2025-07-09
feat(global): INFRA-1097 stupid loggers

SavelevMatthew:
The idea of this change is to obtain core information and auto-generate as much info as possible, so developer just call getLogger and that's it

### DIFF
--- a/packages/keystone/logging/getLogger.js
+++ b/packages/keystone/logging/getLogger.js
@@ -1,8 +1,14 @@
 // NOTE: same as keystone logger
+const path = require('path')
+
+const callsites = require('callsites')
 const { toString } = require('lodash')
 const pino = require('pino')
 const serializers = require('pino-std-serializers')
 
+const conf = require('@open-condo/config')
+
+const { extractPropertiesFromPath, enhanceLogWithAsyncContext } = require('./extractors')
 const { normalizeVariables } = require('./normalize')
 
 const { safeFormatError } = require('../utils/errors/safeFormatError')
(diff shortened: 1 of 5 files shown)


packages/keystone/logging/getLogger.js
```diff
@@ -1,8 +1,14 @@
 // NOTE: same as keystone logger
+const path = require('path')
+
+const callsites = require('callsites')
```
pahaz: https://github.com/sindresorhus/callsites/blob/main/index.js#L1-L16 It's just 15 LOC, so could you please just copy them to avoid extra deps ...

packages/keystone/package.json
```diff
@@ -67,6 +67,7 @@
     "big.js": "^6.2.2",
     "body-parser": "^1.20.3",
     "bull": "^4.8.4",
+    "callsites": "^3.1.0",
```
pahaz: It's just 15 LOC. It's better not to add it as deps

packages/keystone/logging/getLogger.js
```diff
@@ -1,8 +1,14 @@
...
 const { toString } = require('lodash')
 const pino = require('pino')
 const serializers = require('pino-std-serializers')
 
+const conf = require('@open-condo/config')
+
+const { extractPropertiesFromPath, enhanceLogWithAsyncContext } = require('./extractors')
```
pahaz: Do we really need to expose these functions in our public API? They seem internal to me—maybe better keep them private.

packages/keystone/logging/getLogger.js
SavelevMatthew: copied this one...-1 dep

packages/keystone/logging/getLogger.js
SavelevMatthew: marked as private

PR #6422, opened 2025-07-23
feat(condo): DOMA-11976 added access for service user

vovaaxeapolla:


apps/condo/domains/acquiring/access/RecurrentPaymentContext.js
```diff
@@ -5,9 +5,10 @@
 const access = require('@open-condo/keystone/access')
 const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
 
-const { RESIDENT } = require('@condo/domains/user/constants/common')
+const { RESIDENT, SERVICE } = require('@condo/domains/user/constants/common')
 
 const userFilter = (id) => ({ serviceConsumer: { resident: { user: { id } } } })
+const serviceUserFilter = (id) => ({ serviceConsumer: { acquiringIntegrationContext: { integration: { accessRights_some: { user: { id }, deletedAt: null } } } } })
```
AleX83Xpert, line 11: What if serviceConsumer/context/integration is deleted?

vovaaxeapolla: It seems that the integration should still be able to delete the context even if some part of the path is deleted

PR #6434, opened 2025-07-25
fix(condo): DOMA-10589 fixed creating reading duplicates for the same day in register services

abshnko:
MISTAKE: should be DOMA-10859

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplicate detection for meter readings by ensuring that readings submitted on the same day are recognized as duplicates, even if their timestamps differ within that day.

* **Tests**
  * Added new test cases to verify that duplicate meter readings are not created when readings are submitted with slightly different times on the same day.

### DIFF
--- a/apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
+++ b/apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
@@ -176,6 +176,7 @@ const RegisterMetersReadingsService = new GQLCustomSchema('RegisterMetersReading
                     }
 
                     const dateISO = tryToISO(reading.date)
+                    const startOfDayISO = dateISO ? dayjs(dateISO).startOf('day').toISOString() : undefined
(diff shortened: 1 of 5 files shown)


apps/condo/domains/meter/schema/RegisterMetersReadingsService.js
```diff
@@ -247,7 +248,8 @@ const RegisterMetersReadingsService = new GQLCustomSchema('RegisterMetersReading
                     }
 
                     try {
-                        const key = `${meterId}-${dateISO}`
+                        // NOTE: we look for duplicates with the same date, disregarding the time of day
```
YEgorLu, line 251: then you can just do dayjs(...).format('YYYY-MM-DD'), so it becomes clear, that you check for date, not datetime

Also dayjs().startOf('day').toISOString() can give you something like '2025-07-31T20:00:00' for 1st August, don't know if that affects your logic here

dkoviazin (approved):

PR #6493, opened 2025-08-06
fix(condo): DOMA-12014 update pos-integration to main

AleX83Xpert:


SavelevMatthew (commented):
There's no such task as "update pos-integration to main", provide meaningful description of changes instead

PR #6504, opened 2025-08-08
fix(condo): DOMA-11758 Print human understandable message to error wh…

AleX83Xpert:
…en import meters
# Before
image

# After
image


sitozzz (approved):
lgtm

PR #6510, opened 2025-08-08
fix(condo): DOMA-12091 fixed getting DatePicker from locale if it is undefined

abshnko:
ConfigContext is empty if DatePicker component is being called from an iFrame, hence locale can be undefined. Need to pass the correct locale when calling DatePicker from a miniapp, otherwise it will go to default fallback (en)

TODO:
- [ ] fix registry-importer
- [ ] fix callcenter
- [ ] fix pass
- [ ] fix insurance
- [ ] fix new-greenhouse

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of the date picker by preventing errors when locale settings are missing.

### DIFF
--- a/apps/condo/domains/common/components/Pickers/DatePicker.tsx
+++ b/apps/condo/domains/common/components/Pickers/DatePicker.tsx
@@ -39,7 +39,7 @@ const DefaultDatePicker = generatePicker<Dayjs>(dayjsGenerateConfig)
 const generateDatePickerWithLocale = (Picker: PickerType) => (props) => {
     const { locale } = useContext(ConfigProvider.ConfigContext)
 
-    return <Picker css={PickerStyle} locale={locale.DatePicker} {...props} />
+    return <Picker css={PickerStyle} locale={locale?.DatePicker} {...props} />
 }
 
 const DatePicker = generateDatePickerWithLocale(DefaultDatePicker) as DatePickerType


apps/condo/domains/common/components/Pickers/DatePicker.tsx
```diff
@@ -39,7 +39,7 @@ const DefaultDatePicker = generatePicker<Dayjs>(dayjsGenerateConfig)
 const generateDatePickerWithLocale = (Picker: PickerType) => (props) => {
     const { locale } = useContext(ConfigProvider.ConfigContext)
 
-    return <Picker css={PickerStyle} locale={locale.DatePicker} {...props} />
+    return <Picker css={PickerStyle} locale={locale?.DatePicker} {...props} />
```
SavelevMatthew, line 42: Is it even working solution? 
Because right now it seems like in apps they'll be undefined, and you're just escaping them with `?` mark.
And locale in pickers tells a lot about formatting. 

I guess the root of issues is not suppressing the Picker, but identifying why is it missing at first place

abshnko: locales in apps won't be undefined, but i agree i think the best solution is to use locale from Condo itself, but right now i'm at a loss how to do it

PR #6518, opened 2025-08-12
chore(condo): add new feature flag to remove daily ticket limit 

toplenboren:
## Summary by CodeRabbit

- New Features
  - Introduced feature flags to enable or disable daily ticket limit checks, allowing admins to bypass per-day limits for identical tickets and overall daily ticket creation when needed. Default behavior remains unchanged unless toggles are enabled. This adds flexibility for rollouts or urgent operations without code changes; limits continue to apply until a flag is turned on.


pahaz:
Why not increase rather than just remove

PR #6524, opened 2025-08-13
test(ai): DOMA-12111 try to fix random tests

YEgorLu:
Every second test in CI fails because ai tasks did set "error" status or did not update status from "processing"
Idea is, that timeout is too low. Let's call tasks manually, so we wouldn't need to wait for them.
Also that way we will see in CI console exact error, which prevents task from completion

## Summary by CodeRabbit

- Tests
  - Added deterministic controls for AI-flow execution in tests, allowing explicit triggering and verification of scheduling behavior.
  - Expanded coverage across multiple roles and real-life success/failure scenarios with precise assertions.
  - Added setup/teardown hooks to consistently initialize, reset, and restore mocks between tests.
  - Ensured failure paths properly reject during explicit execution; no user-facing changes.

### DIFF
--- a/apps/condo/domains/ai/schema/ExecutionAIFlowTask.test.js
+++ b/apps/condo/domains/ai/schema/ExecutionAIFlowTask.test.js
@@ -61,8 +61,7 @@ describe('ExecutionAIFlowTask', () => {
     beforeEach(async () => {
         userClient = await makeClientWithNewRegisteredAndLoggedInUser()
         userClient2 = await makeClientWithNewRegisteredAndLoggedInUser()
-        executeAIFlowMock.mockReset()
-        executeAIFlowMock.mockImplementation((() => Promise.resolve()))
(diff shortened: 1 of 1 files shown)


toplenboren (approved):

apps/condo/domains/ai/schema/ExecutionAIFlowTask.test.js
```diff
@@ -42,15 +44,24 @@ describe('ExecutionAIFlowTask', () => {
...
+
     beforeAll(async () => {
         adminClient = await makeLoggedInAdminClient()
         supportClient = await makeClientWithSupportUser()
         anonymousClient = await makeClient()
+        executeAIFlowMock = jest.spyOn(aiTasks.executeAIFlow, 'delay')
+    })
+
+    afterAll(async () => {
+        executeAIFlowMock.mockRestore()
```
pahaz, line 58: you should not use any mocks for `test` only for specs

PR #6530, opened 2025-08-13
chore(condo): use native fetch in fetch

toplenboren:


Alexander-Turkin:
maybe replace node-fetch everywhere?

Alexander-Turkin (approved):

PR #6541, opened 2025-08-15
feat(global): DOMA-12131 add ability to change operations weights and disable rate limiting entirely for specific IP / ID

YEgorLu:


toplenboren:
I see the todo left by previous engineer, but ...

If some queries take too much time (when they should not) then the core algo should be examined and fixed? 

Why do we want to store this config in env files? Why not here or in plugin initialization? Are complexity calculation overrides occur because of user data or because of system shenaningans?

packages/keystone/apolloServerPlugins/rateLimiting/plugin.js
```diff
@@ -201,6 +217,9 @@ class ApolloRateLimitingPlugin {
                     allowedQuota = this.#customQuotas[identifier]
                 }
 
+                if (allowedQuota < 0) {
+                    allowedQuota = Infinity
```
pahaz, line 221: ?

PR #6558, opened 2025-08-21
feat(global): INFRA-1156 add ability to override request ip by trusted signed proxies

SavelevMatthew:
## Summary by CodeRabbit

- New Features
  - Public proxying utilities added for signed proxy headers and trusted-proxy IP resolution; integrated IP resolution override into app startup.

- Bug Fixes
  - Improved client IP detection behind proxies; stricter validation of proxy headers, IP formats, timestamps, and signatures; untrusted/malformed headers ignored.

- Tests
  - Comprehensive tests added covering proxy header signing, validation, and IP resolution across many scenarios.

- Chores
  - Package exports and dependencies updated to support the new proxying utilities.

### DIFF
--- a/apps/condo/domains/common/schema/proxying.spec.js
+++ b/apps/condo/domains/common/schema/proxying.spec.js
@@ -189,13 +189,13 @@ describe('Express proxying tests', () => {
                         'x-proxy-ip': ip,
                         'x-proxy-id': 'simple-proxy-name',
                         'x-proxy-timestamp': validTimestampString,
-                        // nosemgrep: javascript.jsonwebtoken.security.jwt-hardcode.hardcoded-jwt-secret
                         'x-proxy-signature': jwt.sign({
                             'x-proxy-ip': ip,
                             'x-proxy-id': 'simple-proxy-name',
                             'x-proxy-timestamp': validTimestampString,
                             method: 'GET',
                             url: '/api/whoami',
+                        // nosemgrep: javascript.jsonwebtoken.security.jwt-hardcode.hardcoded-jwt-secret
                         }, '123'),
                         'x-forwarded-for': proxyIp,
                     }, ip)
@@ -203,13 +203,13 @@ describe('Express proxying tests', () => {
                         'x-proxy-ip': ip,
                         'x-proxy-id': 'simple-proxy-name',
                         'x-proxy-timestamp': invalidTimestampString,
-                        // nosemgrep: javascript.jsonwebtoken.security.jwt-hardcode.hardcoded-jwt-secret
                         'x-proxy-signature': jwt.sign({
                             'x-proxy-ip': ip,
                             'x-proxy-id': 'simple-proxy-name',
                             'x-proxy-timestamp': invalidTimestampString,
                             method: 'GET',
                             url: '/api/whoami',
+                        // nosemgrep: javascript.jsonwebtoken.security.jwt-hardcode.hardcoded-jwt-secret
                         }, '123'),
                         'x-forwarded-for': proxyIp,
                     }, proxyIp)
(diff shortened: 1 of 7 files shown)


packages/miniapp-utils/src/helpers/proxying/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+}
+
+export function getProxyHeadersForIp (method: string, url: string, ip: string, proxyId: string, secret: string): ProxyHeaders {
+    const timestampString = String(Date.now())
+
+    return {
+        [X_PROXY_IP_HEADER]: ip,
+        [X_PROXY_ID_HEADER]: proxyId,
+        [X_PROXY_TIMESTAMP_HEADER]: timestampString,
+        [X_PROXY_SIGNATURE_HEADER]: jwt.sign({
```
sitozzz, line 115: U can add ttl here, so expired signature will be not valid. Maybe after that u can remove lines of your ttl check

packages/keystone/prepareKeystoneApp.js
```diff
@@ -18,6 +21,19 @@ async function prepareKeystoneExpressApp (entryPoint, { excludeApps } = {}) {
     // not a csrf case: used for test & development scripts purposes
     // nosemgrep: javascript.express.security.audit.express-check-csurf-middleware-usage.express-check-csurf-middleware-usage
     const app = express()
+
+    const knownProxies = JSON.parse(conf['TRUSTED_PROXIES_CONFIG'] || '{}')
+    // NOTE: from https://expressjs.com/en/guide/overriding-express-api.html
+    Object.defineProperty(app.request, 'ip', {
```
sitozzz, line 27: What about all other express apps?

packages/miniapp-utils/src/helpers/proxying/utils.ts
sitozzz, line 1: I don't quite like naming of the package. What about just `proxy`?

packages/miniapp-utils/src/helpers/proxying/utils.ts
```diff
@@ -0,0 +1,121 @@
...
+type ProxyId = string
+
+export type KnownProxies = Record<ProxyId, ProxyConfig>
+
+export type TrustProxyFunction = (proxyAddr: string, idx: number) => boolean
+
+const _ipSchema = z.union([z.ipv4(), z.ipv6()])
+const _timeStampBasicRegexp = /^\d+$/
+
+const DEFAULT_PROXY_TIMEOUT_IN_MS = 5_000 // 5 sec to pass request is enough
```
sitozzz, line 21: Could this be moved to ttl of the signature?

packages/miniapp-utils/src/helpers/proxying/utils.ts
SavelevMatthew: 1. Done that
2. I cannot guarantee all external proxy will set this ttl adequately, so I need to check it manually. Double check is better than single check

packages/miniapp-utils/src/helpers/proxying/utils.ts
SavelevMatthew: It's duplicated in TTL signature as well. double check

packages/miniapp-utils/src/helpers/proxying/utils.ts
SavelevMatthew: There will be a Proxy itself to use in our apps, but right now its not about proxy, but about utils to do request proxying, so in my head its all about request-proxying

toplenboren (approved):

packages/keystone/prepareKeystoneApp.js
SavelevMatthew: Currently we're proxying only to KS-based apps

apps/condo/domains/common/schema/proxying.spec.js
```diff
@@ -0,0 +1,329 @@
+jest.resetModules()
```
pahaz, line 1: Why is this file inside the schema?

PR #6581, opened 2025-08-26
fix(condo): DOMA-12182 fix not found message

vovaaxeapolla:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability and timeliness of notification delivery and resends, reducing delays and edge-case failures.
  * Message details and status now appear more completely and consistently immediately after sending or resending.
  * Added input validation for delivery tasks to prevent invalid deliveries.

* **Refactor**
  * Delivery workflow now passes full message data to background tasks, eliminating redundant lookups and improving logging/telemetry.

### DIFF
--- a/apps/condo/domains/notification/schema/SendMessageService.js
+++ b/apps/condo/domains/notification/schema/SendMessageService.js
@@ -16,9 +16,9 @@ const {
     MESSAGE_DELIVERY_DEFAULT_PRIORITY,
     MESSAGE_DELIVERY_PRIORITY_TO_TASK_QUEUE_MAP,
 } = require('@condo/domains/notification/constants/constants')
+const { MESSAGE_FIELDS } = require('@condo/domains/notification/gql')
 const { deliverMessage } = require('@condo/domains/notification/tasks')
 const { Message } = require('@condo/domains/notification/utils/serverSchema')
-const {MESSAGE_FIELDS} = require("../gql");
 
 const logger = getLogger()
 
--- a/apps/condo/domains/notification/schema/SendMessageService.js
+++ b/apps/condo/domains/notification/schema/SendMessageService.js
@@ -18,6 +18,7 @@ const {
 } = require('@condo/domains/notification/constants/constants')
 const { deliverMessage } = require('@condo/domains/notification/tasks')
 const { Message } = require('@condo/domains/notification/utils/serverSchema')
+const {MESSAGE_FIELDS} = require("../gql");
 
 const logger = getLogger()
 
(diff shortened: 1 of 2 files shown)


apps/condo/domains/notification/tasks/deliverMessage.js
```diff
@@ -94,15 +93,11 @@ function getThrottlingCacheKey (message) {
  * Tries to deliver message via available transports depending on transport priorities
  * based on provided message data and available channels. If more prioritized channels fail message delivery,
  * tries to deliver message through less prioritized fallback channels. Updates message status & meta in every case.
- * @param messageId
+ * @param message
  * @returns {Promise<string>}
  */
-async function deliverMessage (messageId) {
+async function deliverMessage (message) {
```
toplenboren, line 98: need to check this api is not used anywhere else

vovaaxeapolla: Checked it. The task is used only by sendMessage mutation

apps/condo/domains/notification/schema/SendMessageService.js
```diff
@@ -202,9 +203,9 @@ const SendMessageService = new GQLCustomSchema('SendMessageService', {
 
                 const message = messageWithSameUniqKey
                     ? messageWithSameUniqKey
-                    : await Message.create(context, messageAttrs, 'id status')
+                    : await Message.create(context, messageAttrs, MESSAGE_FIELDS)
```
sitozzz, line 206: What the idea of change return fields? It's just a more complex db query here

apps/condo/domains/notification/tasks/deliverMessage.js
```diff
@@ -94,17 +92,15 @@ function getThrottlingCacheKey (message) {
...
     const { keystone: context } = getSchemaCtx('Message')
-    const message = await Message.getOne(context,
-        { id: messageId },
-        MESSAGE_FIELDS
-    )
```
sitozzz, line 105: oh I see - u just removed info actualizing and just passing all required info inside task

apps/condo/domains/notification/schema/SendMessageService.js
vovaaxeapolla: The idea is to avoid SELECT queries to the replica. If the error is caused by replication lag, this should resolve the issue. The task first performs a SELECT query to retrieve a message that was just created

toplenboren (approved):

dkoviazin (approved):

PR #6591, opened 2025-09-01
feat(resident-app): DOMA-12185 do not send error message if telegram does not allow to update message with same content

YEgorLu:


SavelevMatthew (approved):
👌

abshnko (approved):

PR #6599, opened 2025-09-03
refactor(condo): DOMA-12211 improve meter reading duplicates detectio…

AleX83Xpert:
…n by comparing values instead of dates

## Summary by CodeRabbit

- New Features
  - Duplicate detection now matches readings by submitted meter values using stable value-based keys for improved deduplication.

- Bug Fixes
  - Re-submitting identical readings (including within one request or same day) returns the existing entry instead of creating duplicates.
  - Readings with different values or from different meters create distinct entries.

- Refactor
  - Duplicate-check flow reorganized for clearer, more stable behavior.

- Tests
  - Expanded integration scenarios and added unit tests covering value extraction and stable key generation.

### DIFF
--- a/apps/condo/domains/meter/utils/serverSchema/registerHelpers.js
+++ b/apps/condo/domains/meter/utils/serverSchema/registerHelpers.js
@@ -189,16 +189,18 @@ function createMeterReadingKey (meterId, sortedValues) {
 }
 
 async function getMeterReadingsForSearchingDuplicates (readings, meters, properties, readingModel) {
+    const orConditions = readings.map((reading) => {
+        const values = getValues(reading)
+        const sanitizedValues = Object.keys(values).reduce((acc, key) => {
+            if (isNil(values[key])) return acc
+            return { ...acc, [key]: values[key] }
+        }, {})
+        return { AND: [sanitizedValues] }
+    })
+
     const plainMeterReadings = await find(readingModel, {
         meter: { id_in: meters.map(meter => meter.id) },
-        OR: readings.map((reading) => {
-            const values = getValues(reading)
-            const sanitizedValues = Object.keys(values).reduce((acc, key) => {
-                if (isNil(values[key])) return acc
-                return { ...acc, [key]: values[key] }
-            }, {})
-            return { AND: [sanitizedValues] }
-        }),
+        OR: orConditions,
         deletedAt: null,
     })
 
@@ -218,6 +220,7 @@ async function getMeterReadingsForSearchingDuplicates (readings, meters, propert
         ...reading,
         meter: metersWithPropertyByIdMap[reading.meter],
     }))
+
     return meterReadings.reduce((acc, meterReading) => {
         const key = createMeterReadingKey(meterReading.meter.id, getSortedValues(meterReading))
 
--- a/apps/condo/domains/meter/utils/serverSchema/registerHelpers.spec.js
+++ b/apps/condo/domains/meter/utils/serverSchema/registerHelpers.spec.js
@@ -2,139 +2,75 @@ const { getSortedValues, createMeterReadingKey } = require('./registerHelpers')
 
 describe('registerHelpers', () => {
     describe('getSortedValues', () => {
-        test('should return sorted values for reading with all meter values', () => {
-            const reading = {
-                value1: '100.5',
-                value2: '200.3',
-                value3: '300.7',
-                value4: '400.1',
-            }
+        const testCases = [
+            {
+                name: 'should return sorted values for reading with all meter values',
+                reading: { value1: '100.5', value2: '200.3', value3: '300.7', value4: '400.1' },
+                expected: ['100.5', '200.3', '300.7', '400.1'],
+            },
+            {
+                name: 'should return sorted values for reading with all unordered meter values',
+                reading: { value3: '300.7', value1: '100.5', value4: '400.1', value2: '200.3' },
+                expected: ['100.5', '200.3', '300.7', '400.1'],
+            },
+            {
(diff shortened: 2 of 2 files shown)


apps/condo/domains/meter/utils/serverSchema/registerHelpers.js
```diff
@@ -155,14 +155,45 @@ function getAddressesKeys (readings, resolvedAddresses) {
...
+        OR: readingsWithValidDates.map((reading) => {
+            const values = getValues(reading)
+            const sanitizedValues = Object.keys(values).reduce((acc, key) => {
+                if (isNil(values[key])) return acc
+                return { ...acc, [key]: values[key] }
+            }, {})
+            return { AND: [sanitizedValues] }
+        }),
         deletedAt: null,
     })
```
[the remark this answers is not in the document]
AleX83Xpert: According to registerMetersReadings' schema, at least the `value1` is required.

apps/condo/domains/meter/utils/serverSchema/registerHelpers.js
```diff
@@ -155,14 +155,53 @@ function getAddressesKeys (readings, resolvedAddresses) {
...
+ */
+function createMeterReadingKey (meterId, sortedValues) {
+    return `${meterId}_${sortedValues.join('_')}`
+}
+
+async function getMeterReadingsForSearchingDuplicates (readings, meters, properties, readingModel) {
     const plainMeterReadings = await find(readingModel, {
         meter: { id_in: meters.map(meter => meter.id) },
-        date_in: uniq(readingsWithValidDates.map(reading => tryToISO(reading.date))),
+        OR: readings.map((reading) => {
```
YEgorLu: Maybe put this block in variable? 

Like "id_in: meters.map(meter => meter.id)" is short and readable, but here you must read it deeply, to understand what is happening

apps/condo/domains/meter/utils/serverSchema/registerHelpers.js
```diff
@@ -179,16 +218,17 @@ async function getMeterReadingByDate (readings, meters, properties, readingModel
         ...reading,
         meter: metersWithPropertyByIdMap[reading.meter],
     }))
-    return meterReadings.reduce((acc, reading) => {
-        const key = `${reading.meter.id}-${dayjs(reading.date).startOf('day').toISOString()}`
-        acc[key] = reading
-        
+    return meterReadings.reduce((acc, meterReading) => {
```
YEgorLu, line 224: as far as I understand, if user registered same readings multiple times, here you will have only earliest reading among duplicates, does it break something?

apps/condo/domains/meter/schema/RegisterMetersReadingsService.test.js
```diff
@@ -894,122 +894,224 @@ describe('RegisterMetersReadingsService', () => {
...
-            organization: { id: organization.id },
-            property: { id: property.id },
+            const meters = await Meter.getAll(adminClient, {
+                organization: { id: organization.id },
+                property: { id: property.id },
+            })
+            expect(meters).toHaveLength(1)
+            expect(meters[0].number).toBe(duplicateReadings[0].meterNumber)
+
+            const metersReadings = await MeterReading.getAll(adminClient, { meter: { id_in: map(meters, 'id') } })
```
YEgorLu, line 1066: Whatever, but maybe would be more readable if you explicitly said { meter: { id: meters[0].id } }. For a minute i thought that you have multiple meters here

apps/condo/domains/meter/utils/serverSchema/registerHelpers.spec.js
```diff
@@ -0,0 +1,285 @@
+const { getSortedValues, createMeterReadingKey } = require('./registerHelpers')
+
+describe('registerHelpers', () => {
+    describe('getSortedValues', () => {
+        test('should return sorted values for reading with all meter values', () => {
```
YEgorLu: It looks like best place to use "test.each", to make it more readable:
```
const testCases = [
    {
      name: 'should return sorted values for reading with all meter values',
      reading: { value1: '100.5', value2: '200.3', value3: '300.7', value4: '400.1' },
      expected: ['100.5', '200.3', '300.7', '400.1']
    },
    ...
]

test.each(testCases)('$name', ({ reading, expected }) => {
    const result = getSortedValues(reading)
    expect(result).toEqual(expected)
})
```
And same for other test cases. Cases where you comparing multiple results maybe wouldn't look so good with test.each

YEgorLu (approved):

apps/condo/domains/meter/utils/serverSchema/registerHelpers.js
AleX83Xpert: okay

apps/condo/domains/meter/utils/serverSchema/registerHelpers.js
AleX83Xpert: This check is only applicable in the context of unified mutation and has no impact on users.

apps/condo/domains/meter/utils/serverSchema/registerHelpers.spec.js
AleX83Xpert: Ok. "test.each" looks nice too :)

apps/condo/domains/meter/schema/RegisterMetersReadingsService.test.js
AleX83Xpert: ¯\\\_(ツ)_/¯

PR #6612, opened 2025-09-08
deps(global): INFRA-1173 bump openid-client to 5.x

SavelevMatthew:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SBBOL OAuth2 client-certificate handling to reduce authentication issues in some environments.

* **Chores**
  * Upgraded openid-client dependency across multiple apps (address-service, condo, miniapp and others) for compatibility.
  * Updated submodule references for many apps to pull in upstream fixes and updates.

* **Tests**
  * Adjusted OIDC-related tests to accommodate updated callback parameter behavior.


apps/address-service/package.json
```diff
@@ -32,7 +32,7 @@
     "graphql": "^16.10.0",
     "graphql-tag": "^2.12.6",
     "lodash": "^4.17.21",
-    "openid-client": "4.7.4",
+    "openid-client": "5.7.1",
```
sitozzz, line 35: That's a lot of breaking change for that update. How do we check if everything is working fine?

sitozzz, line 35: Also it's not a "latest" version. Why not 6.x?

SavelevMatthew: 6.x is ESM-only, so we cannot afford it right now

SavelevMatthew: Overall migration is safe for us except the SBBOL integration, which will be tested on test environment

PR #6662, opened 2025-09-22
feat(condo): DOMA-12309 use undici instead of node-fetch

dkoviazin:
## Summary by CodeRabbit

* **Refactor**
  * Network requests now use the platform fetch API with streamlined timeout handling; public API unchanged.
* **Bug Fixes**
  * Improved abort/timeout behavior and more consistent abort-related error reporting; response headers are preserved when reading raw header values.
* **Tests**
  * Updated tests to assert stringified abort errors and to use the revised raw-headers access.

### DIFF
--- a/packages/keystone/fetch.js
+++ b/packages/keystone/fetch.js
@@ -14,6 +14,27 @@ const FETCH_COUNT_METRIC_NAME = 'fetch.count'
 const FETCH_TIME_METRIC_NAME = 'fetch.time'
 
 
+function wrapResponse (rawResponse) {
+    return Object.assign(
+        Object.create(Object.getPrototypeOf(rawResponse)),
+        rawResponse,
+        {
+            buffer: async () => Buffer.from(await rawResponse.arrayBuffer()),
+
+            // New helper method — do NOT overwrite .headers
+            rawHeaders: () => {
+                const raw = {}
+                // Undici headers is iterable
+                for (const [name, value] of rawResponse.headers) {
+                    if (!raw[name]) raw[name] = []
+                    raw[name].push(value)
+                }
+                return raw
+            },
+        }
+    )
+}
+
 /**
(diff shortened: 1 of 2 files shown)


packages/keystone/fetch.js
```diff
@@ -1,5 +1,5 @@
 const { pickBy } = require('lodash')
-const nodeFetch = require('node-fetch')
+const { fetch: undiciFetch } = require('undici')
```
toplenboren: why not simply remove this line? Fetch is already in Node

dkoviazin: done

packages/keystone/fetch.js
```diff
@@ -149,22 +148,15 @@ const fetchWithRetriesAndLogger = async (url, options = {}) => {
...
-            ])
-
-            if (timeout) {
-                clearTimeout(timeout)
-            }
+            const timeout = setTimeout(() => controller.abort(), abortRequestTimeout)
+
+            const response = await fetchWithLogger(
+                url,
+                { ...fetchOptions, signal: controller.signal },
```
vovaaxeapolla: There is a built-in AbortSignal.timeout(abortRequestTimeout) instead of manual creating timeout controller

dkoviazin: Done

packages/keystone/fetch.spec.js
```diff
@@ -72,9 +72,7 @@ describe('Fetch with retries', () => {
         const options = getOptions({ maxRetries: 0, abortRequestTimeout: 500, timeoutBetweenRequests: 0 })
 
         await catchErrorFrom(async () => await fetch(URL, options), (err) => {
```
toplenboren, line 74: Does it mean that this is breaking change?

toplenboren (approved):

PR #6695, opened 2025-09-29
fix(condo): DOMA-12324 fix requests after logout

nomerdvadcatpyat:
1. Specified `skip` condition in global queries: added user and organization existence and loading checks
2. Added logic with broadcast logout message in `UserMenu`. If user click the logout button in some tab, other tabs also redirects to /auth/signin page
3. Fixed the behavior of AuthProvider: now the user state is stored in a local useState instead of relying directly on the Apollo cache. This ensures the user is properly set and cleared during loading, errors, and logout, eliminates flickering of isAuthenticated, and prevents unnecessary refetch requests after logout.

<details>
  <summary>Queries after logout example</summary>
The behavior is correct: after logout, user is undefined and loading is false:
Screenshot 2025-09-29 at 12 29 15

---
The problematic behavior that occasionally occurred is shown. Initially, after logout, the state was user exists and loading true; then it transitioned to user exists and loading false; and finally to user undefined and loading false. This was caused by the state not being properly tracked across renders. Using a dedicated React state for user fixes the issue by ensuring that user and loading are updated consistently and immediately when logout occurs, preventing intermediate inconsistent states:
 
Screenshot 2025-09-29 at 12 28 46

</details>

## Summary by CodeRabbit

* **New Features**
  * Sign-out syncs across browser tabs; improved cross-tab sign-out messaging.
  * Broadcast API hook accepts optional payloads and safely handles unsupported browsers.

* **Bug Fixes**
  * Unified auth & organization loading guards to avoid premature queries and UI flicker across tours, tasks, mini‑apps, notifications, tickets, and org flows.
  * Tour steps refetching updated for more reliable updates.
  * Client-side auth caching and stricter sign-in/out flows reduce stale state.
  * Organization state now clears persisted cookie when no active employee.


apps/condo/domains/user/components/UserMenu.tsx
```diff
@@ -41,14 +42,41 @@ export const UserMenu: React.FC<UserMenuProps> = ({
         router.push('/user')
     }, [router])
 
+    const authChannel = useMemo(() => {
+        if (typeof window === 'undefined' || typeof window.BroadcastChannel === 'undefined') return null
+        return new BroadcastChannel('auth')
+    }, [])
```
nomerdvadcatpyat: I need to use `useBroadcastChannel` hook.
And check `typeof window.BroadcastChannel === 'undefined'` in it

Alllex202 (approved):
looks good

Issue #6485, opened 2025-08-05
Improve address autocomplete: Augment Google suggestions with DB-stored addresses


paulo-rossy:
**Hello team!**  

I'm experiencing an issue with the address service that uses the Google provider.  
**Expected behavior:**  
When entering a street name (e.g., `Lagos de coronas`), the service should return multiple houses from that street, including the street itself.  

**Problem with Google API:**  
Autocomplete requires users to start typing numbers (house numbers) and returns only up to 5 exact addresses.  
Relevant references:  
- [StackOverflow: Autocomplete house number](https://stackoverflow.com/questions/70984513/autocomplete-house-number-with-google-api)  
- [StackOverflow: Not returning street numbers](https://stackoverflow.com/questions/59807235/google-maps-places-autocomplete-not-returning-street-numbers)  
- [Google Documentation](https://developers.google.com/maps/documentation/places/web-service/legacy/autocomplete)  

**Test queries:**  
1. With types:  
   ```  
   https://maps.googleapis.com/maps/api/place/autocomplete/json?input=Lagos%20de%20Coronas&key=[]&types=street_address  
   ```  
2. Without types (default):  
   ```  
   https://maps.googleapis.com/maps/api/place/autocomplete/json?input=Lagos%20de%20Coronas&key=[]  
   ```  

---

**Proposed solution:**  
For every `item` from `suggest`, search the DB for string matches and augment the results.  

**Example:**  
> Query: `Lagos de coronas`  
> Google’s response:  
> - `Lagos de coronas, Zaragoza`  
> - `Lagos de coronas, Madrid`  
>  
> After DB search:  
> - `Lagos de coronas, Zaragoza` (from Google)  
>   - `Lagos de coronas, Zaragoza 1` (from DB)  
>   - `Lagos de coronas, Zaragoza 2` (from DB)  
>   - `Lagos de coronas, Zaragoza 3` (from DB)  
> - `Lagos de coronas, Madrid` (from Google)  
>   - `Lagos de coronas, Madrid 1` (from DB)  
>   - `Lagos de coronas, Madrid 2` (from DB)  

---

**DB search challenge:**  
- `address_contains` is slow and imprecise.  
- `start_with`/`end_with` are faster, but data in the `Address` schema is normalized (e.g., `C. de los Lagos de Coronas, 23, 50011 Zaragoza, España`), making it incompatible with direct searches for `"Lagos de coronas"`.  

**Proposed fix:**  
[...]

PR #6712, opened 2025-10-01
feat(address-service): added augmented address suggestions with an existing address in the database

paulo-rossy:
according to https://github.com/open-condo-software/condo/issues/6485

## Summary by CodeRabbit

* **New Features**
  * Optional augmentation of address suggestions with matching addresses from the service database via include_db_address, returning enriched suggestions flagged as augmented.
  * Augmented results preserve original suggestion ordering and early-empty-search behavior.

* **Tests**
  * End-to-end tests added for augmentation, deduplication, provider variations, and matching to ensure correct behavior and stability.

### DIFF
--- a/apps/address-service/domains/common/utils/services/appendDbAddressesToSuggestions.js
+++ b/apps/address-service/domains/common/utils/services/appendDbAddressesToSuggestions.js
@@ -49,6 +49,9 @@ async function appendDbAddressesToSuggestions (context, suggestions) {
     const allFoundSources = resultsFromDB.flat()
     const newSuggestionsToAdd = []
 
+    const provider = get(conf, 'PROVIDER')
+    const isGoogleProvider = provider === GOOGLE_PROVIDER
+
     for (const source of allFoundSources) {
         if (!source?.source || !source?.address?.meta) {
             continue
@@ -57,9 +60,6 @@ async function appendDbAddressesToSuggestions (context, suggestions) {
         const normalizedSource = source.source.toLowerCase()
 
         if (!uniqueNormalizedSuggestions.has(normalizedSource)) {
-            const provider = get(conf, 'PROVIDER')
-            const isGoogleProvider = provider === GOOGLE_PROVIDER
-
             const defaultAddressValue = source.address.address
 
             const capitalizedAddressSourceValue = normalizedSource


apps/address-service/domains/common/utils/services/suggest/providers/GoogleSuggestionProvider.js
```diff
@@ -66,7 +137,7 @@ class GoogleSuggestionProvider extends AbstractSuggestionProvider {
     /**
      * @returns {Promise<GooglePredictionObject[]>}
      */
-    async get ({ query, session = '', context = '', language = '', count = 20, helpers = {} }) {
+    async get ({ query, session = '', context = '', language = '', count = 20, helpers = {}, includeDbAddress, godContext }) {
```
AleX83Xpert: Please add the same parameters to AbstractSuggestionProvider.get for better API consistency.

AleX83Xpert: Let's prevent providers from working with databases. The provider's purpose is to work with the external API. Providers should know nothing about databases.
```suggestion
    async get ({ query, session = '', context = '', language = '', count = 20, helpers = {}, includeDbAddress }) {
```

AleX83Xpert (commented):
Hi @paulo-rossy!
Thanks for your useful contribution, and sorry for the delay. You can just tag me when you change the address service.
I'd like you to make more consistent changes. Let's move your `getAugmentedSuggestions()` into SuggestionKeystoneApp. This function should be called optionally, depending on the `includeDbAddress` value.
Your function aims to inject some data into the suggestions array obtained from the provider, like `InjectionsSeeker`. So, you may add calling of your function before or after "injectionsSeeker". Also, there is working sorting in injections, which you may use too.
These changes will gather all data for the suggestions array in a single place.

apps/address-service/domains/common/utils/services/suggest/providers/GoogleSuggestionProvider.js
paulo-rossy: Hello @AleX83Xpert!  I removed this logic from the address providers, now it is at the level of the SuggestionKeystoneApp.

paulo-rossy: I removed this logic from the address providers, now it is at the level of the SuggestionKeystoneApp.

paulo-rossy:
> Hi @paulo-rossy! Thanks for your useful contribution, and sorry for the delay. You can just tag me when you change the address service. I'd like you to make more consistent changes. Let's move your `getAugmentedSuggestions()` into SuggestionKeystoneApp. This function should be called optionally, depending on the `includeDbAddress` value. Your function aims to inject some data into the suggestions array obtained from the provider, like `InjectionsSeeker`. So, you may add calling of your function before or after "injectionsSeeker". Also, there is working sorting in injections, which you may use too. These changes will gather all data for the suggestions array in a single place.

Hello @AleX83Xpert! Fixed the code based on your comments, ready for further review, thanks for the feedback!

apps/address-service/domains/common/utils/services/augmentAddressSuggestions.js
```diff
@@ -0,0 +1,81 @@
...
+    if (uniqueNormalizedSuggestions.size === 0) {
+        return suggestions
+    }
+
+    const searchPromises = Array.from(uniqueNormalizedSuggestions).map(query =>
+        AddressSource.getAll(
+            context,
+            { source_starts_with: query, deletedAt: null },
+            'source address { id key address meta }',
+            { first: 3 }
```
AleX83Xpert: Please move "3" into a constant at the top of this file.

apps/address-service/domains/common/utils/services/augmentAddressSuggestions.js
```diff
@@ -0,0 +1,81 @@
...
+ * and enriches it with matching addresses found in the local AddressSource.
+ * It's designed to solve the issue of missing custom-added addresses in the search results.
+ *
+ * @see https://github.com/open-condo-software/condo/issues/6485
+ * @param {object} context - The context required to execute Keystone queries.
+ * @param {Array<NormalizedSuggestion>} suggestions - The normalized array of suggestions from AddressProvider.
+ * @returns {Promise<Array<object>>} A new array containing both original and augmented suggestions.
+ */
+
+async function getAugmentedSuggestions (context, suggestions) {
```
AleX83Xpert: When I see "getAugmentedSuggestions," I think this function will return only augmented suggestions, and I do not understand why it needs suggestions found before.
What do you think about renaming the function to "pushAugmentedSuggestions"? Or something else to make the reader understand what this function really does.

apps/address-service/domains/common/utils/services/augmentAddressSuggestions.js
```diff
@@ -0,0 +1,81 @@
...
+
+    const combinedSuggestions = [...suggestions, ...newSuggestionsToAdd]
+
+    return combinedSuggestions.sort((a, b) => {
+        if (a.value < b.value) {
+            return -1
+        }
+
+        return 1
+    })
```
AleX83Xpert: Looks like it's possible to simplify
```suggestion
    return combinedSuggestions.sort((a, b) => a.value - b.value)
```

apps/address-service/domains/common/utils/services/appendDbAddressesToSuggestions.js
```diff
@@ -0,0 +1,81 @@
...
+
+    for (const source of allFoundSources) {
+        if (!source?.source || !source?.address?.meta) {
+            continue
+        }
+
+        const normalizedSource = source.source.toLowerCase()
+
+        if (!uniqueNormalizedSuggestions.has(normalizedSource)) {
+            const capitalizedAddressValue = normalizedSource
```
AleX83Xpert: Do you really need to capitalize parts of addresses? There are some countries where people do not capitalize each letter after the comma.
Maybe it's better to use `source.address.address` or directly from `meta` instead? Looks like these values are already country-specific. What do you think?

AleX83Xpert (commented):
Hi @paulo-rossy, thanks for the quick coding. This is much better. I've found only some non-critical points.

apps/address-service/domains/common/utils/services/augmentAddressSuggestions.js
paulo-rossy: Done! Removed to the AUGMENTATION_DB_LIMIT variable

apps/address-service/domains/common/utils/services/augmentAddressSuggestions.js
paulo-rossy: I’ve renamed the function to `appendDbAddressesToSuggestions` — I think this name makes the purpose clearer and more explicit.

apps/address-service/domains/common/utils/services/augmentAddressSuggestions.js
paulo-rossy: Done!

apps/address-service/domains/common/utils/services/appendDbAddressesToSuggestions.js
paulo-rossy: In our case, the value is derived from the source, and the source is always stored in lowercase (see SearchBySource.js and searchServiceUtils.js — both call .toLowerCase() before saving or querying).

For the Google provider specifically, the original suggestions come in properly formatted form — e.g.
España, Zaragoza, Calle de los Lagos de Coronas — as reflected in our tests (GoogleSuggestionProvider.js, normalizedGoogleAddressValue function).

If we mix lowercase addresses from AddressSource with capitalized ones from the provider, the final list would look inconsistent.
I added a small provider check — capitalization is now restored only for Google provider results, while all other providers keep their original casing.

https://github.com/open-condo-software/condo/pull/6712/files#diff-855e7d313c5f0e3b4dd87136c45385c8c981d0a74a40ef756631a7b3449941aaR61-R73

This should make the output more consistent across all sources.

AleX83Xpert: I meant the next. Each AddressSource has a relation to Address, and each Address has the `address` field. This field has capitalization in the right places.
You place a lowercased-and-synthetic-capitalized source as the suggestion's `value`. But what if we use `source.address.address` instead?

paulo-rossy: @AleX83Xpert Hello!

Unfortunately, this is not what I expect. In the address schema, the address field contains the address value in reverse order and with an index. 
Here is an example:
What's in the address - C. de los Lagos de Coronas, 9, 50011 Zaragoza, España
What I expect - España, Zaragoza, Calle de los lagos de coronas, 9

But I agree with you, it really would be more logical. Can I also add a check for the Google provider in this block?

paulo-rossy: https://maps.googleapis.com/maps/api/place/textsearch/json?query=Espa%C3%B1a,%20Zaragoza,%20Calle%20de%20los%20Lagos%20de%20Coronas,%209&key=[]

```
{
   "html_attributions" : [],
   "results" : 
   [
      {
         "formatted_address" : "C. de los Lagos de Coronas, 9, 50011 Zaragoza, Spain",
         "geometry" : 
         {
            "location" : 
            {
               "lat" : 41.6574294,
               "lng" : -0.9333313
            },
            "viewport" : 
            {
               "northeast" : 
               {
                  "lat" : 41.65879592989272,
                  "lng" : -0.9320497201072782
               },
               "southwest" : 
               {
                  "lat" : 41.65609627010728,
                  "lng" : -0.9347493798927224
               }
            }
         },
         "icon" : "https://maps.gstatic.com/mapfiles/place_api/icons/v1/png_71/geocode-71.png",
         "icon_background_color" : "#7B9EB0",
         "icon_mask_base_uri" : "https://maps.gstatic.com/mapfiles/place_api/icons/v2/generic_pinlet",
         "name" : "C. de los Lagos de Coronas, 9",
         "place_id" : "ChIJq38lOG1rWQ0Rhds7zAfX094",
         "reference" : "ChIJq38lOG1rWQ0Rhds7zAfX094",
         "types" : 
         [
            "subpremise"
         ]
      }
   ],
   "status" : "OK"
}
```

paulo-rossy: @AleX83Xpert 
Here is a new comit with these changes -https://github.com/open-condo-software/condo/pull/6712/commits/86b39c6697018db591f72d68416ae4936cb1b725

apps/address-service/domains/common/utils/services/appendDbAddressesToSuggestions.js
```diff
@@ -0,0 +1,88 @@
...
+    for (const source of allFoundSources) {
+        if (!source?.source || !source?.address?.meta) {
+            continue
+        }
+
+        const normalizedSource = source.source.toLowerCase()
+
+        if (!uniqueNormalizedSuggestions.has(normalizedSource)) {
+            const provider = get(conf, 'PROVIDER')
+            const isGoogleProvider = provider === GOOGLE_PROVIDER
```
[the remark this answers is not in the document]
paulo-rossy: done!

apps/address-service/domains/common/utils/services/appendDbAddressesToSuggestions.js
AleX83Xpert: My concern is that this appears to be country-specific or installation-specific logic, but we will deploy it for all.

apps/address-service/domains/common/utils/services/appendDbAddressesToSuggestions.js
```diff
@@ -0,0 +1,88 @@
...
+        )
+    )
+
+    const resultsFromDB = await Promise.all(searchPromises)
+
+    const allFoundSources = resultsFromDB.flat()
+    const newSuggestionsToAdd = []
+
+    const provider = get(conf, 'PROVIDER')
+    const isGoogleProvider = provider === GOOGLE_PROVIDER
```
AleX83Xpert, line 53:
```suggestion
    // This appears to be country-specific or installation-specific logic.
    // TODO: Think about extracting this logic outside of this cute function.
    const provider = get(conf, 'PROVIDER')
    const isGoogleProvider = provider === GOOGLE_PROVIDER
```

AleX83Xpert (approved):

apps/address-service/domains/common/utils/services/appendDbAddressesToSuggestions.js
```diff
@@ -0,0 +1,88 @@
+const get = require('lodash/get')
+
+const conf = require('@open-condo/config')
+
+const { AddressSource } = require('@address-service/domains/address/utils/serverSchema')
+const { BUILDING_ADDRESS_TYPE } = require('@address-service/domains/common/constants/addressTypes')
+
+const { GOOGLE_PROVIDER } = require('../../constants/providers')
```
AleX83Xpert, line 8:
```suggestion
const { GOOGLE_PROVIDER } = require('@address-service/domains/common/constants/providers')
```

apps/address-service/domains/common/utils/services/appendDbAddressesToSuggestions.js.spec.js
```diff
@@ -0,0 +1,381 @@
...
+const {
+    createTestAddress,
+    updateTestAddress,
+    createTestAddressSource,
+    updateTestAddressSource,
+} = require('@address-service/domains/address/utils/testSchema')
+
+const { appendDbAddressesToSuggestions } = require('./appendDbAddressesToSuggestions')
+
+const { GOOGLE_PROVIDER } = require('../../constants/providers')
```
AleX83Xpert, line 16:
```suggestion
const { GOOGLE_PROVIDER } = require('@address-service/domains/common/constants/providers')
```

toplenboren (approved):

AleX83Xpert:
@toplenboren, too early :(

PR #6716, opened 2025-10-01
feat(condo): Experiment: allow usage of markdown in ticket details 

toplenboren:
telegram-cloud-photo-size-2-5404522559225069380-y

- [ ] Checklist item 3
- [x] Checklist item 1
- [ ] Checklist item 2

## Summary by CodeRabbit

* **New Features**
  * Interactive checkboxes in Markdown — toggle task items and propagate updated content.
  * Inline ("lite") Markdown variant for compact text rendering.
  * Editable ticket details rendered as Markdown with in-place editing, optimistic local updates, debounced saving, and error recovery.

* **Documentation / Stories**
  * Storybook examples for interactive checkboxes and the lite Markdown variant.

* **Style**
  * New styling for Markdown task list items for improved layout.

### DIFF
--- a/packages/ui/src/components/Markdown/style.less
+++ b/packages/ui/src/components/Markdown/style.less
@@ -4,8 +4,8 @@
 
 .condo-markdown-task-list-item {
   display: flex;
-  align-items: flex-start;
   gap: 8px;
+  align-items: flex-start;
 }
 
 .condo-markdown {


apps/condo/domains/ticket/components/TicketId/TicketDetailsField.tsx
```diff
@@ -1,22 +1,35 @@
...
+    ticketDetails: string
+    updateTicketDetails?: (newTicketDetails: string) => void
 }
 
-export const TicketDetailsField: React.FC<TicketDetailsFieldProps> = ({ ticket }) => {
+export const TicketDetailsField: React.FC<TicketDetailsFieldProps> = ({ ticketDetails, updateTicketDetails }) => {
     const intl = useIntl()
     const TicketDetailsMessage = intl.formatMessage({ id: 'Problem' })
 
+    // This is for backwards compatibility
```
nomerdvadcatpyat: Don't understand. Backwards compatibility with what?

apps/condo/pages/ticket/[id]/index.tsx
```diff
@@ -617,6 +626,30 @@ export const TicketPageContent = ({ ticket, pollCommentsQuery, refetchTicket, or
...
+                data: {
+                    details: newTicketDetails,
+                    dv: 1,
+                    sender: getClientSideSenderInfo(),
+                },
+            },
+            onError: () => {
+                setTicketDetails(ticket?.details)
+                notification.error({
+                    message: intl.formatMessage({ id: 'ServerErrorPleaseTryAgainLater' }),
```
Comment (nomerdvadcatpyat) [inline apps/condo/pages/ticket/[id]/index.tsx:647]: This message tells the user to visit the page later. I don't think it's appropriate here

packages/ui/src/components/Markdown/markdown.tsx
```diff
@@ -13,27 +14,148 @@ const REMARK_PLUGINS: Array<any> = [
...
+    children,
+    onToggle,
+    disabled = false,
+    node,
+}) => {
+    const position = node.position
+
+    return (
+        <li style={{ listStyle: 'none' }}>
+            <div style={{ display: 'flex', alignItems: 'flex-start', gap: '8px' }}>
```
nomerdvadcatpyat: move styles to css file

packages/ui/src/components/Markdown/markdown.tsx
```diff
@@ -13,27 +14,148 @@ const REMARK_PLUGINS: Array<any> = [
...
+
+                    if (checkboxChild) {
+                        const checked = React.isValidElement(checkboxChild) && checkboxChild?.props?.checked || false
+
+                        const contentChildren = childrenArray.filter(child =>
+                            !(React.isValidElement(child) && child.props?.type === 'checkbox')
+                        )
+
+                        return (
+                            <TaskListItem node={props.node as any} checked={checked} disabled={!hasInteractiveCheckboxes} onToggle={callOnCheckboxChange}>
```
nomerdvadcatpyat, line 150: What is `node` here? Can we avoid using "any" here?

packages/ui/src/components/Markdown/markdown.tsx
```diff
@@ -13,27 +14,148 @@ const REMARK_PLUGINS: Array<any> = [
...
-                p: (props: any) => <Typography.Paragraph {...omit(props, 'ref')} type='secondary' />,
-                a: (props: any) => <Typography.Link {...omit(props, 'ref')} target='_blank'/>,
-                li: ({ children, ...restProps }) => <li {...restProps}><Typography.Text type='secondary'>{children}</Typography.Text></li>,
+                ...MARKDOWN_COMPONENTS_BY_TYPE[type],
+                a: (props: any) => <Typography.Link {...omit(props, 'ref')} rel='noopener noreferrer' target='_blank'/>,
                 pre: (props: any) => <CodeWrapper {...props}/>,
+                input: ({ type, checked, disabled, ...restProps }) => {
+                    if (type !== 'checkbox') {
+                        return <input type={type} {...restProps} />
+                    }
```
nomerdvadcatpyat, line 125: Mb just return void?

toplenboren: No, this would not work.. If this is not a checkbox, it can be another component – we should just render it as it is, because we only customize checkbox component

packages/ui/src/components/Markdown/markdown.tsx
toplenboren: Node is an instance of https://github.com/syntax-tree/hast#element 

I decided to not add another dependency to the project and just use any instead.. But yea, it can be avoided

apps/condo/pages/ticket/[id]/index.tsx
Comment (toplenboren) [reply apps/condo/pages/ticket/[id]/index.tsx:647]: This message will occur in a case when we were not able to update ticket.. I think this happen in only the hardest of outages. 

I think telling user to just visit the page later is better then to just tell him that there was an error

PR #6757, opened 2025-10-09
feat(condo): INFRA-1247 automatically redirect OIDC requests from resident-app

SavelevMatthew:
## Summary by CodeRabbit

* New Features
  * Improved sign-in flow for resident app users: proxy-aware OIDC handling redirects to the correct domain and avoids redirect loops.
  * More accurate client IP detection for requests, enhancing proxy/trust handling and logging.
* Chores
  * Added dependency to support proxy address parsing.
  * Updated Helm subproject reference.


packages/keystone/prepareKeystoneApp.js
```diff
@@ -1,5 +1,6 @@
 const debug = require('debug')('@open-condo/keystone/prepareKeystoneApp')
 const express = require('express')
+const proxyaddr = require('proxy-addr')
```
pahaz, line 3: it's better to inline this util https://github.com/jshttp/proxy-addr/blob/master/index.js

SavelevMatthew: I want it to be as close to express as possible, so if it somehow changes later, to be aligned. 
And its a bit too large to copy for me

SavelevMatthew: And it also has some imports in it, so it will be even bigger

Plus we already has is as dep in miniapp-utils package

PR #6770, opened 2025-10-10
fix(condo): DOMA-12393 added notifications after generate news by incident

Alllex202:
Problem: Browsers can block the opening of new tabs if several seconds pass between clicking on the button and the actual opening of the tab (in our case, this may be due to a long loading time)

Solution: We decided to add a notification with a button to go to the news creation page

## Summary by CodeRabbit

- New Features
  - Customizable completion message after incident create/update, with an optional “Go to News” action when news data is available.
  - Automatic opening of the news creation page with pre-filled news data when generation succeeds.

- Enhancements
  - Completion messaging can be customized or suppressed to better fit different flows.
  - Success notifications now include clearer, contextual descriptions and action buttons.

- Localization
  - Added notification texts and button labels for English, Spanish, and Russian.

### DIFF
--- a/apps/condo/domains/ticket/components/IncidentForm/BaseIncidentForm.tsx
+++ b/apps/condo/domains/ticket/components/IncidentForm/BaseIncidentForm.tsx
@@ -62,7 +62,7 @@ import { searchOrganizationProperty } from '@condo/domains/ticket/utils/clientSc
 import styles from './BaseIncidentForm.module.css'
 
 import type { FormRule as Rule } from 'antd'
-import type { ArgsProps } from 'antd/lib/notification'
+import type { ArgsProps as NotificationProps } from 'antd/lib/notification'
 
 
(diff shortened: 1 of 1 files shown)


apps/condo/domains/ticket/components/IncidentForm/BaseIncidentForm.tsx
```diff
@@ -81,6 +81,7 @@ export type BaseIncidentFormProps = {
     action: (values: IIncidentCreateInput | IIncidentUpdateInput) => Promise<Awaited<ReturnType<CreateIncidentMutationFn | UpdateIncidentMutationFn>>>
     afterAction?: () => Promise<void>
     showOrganization?: boolean
+    onCompletedMessage?: (incident, newsInitialValue) => any
```
nomerdvadcatpyat: Can you add return type instead any?

Alllex202: added

PR #6776, opened 2025-10-11
feat(ui): DOMA-12433 allow Checkbox to receive value and export Checkbox.Group, to make it all work in forms

YEgorLu:
No interface changes

Why? 

1. If you want to render multiple checkboxes in one Form.Item, you must wrap them in Checkbox.Group (otherwise you will receive signle { checked: true } in form values, whatever checkbox you click on. In case of group you will receive array of checked elements)

2. If you want to provide "initialValues" in form for checkbox group, each individual Checkbox inside Checkbox.Group should have "value" prop 

## Summary by CodeRabbit

- **New Features**
  - Added CheckboxGroup for managing multiple checkboxes; available as Checkbox.Group and supports controlled/default values and onChange.
  - Checkbox now accepts a value prop for improved form/group integration.
  - Radio now exposes a Group on the Radio export for grouped usage.

- **Behavior Changes**
  - Radio group props now accept id and className in addition to existing options.

- **Deprecated**
  - A deprecated alias was introduced for the Radio group export.

### DIFF
--- a/packages/ui/src/components/Radio/radio.tsx
+++ b/packages/ui/src/components/Radio/radio.tsx
@@ -18,11 +18,7 @@ type CondoRadioProps = {
 export type RadioProps = Pick<DefaultRadioProps, 'autoFocus' | 'defaultChecked' | 'disabled' | 'onChange' | 'checked' | 'value' | 'children' | 'id'>
 & CondoRadioProps
 
-export interface IRadio {
-    (props: RadioProps): React.ReactElement
-}
-
-const Radio: IRadio = (props) => {
+const Radio: React.FC<RadioProps> = (props) => {
     const { label, icon, labelProps, disabled, onChange, children, id, ...rest } = props
 
     const handleChange = useCallback((event: RadioChangeEvent) => {


packages/ui/src/components/Checkbox/checkbox.tsx
```diff
@@ -66,6 +67,8 @@ const Checkbox: React.FC<CheckboxProps> = (props) => {
     )
 }
 
+Checkbox.Group = DefaultCheckbox.Group
```
SavelevMatthew: It's better to aggregate it in `index.ts` file, like other components

packages/ui/src/components/Checkbox/checkbox.tsx
```diff
@@ -25,9 +25,10 @@ export type CheckboxProps = Pick<DefaultCheckboxProps,
 | 'className'
 | 'onMouseLeave'
 | 'onMouseEnter'
-| 'tabIndex'> & CondoCheckboxProps
+| 'tabIndex'
+| 'value'> & CondoCheckboxProps
 
-const Checkbox: React.FC<CheckboxProps> = (props) => {
+const Checkbox: React.FC<CheckboxProps> & { Group: typeof DefaultCheckbox['Group'] } = (props) => {
```
SavelevMatthew: Are you sure you want 100% props of Default Checkbox?

packages/ui/src/components/Checkbox/checkbox.tsx
```diff
@@ -25,9 +25,10 @@ export type CheckboxProps = Pick<DefaultCheckboxProps,
 | 'className'
 | 'onMouseLeave'
 | 'onMouseEnter'
-| 'tabIndex'> & CondoCheckboxProps
+| 'tabIndex'
+| 'value'> & CondoCheckboxProps
```
SavelevMatthew, line 29: Why not just checked?

YEgorLu: checked is okay if you use one single checkbox

If you need a group of checkboxes (f.e. pick M options out of N options), then you need to wrap it in Group and most likely you will wrap it in Form. And if you want to actually receive selected values / control values or initial values, you must provide "value" prop

packages/ui/src/components/Checkbox/checkbox.tsx
YEgorLu: why would I not want it?

packages/ui/src/components/Checkbox/checkbox.tsx
YEgorLu: moved to another file.

antd internally wraps it in memo(forwardRef(...)), I thing memo is not needed for us, but ref is undocumented and possibly can be used in some antd components, so I keep it

packages/ui/src/components/Checkbox/checkbox.tsx
YEgorLu: Decided to omit prefixCls, styles and options

packages/ui/src/components/Checkbox/checkboxGroup.tsx
```diff
@@ -0,0 +1,27 @@
...
+import React from 'react'
+
+import type { ComponentProps } from 'react'
+
+const DefaultCheckboxGroup = DefaultCheckbox.Group
+type DefaultCheckboxGroupProps = ComponentProps<typeof DefaultCheckboxGroup>
+
+export type CheckboxGroupProps = Pick<DefaultCheckboxGroupProps,
+'children'
+| 'className'
```
SavelevMatthew, line 11: id?

YEgorLu: there is no "id" prop...

YEgorLu:
> ## Quality Gate Failed
He's okay? We have similar code in other components, and very same in new Checkbox update in this pr
image

PR #6806, opened 2025-10-17
feat(condo): INFRA-1236 graceful shutdown

sitozzz:
## Summary by CodeRabbit

* **New Features**
  * Added a health-check endpoint that reports server readiness and returns a "shutting_down" status while the service is draining.

* **Improvements**
  * Implemented graceful shutdown with request draining, active-request tracking, connection closing, and configurable timeouts to ensure cleaner shutdowns and resource cleanup.

* **Chores**
  * Updated deployment submodule pointer.

### DIFF
--- a/bin/run-keystone-app.js
+++ b/bin/run-keystone-app.js
@@ -89,7 +89,7 @@ try {
     logger.warn({ msg: 'load certs error', err })
 }
 
-function setupGracefulShutdown ({ app, keystone, httpServer, httpsServer, tracer }) {
+function setupGracefulShutdown ({ app, keystone, httpServer, httpsServer }) {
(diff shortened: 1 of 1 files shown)


YEgorLu:
Is there a way to potentially know that server is shutting down, so we can, for example, abort request, or some other stuff?

sitozzz:
> Is there a way to potentially know that server is shutting down, so we can, for example, abort request, or some other stuff?

Everything here is designed to avoid interrupting requests. K8s stops sending traffic to the pod after sending a stop signal, but waits for the application to complete requests that started before receiving this signal.

The application responds to these requests, and when the active request counter reaches 0, it terminates as before.

dkoviazin (approved):

Issue #6094, opened 2025-04-18
Build/Migration Python Error on Windows When Running/Migration Condo Locally


oleg-murmur:
**Description:**  
I’m trying to set up Condo locally but encountered errors during the process. I followed the steps in `main/README.md` but ran into issues. Below are my setup details and the steps I took before the error occurred.  

**Environment:**  
- **OS:** Windows 11  
- **Installed prerequisites:**  
  - Docker (running)  
  - Python 3.13.3 (see [screenshot](https://github.com/user-attachments/assets/673aa1af-7da2-49e0-a086-84cb9a6e766e)).  
  - pip 25.0.1 (see [screenshot](https://github.com/user-attachments/assets/c49117a2-3325-457f-b497-2856ed98a1f9)).  
  - Node.js 16.20.2 (see [screenshot](https://github.com/user-attachments/assets/18779843-d1c3-4ab4-8abb-72f86db3f3a2)).

  - VS Code

**Following the doc steps:**
1. Start the databases using docker compose with this command:
`docker compose up -d postgresdb redis`

Result:
![Image](https://github.com/user-attachments/assets/15acd682-bf6a-4456-ba18-b58059dd2380)

2. Environment setup
![Image](https://github.com/user-attachments/assets/63f0e74b-e9a3-4c2a-a401-8b033529aac9)

3. Installing dependencies
Run commands:  
`yarn install`
`npm i -g turbo@^2`
`pip install Django psycopg2-binary`

Result:
![Image](https://github.com/user-attachments/assets/4922472c-4a64-40c2-8542-2c0d07d65e8f)

4. Building @open-condo dependencies

Build Error in @open-condo/bridge on Windows: rm Command Not Found

## **Problem 1:**
When running yarn workspace @app/condo build:deps, the following error occurs:

![Image](https://github.com/user-attachments/assets/2dca9b8e-5b9b-44c5-ad16-925ff75a323f)

@open-condo/bridge:build: 'rm' is not recognized as an internal or external command, operable program or batch file.

This happens because the build script uses the Unix command rm, which is not available in Windows.

Suggested Fixes:
   -  Replace rm with a cross-platform alternative (rimraf or del).
   - Add OS detection in scripts (e.g., use rimraf for all platforms).
   -  Update documentation to mention WSL/Git Bash requirements for Windows users.

Expected Outcome:
The build should work on both Unix-like systems (Linux/macOS) and Windows without errors.

[...]

PR #6816, opened 2025-10-21
fix(condo): #6094 python migrations fix on windows

Karasik876:
On Windows, ./kmigrator.py cannot be executed directly because .py files are not recognized as executables.

Using python ./kmigrator.py works cross-platform and preserves Unix compatibility.

Shouldn't break anything on Unix systems because there shebang takes priority as an interpreter when executing kmigrator.py script BUT on Windows it is ignored causing the error described in #6094 issue. 

## Summary by CodeRabbit

* **Chores**
  * Migration-related npm scripts (kmigrator, makemigrations, migrate, migrate:down, migrate:unlock) updated across multiple apps and templates to invoke the Python interpreter explicitly (python3) for more consistent execution.


pahaz:
> Should it be `python3`, because on Mac and some linux distress there is no `python` binary by default? 🤔

Yes. On macOS and others distros there’s no python (or it points to python3). But we shouldn’t rely on the system interpreter at all. The right way is to use an isolated environment (venv/virtualenv, Poetry, or Pipenv) and run its "own" python with required deps.

Probably, we can change the npm run command for now if it helps to solve a windows compatibility problem.

Karasik876:
> > Should it be `python3`, because on Mac and some linux distress there is no `python` binary by default? 🤔
> 
> Yes. On macOS and others distros there’s no python (or it points to python3). But we shouldn’t rely on the system interpreter at all. The right way is to use an isolated environment (venv/virtualenv, Poetry, or Pipenv) and run its "own" python with required deps.
> 
> Probably, we can change the npm run command for now if it helps to solve a windows compatibility problem.

Totally agree that using virtual environments is the proper way to manage Python dependencies in general.

In this project though Python is only used for migrations (kmigrator.py), and the main stack is JS. Requiring every contributor to set up and activate a venv would add extra overhead for what’s just a lightweight helper script.



On macOS and many Linux distros, `python3` is indeed the default command since `python` may not exist.
On Windows, however, only `python` (not `python3`) is available by default, so using `python3` breaks there.

So the solution I see would be to use this syntax: `python3 ./../../bin/kmigrator.py || python ./../../bin/kmigrator.py`


The only downside I see is that the code becomes a little more verbose:
```json
"kmigrator": "python3 ./../../bin/kmigrator.py || python ./../../bin/kmigrator.py",
"makemigrations": "python3 ./../../bin/kmigrator.py makemigrations || python ./../../bin/kmigrator.py makemigrations",
"migrate": "python3 ./../../bin/kmigrator.py migrate || python ./../../bin/kmigrator.py migrate",
"migrate:down": "python3 ./../../bin/kmigrator.py down || python ./../../bin/kmigrator.py down",
"migrate:unlock": "python3 ./../../bin/kmigrator.py unlock || python ./../../bin/kmigrator.py unlock",
```

Karasik876:
After testing cross-platform behavior, I found that the previous script definitions using `./../../bin/kmigrator.py` or `python ./../../bin/kmigrator.py` or `python3 ./../../bin/kmigrator.py || python ./../../bin/kmigrator.py` do not work consistently across different environments.

- Direct execution (`./kmigrator.py`) works on Unix because the shebang (`#!/usr/bin/env python3`) is respected, **but fails on Windows** where.
- Using `python` **explicitly** works on some Windows machines, but breaks on many modern Unix systems where `python` may not exist by default (most Linux distros and macOS only provide `python3`).
- The fallback syntax `python3 ./../../bin/kmigrator.py down || python ./../../bin/kmigrator.py down` also does not work in all cases.
For [example](https://github.com/open-condo-software/condo/blob/66a75659e25a96e1b93658a29dd2a76ab4404a86/apps/condo/docs/migrations.md?plain=1#L117):
```
docker-compose run app python apps/condo/.kmigrator/manage.py makemigrations --merge
```
fails when used with that fallback approach.

Using `python3` explicitly is the most balanced solution:
- It guarantees execution on all Linux/macOS machines, which make up the majority of contributors.
- It works on Windows systems where `python3` is available in PATH, covering many common installations.
- It avoids forcing Unix users to manually create symlinks (`python` -> `python3`).

apps/address-service/package.json
```diff
@@ -10,10 +10,10 @@
     "build:deps": "turbo build --filter=@app/address-service^...",
     "build": "keystone build",
     "start": "NODE_ENV=production node ./../../bin/run-keystone-app.js",
-    "kmigrator": "./../../bin/kmigrator.py",
-    "makemigrations": "./../../bin/kmigrator.py makemigrations",
-    "migrate": "./../../bin/kmigrator.py migrate",
-    "migrate:down": "./../../bin/kmigrator.py down",
+    "kmigrator": "python3 ./../../bin/kmigrator.py",
```
toplenboren, line 13: We can also create a nodeJS script that will 

0. Init/check python virtual environment? 
1. Check the OS
2. Check the Python execution file
3. Run the correct installer

That way we would be able to cover all the cases and have one specific place to edit configurations

toplenboren:
But overall I agree with changing the command if this fixes windows compatibility 👍

toplenboren (approved):

apps/address-service/package.json
Karasik876: Yes, I also thought about that. I just couldn't test it on all platforms.

This is the script I have, not sure if it works properly:
```javascript
#!/usr/bin/env node
const { spawnSync } = require('child_process');

const args = process.argv.slice(2);
const kmigratorPath = './../../bin/kmigrator.py';

const tryPython = (cmd) => spawnSync(cmd, [kmigratorPath, ...args], {
  stdio: 'inherit'
});

let result = tryPython('python3');

// ENOENT (Error NO ENTry) - python3 executable not found in PATH, try python instead
if (result.error && result.error.code === 'ENOENT') {
  result = tryPython('python');
}

process.exit(result.status ?? 1);
```

And we can execute it like that:
```json
"kmigrator": "node ./../../bin/some_kmigrator_script_name.js",
"makemigrations": "node ./../../bin/some_kmigrator_script_name.js makemigrations",
```
`node` fortunately doesn't have `python`'s problem. It executes the same on windows and unix.

PR #6831, opened 2025-10-27
hotfix(condo): fallback for editor

nomerdvadcatpyat:
## Summary by CodeRabbit

* **Improvements**
  * Enhanced rich-text editor robustness with initialization timeout and health detection.
  * Added graceful fallback to a disabled/plain text area when the editor fails to load.
  * Improved loading UX by showing a loader only while the editor is actively loading and suppressing it on failure.


nomerdvadcatpyat:
> Why didn't use textarea by default instead of forcing other developers to find api key to get things work?

done

PR #6832, opened 2025-10-28
fix(rb): DOMA-12504 added try catch to updatePayment

nomerdvadcatpyat:
## Summary by CodeRabbit

* **Chores**
  * Updated an internal repository pointer. This maintenance change does not modify any user-facing features, exported APIs, or public declarations and should not affect behavior. No functionality changes are expected. Estimated review effort: Low.


dkoviazin (approved):

AleX83Xpert (commented):
If I understand correctly, the problem is that the payment tries to update the invoice's status.
It seems the only thing that's been resolved here is that the server won't crash. But it would be better to fix it on the Condo side. For example, restricting the update of certain ticket fields if there's a published invoice. Or to modify the resolve-hook. The issue of payment and invoice status desynchronization will remain; we're just hiding it from ourselves :). Now we'll only hear about it from a resident: "I paid for the invoice, it was debited, but the invoice still hasn't been paid".

PR #6839, opened 2025-10-28
chore(condo): DOMA-12507 removed a generation of old gql-utils

Alllex202:
Client utilities are currently generated based on `.graphql` files from `<domain>/queries/*.graphql`. 
Therefore, there is no need to generate old GQL utilities for client.

## Summary by CodeRabbit

* **Refactor**
  * Simplified code generation templates by removing auto-generation markers and reorganizing export structure.
  * Removed generated React hooks for domain entity operations, requiring alternative implementation patterns for CRUD interactions.


sitozzz (approved):
okay

PR #6840, opened 2025-10-29
refactor(condo): DOMA-12509 remove jspdf dependency

nomerdvadcatpyat:
Used pdfmake instead of jspdf 
https://github.com/open-condo-software/condo/security/dependabot/413 https://github.com/open-condo-software/condo/security/dependabot/366

1. Migrated `createWrappedPdf` function to pdfmake. It is used in `AcquiringReceipt` (condo) and `Check` (eps) components.
2. Removed `createPdfWithPageBreaks` function with an old analytics pages (reports/detail/report-by-tickets). They are no longer used by anyone.
3. Removed jspdf dependency

## Summary by CodeRabbit

* **Refactor**
  * PDF export simplified to run only in the browser and produce a single-page, image-based download for reports.

* **Chores**
  * Ticket analytics report pages removed; old report URLs now redirect to the general reports area.
  * Removed an unused PDF library dependency.

### DIFF
--- a/apps/condo/next.config.ts
+++ b/apps/condo/next.config.ts
@@ -150,7 +150,7 @@ const nextConfig: NextConfig = {
             },
             {
                 source: '/reports/detail/report-by-tickets/:path*',
-                destination: '/reports/:path*',
+                destination: '/reports',
                 permanent: false,
             },
         ]


apps/condo/pages/reports/detail/report-by-tickets/index.tsx
tolmachev21, line 1: Are we not using these pages?

sitozzz: That page is an old version of analytics page. So it's safe to delete it

sitozzz, line 1: But there is should be some url rewrites, if I remember correctly

nomerdvadcatpyat: Yes

nomerdvadcatpyat: I'll add

nomerdvadcatpyat: Added

sitozzz (approved):
Okay, now looks good

PR #6877, opened 2025-11-06
feat(condo): DOMA-12316 restricted Meter.archiveDate from the future

abshnko:
image

## Summary by CodeRabbit

* **New Features**
  * Archive dates can only be set to past dates — today and future dates are disabled in the date picker; saving a future date is ignored and the previous value is retained.

* **Tests**
  * Added/adjusted tests verifying archiveDate behavior: past dates accepted; future dates rejected and do not overwrite existing values.

* **Documentation**
  * Field descriptions updated to note archiveDate is treated as a flag and cannot be in the future.

### DIFF
--- a/apps/condo/domains/meter/schema/RegisterPropertyMetersReadingsService.test.js
+++ b/apps/condo/domains/meter/schema/RegisterPropertyMetersReadingsService.test.js
@@ -911,7 +911,7 @@ describe('RegisterPropertyMetersReadingsService', () => {
         // create another reading for same meter and change `verificationDate`, `nextVerificationDate` and `archiveDate` fields values
         const nextVerificationDate = dayjs().add(1, 'week').toISOString()
         const verificationDate2 = dayjs().subtract(1, 'week').toISOString()
-        const archiveDate = dayjs().add(2, 'week').toISOString()
+        const archiveDate = dayjs().subtract(2, 'week').toISOString()
         const anotherReadings = [{
             ...readings[0],
             value1: faker.random.numeric(3),


apps/condo/domains/meter/components/Meters/ChangeMeterStatusModal.tsx
```diff
@@ -55,6 +55,7 @@ const ChangeMeterStatusModal = ({
                     format={ARCHIVE_MODAL_DATE_PICKER_DATE_FORMAT}
                     style={ARCHIVE_MODAL_DATE_PICKER_STYLE}
                     onChange={handleChangeSelectedArchiveDate}
+                    disabledDate={current => current && current >= dayjs()}
```
YEgorLu, line 58: why not wrap it in useCallback?

apps/condo/domains/meter/schema/fields/index.js
```diff
@@ -64,8 +66,21 @@ const sealingDate = {
...
-    schemaDoc: 'The date when the meter was disabled and no longer in use',
+    schemaDoc: 'The date when the meter was disabled and no longer in use. Is treated like a flaf and cannot be in the future.',
     type: 'DateTimeUtc',
+    hooks: {
+        resolveInput: ({ resolvedData, fieldPath, existingItem }) => {
+            const value = resolvedData[fieldPath]
+            if (!value) return existingItem?.[fieldPath] || null
+
+            const today = dayjs()
+            if (dayjs(resolvedData[fieldPath]).isAfter(today)){
```
YEgorLu:
```suggestion
            if (dayjs(value).isAfter(today)){
```

apps/condo/domains/meter/schema/fields/index.js
```diff
@@ -64,8 +66,21 @@ const sealingDate = {
...
+                return existingItem?.[fieldPath] || null
+            }
+
+            return resolvedData[fieldPath]
```
YEgorLu:
```suggestion
            return value
```

apps/condo/domains/meter/components/Meters/ChangeMeterStatusModal.tsx
abshnko: it's quite an inexpensive operation and won't affect re-renders of children on creating inline function

AleX83Xpert (approved):

dkoviazin (approved):

PR #6902, opened 2025-11-12
fix(global): INFRA-1308 file upload utils

sitozzz:
## Summary by CodeRabbit

* **New Features**
  * Chunked, metadata-rich batch document uploads with per-file status updates, cancellation and automatic form reset
  * New files package: unified upload/attach/share APIs for browser and server, plus build/export surface
  * Exposes runtime file client ID for configuration
  * File access tokens extended to 1 hour

* **Bug Fixes**
  * Better handling of oversized files, improved error reporting, recovery and robust upload status management

### DIFF
--- a/packages/keystone/fileAdapter/awsFileAdapter.js
+++ b/packages/keystone/fileAdapter/awsFileAdapter.js
@@ -123,7 +123,7 @@ class AwsFileAdapter {
     publicUrl ({ filename, originalFilename, ...props }, user) {
         let folder = this.folder
         let sign
-        if ('meta' in props && props['meta']['appId']) {
+        if ('meta' in props && props['meta']['fileClientId']) {
             folder = props['meta']['sourceFileClientId'] || props['meta']['fileClientId']
             if (!conf['FILE_SECRET']) {
                 throw new Error('FILE_SECRET is not configured')


packages/files/server.js
```diff
@@ -0,0 +1,260 @@
...
+    const raw = conf['FILE_UPLOAD_CONFIG']
+    if (!raw) throw new Error('FILE_UPLOAD_CONFIG is not set')
+    const cfg = typeof raw === 'string' ? JSON.parse(raw) : raw
+    const secret = cfg?.clients?.[fileClientId]?.secret
+    if (!secret) throw new Error(`Unknown fileClientId "${fileClientId}"`)
+    return secret
+}
+
+function getServerUrl () {
+    const serverUrl = conf['SERVER_URL'] || conf['FILE_SERVER_URL']
```
nomerdvadcatpyat: Maybe FILE_SERVER_URL will be a priority?

packages/files/src/client.ts
```diff
@@ -0,0 +1,243 @@
...
+ * Get server URL from environment variables or use provided override.
+ */
+function getServerUrl (override?: string): string {
+    if (override) {
+        return override.replace(/\/$/, '')
+    }
+
+    if (typeof window !== 'undefined') {
+        const envUrl = (typeof process !== 'undefined' && process.env?.NEXT_PUBLIC_SERVER_URL) ||
+                       (window as any).__NEXT_DATA__?.env?.NEXT_PUBLIC_SERVER_URL
```
nomerdvadcatpyat: What if client is not Next app? Mb we need to check just SERVER_URL env too?
Also i think we need to check FILE_SERVER_URL env here, like in other places

packages/files/src/client.ts
```diff
@@ -0,0 +1,243 @@
...
+
+export async function upload ({ serverUrl, cookie, files, meta, attach }: UploadOptions): Promise<UploadResponse> {
+    let FormDataClass: typeof FormData
+    let form: FormData | any
+
+    if (typeof window !== 'undefined') {
+        FormDataClass = FormData
+        form = new FormDataClass()
+    } else {
+        const FormDataModule = require('form-data')
```
nomerdvadcatpyat: Why not use FormData from Node.js?

packages/files/src/client.ts
```diff
@@ -0,0 +1,243 @@
...
+        form = new FormDataClass()
+    }
+
+    if (!Array.isArray(files) || files.length === 0) {
+        throw new Error('upload: "files" must be a non-empty array')
+    }
+
+    for (const f of files) {
+        if (typeof Buffer !== 'undefined') {
+            form.append('file', f, 'file.bin')
```
nomerdvadcatpyat: Is it ok that all files have 'file.bin' name? Mb is it worth giving unique names?

packages/files/src/client.ts
```diff
@@ -0,0 +1,243 @@
...
+        } else if (f && typeof f === 'object' && 'buffer' in f && 'filename' in f) {
+            form.append('file', (f as { buffer: Buffer, filename: string }).buffer, (f as { buffer: Buffer, filename: string }).filename)
+        } else {
+            const file = f as File | Blob
+            form.append('file', file, (file as File).name || 'file.bin')
+        }
```
nomerdvadcatpyat: Can you please describe why we have 3 ways to pass file?

packages/files/src/client.ts
sitozzz: Yep, that's sounds reasonable

packages/files/server.js
sitozzz: Yep, that's a good point though

packages/files/src/client.ts
sitozzz: Because this block of the code is responsible from nodejs (server side) execution. So at the server side in nodejs is no form data bundled by default, so u need to import/require it from that package

packages/files/src/client.ts
sitozzz: That's not really necessary here, because if user is not give us 'originalFileName' on the upload state - he don't might what actually it should be called. But it's okay to change it to the random one

packages/files/src/client.ts
sitozzz: Because there is 3 ways to provide a file to the uploading form.
1) By the buffer
2) By "custom" object with buffer and filename 
3) Finally by Blob object

PR #6963, opened 2025-11-25
feat(miniapp-utils): DOMA-12639 make fingerprint cookie persistent

YEgorLu:
1. It had no "expiration" or "maxAge" -> was clear out after browser closes
2. Since 2022 browsers round maxAge to 400 days max, so let's renew cookie expiration every time we can

## Summary by CodeRabbit

* **Bug Fixes**
  * Made client fingerprint cookies persistent by renewing their expiration on each access and extending their lifespan so fingerprints persist across sessions, improving continuity and reducing repeated sign‑ins.


abshnko (approved):
👍🏻

PR #6973, opened 2025-11-27
feat(condo): DOMA-12648 firebase support many configs

tolmachev21:
## Summary by CodeRabbit

* **New Features**
  * Multi-app Firebase notifications: per-app configuration, routing and batching so multiple apps can send notifications independently.

* **Improvements**
  * Per-app error handling, logging and aggregated per-app send results for clearer delivery diagnostics.
  * Introduced a default Firebase app identifier for fallback behavior.
  * Updated example env config to support per-app Firebase credentials.

* **Tests**
  * Updated notification tests to cover per-app behavior.

* **Chores**
  * Updated submodule reference.

### DIFF
--- a/apps/condo/domains/notification/adapters/firebaseAdapter.js
+++ b/apps/condo/domains/notification/adapters/firebaseAdapter.js
@@ -51,16 +51,15 @@ class FirebaseAdapter {
     messageIdPrefixRegexpByAppId = {}
 
     constructor (config = FIREBASE_CONFIG) {
-        if (!isObject(config) || isEmpty(config)) {
-            throw new Error(`${EMPTY_FIREBASE_CONFIG_ERROR}. Expected a "${FIREBASE_DEFAULT_APP_ID}" app configuration with key "${FIREBASE_DEFAULT_APP_ID}" for fallback purposes.`)
-        }
-
-        for (const [appId, appConfig] of Object.entries(config)) {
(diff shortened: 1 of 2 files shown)


apps/condo/domains/notification/transports/push.spec.js
```diff
@@ -547,6 +547,77 @@ describe('push transport', () => {
 
             })
 
+            it('successfully sends notification to multiple apps with different appIds', async () => {
```
YEgorLu: If this test just checks that you can send messages with different appIds, then ok, but this behaviour always been there. If it checks that it uses different configs for each appId, then I don't see how it is doing so

apps/condo/domains/notification/adapters/firebaseAdapter.js
```diff
@@ -44,23 +45,32 @@ const logger = getLogger()
  * Attempts to send push notifications to devices, connected through different projects will fail.
  */
 class FirebaseAdapter {
-    app = null
-    projectId = null
+    appsByAppId = {}
```
YEgorLu, line 50: I think you must tell here about "default" config, since you have a whole constant for it. Maybe even put it in different field or have a check in constructor. In case of constructor, seems like you can reuse if (isEmpty(config)) throw new Error(EMPTY_FIREBASE_CONFIG_ERROR), just specify that you are expecting a "default" one

Alexander-Turkin (approved):

PR #6977, opened 2025-11-28
feat(condo): DOMA-12666 added check for active acquiring context when calculating BillingReceipt.isPayable

abshnko:
## Summary by CodeRabbit

* **Bug Fixes**
  * Payment eligibility for billing receipts now requires the presence and correct finished-status of related billing and acquiring integration contexts; receipts are marked non-payable when related contexts are missing, deleted, or not finished.

* **Tests**
  * Expanded coverage for receipt payability across context states (including soft-deletes and non-finished statuses); added test utilities to support these scenarios.

### DIFF
--- a/apps/condo/domains/billing/schema/BillingReceipt.test.js
+++ b/apps/condo/domains/billing/schema/BillingReceipt.test.js
@@ -1350,8 +1350,13 @@ describe('BillingReceipt', () => {
         })
         describe('virtual fields check', () => {
             it('checking the completion of virtual fields: canGroupReceipts, hostUrl, acquiringIntegrationId, currencyCode', async () => {
+                const [organization] = await createTestOrganization(admin)
                 const [acquiringIntegration] = await createTestAcquiringIntegration(admin)
-                const [acquiringContext] = await createTestAcquiringIntegrationContext(admin, context.organization, acquiringIntegration)
+                const [billingIntegration] = await createTestBillingIntegration(admin)
+                const [acquiringContext] = await createTestAcquiringIntegrationContext(admin, organization, acquiringIntegration)
+                const [context] = await createTestBillingIntegrationOrganizationContext(admin, organization, billingIntegration)
+                const [property] = await createTestBillingProperty(admin, context)
+                const [account] = await createTestBillingAccount(admin, context, property)
                 await updateTestAcquiringIntegrationContext(admin, acquiringContext.id, { status: CONTEXT_FINISHED_STATUS })
                 const [billingReceipt] = await createTestBillingReceipt(admin, context, property, account)
 


apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -214,6 +214,19 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
 
                 if (!category) return false
 
+                const billingContext = await getByCondition('BillingIntegrationOrganizationContext', {
```
YEgorLu, line 216: Shouldn't you filter for finished context?

apps/condo/domains/billing/schema/BillingReceipt.js
```diff
@@ -214,6 +214,19 @@ const BillingReceipt = new GQLListSchema('BillingReceipt', {
...
+                    deletedAt: null,
+                })
+                if (!billingContext) return false
+
+                const activeAcquiringContexts = await find('AcquiringIntegrationContext', {
+                    organization: { id: get(billingContext, 'organization') },
+                    status: CONTEXT_FINISHED_STATUS,
+                    deletedAt: null,
+                })
+                if (!activeAcquiringContexts || activeAcquiringContexts.length !== 1) return false
```
YEgorLu, line 227: ? Why check for length?

YEgorLu (approved):

apps/condo/domains/billing/schema/BillingReceipt.js
AleX83Xpert, line 227: Why not?
```suggestion
                if (!activeAcquiringContexts || activeAcquiringContexts.length === 0) return false
```

apps/condo/domains/billing/schema/BillingReceipt.js
abshnko: context can be InProgress if integration for getting receipts is turned off for some reason, but already imported receipts should be payable

apps/condo/domains/billing/schema/BillingReceipt.js
abshnko: in our system we have a limit of only one active acquiring context

apps/condo/domains/billing/schema/BillingReceipt.js
AleX83Xpert: Could you please import this constant from the billing domain?

apps/condo/domains/billing/schema/BillingReceipt.js
abshnko: i wanted to, but at the end decided to be on the safe side and if for some reason there are more than 1 active contexts, to forbid paying. in case there was a mistake

apps/condo/domains/billing/schema/BillingReceipt.js
abshnko: fixed

PR #7006, opened 2025-12-05
fix(condo): DOMA-12689 fix empty email

Alexander-Turkin:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved email validation during employee import to more accurately identify invalid emails, reducing false error messages when email fields are legitimately empty.


apps/condo/domains/organization/hooks/useEmployeeImporterFunctions.tsx
```diff
@@ -166,7 +166,8 @@ export const useEmployeeImporterFunctions = (): [Columns, RowNormalizer, RowVali
         }
 
         const email = row?.addons?.email
-        if (email === null) {
+        const emailValue = String(row?.row?.[5]?.value ?? '').trim()
+        if (emailValue && email === null) {
```
toplenboren, line 170: this is really hard to read:

1. why email is different from emailValue?
2. why is the code to get emailValue is so complex? row.row[5] (>_<) .value or '' (?) with .trim()

Let's rename these vars and make the code to get emailValue more simple – maybe use a ternary and move trim() into another place?

If actual email value can be obtained from multiple places it is best to just write it as is, like in here: emailFromRowData, emailFromAuthedUser

PR #7009, opened 2025-12-08
fix(condo): fix webhook ci

Alexander-Turkin:
## Summary by CodeRabbit

* **Chores**
  * Updated internal configuration settings with no impact on application functionality or user experience.

### DIFF
--- a/packages/webhooks/tasks/sendWebhook.spec.js
+++ b/packages/webhooks/tasks/sendWebhook.spec.js
@@ -42,7 +42,7 @@ const { getWebhookTasks } = require('@open-condo/webhooks/tasks')
 const SendWebhookTests = (appName, actorsInitializer, userCreator, userDestroyer, entryPointPath) => {
     describe(`sendWebhook task basic tests for ${appName} app`, () => {
         const appEntryPoint = require(entryPointPath)
-        setFakeClientMode(appEntryPoint, { excludeApps: ['OIDCMiddleware', 'NextApp', 'AdminUIApp'] })
+        setFakeClientMode(appEntryPoint, { excludeApps: ['OIDCMiddleware'] })
 
         let sendWebhook
         let actors
--- a/apps/condo/next.config.ts
+++ b/apps/condo/next.config.ts
@@ -75,6 +75,7 @@ const nextConfig: NextConfig = {
         '@open-condo/featureflags',
         '@condo/domains',
     ],
+    experimental: {},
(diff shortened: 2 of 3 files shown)


packages/webhooks/tasks/sendWebhook.spec.js
```diff
@@ -42,7 +42,7 @@ const { getWebhookTasks } = require('@open-condo/webhooks/tasks')
 const SendWebhookTests = (appName, actorsInitializer, userCreator, userDestroyer, entryPointPath) => {
     describe(`sendWebhook task basic tests for ${appName} app`, () => {
         const appEntryPoint = require(entryPointPath)
-        setFakeClientMode(appEntryPoint, { excludeApps: ['OIDCMiddleware'] })
+        setFakeClientMode(appEntryPoint, { excludeApps: ['OIDCMiddleware', 'NextApp', 'AdminUIApp'] })
```
SavelevMatthew: Why does it matter here?

Alexander-Turkin: It seems there is problem with compatibility of next and node 24, ui exception fixes the problem, at least locally)

in setFakeClientMode we are rewriting excludeApps
```
const res = await prepareKeystoneExpressApp(entryPoint, { excludeApps: ['NextApp', 'AdminUIApp'], ...prepareKeystoneOptions })
```

SavelevMatthew: what exactly broke? its strange that you try to hack something in test instead of understanding and fixing the root of the issue

SavelevMatthew: Does it mean prod if affected or what?

vovaaxeapolla: It's sort of bug in jest and next.js and chromium together. There are some changes in next.js config file. They changed cloneObject func and now it fails on loading config file during tests due to broken dynamic import system for vm used by jest. So prod and dev doesn't use jest so there is no error on these stands.

This fragment of code from next.js config loader file works improperly in tests
```
    const proto = Object.getPrototypeOf(obj);
    const isPlainObject = proto === Object.prototype || proto === null;
```

So I found 3 solutions of the current problem
Don't use next.js in our tests
Add empty experimental field in config
Add inner next env __NEXT_TEST_MODE = 'jest', or figure out how to set it properly in tests

PR #7015, opened 2025-12-09
fix(condo): DOMA-12533 turn off cache in contact page and fix offset query type

tolmachev21:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed table state reset on Contacts page when navigating without query parameters; search input and row selections are cleared.
  * Stabilized table state change callbacks to prevent missed or stale updates.

* **Improvements**
  * Global search now syncs with the table and export actions so exports reflect current search.
  * Navigation uses the actual route path to avoid unexpected fallbacks.
  * Data fetching always uses network-only to ensure up-to-date results.

* **Style**
  * Filter controls updated to use consistent inline styling.

### DIFF
--- a/packages/ui/src/components/Table/table.tsx
+++ b/packages/ui/src/components/Table/table.tsx
@@ -233,16 +233,16 @@ function TableComponent<TData extends RowData = RowData> (
         return tableColumns
     }, [tableColumns, columnOrder])
 
-    const stableOnTableStateChange = useRef(onTableStateChange)
+    const onTableStateChangeRef = useRef(onTableStateChange)
     
     useEffect(() => {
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/Table/table.tsx
```diff
@@ -233,18 +233,24 @@ function TableComponent<TData extends RowData = RowData> (
         return tableColumns
     }, [tableColumns, columnOrder])
 
+    const stableOnTableStateChange = useRef(onTableStateChange)
```
nomerdvadcatpyat:
```suggestion
    const onTableStateChangeRef = useRef(onTableStateChange)
```

Alexander-Turkin (approved):

PR #7026, opened 2025-12-11
feat(condo): update ticketComment access

StalinidzeCorp:
hello everyone! we are making a helpdesk system based on your application, we want to use comments to communicate with residents by email, for this we have made our own miniapp, and we check and create comments through the service account, but there is a problem with the fact that we cannot receive data along this path, so I propose changes in access

## Summary by CodeRabbit

**Bug Fixes**
- Enhanced ticket comment visibility for service users through updated access permissions
- Removed comment type restrictions, expanding comment creation and modification capabilities for service users

### DIFF
--- a/apps/condo/domains/ticket/access/TicketComment.js
+++ b/apps/condo/domains/ticket/access/TicketComment.js
@@ -43,7 +43,6 @@ async function canReadTicketComments (args) {
 
         return {
             ...accessFilter,
-            type: ORGANIZATION_COMMENT_TYPE,
         }
     }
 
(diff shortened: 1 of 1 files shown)


toplenboren (changes requested):
In this PR you are allowing service users to read resident comments. 

If we were to proceed with this PR, it would mean that all service accouts would gain access to resident comments, which is a blocker for us.

Can you please elaborate, why do you need to make this change? This way we'd be able to come up with an acceptable solution :-)

PR #7095, opened 2026-01-13
fix(condo): DOMA-12772 optimized SyncRemoteClientService

abshnko:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved push-token and device collision handling during remote client sync to avoid accidental token reassignment.
  * Ensures records are created or updated only when actual changes exist, preventing unnecessary writes.

* **Refactor**
  * Sync logic rewritten to reliably detect existing clients and apply precise create/update behavior.

* **Tests**
  * Added tests verifying no update occurs when incoming device data is unchanged.

### DIFF
--- a/apps/condo/domains/notification/schema/SyncRemoteClientService.js
+++ b/apps/condo/domains/notification/schema/SyncRemoteClientService.js
@@ -91,7 +91,19 @@ const SyncRemoteClientService = new GQLCustomSchema('SyncRemoteClientService', {
                 } else {
                     const diff = {}
                     for (const [key, value] of Object.entries(attrs)) {
-                        if (!isEqual(existing[key], value) && key !== 'sender') {
+                        if (key === 'sender') continue
+
+                        if (key === 'owner') {
+                            const existingOwnerId =  existing?.owner?.id || existing?.owner
+                            const newOwnerId = value?.connect?.id ?? null
+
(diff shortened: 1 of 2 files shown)


apps/condo/domains/notification/schema/SyncRemoteClientService.js
```diff
@@ -43,43 +45,68 @@ const SyncRemoteClientService = new GQLCustomSchema('SyncRemoteClientService', {
...
 
-                const userId = get(context, 'authedItem.id', null)
-                const owner = userId ? { disconnectAll: true, connect: { id: userId } } : null
-                const attrs = {
-                    dv, sender, deviceId, appId, owner,
+                const attrs = pickBy({
+                    dv, sender, deviceId, appId,
                     pushToken, pushTransport, devicePlatform, pushType, meta,
                     pushTokenVoIP, pushTransportVoIP, pushTypeVoIP,
+                    owner: userId ? { connect: { id: userId } } : undefined,
```
YEgorLu: 1. you forgot "disconnectAll: true" (I dunno why it was here, maybe important)

2. I think this will always differ from existring owner. Existing owner will be just { id } isn't it

abshnko: fixed, thanks

PR #7126, opened 2026-01-22
feat(condo): DOMA-12797 try to parallel push messages sending

YEgorLu:
Around 50% of push sendings end up taking >=1s of time.
To send 1 message (in adapter) we must wait 100-1000 ms from what I've seen.
So let's try and parallel these requests.
From what I've gathered this kind of load must not be blocked by providers

## Summary by CodeRabbit

* **New Features**
  * Parallelized push dispatch across apps and transports for faster delivery.

* **Bug Fixes**
  * Improved aggregation of per-item and per-app delivery results, fixing partial-counting issues.
  * Automatically clears invalid/expired push tokens to reduce repeated failures.

* **Chores**
  * Refactored notification delivery flow for clearer per-app/per-item error reporting and more robust result composition.

### DIFF
--- a/apps/condo/domains/notification/adapters/hcmAdapter.js
+++ b/apps/condo/domains/notification/adapters/hcmAdapter.js
@@ -304,7 +304,7 @@ class HCMAdapter {
                 } catch (error) {
                     const safeError = safeFormatError(error, false)
                     logger.error({ msg: 'sendNotification error', error: safeError })
-                    return { state: 'error', err: safeError }
+                    return { state: 'error', error: safeError }
                 }
             }))
 
@@ -317,7 +317,7 @@ class HCMAdapter {
                 const response = p.value
                 hcmResult.responses.push(response)
 
-                if (response.error) {
+                if (response.state === 'error') {
                     hcmResult.failureCount++
                     continue
                 }


apps/condo/domains/notification/adapters/hcmAdapter.js
```diff
@@ -281,7 +281,7 @@ class HCMAdapter {
         if (!isNull(this.apps[APP_MASTER_KEY]) && !isNull(this.apps[APP_RESIDENT_KEY]) && !isEmpty(notifications)) {
             const hcmResult = HCMAdapter.getEmptyResult()
 
-            for (let idx = 0; idx < notifications.length; idx++) {
+            const promises = await Promise.allSettled(notifications.map(async (_, idx) => {
```
AleX83Xpert, line 284: How much can the `notifications.length` be? Do we need to split into chunks?

apps/condo/domains/notification/adapters/hcmAdapter.js
```diff
@@ -293,35 +293,45 @@ class HCMAdapter {
...
                         pushToken: notification.token,
                         pushType: pushTypes[notification.token],
                         ...sendResult,
-                    })
-
-                    if (sendResult.code === PUSH_SUCCESS_CODE) hcmResult.successCount += 1
+                    }
+                } catch (error) {
+                    const safeError = safeFormatError(error, false)
+                    logger.error({ msg: 'sendNotification error', error: safeError })
```
AleX83Xpert, line 306:
```suggestion
                    logger.error({ msg: 'sendNotification error', err: safeError })
```

apps/condo/domains/notification/adapters/appleAdapter.js
```diff
@@ -207,34 +207,37 @@ class AppleAdapter {
...
             }
-            for (const [appId, notificationsBatchForApp] of Object.entries(notificationsByAppId)) {
+
+            const promises = await Promise.allSettled(Object.entries(notificationsByAppId).map(async ([appId, notificationsBatchForApp]) => {
                 const configForApp = this.#config[appId]
                 if (!configForApp) {
                     logger.error({ msg: 'unknown appId. Config was not found', data: { appId } })
-                    continue
+                    return {
+                        state: 'error',
```
AleX83Xpert, line 225: Maybe move `'error'` into a constant? This string is used in some places.

apps/condo/domains/notification/adapters/hcmAdapter.js
YEgorLu: there is 1 Message.
We take all RemoteClients we can and map them to { pushToken, ...notificationData } array. That array is notifications. From what I've seen its 10-20 at max

YEgorLu:
> Would you like to add tests? At least for successful parallel execution and for the erroneous one (one of the promises is not fulfilled).

The thing is, that there are no tests basically. Existing tests are using fake tokens, and these have different logic, then normal tokens. So we can say that push sending was never actually tested.

apps/condo/domains/notification/adapters/appleAdapter.js
YEgorLu: Here and there are a lot of things, which we can refactor. F.e. all adapters are basically same, we can make 1-2 classes and just extend them to calculate success/failure count and to use needed messaging service. So I think it all must be refactored at some point.

So I choose not to change anything like that right now

apps/condo/domains/notification/adapters/hcmAdapter.js
AleX83Xpert: Looks like an uncontrolled amount. Maybe split into chunks with controlled length?

apps/condo/domains/notification/adapters/appleAdapter.js
AleX83Xpert: ok, understood

apps/condo/domains/notification/adapters/hcmAdapter.js
YEgorLu: I guess naaah. Providers say we can send thousands push messages for same push token in parallel

AleX83Xpert: ok

PR #7134, opened 2026-01-26
hotfix(condo): pass initial height to global apps iFrame

abshnko:
## Summary by CodeRabbit

**Bug Fixes**
- Fixed visibility behavior for embedded content containers to ensure they are fully hidden when toggled off, preventing partial visibility and layout issues.


apps/condo/domains/miniapp/components/GlobalApps/GlobalAppsContainer.tsx
```diff
@@ -126,6 +126,7 @@ export const GlobalAppsContainer: React.FC = () => {
                     reloadScope='user'
                     ref={el => void (iframeRefs.current[index] = el as HTMLIFrameElement)}
                     hidden={!isDebug}
+                    initialHeight={0}
```
SavelevMatthew: why? you should instead refer to hidden prop

PR #7148, opened 2026-02-02
feat(global): INFRA-1461 extend kmigrator to support analytics views

SavelevMatthew:
## Release Notes

* **New Features**
  * Enabled analytics functionality across the platform for enhanced insights and reporting capabilities
  * Implemented sensitive data classification to protect user personal information including name, email, password, and contact details
  * Established database analytics infrastructure with dedicated schema and views to support data analysis


packages/keystone/plugins/analytical.js
```diff
@@ -0,0 +1,38 @@
+const { plugin } = require('./utils/typing')
+
+const SENSITIVE_FIELDS = new Set([
+    'phone',
+    'email',
+    'password',
+    'secret',
+    'token',
+    'settings',
```
dkoviazin: settings and state may be not sensitive

bin/kmigrator.py
```diff
@@ -673,12 +753,67 @@ def _repl1(x):
     if source != code:
         item.write_text(code, encoding='utf-8')
 
+def _generate_views_migration(ctx, fwd=True):
+    old_state = json.loads(ctx['__KNEX_VIEWS_MIGRATION_STATE__'])
+    new_state = json.loads(ctx['__KNEX_VIEWS_DATA__'])
+    if old_state == new_state:
+        return None
+    if not fwd:
```
sitozzz, line 761: What does `fwd` flag mean here?

bin/kmigrator.py
```diff
@@ -673,12 +753,67 @@ def _repl1(x):
...
+    def add_line(line):
+        nonlocal migration
+        migration += line + '\n'
+
+    def add_comment(comment):
+        add_line('--')
+        add_line('-- ' + comment)
+        add_line('--')
+
+    def generate_view_sql(lst_key, lst):
```
sitozzz, line 775: If this is an abbreviation of the word "list", then it seems redundant and I would not do it that way.
```suggestion
    def generate_view_sql(list_key, list):
```

bin/kmigrator.py
sitozzz, line 1: U also should change `VERSION` from (1, 6, 0) to (1, 7, 0)

bin/kmigrator.py
SavelevMatthew: It has conflict with python internal key, and arg from outer scope, so it was suggested by IDE to rename it. I've chose go-styled lst (since there's already fwd / bwd)

bin/kmigrator.py
SavelevMatthew: Direction of migration. forward = from previous state to current, backward = from current state to previous.

Since logic is the same...

bin/kmigrator.py
```diff
@@ -462,6 +464,50 @@ def main():
...
+    const config = {
+        dv: 1,
+        lists: {},
+    }
+
+    for (const [listKey, list] of Object.entries(keystone.lists)) {
+        if (list?.createListConfig?.analytical) {
+            // Exclude virtual fields
+            const fields = Object.keys(list.fieldsByPath).filter(field => list.createListConfig.fields[field]?.type?.type !== 'Virtual')
+            const sensitiveFields = fields.filter(field => list.createListConfig.fields[field]?.sensitive)
```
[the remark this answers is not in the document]
SavelevMatthew: It's already processed schema, so type is resolved to object { type: 'TypeName', implementation: ... }

bin/kmigrator.py
```diff
@@ -462,6 +464,53 @@ def main():
...
+            // Exclude virtual fields
+            const fields = Object.keys(list.fieldsByPath)
+                .filter(field => list.createListConfig.fields[field]?.type?.type !== 'Virtual')
+                // Important to check diffs in python
+                .toString()
+            const sensitiveFields = fields.filter(field => list.createListConfig.fields[field]?.sensitive)
+
+            config.lists[listKey] = {
+                fields,
+                sensitiveFields,
```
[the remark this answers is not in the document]
SavelevMatthew: My bad, it supposed to be `toSorted`, not `toString`

bin/kmigrator.py
```diff
@@ -699,15 +853,27 @@ def _4_1_makemigrations(ctx, merge=False, check=False, empty=False):
...
             print('WARN: filename={}'.format(filename))
             print('WARN: logfile={}\n'.format(log_file.resolve()))
             log_file.write_bytes(e.stderr)
-            text = KNEX_MIGRATION_TPL_NO_DOWN.format(**locals())
+            template = KNEX_MIGRATION_TPL_NO_DOWN
+            if not views_inserted and bwd_views_sql:
+                bwd_sql = bwd_views_sql
+                template = KNEX_MIGRATION_TPL
+                views_bwd_inserted = True
+            text = template.format(**locals())
```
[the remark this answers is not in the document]
SavelevMatthew: yep, good catch

bin/kmigrator.py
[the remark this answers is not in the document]
SavelevMatthew: fixed

bin/kmigrator.py
sitozzz: okay, now I get it

bin/kmigrator.py
sitozzz: Yes, I didn't think about that. Then let it remain so.

sitozzz (approved):
btw, looks good

packages/keystone/plugins/analytical.js
SavelevMatthew: it might not be, but it might contain some non-analytical info (at least from our architectural decisions), so I'll keep them here

apps/condo/migrations/20260202130159-0507_auto_20260202_0802.js
```diff
@@ -0,0 +1,33 @@
...
+    await knex.raw(`
+    
+--
+-- Create analytics schema
+--
+CREATE SCHEMA IF NOT EXISTS "analytics";
+--
+-- Create view for "User" table
+--
+CREATE OR REPLACE VIEW "analytics"."User" AS SELECT "name", NULL as "password", "type", "isAdmin", "isSupport", NULL as "email", "isEmailVerified", NULL as "phone", "isPhoneVerified", "externalPhone", "isExternalPhoneVerified", "externalEmail", "isExternalEmailVerified", "externalSystemName", "hasMarketingConsent", "avatar", "meta", "locale", "customAccess", "showGlobalHints", "rightsSet", "isTwoFactorAuthenticationEnabled", "id", "v", "createdAt", "updatedAt", "createdBy", "updatedBy", "deletedAt", "newId", "dv", "sender" FROM "public"."User";
```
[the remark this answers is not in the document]
SavelevMatthew: adjusted default analytical plugin search for this cases

PR #7204, opened 2026-02-16
feat(condo): DOMA-12905 add mutation to send voip start message to all verified residents on address + unit

YEgorLu:
## Release Notes

**New Features**
- Added a `sendVoIPStartMessage` GraphQL mutation to initiate VoIP calls with support for both B2C app and native call handling
- Provides message delivery statistics including verified contact count, created message count, and error count
- Includes VoIP call status tracking functionality
- Added new GraphQL input/output types and unit type enums for VoIP messaging
- Added multi-language error messages for VoIP API operations


apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,447 @@
...
+
+const logInfo = ({ b2cAppId, callId, stats, errors }) => {
+    logger.info({ msg: `${SERVICE_NAME} stats`, entityName: 'B2CApp', entityId: b2cAppId, data: { callId, stats }, err: errors })
+}
+
+const SendVoIPStartMessageService = new GQLCustomSchema('SendVoIPStartMessageService', {
+    types: [
+        {
+            access: true,
+            type: 'enum VoIPType {' +
```
dkoviazin: not readable format for enum type

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,447 @@
...
+                'dtfmCommand: String!' +
+                '"""' +
+                'Name of a panel to be displayed' +
+                '"""' +
+                'name: String!' +
+                '}',
+        },
+        {
+            access: true,
+            type: 'input SendVoIPStartMessageData { ' +
```
dkoviazin: you can use `
to write multi line code
`

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,447 @@
...
+                    createMessageErrors: [],
+                    isStatusCached: false,
+                    isPropertyFound: false,
+                    isAppFound: false,
+                }
+
+                // 1) Check B2CApp and B2CAppProperty
+                const b2cAppId = app.id
+
+                const [b2cAppProperty] = await B2CAppProperty.getAll(context, {
```
dkoviazin: maybe itemsQuery ?

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,447 @@
...
+                    logInfo({ b2cAppId, callId: data.callId, stats: logInfoStats })
+                    throw new GQLError(ERRORS.APP_NOT_FOUND, context)
+                }
+                logInfoStats.isAppFound = true
+                
+                const b2cAppName = b2cAppProperty.app.name
+                
+                // 2) Find Users
+                const verifiedContactsOnUnit = await find('Contact', {
+                    property: { addressKey: addressKey, deletedAt: null },
```
dkoviazin: maybe, better to add user: { deletedAt: null } here

I know that later on you will filter this residents by not deleted users

apps/condo/domains/miniapp/utils/voip.js
```diff
@@ -0,0 +1,59 @@
...
+function validateCallId (callId) {
+    return typeof callId === 'string'
+        && callId.length > 0
+        && callId.length < 300
+}
+
+function normalizeCallId (callId) {
+    return callId
+        // eslint-disable-next-line no-control-regex
+        .replace(/[\u0000-\u001F\s]/g, '')
```
dkoviazin: better to move to constant and name it

dkoviazin (approved):

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: Don't understand what do you mean tbh

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: Will try.
Last time I checked generated docs had invalid spacings because of that

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: We kinda need to check accesses here, so with itemsQuery I would need to make different things depending on the user type: if admin, do not fetch access rights. So looks the same I guess

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: 'Contact' does not have "user" field, not directly related to "User".
But maybe in find('Resident', ...) it would be good

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
dkoviazin:
```
`
enum VoIPType {
  """Makes mobile app use it's call app instead of B2CApp's"""
  sip
}
`
```

apps/condo/domains/miniapp/access/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,19 @@
+const { throwAuthenticationError } = require('@open-condo/keystone/apolloErrorFormatter')
+
+const { SERVICE } = require('@condo/domains/user/constants/common')
+
+async function canSendVoIPStartMessage ({ args: { data }, authentication: { item: user } }) {
+    if (!user) return throwAuthenticationError()
+    if (user.deletedAt) return false
+    if (user.isAdmin) return true
+
+    if (user.type === SERVICE) {
```
SavelevMatthew, line 10: I guess its time to rebase?)

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,537 @@
...
+const CACHE_TTL = {
+    DEFAULT: DEFAULT_NOTIFICATION_WINDOW_DURATION_IN_SECONDS,
+    [VOIP_INCOMING_CALL_MESSAGE_TYPE]: 2,
+}
+
+/**
+ * If debug app is set and debug app settings are configured, then user can send push messages without creating B2CApp first.
+ * This is useful for testing and development, but it should be turned off on production
+ */
+const DEBUG_APP_ID = conf.MINIAPP_PUSH_MESSAGE_DEBUG_APP_ID
```
SavelevMatthew, line 40: should it be here? I thought it was some legacy or smth

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,537 @@
...
+const DEBUG_APP_ENABLED = !!DEBUG_APP_ID
+const DEBUG_APP_SETTINGS = DEBUG_APP_ENABLED ? Object.freeze(JSON.parse(conf.MINIAPP_PUSH_MESSAGE_DEBUG_APP_SETTINGS)) : {}
+const ALLOWED_UNIT_TYPES = [FLAT_UNIT_TYPE, APARTMENT_UNIT_TYPE]
```
SavelevMatthew, line 43: why COMMERCIAL_UNIT_TYPE is not allowed?

SavelevMatthew: why do you even want to restrict it?

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,537 @@
...
+const VOIP_TYPE_CUSTOM_FIELD_ID = conf.VOIP_TYPE_CUSTOM_FIELD_ID
```
SavelevMatthew, line 44: I thought we'll do something like this:
```
const customValues = await CustomValue.getAll(where: { field: { name_starts: "voip" } })
const overrides = { ... } // converts customValues to { voipType: "...", [key]: value }
```

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,537 @@
...
+
+const redisGuard = new RedisGuard()
+
+const logger = getLogger()
+
+const SERVICE_NAME = 'sendVoIPStartMessage'
+const ERRORS = {
+    PROPERTY_NOT_FOUND: {
+        query: SERVICE_NAME,
```
SavelevMatthew, line 53:
```suggestion
        mutation: SERVICE_NAME,
```

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,537 @@
...
+    WRONG_SENDER_FORMAT: {
+        ...COMMON_ERRORS.WRONG_SENDER_FORMAT,
+        mutation: SERVICE_NAME,
+    },
+}
+
+const B2CAPP_VOIP_TYPE = 'b2cApp'
+
+function logInfo ({ b2cAppId, callId, stats, errors }) {
+    logger.info({ msg: `${SERVICE_NAME} stats`, entityName: 'B2CApp', entityId: b2cAppId, data: { callId, stats }, err: errors })
```
SavelevMatthew, line 88: I thought err is a single field 🤔

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,537 @@
...
+}
+
+const SendVoIPStartMessageService = new GQLCustomSchema('SendVoIPStartMessageService', {
+    types: [
+        {
+            access: true,
+            type: `input VoIPPanel {
+                """
+                Dtfm command for panel
```
SavelevMatthew, line 97:
```suggestion
                Dtfm command used to open the panel
```

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,537 @@
...
+                """
+                dtfmCommand: String!
+                """
+                Name of a panel to be displayed
+                """
+                name: String!
```
SavelevMatthew, line 103: Should it be optional, I don't think we have it in all cases

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,537 @@
...
+
+function logInfo ({ b2cAppId, callId, stats, errors }) {
+    logger.info({ msg: `${SERVICE_NAME} stats`, entityName: 'B2CApp', entityId: b2cAppId, data: { callId, stats }, err: errors })
+}
+
+const SendVoIPStartMessageService = new GQLCustomSchema('SendVoIPStartMessageService', {
+    types: [
+        {
+            access: true,
+            type: `input VoIPPanel {
```
SavelevMatthew, line 95: It's better to keep naming a bit more detailed, in case we'll decided to add VoipPanel as model later.

Maybe something like:
```suggestion
            type: `input VoIPPanelParamters {
```

SavelevMatthew: why its not prefixed with service name like other inputs?

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,537 @@
...
+                """
+                callId: String!,
+                """
+                If you want your B2CApp to handle incoming VoIP call, provide this argument.
+                """
+                b2cAppCallData: SendVoIPStartMessageDataForCallHandlingByB2CApp,
+                """
+                If you want mobile app to handle call (without your B2CApp), provide this argument. If "b2cAppCallData" and "nativeCallData" are provided together, native call is prioritized. 
+                """
+                nativeCallData: SendVoIPStartMessageDataForCallHandlingByNative
```
SavelevMatthew, line 164: Hm, I'm not sure about this separation. Does it mean I cannot fill both or?

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,537 @@
...
+                        isAppFound: false,
+                    },
+                    b2cAppId: app.id,
+                    callId: callData.callId,
+                }
+
+                // 1) Check B2CApp and B2CAppProperty
+                const b2cAppId = app.id
+
+                const [b2cAppProperty] = await B2CAppProperty.getAll(context, {
```
SavelevMatthew, line 253: why not just getOne?

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,537 @@
...
+                    logInfo(logContext)
+                    return {
+                        verifiedContactsCount: verifiedContactsOnUnit.length,
+                        createdMessagesCount: 0,
+                        erroredMessagesCount: 0,
+                    }
+                }
+
+                // NOTE(YEgorLu): doing same as Resident.isVerifiedByManagingCompany virtual field
+                const usersOfContacts = await find('User', {
```
SavelevMatthew, line 321: I don't get it, why not just fetching residents on address?

Btw, fetching users by phone seems strange, you have ids from residents, why phones

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,537 @@
...
+
+                if (!DEBUG_APP_ENABLED || b2cAppId !== DEBUG_APP_ID) {
+                    appSettings = await getByCondition('AppMessageSetting', {
+                        b2cApp: { id: b2cAppId }, type: VOIP_INCOMING_CALL_MESSAGE_TYPE, deletedAt: null, 
+                    })
+                }
+
+                else {
+                    appSettings = { ...DEBUG_APP_SETTINGS }
+                }
```
SavelevMatthew, line 369: 1. why else is so separated
2. seems like duplication of legacy code, do we need it? maybe move it to util?

apps/condo/domains/miniapp/utils/voip.js
SavelevMatthew: yep

apps/condo/domains/miniapp/constants.js
```diff
@@ -131,6 +132,9 @@ const ACCESS_TOKEN_SESSION_ID_PREFIX = `${Buffer.from('b2bAccessToken').toString
 
 const ACCESS_TOKEN_UPDATE_MANY_CHUNK_SIZE = 500
 
+const MAGIC_VOIP_TYPE_CONSTANT_FOR_OLD_VERSIONS_COMPATIBILITY = 'sip' // without this constant mobile app will not try to make native call
```
SavelevMatthew, line 135: Yo, we dont name our constants like this...There's no magic in IT :)

apps/condo/domains/miniapp/constants.js
```diff
@@ -131,6 +132,9 @@ const ACCESS_TOKEN_SESSION_ID_PREFIX = `${Buffer.from('b2bAccessToken').toString
...
+const DEFAULT_VOIP_TYPE = 0
```
SavelevMatthew, line 136: what's this magic number?

apps/condo/domains/miniapp/access/SendVoIPStartMessageService.js
YEgorLu: Yeah

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: I don't exactly know, looks like it is being used for local development of miniapps

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: I remember we had problems with other systems as they would send voip push for (f.e.) parking lot, which is wrong. So we made a restriction for these unit types in them

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: In this part that you specified comment says that you can fill both.

We have different validation (required fields etc) for b2c data and for native call data. And there is 2 ways:
1) Lets allow everything as in sendB2CAppPushMessage, and throw a bunch of errors
2) Lets make fields required in schema, so caller can't miss

I guess 2) is better, thats why I separate them

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: I don't need ALL residents. I need only VERIFIED residents.
Verified means that I must find relations Resident <-> Contact
They relate only with user: Resident.user <-> User.id,  User.phone <-> Contact.phone, Resident.address <-> Contact.address.

It can be started from Contacts, it can be started from Residents, but anyway I must in the end relate one to other using phone number and User.

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: Yeah it Ctrl C + Ctrl V. 

It looks as it is needed for mobile app local development, tough I will recheck this

apps/condo/domains/miniapp/constants.js
YEgorLu: It's literally magic bs which does something unrelated to it's value

apps/condo/domains/miniapp/constants.js
YEgorLu: So we have voipType.

Right now supported values are: 'sip' | anything other than 'sip', which means 2 clients

Now we are going to add N clients. Each app can treat client ('voipType') differently as it chooses. So giving voipType a meaningful name looks useless, same value will be meaningful for one app and random string for the other.

So I decided that voipType must just be a number. As all data in pushes it will be stringified anyway, so it is string with a number in it. So the default is 0, and we treat it like an old one (being replaced to 'sip')

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: There is no uniqueness in the name, so it looks even more strange?

Like, voipType depends on it's value, which can be anything with zero arrangements

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: 🤔

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: Ok now I know that we don't have that restriction product-wise, just external errors related to it. So I will allow all unitTypes

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
SavelevMatthew: maybe its worth moving into util, so it can be reused across multiple mutations / services

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
SavelevMatthew: its better to clear it here

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
SavelevMatthew: addressKey / unitName / unitType

-> Residents(where: { addressKey, unitName, unitType}) -> const usersIds = residents.map(r => r.user)
-> Users(where: { id_in: [] }) -> ids (for push), + phones 
-> Find single property (Property util)
-> Contacts(where: { property: { id }, unitName, unitType, addressKey })

apps/condo/domains/miniapp/constants.js
SavelevMatthew: const NATIVE_VOIP_TYPE = 'sip'
const B2C_APP_VOIP_TYPE = 'b2cApp'

SavelevMatthew: CustomField = "voipPreferredProtol"

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: Lol I just understood that in this mutation, unlike sendB2CAppMessage, existence of b2c app and related items is required, so there is no way that non existent debug app would work. Removing it

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: Yeah, for some reason I named it as that and put "err" anyway

apps/condo/domains/miniapp/schema/B2CAppAccessRightSet.test.js
```diff
@@ -4,20 +4,22 @@
...
 
 
-const { B2CAppAccessRightSet, createTestB2CAppAccessRightSet, updateTestB2CAppAccessRightSet, createTestB2CAppProperty, updateTestB2CAppProperty, B2CAppProperty } = require('@condo/domains/miniapp/utils/testSchema')
+const { B2CAppAccessRightSet, createTestB2CAppAccessRightSet, updateTestB2CAppAccessRightSet, createTestB2CAppProperty, updateTestB2CAppProperty, B2CAppProperty, sendVoIPStartMessageByTestClient, updateTestB2CAppAccessRight } = require('@condo/domains/miniapp/utils/testSchema')
 const { createTestB2CApp, createTestB2CAppAccessRight } = require('@condo/domains/miniapp/utils/testSchema')
 const { buildFakeAddressAndMeta } = require('@condo/domains/property/utils/testSchema/factories')
 const { makeClientWithNewRegisteredAndLoggedInUser, makeClientWithSupportUser } = require('@condo/domains/user/utils/testSchema')
 const { makeClientWithServiceUser } = require('@condo/domains/user/utils/testSchema')
 
+const { FLAT_UNIT_TYPE } = require('../../property/constants/common')
```
SavelevMatthew:
```suggestion
const { FLAT_UNIT_TYPE } = require('@condo/domains/property/constants/common')
```

apps/condo/domains/miniapp/schema/SendVoIPCallStartMessageService.js
```diff
@@ -0,0 +1,573 @@
...
+        ...COMMON_ERRORS.WRONG_SENDER_FORMAT,
+        mutation: SERVICE_NAME,
+    },
+}
+
+function logInfo ({ b2cAppId, callId, stats, err }) {
+    logger.info({ msg: `${SERVICE_NAME} stats`, entityName: 'B2CApp', entityId: b2cAppId, data: { callId, stats }, err: err })
+}
+
+async function checkLimits ({ context, b2cAppId, logContext }) {
```
SavelevMatthew, line 94: should this be shared with sendB2CAppPushMessageService (moved into utils)?

apps/condo/domains/miniapp/schema/SendVoIPCallStartMessageService.js
```diff
@@ -0,0 +1,573 @@
...
+    const appSettings = await getByCondition('AppMessageSetting', {
+        b2cApp: { id: b2cAppId }, 
+        type: VOIP_INCOMING_CALL_MESSAGE_TYPE, 
+        deletedAt: null, 
+    })
+                
+    const searchKey = `${VOIP_INCOMING_CALL_MESSAGE_TYPE}-${b2cAppId}`
+    const ttl = CACHE_TTL[VOIP_INCOMING_CALL_MESSAGE_TYPE] || CACHE_TTL['DEFAULT']
```
SavelevMatthew, line 102: What's the current limitations? 1 call in 2 seconds? What if multiple panels calls at the same time?

Since, its not per user, rate limit it trickier this time

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,573 @@
...
+            isStatusCached: false,
+            isPropertyFound: false,
+            isAppFound: false,
+        },
+        b2cAppId: app.id,
+        callId: callData.callId,
+    }
+}
+
+async function sendMessageToActor ({
```
SavelevMatthew: We don't have word actor in our domains....we have user / contact / resident / address

So utils is confusing to read

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+    const needToPasteB2CAppCallData = !callData.nativeCallData || (callData.b2cAppCallData && customVoIPValues.voipType === B2C_APP_VOIP_TYPE) 
+
+    if (!needToPasteB2CAppCallData) {
+        let voipType = customVoIPValues.voipType || NATIVE_VOIP_TYPE
+        // CustomValue says to use B2C_APP_VOIP_TYPE, but we have no b2cAppCallData
+        if (voipType === B2C_APP_VOIP_TYPE) {
+            voipType = NATIVE_VOIP_TYPE
+        }
```
SavelevMatthew: what is going on right here....Why do you insert mess into new code?

SavelevMatthew: Why not just spread custom values

apps/condo/domains/miniapp/schema/SendVoIPCallStartMessageService.js
YEgorLu: I don't think so. In sendB2C it also checks for debug app, which can't be used here... Limit is being checked differently, it's like, no use I guess to move this

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: above in comment it literally says what is happening. 

Somebody configured CustomValue for this contact to use miniapp as call app. But in mutation they did not provide data for miniapp call. In that case we falling back to standard native call. Otherwise we will send empty voip push -> user will see ringing, but after answer nothing would happen

YEgorLu: And custom values are being spread just a few lines below

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: resident, contact and user needs to be grouped. Otherwise this file becomes pushAdapter v2. So I decided to name such group somehow

apps/condo/domains/miniapp/utils/voip.js
```diff
@@ -0,0 +1,59 @@
+const { getKVClient } = require('@open-condo/keystone/kv')
+
+const KEY_PREFIX = 'voipCallStatus'
+const kv = getKVClient(KEY_PREFIX)
+const CALL_STATUS_TTL_IN_SECONDS = 60 * 3
+
+const CALL_STATUS_START_SENT = 'CALL_STATUS_START_SENT'
```
SavelevMatthew:
```suggestion
const CALL_STATUS_STARTED = 'CALL_STATUS_STARTED'
```

apps/condo/domains/miniapp/utils/voip.js
```diff
@@ -0,0 +1,59 @@
...
+const CALL_STATUS_CANCEL_SENT = 'CALL_STATUS_CANCEL_SENT'
```
SavelevMatthew:
```suggestion
const CALL_STATUS_ANSWERED = 'CALL_STATUS_ANSWERED'
```

apps/condo/domains/miniapp/utils/voip.js
SavelevMatthew: (optional) think about readability in call_state context

apps/condo/domains/miniapp/utils/voip.js
```diff
@@ -0,0 +1,59 @@
...
+const CALL_STATUSES = {
+    CALL_STATUS_START_SENT: CALL_STATUS_START_SENT,
+    CALL_STATUS_CANCEL_SENT: CALL_STATUS_CANCEL_SENT,
+}
+
+function validateCallId (callId) {
+    return typeof callId === 'string'
+        && callId.length > 0
+        && callId.length < 300
```
SavelevMatthew:
```
/[a-z0-9-\._]{1, 300}/gi
```

apps/condo/domains/miniapp/utils/voip.js
```diff
@@ -0,0 +1,59 @@
+const { getKVClient } = require('@open-condo/keystone/kv')
+
+const KEY_PREFIX = 'voipCallStatus'
+const kv = getKVClient(KEY_PREFIX)
+const CALL_STATUS_TTL_IN_SECONDS = 60 * 3
```
SavelevMatthew:
```suggestion
const CALL_STATUS_TTL_IN_SECONDS = 60 * 3 // 3 minutes
```

apps/condo/domains/miniapp/utils/voip.js
```diff
@@ -0,0 +1,59 @@
...
+        'EX',
+        CALL_STATUS_TTL_IN_SECONDS,
+    )
+}
+
+async function getCallStatus ({ b2cAppId, callId }) {
+    if (!validateCallId(callId)) return null
+    const res = await kv.get(buildKey(b2cAppId, normalizeCallId(callId)))
+    try {
+        return JSON.parse(res)
```
SavelevMatthew: zod?

apps/condo/domains/miniapp/schema/SendVoIPCallStartMessageService.js
```diff
@@ -0,0 +1,572 @@
...
+    MESSAGE_META,
+    VOIP_INCOMING_CALL_MESSAGE_TYPE, MESSAGE_SENDING_STATUS,
+} = require('@condo/domains/notification/constants/constants')
+const { sendMessage } = require('@condo/domains/notification/utils/serverSchema')
+const { UNIT_TYPES } = require('@condo/domains/property/constants/common')
+const { getOldestNonDeletedProperty } = require('@condo/domains/property/utils/serverSchema/helpers')
+const { Resident } = require('@condo/domains/resident/utils/serverSchema') 
+const { RedisGuard } = require('@condo/domains/user/utils/serverSchema/guards')
+
+const CACHE_TTL = {
```
SavelevMatthew, line 33: btw, this should be shared with cancel mutation

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,572 @@
...
+        residentId: resident.id,
+        callId: callData.callId,
+    }
+
+    const needToPasteB2CAppCallData = !callData.nativeCallData || (callData.b2cAppCallData && customVoIPValues.voipType === B2C_APP_VOIP_TYPE) 
+
+    if (!needToPasteB2CAppCallData) {
+        let voipType = customVoIPValues.voipType || NATIVE_VOIP_TYPE
+        // CustomValue says to use B2C_APP_VOIP_TYPE, but we have no b2cAppCallData
+        if (voipType === B2C_APP_VOIP_TYPE) {
```
SavelevMatthew:
```suggestion
        if (customVoIPValues.voipType === B2C_APP_VOIP_TYPE && !b2cAppCallData) {
```

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,572 @@
...
+    if (!oldestProperty?.id) {
+        logInfo(logContext)
+        return { 
+            verifiedResidentsWithContacts,
+            earlyReturnValue: {
+                verifiedContactsCount: 0,
+                createdMessagesCount: 0,
+                erroredMessagesCount: 0,
+            },
+        }
```
SavelevMatthew: util?

isEarlyReturn()
finishProcessing() // +log?


```suggestion
        return finishProcessing({ ctx })
```

SavelevMatthew:
```
class RejectCallError extends Error {
     constructor(returnData) {}
}


resolver: {
   try {
   } catch (e) {
      if (e instance RejectCallError) {
         return e.returnData
      }
   }
}
```

apps/condo/domains/miniapp/schema/SendVoIPCallStartMessageService.js
```diff
@@ -0,0 +1,572 @@
...
+                verifiedContactsCount: 0,
+                createdMessagesCount: 0,
+                erroredMessagesCount: 0,
+            },
+        }
+    }
+
+    const allVerifiedContactsOnUnit = await find('Contact', {
+        property: { id: oldestProperty.id, deletedAt: null },
+        unitName_i: unitName,
```
SavelevMatthew, line 230: is `_i` required?

apps/condo/domains/miniapp/schema/SendVoIPCallStartMessageService.js
```diff
@@ -0,0 +1,572 @@
...
+                verifiedContactsCount: 0,
+                createdMessagesCount: 0,
+                erroredMessagesCount: 0,
+            },
+        }
+    }
+
+    const allResidentsOnUnit = await Resident.getAll(context, {
+        addressKey,
+        unitName,
```
SavelevMatthew: inconsistent `_i` behaviour?

apps/condo/domains/miniapp/schema/SendVoIPCallStartMessageService.js
```diff
@@ -0,0 +1,572 @@
...
+            },
+        }
+    }
+
+    const verifiedResidentsWithContactsByPhone = {}
+
+    for (const contact of allVerifiedContactsOnUnit) {
+        const verifiedResidentWithContact = { contact, resident: null, user: null }
+        verifiedResidentsWithContacts.push(verifiedResidentWithContact)
+        verifiedResidentsWithContactsByPhone[verifiedResidentWithContact.contact.phone] = verifiedResidentWithContact
```
SavelevMatthew, line 270: nullish?

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,572 @@
...
+        deletedAt: null,
+    }, 'id app { id name }')
+
+    if (!b2cAppProperty) {
+        logInfo(logContext)
+        throw new GQLError(ERRORS.PROPERTY_NOT_FOUND, context)
+    }
+    logContext.logInfoStats.isPropertyFound = true
+                
+    if (!b2cAppProperty.app) {
```
SavelevMatthew: is this possible?

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,572 @@
...
+        verifiedResidentsWithContacts,
+        earlyReturnValue: {
+            verifiedContactsCount: allVerifiedContactsOnUnit.length,
+            createdMessagesCount: 0,
+            erroredMessagesCount: 0,
+        },
+    }
+}
+
+async function verifyApp ({ context, logContext, addressKey, app }) {
```
SavelevMatthew: do you need this util if you have access

apps/condo/domains/miniapp/schema/SendVoIPCallStartMessageService.js
```diff
@@ -0,0 +1,572 @@
...
+        logInfo(logContext)
+        throw new GQLError(ERRORS.APP_NOT_FOUND, context)
+    }
+    logContext.logInfoStats.isAppFound = true
+
+    return { b2cAppId, b2cAppName: b2cAppProperty.app.name }
+}
+
+async function getCustomVoIPValuesByContacts ({ context, contactIds }) {
+    const customVoIPValues = await CustomValue.getAll(context, {
```
SavelevMatthew, line 321: filter on `Contact`

🆘 🆘 🆘 🆘 🆘

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
```diff
@@ -0,0 +1,572 @@
...
+        {
+            access: true,
+            type: `enum SendVoIPStartMessageUnitType { ${UNIT_TYPES.join('\n')} }`,
+        },
+    ],
+
+    mutations: [
+        {
+            access: access.canSendVoIPStartMessage,
+            schema: 'sendVoIPStartMessage(data: SendVoIPStartMessageInput!): SendVoIPStartMessageOutput',
```
SavelevMatthew:
```suggestion
            schema: 'sendVoIPCallStartMessage(data: SendVoIPStartMessageInput!): SendVoIPStartMessageOutput',
```

apps/condo/domains/miniapp/schema/SendVoIPCallStartMessageService.js
SavelevMatthew: 🆘 🆘 🆘 🆘 🆘

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
SavelevMatthew: 🆘

apps/condo/domains/miniapp/utils/voip.js
SavelevMatthew: 🆘

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: Originally I thought about checking access for B2CAppProperty in here with caller's auth.

Now we decided that there is no need, as "accesses" do check that.

I still need to get B2CAppName for push payload thou and maybe check what admin provided.

So option 1 - just get b2cappname. option 2 - do not change anything, so admin also has some type of validation.

I think I will keep it for admin

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: no, removed

apps/condo/domains/miniapp/schema/SendVoIPCallStartMessageService.js
YEgorLu: filtered

apps/condo/domains/miniapp/schema/SendVoIPCallStartMessageService.js
YEgorLu: for cancel mutation like half of this file will go into utils. So will do that all at once with cancel mutation

apps/condo/domains/miniapp/utils/voip.js
YEgorLu: I decided to add smth like this and remove "normalizeCallId". So we can validate "callId" at the start

apps/condo/domains/miniapp/utils/voip.js
YEgorLu: So we thought about giving "reason" to cancel pushes. And all possible reasons are basically: somebody answered or panel / devices canceled or did not answer in time. So we can have call as "answered" / "ended" / "started", why not keep all 3

apps/condo/domains/miniapp/schema/SendVoIPStartMessageService.js
YEgorLu: Went with error way

apps/condo/domains/miniapp/schema/SendVoIPCallStartMessageService.js
```diff
@@ -0,0 +1,621 @@
...
+    PROPERTY_NOT_FOUND: {
+        mutation: SERVICE_NAME,
+        variable: ['data', 'addressKey'],
+        code: BAD_USER_INPUT,
+        type: PROPERTY_NOT_FOUND_ERROR,
+        message: 'Unable to find Property or B2CAppProperty by provided addressKey',
+    },
+    CALL_DATA_NOT_PROVIDED: {
+        mutation: SERVICE_NAME,
+        variable: ['data, callData'],
```
SavelevMatthew, line 54:
```suggestion
        variable: ['data', 'callData'],
```

apps/condo/domains/miniapp/utils/voip.js
```diff
@@ -0,0 +1,107 @@
...
+    const length = typeof callMeta === 'string' ? callMeta.length : JSON.stringify(callMeta).length
+    return length <= MAX_CALL_META_LENGTH
+}
+
+function buildKey (b2cAppId, organizationId, propertyId, callId) {
+    return [KEY_PREFIX, b2cAppId, organizationId, propertyId, Buffer.from(callId).toString('base64')].join(':')
+}
+
+function generateCallStatusToken () {
+    return randomBytes(32).toString('base64')
```
SavelevMatthew, line 51: why not just random uuid?

SavelevMatthew (approved):
ok

apps/condo/domains/miniapp/utils/voip.js
YEgorLu: welll, maybe there is no difference betwen random bytes and uuid in this case. We build keys by 3 publicly known static uuids and one non-secure temporary random string. So we are adding one dynamic hidden (means secure) temporary string. I guess yah, it does need to be randomBytes

PR #7412, opened 2026-03-30
feat(condo): DOMA-13081 field type for store data externally

AleX83Xpert:
# Code Review Guide - ExternalContent Field

## 🎯 Start Here

This PR adds a new `ExternalContent` field type that stores large data in external files instead of the database. It also includes a generic backfill script to migrate any existing inline JSON field to file storage.

---

### 1. Security - Path Traversal Protection ⚠️ CRITICAL

**File:** `packages/keystone/fieldsUtils/ExternalContent/validateFilePath.js`

**What to check:**
- Empty string and non-string validation at the start of `validateFilePath()` (for both `basePath` and `filename`)
- Absolute path blocking with `path.isAbsolute(filename)` before any join
- `path.join()` + `path.normalize()` on both the full path and the base path
- Guard against `normalized === normalizedBase` (resolving to base dir itself)
- Final check: `normalized.startsWith(normalizedBase + path.sep)` - returns `normalized`, not `fullPath`

**Why it matters:** Prevents attackers from reading arbitrary files from the server.

---

### 2. Race Condition Fix - DataLoader ⚠️ CRITICAL

**File:** `packages/keystone/fieldsUtils/ExternalContent/FileContentLoader.js`

**What to check:**
- Constructor initializes three structures: `cache` (Map), `pending` (Map), and `queue` (Array)
- `load()` checks `cache` first, then `pending` map **before** creating a new promise
- New promise is stored in `pending` immediately after creation, keyed by `cacheKey`
- `promise.finally()` moves result from `pending` to `cache` and cleans up `pending`
- Batch scheduling uses `_schedulingBatch` flag to prevent a second timer from being set between the `!this.batchTimer` check and the timer assignment
- `batchDelayMs: 0` uses `setImmediate` instead of `setTimeout` for immediate execution

**Why it matters:** Without `pending` map, concurrent GraphQL resolver calls for the same file create duplicate queue entries and duplicate reads.

---

### 3. Backfill Script - Migration Safety ⚠️ CRITICAL

**File:** `apps/rb/bin/backfillExternalContentField.js`

**What to check:**
- `isFileMeta(fieldValue)` check skips records that are already migrated (idempotent)
- `null`/`undefined` raw values are skipped (falsy check is intentional - `false`/`0`/`''` are valid JSON payloads)
- Migration is delegated to `serverUtils[model].update()` which triggers `resolveInput` — the field's own save logic handles file creation and DB update atomically within Keystone's mutation pipeline
- `--dry-run` flag logs what would happen without writing files or updating DB
- `--continue-on-error` flag allows partial runs; failed record IDs are collected and printed at the end
- Offset-based pagination (`skip`) — safe because processed records change their shape (become file-meta), so they won't appear in a re-run

**Why it matters:** Prevents re-migrating already-migrated records and provides a safe recovery path on partial failures.

---

### 4. Dual GraphQL Fields - `field` vs `fieldResolved`

**File:** `packages/keystone/fields/ExternalContent/Implementation.js`

**What to check:**
- `gqlOutputFields()` returns two fields: `${path}: String` (raw file-meta JSON for Admin UI) and `${path}Resolved: <format type>` (deserialized content for API clients)
- `gqlOutputFieldResolvers()` only registers a resolver for `${path}Resolved` — the raw field is served directly from the DB column
- `resolveInput()` receives `context` and passes `context?.authedItem` as `user` to `adapter.publicUrl()` — needed for adapters that sign URLs per-user
- Returned object always includes `_type: FILE_META_TYPE` and `meta: { format }` so `resolveExternalContentValue` can deserialize without guessing the format

---

### 5. Admin UI - Field Component

**File:** `packages/keystone/fields/ExternalContent/views/Field.js`

**What to check:**
- `isReadOnly = field.isReadOnly !== false` — reads directly from the controller instance, not from `field.controller`
- `initialEditValue` stringify logic and `useEffect` both use `typeof resolvedContent === 'string' ? resolvedContent : JSON.stringify(resolvedContent, null, 2)` — consistent handling prevents `[object Object]` in textarea
- `useEffect` dependency is `[resolvedContent]` only — triggers when async resolved content arrives after initial render

---

### 6. Input Validation

**File:** `packages/keystone/fieldsUtils/ExternalContent/createExternalDataField.js`

**What to check:**
- `adapter` is required — throws immediately with clear message if missing
- `maxSizeBytes` validated as a positive finite number
- `batchDelayMs` validated as a non-negative finite number
- `isAdminUIReadOnly` mapped to `adminConfig.isReadOnly` and **not** leaked into the returned field object
- `type: 'ExternalContent'` is placed **after** `...otherProps` to prevent the caller from overriding it

**Why it matters:** Fails fast with clear errors if misconfigured.

## Summary by CodeRabbit

* **New Features**
  * ExternalContent field: file-backed storage with adapters, configurable formats (json/xml/text), size limits, batching/DataLoader, safe path validation, admin UI editor/cell/controller, GraphQL raw + resolved virtuals.

* **Documentation**
  * Detailed ExternalContent README covering usage, config, migration, security, and troubleshooting.

* **Tests**
  * Extensive tests for processors, loaders, validation, serialization, resolution, batching, and error handling.

* **Chores**
  * Submodule pointer updated; package exports and historical handling extended to include ExternalContent.


packages/keystone/fields/ExternalContent/views/Controller.js
```diff
@@ -0,0 +1,107 @@
...
+        if (typeof value === 'string') {
+            try {
+                // Try to parse as JSON (handles legacy JSON string metadata)
+                const parsed = JSON.parse(value)
+                // If it's an object with file metadata properties, return it
+                if (parsed && typeof parsed === 'object') {
+                    return parsed
+                }
+                // Otherwise return the parsed value as-is
+                return parsed
```
YEgorLu: ?

if (...) return parsed
else return parsed

packages/keystone/fields/ExternalContent/Implementation.js
```diff
@@ -0,0 +1,307 @@
...
+            throw new Error(`ExternalContent: "adapter" is required for ${path}`)
+        }
+
+        // Validate adapter is properly configured (not NoFileAdapter)
+        if (adapter.constructor.name === 'NoFileAdapter' || adapter.error?.message?.includes('NoFileAdapter')) {
+            throw new Error(`ExternalContent: adapter is not properly configured for ${path}. Check FILE_FIELD_ADAPTER and storage configuration.`)
+        }
+
+        // Pass resolved values to parent class
+        super(path, {
```
YEgorLu: "format" and "processors" does not change and are in "options", so a little misleading to paste them again, like they've changed somewhere and not just exactly what user pasted in field config

packages/keystone/fields/ExternalContent/Implementation.js
```diff
@@ -0,0 +1,307 @@
...
+            ...options,
+            format,
+            processors,
+            graphQLInputType: cfg.graphQLInputType,
+            graphQLReturnType: cfg.graphQLReturnType,
+            mimetype: cfg.mimetype,
+            fileExt: cfg.fileExt,
+        }, meta)
+
+        this.adapter = adapter
```
YEgorLu, line 72: everything what passed into "options" via super(..., options, ...) already exists in "this.config" object

packages/keystone/fields/ExternalContent/Implementation.js
```diff
@@ -0,0 +1,307 @@
...
+    /**
+     * Provides both raw file metadata and deserialized content via GraphQL.
+     * 
+     * - `fieldName` - Returns raw file metadata (JSON string from DB) for admin UI
+     * - `fieldNameResolved` - Virtual field that returns deserialized content for API clients
+     * 
+     * @returns {Object} Field resolver mapping
+     */
+    gqlOutputFieldResolvers () {
+        const { resolveExternalContentValue } = require('@open-condo/keystone/fieldsUtils/ExternalContent/resolveExternalContentValue')
```
YEgorLu: place require on top of file

packages/keystone/fields/ExternalContent/Implementation.js
```diff
@@ -0,0 +1,307 @@
...
+                    return await resolveExternalContentValue(value, {
+                        adapter: this.adapter,
+                        formatProcessors: this.formatProcessors,
+                        context,
+                        batchDelayMs: this.batchDelayMs,
+                    })
+                } catch (err) {
+                    const itemId = item?.id || 'unknown'
+                    logger.warn({ msg: 'Error resolving ExternalContent field', field: this.path, itemId, err })
+                    throw err
```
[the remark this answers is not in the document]
YEgorLu: Yeah, there is no "field" and "itemId" fields in logger

I guess you should be able to get listKey and then paste something like 

logger.warn({ ..., entity: listKey, entityId: itemId, data: { field: this.path } })

packages/keystone/fields/ExternalContent/views/Controller.js
AleX83Xpert: Hm, I've cut the statements but not edited the structure. Will fix it.

packages/keystone/fields/ExternalContent/Implementation.js
```diff
@@ -0,0 +1,307 @@
...
+     * 2. If setting new value: saves new file, then deletes old file
+     * 
+     * Save-before-delete pattern prevents data loss if save() fails.
+     * 
+     * Known limitation: If database update fails after save() succeeds but before transaction commits,
+     * the new file will be orphaned in storage while the old file reference remains in DB.
+     * This is an acceptable trade-off vs losing the previous file.
+     * 
+     * Potential mitigation strategies for future implementation:
+     * 1. Transaction hooks: Use Keystone afterChange hooks to delete orphaned files on transaction rollback
```
YEgorLu: I guess you really should do delete in afterChange

Right now you are saving new and deleting old file in "resolveInput". If saving is failed on database step as you say, then not just new file is orphaned, but old file is also deleted.

packages/keystone/fields/ExternalContent/Implementation.js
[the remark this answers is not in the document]
AleX83Xpert: Refactored

packages/keystone/fields/ExternalContent/Implementation.js
AleX83Xpert: done

packages/keystone/fields/ExternalContent/Implementation.js
AleX83Xpert: refactored

packages/keystone/fields/ExternalContent/Implementation.js
AleX83Xpert: refactored

packages/keystone/fields/ExternalContent/views/Controller.js
AleX83Xpert: Done

packages/keystone/fields/ExternalContent/Implementation.js
AleX83Xpert: Refactored

packages/keystone/fieldsUtils/ExternalContent/createExternalDataField.js
```diff
@@ -0,0 +1,62 @@
...
+        throw new Error('createExternalDataField: adapter is required')
+    }
+    
+    if (maxSizeBytes !== undefined && (typeof maxSizeBytes !== 'number' || maxSizeBytes <= 0)) {
+        throw new Error('createExternalDataField: maxSizeBytes must be a positive number')
+    }
+    
+    if (batchDelayMs !== undefined && (typeof batchDelayMs !== 'number' || batchDelayMs < 0)) {
+        throw new Error('createExternalDataField: batchDelay must be a non-negative number')
+    }
```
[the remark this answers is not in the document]
AleX83Xpert: fixed

packages/keystone/fields/ExternalContent/views/Controller.js
```diff
@@ -0,0 +1,107 @@
...
+    getProcessor = () => {
+        const allProcessors = { ...this.defaultProcessors, ...this.processors }
+        return allProcessors[this.format]
+    }
+
+    deserialize = data => {
+        const { path } = this
+        if (!data?.[path]) {
+            return null
+        }
```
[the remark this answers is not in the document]
AleX83Xpert: For ExternalContent, the field stores either a JSON string of file metadata or legacy content - an empty string '', 0, or false are not valid values for this field. The `!data?.[path]` guard is actually fine here in practice.

packages/keystone/fields/ExternalContent/views/Controller.js
```diff
@@ -0,0 +1,97 @@
+import FieldController from '@open-keystone/fields/Controller'
+
+const { DEFAULT_PROCESSORS } = require('@open-condo/keystone/fieldsUtils/ExternalContent/defaultProcessors')
```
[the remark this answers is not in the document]
AleX83Xpert: `@open-keystone/fields/Controller` is used in all the other field controllers in the same package (Json, Options, ...) - it's the standard pattern here, not a forbidden import. And `@open-condo/keystone/fieldsUtils/ExternalContent/defaultProcessors` is already the internal path.

packages/keystone/fields/ExternalContent/Implementation.js
```diff
@@ -0,0 +1,288 @@
...
+const { Readable } = require('node:stream')
+
+const { Implementation } = require('@open-keystone/fields')
+
+const { ExternalContent } = require('@open-condo/keystone/fieldsUtils')
+const logger = require('@open-condo/keystone/logging')
+
+const { FILE_META_TYPE, DEFAULT_PROCESSORS, isFileMeta, resolveExternalContentValue } = ExternalContent
+
+const DEFAULT_MAX_SIZE_BYTES = 10 * 1024 * 1024 // Default max size: 10MB
```
sitozzz, line 11: Why not 100mb as "max size" that could be uploaded to platform?

packages/keystone/fields/ExternalContent/Implementation.js
```diff
@@ -0,0 +1,288 @@
...
+     * @returns {Promise<Object|null|undefined>} File-meta object to store in DB
+     */
+    async resolveInput ({ resolvedData, existingItem, listKey, context }) {
+        const nextValue = resolvedData[this.path]
+
+        if (nextValue === undefined) return undefined
+
+        if (nextValue === null) {
+            return null
+        }
```
sitozzz: Is that original code of resolveInput hook? Looks a little strange)
```suggestion
        if (nextValue === undefined || nextValue === null) return nextValue
```

packages/keystone/fields/ExternalContent/Implementation.js
```diff
@@ -0,0 +1,288 @@
...
+
+        // Old file should be deleted if field was cleared or replaced with a new file
+        const prevFilename = prevValue.filename
+        const nextFilename = isFileMeta(nextValue) ? nextValue.filename : null
+        if (prevFilename === nextFilename) return
+
+        try {
+            await this.adapter.delete(prevValue)
+        } catch (err) {
+            logger.debug({ msg: 'Failed to delete old file after change', err, data: { filename: prevFilename } })
```
sitozzz: I think this should be not a debug log. At least info or warning

packages/keystone/fields/ExternalContent/Implementation.js
AleX83Xpert: The content of the file is loaded into memory. So, I set the default maxSize less than on the platform.
Anyway, this is a tunable option.

packages/keystone/fields/ExternalContent/Implementation.js
sitozzz, line 1: It'll be great to have a tool for migration from "old format" to new one

packages/keystone/fields/ExternalContent/Implementation.js
AleX83Xpert: Fixed

packages/keystone/fields/ExternalContent/Implementation.js
AleX83Xpert: Agree, now it is warn.

packages/keystone/fieldsUtils/ExternalContent/FileContentLoader.js
```diff
@@ -0,0 +1,281 @@
...
+            
+            try {
+                // Determine adapter type
+                const isLocalAdapter = typeof this.adapter?.src === 'string'
+                
+                if (isLocalAdapter) {
+                    await this._executeBatchLocal(batch)
+                } else {
+                    await this._executeBatchCloud(batch)
+                }
```
nomerdvadcatpyat, line 184: I think it will be better if we move this logic inside adapters.

Something like this:

```
async _executeBatch(batch) {
    const promises = batch.map(async ({ fileMeta, resolve, reject }) => {
        try {
            const buffer = await this.adapter.getBuffer(fileMeta)
            resolve(buffer)
        } catch (err) {
            reject(err)
        }
    })

    await Promise.allSettled(promises)
}
```

packages/keystone/fieldsUtils/ExternalContent/createExternalDataField.js
```diff
@@ -0,0 +1,64 @@
...
+        schemaDoc: `External content field storing ${format} data in external files`,
+        ...otherProps,
+        type: 'ExternalContent',
+        adapter,
+        format,
+        processors,
+        maxSizeBytes,
+        batchDelayMs,
+        adminConfig: {
+            isReadOnly: true,
```
nomerdvadcatpyat, line 56: It's not about `access: { read: true, create: userIsAdmin, update: userIsAdmin }`?

apps/rb
nomerdvadcatpyat, line 1: Why migrate script not in condo bin?

packages/keystone/fieldsUtils/ExternalContent/createExternalDataField.js
AleX83Xpert: No, it's about the ability to edit payload in the admin interface.

apps/rb
AleX83Xpert: Since the first place this new type is used is in RB's model, I placed it in the same app where the field type is used. The fallback script is server side script.

nomerdvadcatpyat: It looks like a common script for migrating a field to another type. Maybe we should add this to `packages/migrator`?

packages/keystone/fields/ExternalContent/Implementation.js
```diff
@@ -0,0 +1,288 @@
...
+
+    // Admin
+    extendAdminMeta (meta) {
+        return {
+            ...meta,
+            graphQLAdminFragment: this.config.graphQLAdminFragment || '',
+            format: this.config.format,
+            processors: this.formatProcessors,
+            adminConfig: this.adminConfig || {},
+        }
```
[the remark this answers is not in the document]
AleX83Xpert: Nope! See node_modules/@open-keystone/fields/dist/fields.cjs.dev.js:64

packages/keystone/fields/ExternalContent/README.md
```diff
@@ -0,0 +1,428 @@
...
+const MySchema = {
+    fields: {
+        myData: {
+            type: 'ExternalContent',
+            adapter: MyFieldFileAdapter,
+            format: 'json', // or 'xml', 'text'
+            maxSizeBytes: 10 * 1024 * 1024, // 10MB (optional, default: 10MB)
+        },
+    },
+}
```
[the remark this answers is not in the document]
AleX83Xpert: Will fix soon

packages/keystone/fields/ExternalContent/Implementation.js
AleX83Xpert: Yes, it exists. See apps/rb/bin/backfillExternalContentField.js.

packages/keystone/fieldsUtils/ExternalContent/FileContentLoader.js
AleX83Xpert: The read logic is too ExternalContent-specific (ENOENT/404, path validating, logging). In principle, it's possible (add getBuffer to all adapters), but the trade-off matters.

apps/rb
AleX83Xpert: The migrator is a published npm package (npx @open-condo/migrator) - it doesn't have @open-condo/keystone, prepareKeystoneExpressApp, or itemsQuery as dependencies. And looks like it shouldn't.

packages/keystone/fields/ExternalContent/Implementation.js
```diff
@@ -0,0 +1,284 @@
+const { randomUUID } = require('node:crypto')
+const { Readable } = require('node:stream')
+
+const { Implementation } = require('@open-keystone/fields')
+
+const { ExternalContent } = require('@open-condo/keystone/fieldsUtils')
+const logger = require('@open-condo/keystone/logging')
```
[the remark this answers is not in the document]
AleX83Xpert: Fixed

apps/rb
nomerdvadcatpyat: ok

packages/keystone/fieldsUtils/ExternalContent/FileContentLoader.js
nomerdvadcatpyat: For me ENOENT/404, path validating and logging it's adapter specific logic. Inside the adapter, we determine how to read the file and how to validate it, and give some common interface to the outside.

Maybe it's ok to leave it as is, but i think that `this.adapter?.src` check may be weak

nomerdvadcatpyat (approved):

packages/keystone/fieldsUtils/ExternalContent/FileContentLoader.js
AleX83Xpert: I've enhanced adapter type detection by verifying cloud adapter capabilities rather than local adapter properties.

dkoviazin (approved):
Please, remove submodule update from this PR

AleX83Xpert:
> Please, remove submodule update from this PR

Done

packages/keystone/fields/ExternalContent/Implementation.js
```diff
@@ -0,0 +1,299 @@
...
+            try { return JSON.parse(v) } catch { return null }
+        }
+
+        const prevValue = parseDbValue(existingItem?.[this.path])
+        const nextValue = parseDbValue(updatedItem?.[this.path])
+
+        // Update file metadata with actual item id for new files
+        // This allows OBSFilesMiddleware to verify access control
+        const prevFilename = isFileMeta(prevValue) ? prevValue.filename : null
+        const nextFilename = isFileMeta(nextValue) ? nextValue.filename : null
```
YEgorLu, line 268: ~~1) Above filename is being made with random ID every time. Does that mean that nextFilename === prefFilename situation impossible?~~

~~2) If filename changes, we do not delete old file. But if filename did not change, we delete old file.... I don't really get it. We delete same file after it's update or something?~~

It seems I was mistaken. For some reason I read all "if"s backwards

YEgorLu (approved):

PR #7517, opened 2026-04-23
fix(condo): fix random tests

nomerdvadcatpyat:
Fix several flaky tests in notification, marketplace, and webhooks domains
image

## RemoteClientPushToken › CRUD › create › admin can

`beforeAll` creates a push token for `globalAdminRemoteClient` (line 51) with a random
`provider`. The "admin can" test then tried to create a second token for the same
`remoteClient`, and with 1-in-6 probability the same `provider` was picked — hitting the
partial unique index `(remoteClient, provider, isVoIP=true)`.

Fix: reuse the already-created `globalAdminRemoteClientPushToken` to derive opposite
`isVoIP`/`isPush` values. The two partial unique indexes cover `isVoIP=true` and
`isPush=true` separately, so creating a token with the flipped flag is guaranteed not to
collide regardless of which provider is drawn.

Also fixed a latent bug in `getRandomPushTokenData`: the original
`isVoIP = extraAttrs.isVoIP || !extraAttrs.isPush || faker.datatype.boolean()` always
evaluated to `true` when called without arguments (`!undefined === true`), making every
generated token a VoIP token.

## SyncRemoteClientService › Push tokens cleaning › normal behaviour

The test called `getRandomPushTokenData()` twice without arguments. Due to the bug above
both tokens always had `isVoIP: true, isPush: false`. When both happened to draw the same
`transport` (~1-in-6 chance), `buildConflictResolutionJobs` treated the first token as a
conflict with the new winner on that provider and deleted it, breaking the assertion.

Fix: pass explicit different providers (`firebase` and `apple`) so the two tokens can
never conflict, regardless of the `isVoIP` value.

## Invoice › webhook signature can be verified after invoice payment

The test started a local **HTTPS** server for webhook delivery. In CI a real Bull worker
runs as a separate process alongside the test suite. That worker calls the webhook URL
with a real HTTP client — no Jest mock, no way to bypass the self-signed TLS certificate —
so the request fails with a certificate error. The test passed only when `delay.fn` ran
fast enough to acquire the Redis lock before the worker did.

Fix: switch the test server to **HTTP**. The worker can reach it without any certificate
issues, and the `jest.mock('@open-condo/keystone/fetch', …)` SSL-bypass block is no
longer needed.

## sendWebhookPayload › Must handle timeout errors correctly

`192.0.2.1` (RFC 5737 TEST-NET-1) is non-routable by design, but Linux behavior in Docker
is non-deterministic: if a default gateway exists the TCP SYN packet is forwarded and
silently dropped, triggering a real 10 s `TimeoutError`; if no route exists the kernel
returns `ENETUNREACH` immediately. The assertion `toContain('timeout')` failed in the
second case.

Fix: replace `toContain('timeout')` with `toMatch(/timeout|ENETUNREACH/i)`. Both errors
mean the host is unreachable — the important invariant (retry scheduled, no HTTP status
code) is still fully verified.

## Summary by CodeRabbit

* **Tests**
  * Improved webhook and push-notification test reliability: switched webhook tests to plain HTTP test servers (removed TLS bypass), rely on hooks rather than manual task triggers, use deterministic push-provider selection, and added a delayed local endpoint for timeout scenarios.
  * Made push-token generation deterministic for VoIP/App flags.

* **Chores**
  * Test setup and timeout handling hardened to accommodate varied network outcomes.

### DIFF
--- a/apps/condo/domains/marketplace/schema/invoice.spec.js
+++ b/apps/condo/domains/marketplace/schema/invoice.spec.js
@@ -32,25 +32,6 @@ const {
 } = require('@condo/domains/organization/utils/testSchema')
 
 
-// Mock fetch to allow self-signed certificates for HTTPS in tests
-jest.mock('@open-condo/keystone/fetch', () => {
-    const originalModule = jest.requireActual('@open-condo/keystone/fetch')
-    const https = require('https')
-
-    return {
-        ...originalModule,
-        fetch: jest.fn((url, options = {}) => {
-            // For HTTPS URLs, add agent with rejectUnauthorized: false
-            if (url.startsWith('https://')) {
-                //nosemgrep: problem-based-packs.insecure-transport.js-node.bypass-tls-verification.bypass-tls-verification
-                const agent = new https.Agent({ rejectUnauthorized: false })
-                return originalModule.fetch(url, { ...options, agent })
-            }
-
-            return originalModule.fetch(url, options)
-        }),
-    }
-})
 
 const { keystone } = index
 
@@ -95,7 +76,7 @@ describe('Invoice', () => {
                 })
                 res.status(200).json({ received: true })
             })
-            initTestExpressApp('InvoiceWebhookServer', webhookApp, 'https')
+            initTestExpressApp('InvoiceWebhookServer', webhookApp, 'http')
 
             test('webhook signature can be verified after invoice payment using real HTTP server', async () => {
                 // Clear previous webhook requests
--- a/apps/condo/domains/marketplace/schema/invoice.spec.js
+++ b/apps/condo/domains/marketplace/schema/invoice.spec.js
@@ -12,7 +12,6 @@ const {
     setFakeClientMode,
 } = require('@open-condo/keystone/test.utils')
 const { WebhookPayload } = require('@open-condo/webhooks/schema/utils/serverSchema')
-const { getWebhookRegularTasks } = require('@open-condo/webhooks/tasks')
 
 const { CONTEXT_FINISHED_STATUS } = require('@condo/domains/acquiring/constants/context')
 const { PAYMENT_DONE_STATUS } = require('@condo/domains/acquiring/constants/payment')
@@ -30,7 +29,6 @@ const {
 const {
     createTestOrganization,
 } = require('@condo/domains/organization/utils/testSchema')
-
 
 
 const { keystone } = index
@@ -125,9 +123,6 @@ describe('Invoice', () => {
                     })
                     expect(webhookPayload).toBeTruthy()
                 }, { timeout: 5000, interval: 100 })
-
-                const sendWebhookPayloadTask = getWebhookRegularTasks().sendWebhookPayload
-                await sendWebhookPayloadTask.delay.fn(webhookPayload.id)
 
                 // Wait for webhook to be sent to our test server (with longer timeout for async task processing)
                 await waitFor(async () => {
--- a/apps/condo/domains/notification/schema/RemoteClientPushToken.test.js
+++ b/apps/condo/domains/notification/schema/RemoteClientPushToken.test.js
@@ -15,6 +15,10 @@ const {
     expectToThrowUniqueConstraintViolationError,
 } = require('@open-condo/keystone/test.utils')
 
+const {
+    PUSH_TRANSPORT_FIREBASE,
+    PUSH_TRANSPORT_APPLE,
+} = require('@condo/domains/notification/constants/constants')
 const { RemoteClientPushToken, createTestRemoteClientPushToken, updateTestRemoteClientPushToken, createTestRemoteClient,
     updateTestRemoteClient,
 } = require('@condo/domains/notification/utils/testSchema')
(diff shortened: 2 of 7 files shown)


apps/condo/domains/notification/schema/RemoteClientPushToken.test.js
```diff
@@ -59,7 +59,10 @@ describe('RemoteClientPushToken', () => {
         describe('create', () => {
 
             test('admin can', async () => {
-                const [obj, attrs] = await createTestRemoteClientPushToken(admin, globalAdminRemoteClient)
+                const [obj, attrs] = await createTestRemoteClientPushToken(admin, globalAdminRemoteClient, {
+                    isVoIP: !globalAdminRemoteClientPushToken.isVoIP,
+                    isPush: globalAdminRemoteClientPushToken.isVoIP,
```
AleX83Xpert: `isVoIP: !isVoIP` ?

AleX83Xpert: What about the mirrored test?
```
isVoIP: globalAdminRemoteClientPushToken.isVoIP,
isPush: !globalAdminRemoteClientPushToken.isVoIP
```

apps/condo/domains/marketplace/schema/invoice.spec.js
```diff
@@ -32,25 +32,6 @@ const {
...
-            if (url.startsWith('https://')) {
-                //nosemgrep: problem-based-packs.insecure-transport.js-node.bypass-tls-verification.bypass-tls-verification
-                const agent = new https.Agent({ rejectUnauthorized: false })
-                return originalModule.fetch(url, { ...options, agent })
-            }
-
-            return originalModule.fetch(url, options)
-        }),
-    }
-})
```
AleX83Xpert, line 53: Why have you removed https?

packages/webhooks/tasks/sendWebhookPayload.spec.js
```diff
@@ -337,7 +338,7 @@ const SendWebhookPayloadTests = (appName, actorsInitializer) => {
             expect(updated).toHaveProperty('status', WEBHOOK_PAYLOAD_STATUS_PENDING)
             expect(updated).toHaveProperty('attempt', 1)
             expect(updated).toHaveProperty('lastErrorMessage')
-            expect(updated.lastErrorMessage).toContain('timeout')
+            expect(updated.lastErrorMessage).toMatch(/timeout|ENETUNREACH/i)
```
AleX83Xpert: `E NET UNREACH` is not the same as a timeout. But maybe it's ok for tests ¯\\\_(ツ)_/¯

apps/condo/domains/marketplace/schema/invoice.spec.js
nomerdvadcatpyat: Because this mock works only in test process, not in worker. In worker (task created in afterChange) requests were failed. I wrote about it it pr description.

You can also reproduce it locally with no TESTS_FAKE_WORKER_MODE env. With https test will fail randomly

packages/webhooks/tasks/sendWebhookPayload.spec.js
nomerdvadcatpyat: We can check only that error exists instead of check that it is timeout. What do you think?

apps/condo/domains/marketplace/schema/invoice.spec.js
AleX83Xpert: This test file uses fakeClientMode. Try to run the worker before the tests because the worker starts in CI.

apps/condo/domains/notification/schema/RemoteClientPushToken.test.js
nomerdvadcatpyat: It is needed to avoid RemoteClientPushToken constraints which randomly triggered:

```
            {
                type: 'models.UniqueConstraint',
                fields: ['provider', 'remoteClient'],
                condition: 'Q(deletedAt__isnull=True) & Q(isVoIP__exact = True)',
                name: 'remote_client_push_token_unique_remoteclient_provider_isvoip',
            },
            {
                type: 'models.UniqueConstraint',
                fields: ['provider', 'remoteClient'],
                condition: 'Q(deletedAt__isnull=True) & Q(isPush__exact = True)',
                name: 'remote_client_push_token_unique_remoteclient_provider_ispush',
            },
```


Changed it to passing different providers to make it more explicit

nomerdvadcatpyat: ^

apps/condo/domains/marketplace/schema/invoice.spec.js
AleX83Xpert: Looks like here we have the race between `await sendWebhookPayloadTask.delay.fn(webhookPayload.id)` in test and the worker's task (started after payment is done). If yes, it does not matter http or https here.

packages/webhooks/tasks/sendWebhookPayload.spec.js
AleX83Xpert: Hm, what if we add a real route in the test express app, but this route will answer after `WEBHOOK_PAYLOAD_TIMEOUT_IN_MS + 5000`ms delay?

apps/condo/domains/marketplace/schema/invoice.spec.js
nomerdvadcatpyat: > This test file uses fakeClientMode. Try to run the worker before the tests because the worker starts in CI.

Yes. `TESTS_FAKE_CLIENT_MODE` is true, but not `TESTS_FAKE_WORKER_MODE`.

Because of the worker is a separate process, the fetch mock does not work there and https requests to send a webhook fail with the self-signed certificate error. Due to `https`, requests in the worker will always fall.

The test with `https` only passed when `sendWebhookPayloadTask.delay.fn(webhookPayload.id)` manages to take the lock and send the webhook. Before sendWebhookPayload does this in the worker.

packages/webhooks/tasks/sendWebhookPayload.spec.js
nomerdvadcatpyat: Done

apps/condo/domains/notification/utils/testSchema/utils.js
YEgorLu, line 1: I don't get why you need this change

YEgorLu (approved):

apps/condo/domains/notification/utils/testSchema/utils.js
nomerdvadcatpyat: To explicitly set false variables

apps/condo/domains/marketplace/schema/invoice.spec.js
AleX83Xpert: So, for now, it does not matter where the hook will be called from: delay.fn() or the real worker process. It means we are able to remove the calling of delay.fn() from the test. What do you think?

nomerdvadcatpyat: Yes, we can. Removed

AleX83Xpert (approved):

PR #7553, opened 2026-05-03
fix(condo): DOMA-13219 disable next image cache for submodules

dkoviazin:


AleX83Xpert (approved):
Why did we decide to disable cache instead of trying to fix it?

abshnko (approved):

PR #7577, opened 2026-05-06
fix(condo): fix not updatable customValue if unique per object

YEgorLu:
## Summary by CodeRabbit

* **New Features**
  * Added unique-per-object constraint support for custom fields, enabling administrators to configure fields that enforce a single value per parent object with comprehensive validation.

* **Bug Fixes**
  * Fixed custom field uniqueness validation to correctly handle updates to existing custom field values when marked with unique-per-object constraints, while preventing false duplicate errors.


SavelevMatthew (approved):
good catch

abshnko (approved):

PR #7584, opened 2026-05-07
fix(condo): DOMA-13235,DOMA-13199 fixed a news files view. Fixed files sending to telegram news

Alllex202:
- [x] Hide files for sharing step if files is empty
- [x] Use feature flag for files on news item card page 
- [x] Fix files sending to telegram news

## Summary by CodeRabbit

* **New Features**
  * Added a feature flag to enable/disable mini-app integration for news item files and to control file-preview behavior.

* **Bug Fixes**
  * Files field now respects feature flags and won’t display during the sharing step if files are missing/invalid or the mini-app flag is off.
  * Preview updates now conditionally include prepared files based on the flag.

* **Chores**
  * Updated related subproject references.


AleX83Xpert (approved):
What about deleting messages from the technical proxy-channel?

nomerdvadcatpyat (approved):

Alllex202:
> What about deleting messages from the technical proxy-channel?

That's a good idea, but we don't know what might happen to the file. 
We're concerned that after deleting a file (message) from the proxy channel, the file in the target channel might also be deleted after some time.
We'll consider your suggestion.

PR #7642, opened 2026-05-20
fix(condo): DOMA-12256 fix invoice update error when only invoice status updated and ticket fields not matches with invoice

nomerdvadcatpyat:
Fixed two bugs with invoices linked to tickets:

- Cancelling or paying a published invoice was failing with `CLIENT_DATA_DOES_NOT_MATCH_TICKET` when invoice and ticket had mismatched client data — different unit name casing, different addresses, or different contact records.

- When a dispatcher updated ticket fields, the automatic sync to the linked published invoice was silently failing with the same error.

**Root cause:** getUnitTypeFieldResolveInput was always returning a value, which caused Keystone to inject unitType into resolvedData on every invoice update — even status-only changes like cancel or paid. This made the validation think client data was being changed and triggered a full comparison against the ticket.

Fixed by returning undefined from getUnitTypeFieldResolveInput (and getSectionTypeFieldResolveInput) when neither field is present in the update payload. Also narrowed the client data validation in Invoice.validateInput to only compare fields that are actually being changed in the current request, which fixes the ticket sync failing due to unrelated field mismatches.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invoice validation for ticket-linked invoices now only compares the subset of client fields that changed.
  * Resolved edge-case in server-side data resolution: functions now return undefined early when expected type/name keys are missing, preventing incorrect merges.

* **Tests**
  * Added tests covering invoice status updates for ticket-linked scenarios where invoice client data differs from the connected ticket.

[Review Change Stack](https://app.coderabbit.ai/change-stack/open-condo-software/condo/pull/7642?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/condo/domains/common/utils/serverSchema/resolveHelpers.js
+++ b/apps/condo/domains/common/utils/serverSchema/resolveHelpers.js
@@ -1,9 +1,13 @@
-const { get } = require('lodash')
+const get = require('lodash/get')
 
 const { FLAT_UNIT_TYPE, SECTION_SECTION_TYPE } = require('@condo/domains/property/constants/common')
 
 
 const getUnitTypeFieldResolveInput = ({ unitTypeFieldName = 'unitType', unitNameFieldName = 'unitName' } = {}) => ({ resolvedData, existingItem }) => {
+    if (!(unitTypeFieldName in resolvedData) && !(unitNameFieldName in resolvedData)) {
+        return undefined
+    }
+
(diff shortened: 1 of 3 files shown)


sitozzz (approved):
looks okay

dkoviazin (approved):

AleX83Xpert (approved):

abshnko (approved):

PR #7667, opened 2026-05-25
feat(registry-importer): DOMA-13293 added analytics in registry app

nomerdvadcatpyat:
## Summary by CodeRabbit

* **Chores**
  * Updated the referenced Helm chart to a newer revision.
  * Updated the registry-importer component reference to a newer revision.
  * No other code or configuration changes are included in this update; behavior and public interfaces remain unchanged.

[Review Change Stack](https://app.coderabbit.ai/change-stack/open-condo-software/condo/pull/7667?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1123,6 +1123,7 @@ __metadata:
     nanoid: ^3.3.4
     next: ^15.5.7
     next-with-less: ^3.0.1
+    nunjucks: ^3.2.4
     react: ^18.3.1
     react-dom: ^18.3.1
     react-router-dom: 5.2.0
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -1650,7 +1650,7 @@ __metadata:
     "@open-condo/miniapp-utils": "workspace:^"
     "@open-condo/next": "workspace:^"
     "@open-condo/ui": "workspace:^"
-    "@open-keystone/app-graphql": ^7.0.6
```
abshnko, line 1654: downgrading here, maybe rebase

.helm
abshnko, line 1: i'm not sure about this but there might be some problems if we don't pass _default value to helm values. i might be wrong but i think it's better to specify _default

abshnko (approved):
👍🏻

.helm
nomerdvadcatpyat: In other apps there is the same configs. Maybe problems will be if we specify empty string value, but no _default value it's ok

Alexander-Turkin (approved):

PR #7711, opened 2026-06-08
fix: INFRA-1678 return mimetype to field resolver

sitozzz:
## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened upload and sharing validation by requiring and verifying file MIME types in signatures, reducing risk from incorrect or maliciously labeled files.
  * Input handling now uses the verified MIME type instead of trusting user-supplied values.

* **Tests**
  * Expanded tests to assert that decrypted upload/share signatures include the expected MIME type.


SavelevMatthew (approved):
ok

PR #7765, opened 2026-06-19
refactor(condo): DOMA-13373 added skipAccessControl to uploadFileFromServer and attachFileFromServer

nomerdvadcatpyat:
Migrated server-side export file uploads from the legacy graphql-upload multipart flow to the new signature-based flow.

**Problem:** Worker tasks had no way to use the new file upload flow without making authenticated HTTP requests — workers have no session, so calling `/api/files/upload` or `/api/files/attach` would fail with 401.

**Solution:** Added `skipAccessControl` option to `uploadFilesFromServer` and `attachFileFromServer` in `@open-condo/files/server`. When set, the functions bypass HTTP entirely and call Keystone directly — the same guarantee as `keystone.createContext({ skipAccessControl: true })`, which only works within the same process.

Used this api in `export.js`, which used in many export tasks.
Also updated `UserHelpRequest` requests with file from frontend.

## Release Notes

* **New Features**
  * Added onboarding file upload support for user help requests (client-side upload when configured/authenticated).

* **Improvements**
  * Upgraded export and document generation uploads to use server-side file handling with accurate payload sizing.
  * Improved meter import error-file generation by uploading failed-data attachments via the unified upload utility.
  * Added a faster “no-HTTP” upload/attach mode for file middleware and custom file field handling.

* **Tests**
  * Expanded coverage for upload/attach behaviors, including skip-access-control scenarios.


apps/condo/domains/common/utils/serverSchema/export.js
```diff
@@ -180,9 +225,13 @@ const exportRecordsAsCsvFile = async ({ context, loadRecordsBatch, convertRecord
     writeStream.close()
 
     const stream = fs.createReadStream(filename, { encoding: 'utf8' })
-    const file = buildUploadInputFrom({
-        stream, filename: `export_${dayjs().format('DD_MM')}.csv`, mimetype: 'text/csv', encoding: 'utf8',
-        meta: { listkey: taskServerUtils.gql.SINGULAR_FORM, id: taskId },
+    const { size } = fs.statSync(filename)
+    const file = await uploadExportFile({
+        context,
+        file: { stream, filename: `export_${dayjs().format('DD_MM')}.csv`, mimetype: 'text/csv', encoding: 'utf8', size },
```
sitozzz, line 197: Mimetype now detects on server side. U don't need to set it manually

packages/files/utils.js
```diff
@@ -332,127 +332,153 @@ function parserHandler ({ processRequestOptions } = {}) {
...
+                fileSize: String(files[index].size),
+                fileMimeType: files[index].mimetype,
+            },
+        })),
+        `id fileMeta ${FILE_RECORD_META_FIELDS}`
+    )
+    const fileRecords = await FileRecord.updateMany(context,
+        createdFiles.map(e => ({
+            id: e.id, data: { fileMeta: { ...e.fileMeta, recordId: e.id }, dv: meta.dv, sender: meta.sender },
+        })), `id fileMeta { originalFilename meta ${FILE_RECORD_USER_META} }`)
```
sitozzz, line 406: I don't get it. What do u try to do here?

sitozzz: Oh okay, it's strange diff rendering. I check this one more time

sitozzz (commented):
It would be great to see tests alongside with existing ones, that checks "server side upload" without model context

packages/files/utils.js
nomerdvadcatpyat: Yea, i'm just move this logic in other function

apps/condo/domains/common/schema/file.spec.js
sitozzz, line 1: why do u change it to spec?

nomerdvadcatpyat: Because i need `setFakeClientMode` for server-side file upload tests with skipAccessControl. 
`setFakeClientMode` sets only in `spec.js` files

sitozzz (approved):
looks pretty good

abshnko (approved):
good job 👍🏻

PR #7784, opened 2026-06-24
fix(condo): DOMA-13378 fix file list in ticket comments

nomerdvadcatpyat:
**Problem:** Embedding the uploader into the Input.TextArea toolbar squeezed the file list with no scroll, so it overflowed onto the input field.

https://github.com/user-attachments/assets/8ef28930-d4f2-4c4f-bd80-acd6fd99ea39

**Solution:** Moved the list into its own block below the input with its own scroll, keeping the paperclip button in the toolbar (it triggers the uploader's hidden input via a ref).

https://github.com/user-attachments/assets/e422185b-a648-4b42-93b0-549e974cf8d9

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved comment attachments so the upload control opens reliably from the toolbar.
  * Updated the comment layout to better handle attached files without affecting the text area spacing.
  * Added a dedicated file list area with scrolling and cleaner overflow handling for uploaded items.

### DIFF
--- a/apps/condo/domains/common/components/Comments/CommentForm.tsx
+++ b/apps/condo/domains/common/components/Comments/CommentForm.tsx
@@ -1,7 +1,7 @@
 import { Form, FormInstance, InputRef } from 'antd'
 import classNames from 'classnames'
 import cookie from 'js-cookie'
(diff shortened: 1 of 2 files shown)


sitozzz (approved):
lgtm

PR #7815, opened 2026-07-02
fix(condo): reduce Message.processingMeta size

YEgorLu:
So there is some duplicates in Message.processingMeta json. Here I am removing them:

1) { messageContext } is useless, as it is duplicate of latest { transportsMeta: [{ messageContext }] }
2) push adapter pasted whole current Message in messageContext, which at least duplicates big { meta: { data } } (useless)

So this should reduce Message.processingMeta's size for something like 50% based on what I tested -> reduce size of Message / MessageHistoryRecord

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved notification delivery handling so sensitive message context is no longer included in processing metadata.
  * Updated push notification handling to keep internal message details out of stored delivery data while preserving delivery behavior.


sitozzz (commented):
Looks like PR title does not match the changes

YEgorLu:
> Looks like PR title does not match the changes

It does, as here repeated cyclic info about Message is being removed from "processingMeta" field

PR #7830, opened 2026-07-08
refactor(condo): DOMA-13429 reduce ticket pages complexity

nomerdvadcatpyat:
**Deduplicate count queries on ticket list** — `TicketStatusFilterContainer` and
`TicketTypeFilterSwitch` were each firing their own `getTicketsCount` queries.
They now receive counts as props from the parent instead.

**Delta sync polling** — instead of unconditional full refetch every N seconds,
each tick first fires a cheap `getLatestTicketUpdatedAt` query (fetches 1 record).
Full refetch only happens if something actually changed. Cursor is persisted in
`localStorage` so page reload doesn't trigger an unnecessary refetch.

**Polling extracted to `useTicketListPolling` hook** — reusable, keeps the component clean.

**Removed background counter refetch** — ticket type/status counters no longer
refetch in the polling loop; they update naturally on user actions.

**Skip status counters on complex filters** — `getTicketsCountersByStatus` fires
6 `_allTicketsMeta` queries in parallel. When the current filter's estimated
complexity exceeds 1000 units, these queries are skipped entirely and the status
bar is hidden. Controlled by feature flag `ticket-status-counters-limit`.

**Skip queries in ticket form** — all 5 `useAllObjects` calls in `TicketAssignments`
are skipped until a property is selected (fields are disabled anyway).
Saves 5 queries on every ticket create form open.

**Feature flag** `refetch-tickets-interval-in-seconds` — polling interval is now
a number (seconds) instead of boolean, configurable per environment. Default 60s.

---

### Complexity savings

Each `_allTicketsMeta` costs `whereFactor × 10` units. With a typical filter
`whereFactor` of 4–8, removing 2 duplicate count queries per tick saves
**~160 complexity/tick**.

Delta sync replaces a full `getTickets` + `_allTicketsMeta` (~96 units) with a
cheap single-record check (~9 units) on ticks where nothing changed. Assuming
90% of ticks are idle, that's **~87% reduction in polling cost**.

At a 60s interval over 1 hour: polling complexity drops from ~15 000 to ~1 500.

`getTicketsCountersByStatus` is 6 `_allTicketsMeta` queries per render. On a
complex filter (`whereFactor` ~8) each costs 80 units — **480 units saved** when
the complexity guard kicks in.

Skipping 5 `useAllObjects` calls in the ticket form saves ~30 units per form open.

## Summary by CodeRabbit

* **New Features**
  * Ticket lists now refresh using a polling hook that tracks the latest `updatedAt` and resumes immediately when the tab becomes visible.
  * Favorites retrieval now supports configurable pagination and uses a dedicated count query for more accurate totals.
  * Added an optional cap on ticket status counters by estimating filter complexity; counters may hide when limits are exceeded.

* **Bug Fixes**
  * Improved loading/skipping behavior for scoped ticket assignment and employee data.
  * Ticket status counter UI no longer shows the counter tag when the count is missing.

* **Chores**
  * Updated refetch interval behavior to be feature-flag configurable.

### DIFF
--- a/apps/condo/domains/ticket/utils/complexity.ts
+++ b/apps/condo/domains/ticket/utils/complexity.ts
@@ -3,7 +3,8 @@
  * packages/keystone/apolloServerPlugins/rateLimiting/query.utils.js
  *
  * Used to estimate _allTicketsMeta query complexity before sending it,
- * so we can skip getTicketsCountersByStatus when it would be too expensive.
+ * so we can hide counts in getTicketsCountersByStatus when it would be too expensive.
+ * The limit is configured via the ticket-status-counters-limit feature flag (number).
  */
 
 const WHERE_SCALING_FACTOR = 2
--- a/apps/condo/domains/ticket/utils/complexity.ts
+++ b/apps/condo/domains/ticket/utils/complexity.ts
@@ -1,0 +1,39 @@
+/**
+ * Frontend mirror of extractWhereComplexityFactor from
+ * packages/keystone/apolloServerPlugins/rateLimiting/query.utils.js
+ *
+ * Used to estimate _allTicketsMeta query complexity before sending it,
+ * so we can skip getTicketsCountersByStatus when it would be too expensive.
+ */
+
+const WHERE_SCALING_FACTOR = 2
+const WHERE_PAGE_LIMIT = 100
+const WHERE_MAX_TOTAL_RESULTS = 1000
+const WHERE_QUERY_WEIGHT = 2
+// getTicketsCountersByStatus fires one _allTicketsMeta per status type
+const STATUS_QUERY_COUNT = 6
+
+function extractWhereComplexityFactor (where: Record<string, unknown>): number {
+    const fieldsComplexity = Object.entries(where).reduce<number>((acc, [fieldName, fieldValue]) => {
+        if (Array.isArray(fieldValue)) {
+            if (fieldName === 'AND' || fieldName === 'OR') {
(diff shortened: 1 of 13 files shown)


apps/condo/domains/ticket/utils/complexity.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Frontend mirror of extractWhereComplexityFactor from
+ * packages/keystone/apolloServerPlugins/rateLimiting/query.utils.js
+ *
+ * Used to estimate _allTicketsMeta query complexity before sending it,
+ * so we can skip getTicketsCountersByStatus when it would be too expensive.
```
toplenboren: Why should we skip it tho?

toplenboren (approved):
Do you want to just skip refetching when creating new ticket? And refetch only on interval?

apps/condo/domains/ticket/utils/complexity.ts
nomerdvadcatpyat: getTicketsCountersByStatus fires 6 _allTicketsMeta queries in parallel. With a heavy filter (nested relations, large id_in arrays, OR conditions) each can cost 80–200 complexity units — that's potentially 960+ units in a single render, which hits the rate limit

nomerdvadcatpyat:
> Do you want to just skip refetching when creating new ticket? And refetch only on interval?

No — cache eviction after ticket creation already triggers an immediate refetch when navigating back to /ticket. The polling interval handles background updates from other users, not post-create refresh

PR #7861, opened 2026-07-17
feat(eps): DOMA-13456 remove urls from req

vovaaxeapolla:
## Summary by CodeRabbit

* **Chores**
  * Updated the embedded EPS component to the latest available revision.


dkoviazin (approved):

YEgorLu (commented):
Ok but:

Better to put kv.get / kv.set requests close to each other, f.e. in some util.
Add a comment about actual seconds/minutes/hours 60*60 is, so reader will not calculate it himself every time

AleX83Xpert (approved):

PR #7880, opened 2026-07-24
fix(condo): DOMA-11344 qr-payment - determine needed receipt by recipient account number also

YEgorLu:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QR-code validation for properties with multiple billing accounts by deriving billing and acquiring contexts from all applicable accounts.
  * Correctly selects the last receipt using recipient account details and matching identifiers.
  * Uses the billing context to compute the correct billing period during comparison.
  * Provides clear errors when billing or acquiring context cannot be determined.
  * Updated and expanded automated test coverage for multi-account and receipt-selection scenarios.


apps/condo/domains/billing/schema/ValidateQRCodeService.js
dkoviazin, line 160: Better to use BillingRecipient first, as organization can have multiple BillingRecipients with different tins, also check if recipient is approved in case it uses different tin

dkoviazin, line 157: use bankAccount as a first filter

apps/condo/domains/billing/schema/ValidateQRCodeService.js
```diff
@@ -170,38 +171,38 @@ const ValidateQRCodeService = new GQLCustomSchema('ValidateQRCodeService', {
...
                     integration: { type: ACQUIRING_INTEGRATION_ONLINE_PROCESSING_TYPE, deletedAt: null },
                     deletedAt: null,
                 })
 
-                if (!acquiringContext) throw new GQLError(ERRORS.NO_ACQUIRING_CONTEXT, context)
+                if (acquiringContexts.length === 0) throw new GQLError(ERRORS.NO_ACQUIRING_CONTEXT, context)
 
-                qrCodeFields['PaymPeriod'] = getQRCodePaymPeriod(qrCodeFields, billingContext)
 
                 // NOTE (YEgorLu): We just take tin, bic, bankAccount from QR-code
```
dkoviazin, line 201: 3 cases:
1. we do not have any receipt and no PaymPeriod information
2. we have unpaid receipt - we need to set the same period 
3. we have paid receipt - we need to set next period

dkoviazin (approved):

vovaaxeapolla (approved):
lgtm

PR #7888, opened 2026-07-29
fix: INFRA-1724 add cc and bcc to sendsay email adapter

sitozzz:
## Summary by CodeRabbit

* **New Features**
  * Added support for **CC** recipients in Sendsay email sends by delivering CC alongside primary recipients.
  * Added support for **BCC** recipients by sending them separately, returning per-recipient BCC results in the operation context and reporting partial delivery when applicable.
  * Email delivery is now sequential and avoids duplicate sends across **to/cc/bcc**; sending stops after the first primary failure.
* **Tests**
  * Expanded adapter tests to verify the new request flow, payload structure, deduplication, failure/partial-delivery behavior, and correct BCC result collection.

### DIFF
--- a/packages/keystone/emailAdapter.js
+++ b/packages/keystone/emailAdapter.js
@@ -4,6 +4,7 @@ const https = require('https')
 const FormData = require('form-data')
 const get = require('lodash/get')
 const isEmpty = require('lodash/isEmpty')
+const omit = require('lodash/omit')
 const uniq = require('lodash/uniq')
 const { z } = require('zod')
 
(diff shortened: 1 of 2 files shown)


packages/keystone/emailAdapter.js
```diff
@@ -703,7 +704,7 @@ class SendsayEmail {
                     group: 'personal',
                     sendwhen: 'now',
                     letter,
-                    'users.list': toEmails.join('\n'),
+                    'users.list': [...toEmails, ...ccEmails, ...bccEmails].join('\n'),
```
dkoviazin: you need to create a separated letter for bcc addresses

dkoviazin (approved):

packages/keystone/emailAdapter.js
sitozzz: yep, good catch here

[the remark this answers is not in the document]
sitozzz: yep, fixed that

sitozzz:
@coderabbitai review

PR #7944, opened 2026-08-20
fix(condo): DOMA-12178 fix multiple ticket creation from ticket form

nomerdvadcatpyat:
## What

Fixes duplicate ticket creation when the "Create ticket" button is clicked
multiple times rapidly, especially noticeable on slow/throttled connections.

## Why

Two independent race conditions allowed a second click to fire a second
`createTicket` mutation:

1. `FormWithAction.handleSave` had no re-entrancy guard — a click during the
   (async) field-validation phase, before `isLoading` became `true`, could
   start a second `form.submit()` in parallel.
2. In `CreateTicketForm`, the submit button was re-enabled as soon as the
   `createTicket` mutation resolved, without waiting for the subsequent
   `router.push` navigation away from the page to finish. On a slow
   connection that navigation can take seconds, during which the form stayed
   mounted and clickable, letting a second click create a second ticket.

## Changes

- `FormList.tsx`: add a synchronous `useRef` guard in `handleSave` to block
  overlapping submits.
- `TicketSubmitButton.tsx`: `disabled` now also reflects `isLoading`.
- `CreateTicketForm.tsx` (condo + callcenter): keep the submit button
  locked after a successful creation until the page actually navigates
  away; only unlock on error, so the user can retry.

## Test plan

- Throttle network to Slow 3G, fill required fields, rapidly click "Create
  ticket" — only one ticket/notification is created.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented duplicate form submissions while requests are in progress.
  - Submit buttons are now disabled during ticket submission.
  - Submission state resets correctly after validation failures or interrupted requests.
  - Improved error handling during ticket creation, ensuring failures are surfaced and forms can be submitted again.
  - Improved consistency of submission behavior across supported forms and ticket workflows.

### DIFF
--- a/apps/condo/domains/common/components/containers/FormList.tsx
+++ b/apps/condo/domains/common/components/containers/FormList.tsx
@@ -321,13 +321,9 @@ const FormWithAction: React.FC<IFormWithAction> = (props) => {
                 isSubmittingRef.current = false
                 throw err
             }
-            if (result && typeof result.finally === 'function') {
-                return result.finally(() => {
-                    isSubmittingRef.current = false
-                })
-            }
-            isSubmittingRef.current = false
-            return result
+            return Promise.resolve(result).finally(() => {
+                isSubmittingRef.current = false
+            })
         }
         if (values.hasOwnProperty(NON_FIELD_ERROR_NAME)) delete values[NON_FIELD_ERROR_NAME]
         let data
@@ -422,8 +418,6 @@ const FormWithAction: React.FC<IFormWithAction> = (props) => {
         <Form
             form={form}
             layout={layout}
-            onFinish={_handleSubmit}
-            onFinishFailed={handleSubmitFailed}
             initialValues={initialValues}
             validateTrigger={validateTrigger}
             onValuesChange={handleChange}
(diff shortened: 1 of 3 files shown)


sitozzz (commented):
why not just use debounce on action call here?

nomerdvadcatpyat:
> why not just use debounce on action call here?

Debounce only blocks clicks within its own short window, not for as long as the request actually takes - so a click seconds later, while the first request is still pending, still gets through.
If you set the debounce window large enough to safely cover slow 3G (several seconds), every user, including ones on fast connections, gets locked out of retrying for that full duration even after a quick, already-finished failure, and there's still no value you can pick that's guaranteed long enough for an even slower connection.

sitozzz:
> > why not just use debounce on action call here?
> 
> Debounce only blocks clicks within its own short window, not for as long as the request actually takes - so a click seconds later, while the first request is still pending, still gets through. If you set the debounce window large enough to safely cover slow 3G (several seconds), every user, including ones on fast connections, gets locked out of retrying for that full duration even after a quick, already-finished failure, and there's still no value you can pick that's guaranteed long enough for an even slower connection.

okay, in that way solution is acceptable

sitozzz (approved):

PR #7953, opened 2026-08-25
fix: INFRA-1724 email provider tweaks

sitozzz:
## Summary by CodeRabbit

- **New Features**
  - Added support for inline email attachments with Unisender Go while preserving content references in message HTML.
  - Long email message tags are now safely limited to 50 characters.

- **Bug Fixes**
  - Improved email delivery failure handling by consistently reporting valid numeric error statuses.
  - Prevented successful HTTP responses from being misreported as delivery errors.

- **Documentation**
  - Clarified provider-specific inline attachment behavior and Unisender Go requirements, including tag limits and BCC domain verification.

### DIFF
--- a/packages/keystone/emailAdapter.js
+++ b/packages/keystone/emailAdapter.js
@@ -211,7 +211,8 @@ const resolveInlineAttachments = async (meta, attachmentOptions) => {
 }
 
 /**
- * Rewrites `cid:filename` references in HTML to data URIs (for non-Mailgun providers).
+ * Rewrites `cid:filename` references in HTML to data URIs.
+ * Used by providers without native CID inline parts (e.g. Sendsay).
  * @param {string|null|undefined} html
  * @param {Array<{ originalFilename: string, mimetype: string, buffer: Buffer }>} inlines
  * @returns {string|null|undefined}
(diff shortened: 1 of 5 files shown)


packages/keystone/emailAdapter.js
```diff
@@ -361,13 +362,28 @@ class MailgunEmail {
 class UnisenderGoEmail {
     static type = 'unisendergo'
 
+    /** @see https://godocs.unisender.ru/web-api-ref#email-send */
+    static MAX_TAG_LENGTH = 50
```
AleX83Xpert, line 366: Why a static var instead of a regular constant? It looks a little strange when this static constant is used within the same class.

packages/keystone/emailAdapter.js
AleX83Xpert, line 531: Why "context"? This is actually "responseJson"

AleX83Xpert, line 443: It's good to add JSdoc with a return type, because it's useful to understand the context (or response) structure.

AleX83Xpert (approved):
lgtm

packages/keystone/emailAdapter.js
sitozzz: I just like to write static fields as constants inside class syntax, no special meaning here

packages/keystone/emailAdapter.js
sitozzz: Yep, but return type might be different from one adapter to another

PR #7958, opened 2026-08-27
feat(registry-importer): DOMA-13514 added payments file webhook

vovaaxeapolla:
## Summary by CodeRabbit

* **New Features**
  * Added webhook notifications when a payments file reaches the downloaded status.
  * Notifications support configured acquiring integrations and include file status details.
  * Added the `paymentsFile.status.updated` webhook event.
* **Improvements**
  * Application preparation now configures payments file webhook settings when needed.
* **Documentation**
  * Documented the new payments file status webhook event.
* **Tests**
  * Added coverage for successful notifications and cases where no notification is sent.


apps/condo/domains/acquiring/schema/PaymentsFile.js
```diff
@@ -132,7 +135,39 @@ const PaymentsFile = new GQLListSchema('PaymentsFile', {
...
         },
-        afterChange: async ({ operation, updatedItem }) => {
+        afterChange: async ({ context, operation, existingItem, updatedItem }) => {
+            const previousStatus = existingItem?.status
+            const newStatus = updatedItem?.status
+            const statusChanged = operation === 'update' &&
+                previousStatus !== newStatus &&
+                newStatus === PAYMENTS_FILE_DOWNLOADED_STATUS
+
+            if (statusChanged) {
```
YEgorLu: I think somewhere around here you should check that ACQUIRING_INTEGRATION_ID is provided

apps/condo/domains/acquiring/schema/PaymentsFile.js
```diff
@@ -132,7 +135,39 @@ const PaymentsFile = new GQLListSchema('PaymentsFile', {
             }
             return resolvedData
         },
-        afterChange: async ({ operation, updatedItem }) => {
+        afterChange: async ({ context, operation, existingItem, updatedItem }) => {
+            const previousStatus = existingItem?.status
+            const newStatus = updatedItem?.status
+            const statusChanged = operation === 'update' &&
```
AleX83Xpert: This is not just a regular change of status. The status changed to a particular value.
```suggestion
            const isDownloaded = operation === 'update' &&
```

YEgorLu (approved):
I don't understand how these webhooks work, but okay I guess.

I seen here plugin webHooked, in afterChange it checks for "WebhookSubscription" models for that schema and takes url from it. Can "WebhookSubscription" be used here instead of env?

apps/condo/domains/acquiring/schema/PaymentsFile.js
dkoviazin: better to name it to fileWasDownloaded as you check the final state PAYMENTS_FILE_DOWNLOADED_STATUS

apps/condo/domains/acquiring/schema/PaymentsFile.js
```diff
@@ -15,9 +17,10 @@ const {
     PAYMENTS_FILE_NEW_STATUS,
     PAYMENTS_FILE_DOWNLOADED_STATUS,
 } = require('@condo/domains/acquiring/constants/constants')
+const { WEBHOOK_EVENT_PAYMENTS_FILE_STATUS_UPDATED } = require('@condo/domains/common/constants/webhooks')
 const {  MONEY_AMOUNT_FIELD } = require('@condo/domains/common/schema/fields')
 const { convertFileNameToUTF8 } = require('@condo/domains/common/utils/fixFileNameEncoding')
-
+const { url: URL, secret: SECRET, acquiringIntegrationId: ACQUIRING_INTEGRATION_ID = null } = conf.PAYMENTS_FILE_WEBHOOK_CONFIG ? JSON.parse(conf.PAYMENTS_FILE_WEBHOOK_CONFIG) : {}
```
dkoviazin, line 23: Maybe, better to use additional field in acquiringIntegration model and not to use env

dkoviazin (approved):

AleX83Xpert (approved):
LGTM

abshnko (approved):
👍🏻

apps/condo/domains/acquiring/schema/PaymentsFile.js
vovaaxeapolla: I'll look at it. If ACQUIRING_INTEGRATION_ID is not provided it will be null. It's not normal to find some context with null integration

vovaaxeapolla:
> I don't understand how these webhooks work, but okay I guess.
> 
> I seen here plugin webHooked, in afterChange it checks for "WebhookSubscription" models for that schema and takes url from it. Can "WebhookSubscription" be used here instead of env?

I use new webhook transport coz it has signature of payload. Old WebhookSubscription doesn't use new transport and has no signature. Best is combo of these two but it is another problem

PR #8028, opened 2026-09-17
fix(condo): DOMA-13529 fixed updating Meter.archiveDate to null

abshnko:
## Summary by CodeRabbit

- **Bug Fixes**
  - Existing meter archive dates can now be cleared by explicitly setting the archive date to null.
  - Cleared archive dates remain cleared when meters are updated.
  - Existing archive dates are preserved when the archive date is omitted during an update.

- **Tests**
  - Added coverage for clearing archive dates and preserving them when the field is omitted.


apps/condo/domains/meter/schema/Meter.test.js
```diff
@@ -1405,6 +1405,19 @@ describe('Meter', () => {
                     })
                     expect(updatedMeter3).toHaveProperty('archiveDate', pastDate)
                 })
+
+                test('should clear an existing archiveDate', async () => {
```
YEgorLu, line 1409: Is there a test which checks that archiveDate will not clear if provided nullable / undefined or smth else?

Field resolver may change

abshnko: added a test for that case

dkoviazin (approved):

YEgorLu (approved):
ok

PR #8038, opened 2026-09-22
feat(condo): DOMA-13604 added soft delete cascade for Meters on Property deletion

abshnko:
## Summary by CodeRabbit

- **Bug Fixes**
  - Soft-deleting a property now also soft-deletes its associated meter records.
  - Existing property cleanup behavior remains supported.

- **Tests**
  - Added coverage to verify that associated meters are soft-deleted when a property is soft-deleted.


apps/condo/domains/meter/utils/serverSchema/resolveHelpers.js
```diff
@@ -23,6 +24,19 @@ async function connectContactToMeterReading (context, resolvedData, existingItem
     return contactId
 }
 
+async function softDeleteMetersByProperty (context, updatedItem) {
+    const { dv, sender, id } = updatedItem
+    const meterContext = context.createContext({ skipAccessControl: true })
+    const meters = await Meter.getAll(meterContext, {
```
YEgorLu: 1) why not find() ?

2) This funk is in "meter" domain, but expects updatedItem from "property" domain. I suggest move this or rename "updatedItem"

YEgorLu (approved):

